PR #1491, opened 2019-08-28
Adding option to apply git tags when using --pack and publish

chaseholland:
Adding CLI option to apply git tags when publishing and using the --pack option.


apps/rush-lib/src/cli/actions/PublishAction.ts
```diff
@@ -180,6 +181,12 @@ export class PublishAction extends BaseRushAction {
       parameterShortName: undefined,
       description: 'If this flag is specified with --publish, packages will be published with --force on npm'
     });
+    this._applyGitTagsOnPack = this.defineFlagParameter({
+      parameterLongName: '--apply-git-tags-on-pack',
+      parameterShortName: undefined,
```
iclanton: Just omit this.

```suggestion
```

apps/rush-lib/src/cli/actions/PublishAction.ts
```diff
@@ -180,6 +181,12 @@ export class PublishAction extends BaseRushAction {
...
+      description: `If specified with --publish and --pack, git tags will be applied for packages on pack` +
```
iclanton:
```suggestion
      description: `If specified with --publish and --pack, git tags will be applied for packages` +
```

common/changes/@microsoft/rush/master_2019-08-28-18-32.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Adding --apply-git-tags-on-pack flag to publish to apply git tags when using --pack",
```
iclanton:
```suggestion
      "comment": "Adding --apply-git-tags-on-pack flag to the publish command to apply git tags when using --pack",
```

apps/rush-lib/src/cli/actions/PublishAction.ts
```diff
@@ -180,6 +181,12 @@ export class PublishAction extends BaseRushAction {
       parameterShortName: undefined,
       description: 'If this flag is specified with --publish, packages will be published with --force on npm'
     });
+    this._applyGitTagsOnPack = this.defineFlagParameter({
```
iclanton: We should add a check to `run` to cause an error if `--publish` and `--pack` aren't also specified.

chaseholland: Shouldn't we only require `--pack`? That way it will spit out a dry run of the tags it's going to add.

iclanton: Ah yeah good point.

PR #1509, opened 2019-09-05
[rush] WIP: Add toggle to disable RushVersionManager when working on rush itself

VincentSurelle:
Testing Rush when working on it can be hard because of the `RushVersionManager`.

Actually, I put an environment variable toggle to disable this behavior and make it easy to use `rush` custom builds


iclanton:
It looks like you're running prettier or something in your editor. Can you revert the files without code changes to reduce churn?

Issue #1494, opened 2019-08-29
[rush] Add `--from-version-policy` flag


domarmstrong:
**Is this a feature or a bug?**

- [x] Feature
- [ ] Bug

**What is the expected behavior?**

In the same way that we have --to and --from. It would be useful to have an opposite of --to-version-policy. We use the monorepo for applications and dependencies, --to-version-policy is very useful to be able to just build our dependencies without the apps. There are a few times when I've equally just wanted to build the apps, --from-version-policy=app would allow us to do this, and it feels like a pairing of commands already available with -f and -t.

iclanton:
This seems useful. Want to take a crack at it?

domarmstrong:
I could in a few weeks

domarmstrong:
https://github.com/microsoft/web-build-tools/pull/1514

iclanton:
This has been released in Rush 5.13.0.

PR #1514, opened 2019-09-06
[rush] Add --from-version-policy build flag  #1494

domarmstrong:
A a mirror of --to-version-policy flag the same way --to has --from. #1494


common/changes/@microsoft/rush/master_2019-09-06-22-37.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Add --from-version-policy flag for \"rush rebuild\" to allow building from a version policy with direct and indirect dependant projects",
```
iclanton:
```suggestion
      "comment": "Add a --from-version-policy option for bulk commands to allow running the command (like build) from packages with a version policy and their direct and indirect dependent projects",
```

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -156,6 +157,12 @@ export class BulkScriptAction extends BaseScriptAction {
       argumentName: 'PROJECT1',
       description: 'Run command in the specified project and all of its dependencies'
     });
+    this._fromVersionPolicy =  this.defineStringListParameter({
+      parameterLongName: '--from-version-policy',
+      argumentName: 'VERSION_POLICY_NAME',
+      description: 'Run command in all projects with the specified version policy '
+        + 'and all projects that directly or indirectly depend on them'
```
iclanton:
```suggestion
        + 'and all projects that directly or indirectly depend on projects with the specified version policy'
```

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -184,11 +191,13 @@ export class BulkScriptAction extends BaseScriptAction {
...
   }
 
-  private _mergeToProjects(): string[] {
-    const projects: string[] = [...this._toFlag.values];
-    if (this._toVersionPolicy.values && this._toVersionPolicy.values.length) {
+  private _mergeProjectsWithVersionPolicy(flags: CommandLineStringListParameter,
+    versionPolicies: CommandLineStringListParameter): string[] {
+
+    const projects: string[] = [...flags.values];
+    if (versionPolicies.values && versionPolicies.values.length) {
```
iclanton: Let's make this check more explicit.

```suggestion
    if (versionPolicies.values && versionPolicies.values.length > 0) {
```

domarmstrong: 👍

domarmstrong:
Should I squash these, or do you keep the full history?

iclanton:
Keeping the full history is fine.

PR #1517, opened 2019-09-09
[rush-lib] Fixing issue in publishing when using --pack and --apply-git-tags-on-pack

chaseholland:
Resolving an issue where publish would fail when using `--pack` and `--apply-git-tags-on-pack` if the git tag already existed. Now skipping tags for packages that are already published using the same check as non-pack publish flow.

### DIFF
--- a/common/changes/@microsoft/rush/master_2019-09-09-18-25.json
+++ b/common/changes/@microsoft/rush/master_2019-09-09-18-25.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Fixing apply git tags during publish when using --pack and the tag already exists (existing tags not re-tagged, same as non --pack publish)",
+      "packageName": "@microsoft/rush",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush",
+  "email": "chaseholland@users.noreply.github.com"
+}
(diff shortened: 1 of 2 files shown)


common/changes/@microsoft/rush/master_2019-09-09-18-25.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Fixing apply git tags during publish when using --pack and the tag already exists (existing tags not re-tagged, same as non --pack publish)",
```
iclanton:
```suggestion
      "comment": "Fix an issue where Rush attempted to add Git tags for packages that had already been published when the publish command is run with the --pack flag. This caused a fatal error when tags already existed.",
```

chaseholland: Updated the comment with a commit and mentioned the `--apply-git-tags-on-pack` flag. Thanks for the suggestion!

common/changes/@microsoft/rush/master_2019-09-09-18-25.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Fix an issue where Rush attempted to add Git tags for packages that had already been published when the publish command is run with the --pack and --apply-git-tags-on-pack flag. This caused a fatal error when tags already existed.",
```
iclanton:
```suggestion
      "comment": "Fix an issue where Rush attempted to add Git tags for packages that had already been published when the publish command is run with the --pack and --apply-git-tags-on-pack flags. This caused a fatal error when tags already existed.",
```

PR #1580, opened 2019-10-10
[rush] Fix an issue where "rush build" will fail because a git command can be too long.

iclanton:
Fixes https://github.com/microsoft/rushstack/issues/1579

### DIFF
--- a/libraries/package-deps-hash/src/getPackageDeps.ts
+++ b/libraries/package-deps-hash/src/getPackageDeps.ts
@@ -2,10 +2,9 @@
 // See LICENSE in the project root for license information.
 
 import * as child_process from 'child_process';
-import { IPackageDeps } from './IPackageDeps';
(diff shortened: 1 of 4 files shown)


libraries/package-deps-hash/src/getPackageDeps.ts
```diff
@@ -92,41 +94,74 @@ export function parseGitStatus(output: string, packagePath: string): Map<string,
...
     {
-      cwd: path,
-      stdio: 'pipe'
-    }).toString();
+      currentWorkingDirectory: path
+    }
+  );
+
+  if (result.status !== 0) {
+    throw new Error(`git ls-tree exited with status ${result}: ${result.stderr}`);
```
sachinjoseph: `${result.status}`?

iclanton: Good catch.

PR #1591, opened 2019-10-17
Fix CommandLineChoiceParameter error when alternatives array contains a single value

KevinTCoughlin:
Fixes an issue where an error is thrown if the supplied alternative values array contains only a single value. This scenario is supported in the schema.


sdalonzo:
I'm a little late to the party after reporting this on gitter, but thank you!

PR #1612, opened 2019-11-01
[rush] Fix the issue: "rush add" should be able to add unpublished locally-linked projects 

Claudiazhaoya:
This address the issue: https://github.com/microsoft/rushstack/issues/991


.vscode/launch.json
```diff
@@ -1,6 +1,18 @@
 {
 	"version": "0.2.0",
 	"configurations": [
+		{
```
iclanton: Can you revert the changes to this file?

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -355,18 +355,28 @@ export class PackageJsonUpdater {
       }
       console.log(`Querying NPM registry for latest version of "${packageName}"...`);
 
+      //if the project exists in the local
+      const rushConfigProjects: RushConfigurationProject[] = this._rushConfiguration.projects;
+      rushConfigProjects.forEach(rushConfigProject => {
```
iclanton: Can you use an iterative loop here instead?

```suggestion
      for (const rushConfigProject of rushConfigProjects) {
```

common/config/rush/pnpm-lock.yaml
```diff
@@ -106,7 +106,7 @@ dependencies:
   '@types/through2': 2.0.32
```
iclanton: You can revert the changes to this file.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -355,18 +355,28 @@ export class PackageJsonUpdater {
       }
       console.log(`Querying NPM registry for latest version of "${packageName}"...`);
```
iclanton: This log message is no longer accurate. This should go inside the condition after the loop.

Same with the code [here](https://github.com/microsoft/rushstack/pull/1612/files#diff-51550989f206f917116c6152dd70b1e9R366-R371)

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -355,18 +355,28 @@ export class PackageJsonUpdater {
...
 
+      //if the project exists in the local
+      const rushConfigProjects: RushConfigurationProject[] = this._rushConfiguration.projects;
+      rushConfigProjects.forEach(rushConfigProject => {
+        if(rushConfigProject.packageName === packageName) {
+          selectedVersion = rushConfigProject.packageJson.version;
```
iclanton: You should break out of the loop here.

We also probably want to handle the case where a version is specified on the command-line and it doesn't match the version in the package.json

We also need to handle the case where the dependency that we're adding is marked as a cyclic dependency for the project in `rush.json`.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -355,18 +355,28 @@ export class PackageJsonUpdater {
       }
       console.log(`Querying NPM registry for latest version of "${packageName}"...`);
 
+      //if the project exists in the local
```
iclanton:
```suggestion
      // determine if the package is a project in the local repository
```

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -355,18 +355,28 @@ export class PackageJsonUpdater {
...
+      const rushConfigProjects: RushConfigurationProject[] = this._rushConfiguration.projects;
+      rushConfigProjects.forEach(rushConfigProject => {
+        if(rushConfigProject.packageName === packageName) {
+          selectedVersion = rushConfigProject.packageJson.version;
+        }
+      });
+
       let commandArgs: string[];
```
iclanton: This code should go inside the below condition.

common/changes/@microsoft/rush/zhas-rush-add-test_2019-11-01-21-43.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "rush add now can add unpublished locally-linked projects",
```
iclanton:
```suggestion
      "comment": "Fix an issue with Rush add, where Rush was unable to add unpublished local projects as dependencies.",
```

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +306,98 @@ export class PackageJsonUpdater {
...
 
-      let commandArgs: string[];
-      if (this._rushConfiguration.packageManager === 'yarn') {
-        commandArgs = ['info', packageName, 'versions', '--json'];
-      } else {
-        commandArgs = ['view', packageName, 'versions', '--json'];
+      // determine if the package is a project in the local repository and if the version exists
+      for (const rushConfigProject of rushConfigProjects) {
+        const rushConfigProjectVersion: string = rushConfigProject.packageJson.version;
+        if(rushConfigProject.packageName === packageName && rushConfigProject.cyclicDependencyProjects.size === 0) {
```
iclanton: The cyclic dependency check should probably happen before the loop starts.

I think this logic is also backwards: we need to check if the project we're adding the dependency *to* has the dependency listed as a cyclic. We also need to see if the dependency is in the set of cyclics, not just if there are any cyclic dependencies listed.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +306,98 @@ export class PackageJsonUpdater {
...
+          if(initialSpec === rushConfigProjectVersion) {
```
iclanton: This should use `semver.satisfies(...)`. If the initialspec is anything other than a single version (like `~1.2.3`), this will always cause an exception to be thrown.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +306,98 @@ export class PackageJsonUpdater {
 
     await installManager.ensureLocalPackageManager();
     let selectedVersion: string | undefined;
+    const rushConfigProjects: RushConfigurationProject[] = this._rushConfiguration.projects;
 
     if (initialSpec && initialSpec !== 'latest') {
```
iclanton: We probably want the local package check to happen even if a version specification isn't provided.

iclanton:
Can you list out the scenarios you tested? There are a lot of them and we should be sure the behavior is expected.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +309,98 @@ export class PackageJsonUpdater {
...
       }
-      console.log(`Querying NPM registry for latest version of "${packageName}"...`);
 
-      let commandArgs: string[];
-      if (this._rushConfiguration.packageManager === 'yarn') {
-        commandArgs = ['info', packageName, 'dist-tags.latest', '--silent'];
-      } else {
-        commandArgs = ['view', `${packageName}@latest`, 'version'];
+      // determine if the package is a project in the local repository
+      for (const rushConfigProject of rushConfigProjects) {
```
apostolisms: This seems to be duplicated above (very similar). Better to move to its own function

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +309,98 @@ export class PackageJsonUpdater {
...
       console.log(colors.gray('Finding versions that satisfy the selector: ') + initialSpec);
       console.log();
-      console.log(`Querying registry for all versions of "${packageName}"...`);
 
-      let commandArgs: string[];
-      if (this._rushConfiguration.packageManager === 'yarn') {
-        commandArgs = ['info', packageName, 'versions', '--json'];
-      } else {
-        commandArgs = ['view', packageName, 'versions', '--json'];
+      // determine if the package is a project in the local repository and if the version exists
```
apostolisms: Shouldn't we check for local packages after we checked the npm registry ?

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -268,13 +269,15 @@ export class PackageJsonUpdater {
    * Selects an appropriate version number for a particular package, given an optional initial SemVer spec.
    * If ensureConsistentVersions, tries to pick a version that will be consistent.
    * Otherwise, will choose the latest semver matching the initialSpec and append the proper range style.
+   * @param projects - the project which will have its package.json updated
```
iclanton: >the project which will have its package.json updated [](start = 23, length = 52)

This comment doesn't seem correct. It looks like all of the projects in the rushconfig.

iclanton:
```suggestion
   * @param projects - the projects which will have their `package.json`s updated
```

apps/rush-lib/src/logic/PackageJsonUpdater.ts
iclanton: If we find the local package, we don't need to check the registry.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +309,98 @@ export class PackageJsonUpdater {
...
 
-      let commandArgs: string[];
-      if (this._rushConfiguration.packageManager === 'yarn') {
-        commandArgs = ['info', packageName, 'versions', '--json'];
-      } else {
-        commandArgs = ['view', packageName, 'versions', '--json'];
+      // determine if the package is a project in the local repository and if the version exists
+      for (const rushConfigProject of rushConfigProjects) {
+        const rushConfigProjectVersion: string = rushConfigProject.packageJson.version;
+        if(rushConfigProject.packageName === packageName && !projects[0].cyclicDependencyProjects.has(packageName)) {
```
iclanton: We don't just want to check the first project in the `projects` array. It should probably check all of them.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -268,13 +269,15 @@ export class PackageJsonUpdater {
...
    * Otherwise, will choose the latest semver matching the initialSpec and append the proper range style.
+   * @param projects - the project which will have its package.json updated
    * @param packageName - the name of the package to be used
    * @param initialSpec - a semver pattern that should be used to find the latest version matching the spec
    * @param implicitlyPinnedVersion - the implicitly preferred (aka common/primary) version of the package in use
    * @param rangeStyle - if this version is selected by querying registry, then this range specifier is prepended to
    *   the selected version.
    */
   private async _getNormalizedVersionSpec(
+    projects: RushConfigurationProject[],
```
iclanton: Do we support multiple projects?

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +309,98 @@ export class PackageJsonUpdater {
...
-        commandArgs = ['view', packageName, 'versions', '--json'];
+      // determine if the package is a project in the local repository and if the version exists
+      for (const rushConfigProject of rushConfigProjects) {
+        const rushConfigProjectVersion: string = rushConfigProject.packageJson.version;
+        if(rushConfigProject.packageName === packageName && !projects[0].cyclicDependencyProjects.has(packageName)) {
+          if(semver.satisfies(rushConfigProjectVersion, initialSpec)) {
+            selectedVersion = initialSpec;
+            break;
+          } else {
+            throw new Error(`Unable to find a version of "${packageName}" that satisfies`
```
iclanton: Let's be a little more elaborate with this error:

```suggestion
            throw new Error(`The dependency being added ("${packageName}") is a project in the local Rush repository, but the version specifier provided (${initialSpec}) does not match the local project version (${rushConfigProjectVersion}). Correct the version specifier, omit a version specifier, or include "${packageName}" as a cyclicDependencyProject in rush.json for ${<project-being-added-to>} if it is intended for "${packageName}" to come from an external feed and not from the local Rush repository.`);
```

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +309,98 @@ export class PackageJsonUpdater {
...
-      let commandArgs: string[];
-      if (this._rushConfiguration.packageManager === 'yarn') {
-        commandArgs = ['info', packageName, 'dist-tags.latest', '--silent'];
-      } else {
-        commandArgs = ['view', `${packageName}@latest`, 'version'];
+      // determine if the package is a project in the local repository
+      for (const rushConfigProject of rushConfigProjects) {
+        if(rushConfigProject.packageName === packageName && !projects[0].cyclicDependencyProjects.has(packageName)) {
+          if(selectedVersion === undefined) {
+            selectedVersion = rushConfigProject.packageJson.version;
```
iclanton: This check should go outside of the loop.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +309,98 @@ export class PackageJsonUpdater {
...
 
-      let commandArgs: string[];
-      if (this._rushConfiguration.packageManager === 'yarn') {
-        commandArgs = ['info', packageName, 'dist-tags.latest', '--silent'];
-      } else {
-        commandArgs = ['view', `${packageName}@latest`, 'version'];
+      // determine if the package is a project in the local repository
+      for (const rushConfigProject of rushConfigProjects) {
+        if(rushConfigProject.packageName === packageName && !projects[0].cyclicDependencyProjects.has(packageName)) {
+          if(selectedVersion === undefined) {
```
iclanton: This check should go outside of the loop.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
Claudiazhaoya: I don't think so.. That's why I'm curious about why it has an array here. The statement is like this: projects = [currentProject]; Don't you think we actually don't need this array here?

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -264,17 +265,65 @@ export class PackageJsonUpdater {
     project.addOrUpdateDependency(packageName, newVersion, dependencyType!);
   }
 
+  /**
+   * Check if the package is the cyclic dependency in the project dependencies in package.json
+   */
+  public isCyclicDependency(
```
iclanton: Let's make these `private` to begin with.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -264,17 +265,65 @@ export class PackageJsonUpdater {
...
+      if(project.cyclicDependencyProjects.has(packageName)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Check if the project is local project
+   * If it is found in rush.json and is not a cyclic dependency, return true
```
iclanton: This function returns a version specifier, not true or false.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -264,17 +265,65 @@ export class PackageJsonUpdater {
...
+   * Otherwise, return false
+   * @param initialSpec - a semver pattern that should be used to find the latest version matching the spec
+   * @param packageName - the name of the package to be used
+   * @param projects - the projects which will have their package.json's updated
+   * @param rushConfigProjects - the projects which are in the rush.json
+   */
+  public isLocalProject(
```
iclanton: A better name for this would be `getVersionForLocalDependency`.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -306,67 +355,91 @@ export class PackageJsonUpdater {
...
-      let commandArgs: string[];
-      if (this._rushConfiguration.packageManager === 'yarn') {
-        commandArgs = ['info', packageName, 'versions', '--json'];
+    // determine if the package is a project in the local repository and if the version exists
+      const version: string | undefined = this.isLocalProject(initialSpec, packageName, projects, rushConfigProjects);
+      if(version !== undefined) {
+        selectedVersion = version
       } else {
-        commandArgs = ['view', packageName, 'versions', '--json'];
+        throw new Error(`The dependency being added ("${packageName}") is a project in the local Rush repository,
```
iclanton: Split this onto multiple lines by concatenating strings. The way this is written, there will be line breaks and indentation in the error message.

iclanton: This will throw if the dependency being added isn't a local dependency. In that case, we want to just query the feed.

apps/rush-lib/src/logic/PackageJsonUpdater.ts
iclanton: Looks like this is used when the `--all` flag is specified.

PR #1627, opened 2019-11-13
[gulp-core-build-serve] Support additional static paths in `gulp serve` task

Js-Brecht:
This adds a small bit of functionality to add support for serving additional paths as static files via `gulp-connect`.

In various mono-repos (including `rush`, `yarn`, `pnpm`, and `lerna`), dependencies are symlinked from all of the workspaces to a central location.  This causes problems with `gulp serve`, because it is not serving files from that location.

For example, say I have a `rush` repo named `foo`, and a workspace named `bar`, which requires dependency `baz`; it would be structured like this:
```
- foo
|--> /common/temp/node_modules
   |--> /.pnpm/registry/baz
   |--> /baz (-> ./.pnpm/registry/baz)
|--> /workspaces
   |--> /bar
      |--> /node_modules
         |--> /baz (-> ../../../common/temp/node_modules/baz)
```

So, as you can see, `baz` will be resolved to `/foo/common/temp/node_modules/.pnpm/registry/baz`.  Meanwhile, `gulp serve` is serving files underneath `/foo/workspaces/bar`, so it doesn't have access to the `baz` dependency.  Because of this type of circumstance the SPFx local workbench will fail; dependencies vital to how the workbench runs will be missing/inaccessible.

With this PR, developers will be able to write a `gulpfile.js` that looks like this:
```js
const build = require('@microsoft/sp-build-web');
const config = build.getConfig();
config.staticPaths = [
   {
      path: path.resolve('../../common/temp/node_modules'),
      url: '/common/temp/node_modules',
   }
];
```
and successfully run their local workbench.

---

I extended the public API in the build config because I thought it would be a good idea to have clearly defined type defs, instead of just using the `properties` bag to pass the desired values around.

I have tested with SPFx 1.9.1, using `pnpm` version 4.1.7.  It works as expected.  I have yet to do any comprehensive testing with other package managers/SPFx versions.

Of course, to benefit from this, it requires coercing `@microsoft/gulp-core-build` and `@microsoft/gulp-core-build-serve` to a higher version, but I didn't experience any issues with doing that.  It's possible that earlier SPFx versions might.

Should fix SharePoint/sp-dev-docs#3736, but I need to test with version 1.8.2 before I know for sure.

### DIFF
--- a/common/changes/@microsoft/gulp-core-build-serve/feature-static-paths_2019-11-13-00-26.json
+++ b/common/changes/@microsoft/gulp-core-build-serve/feature-static-paths_2019-11-13-00-26.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/gulp-core-build-serve",
+      "comment": "Implemented functionality in `gulp-connect` for serving additional static paths defined via buildConfig.staticPaths.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@microsoft/gulp-core-build-serve",
+  "email": "1935258+Js-Brecht@users.noreply.github.com"
+}
--- a/common/changes/@microsoft/gulp-core-build-serve/feature-static-paths_2019-11-13-00-26.json
+++ b/common/changes/@microsoft/gulp-core-build-serve/feature-static-paths_2019-11-13-00-26.json
@@ -3,7 +3,7 @@
     {
       "packageName": "@microsoft/gulp-core-build-serve",
       "comment": "Implemented functionality in `gulp-connect` for serving additional static paths defined via buildConfig.staticPaths.",
-      "type": "patch"
+      "type": "minor"
     }
   ],
   "packageName": "@microsoft/gulp-core-build-serve",
--- a/common/changes/@microsoft/gulp-core-build/feature-static-paths_2019-11-13-00-26.json
+++ b/common/changes/@microsoft/gulp-core-build/feature-static-paths_2019-11-13-00-26.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/gulp-core-build",
+      "comment": "Added API endpoint for serving additionnal static paths with `gulp serve`",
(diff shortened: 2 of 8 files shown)


halfnibble:
Hello @Js-Brecht, I am talking with the SPFx people about https://github.com/SharePoint/sp-dev-docs/issues/3736

We have a rush monorepo with webparts based on SPFx. And I can `gulp serve` content from `node_modules` (which is actually a symlink, as you stated) just fine. So I'm curious to learn more about the issue you are having.

Can you provide a minimal repro that demonstrates the problem with `gulp serve`? Also, which OS are you using? Thanks.

Js-Brecht:
@halfnibble I'm curious to know how yours is able to work.  I've had the `gulp serve` issue for quite a while now.  I think it used to work in SPFx v1.7, but since v1.8 it's been broken.  Not sure what changed, or how it was able to work before.

If I remember correctly, it has to do with `@microsoft/sp-webpart-workbench`, and some other dependencies... they aren't included in the bundle, so when the server is started, the browser tries to retrieve their source... which winds up being outside of the currently served root directory because of the symlink.  If it makes a difference, it's actually a Junction on Windows, because of how Windows handles symlinks.

This problem exists in pnpm workspaces, too; not just rush monorepo.  Oddly enough, I was able to use pnpm workspaces for a while, so I just didn't use rush.  However, it wound up breaking, too; I think it was after the upgrade to v1.9.1 (been a little while now).  Again, not really sure why it was able to work before.

If you'd like, I can include some of the console output; it's what led me to this solution.

halfnibble:
@Js-Brecht This interesting. I was able to repro the issue using your example.

The biggest difference I've noticed so far is that our manifests file (for serve, located in `<project>/temp/manifests.js`) resolved all packages via `node_modules`, whereas running in your repro, I get urls with `common/temp/node_modules`, which should not occur.

Our configurations look very similar. I'll dig deeper and let you know what's going on.

iclanton:
GCB isn't in this repo anymore

PR #1633, opened 2019-11-20
Removing check for --registry when using --pack

chaseholland:
Removing the `--registry` check when using `--pack`, since some pack flows will run `npm run view`, which requires the registry component.

### DIFF
--- a/common/changes/@microsoft/rush/master_2019-11-20-02-32.json
+++ b/common/changes/@microsoft/rush/master_2019-11-20-02-32.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Removing registry check from --pack, since --pack might need to run view against a registry to check if a package exists",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush",
+  "email": "chaseholland@users.noreply.github.com"
+}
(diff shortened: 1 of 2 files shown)


iclanton:
Can you run `rush change` and include a changelog?

chaseholland:
There is a change file?

common/changes/@microsoft/rush/master_2019-11-20-02-32.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Removing registry check from --pack, since --pack might need to run view against a registry to check if a package exists",
```
iclanton:
```suggestion
      "comment": "Remove an error thrown when the --registry and --pack arguments are used on rush publish, because --pack might be required to check if a package has been published against a custom registry.",
```

common/changes/@microsoft/rush/master_2019-11-20-02-32.json
```diff
@@ -0,0 +1,11 @@
...
+      "type": "patch"
```
iclanton: Rush is versioned with a version policy, so this should stay empty.

```suggestion
      "type": ""
```

common/changes/@microsoft/rush/master_2019-11-20-02-32.json
chaseholland: Amended with "Remove an error thrown when the --registry and --pack arguments are used on rush publish, because **--registry** might be required to check if a package has **already** been published against a custom registry." -- Edits in bold.

PR #1639, opened 2019-11-22
[rush] Fix applying git tags when using --pack --apply-git-tags-on-pack and using publish --include-all

chaseholland:
Resolving an issue where git tags would not be applied when using `--pack --apply-git-tags-on-pack` and using `publish --include-all`.

The issue was, publish was checking if a package was already published before adding a tag. In the `publish --include-all` flow, rush publishes and then goes to add the tag, so the package reads as published and the tag is not applied. 

This check has been moved to checking whether not a tag already exists locally via a new utility method in `PublishGit.ts`.

### DIFF
--- a/apps/rush-lib/src/cli/actions/PublishAction.ts
+++ b/apps/rush-lib/src/cli/actions/PublishAction.ts
@@ -324,13 +324,15 @@ export class PublishAction extends BaseRushAction {
             return;
           }
 
+          const packageVersion: string = packageConfig.packageJson.version;
+
           // Do not create a new tag if one already exists, this will result in a fatal error
           if (git.hasTag(packageConfig)) {
-            console.log(`Not tagging ${packageName}@${packageConfig.packageJson.version}. A tag already exists for this version.`);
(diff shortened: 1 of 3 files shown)


apps/rush-lib/src/cli/actions/PublishAction.ts
```diff
@@ -324,18 +324,22 @@ export class PublishAction extends BaseRushAction {
...
 
-          git.addTag(!!this._publish.value && !this._registryUrl.value, packageName, packageConfig.packageJson.version);
+          git.addTag(!!this._publish.value, packageName, packageVersion);
           updated = true;
         };
 
         if (this._pack.value) {
           // packs to tarball instead of publishing to NPM repository
           this._npmPack(packageName, packageConfig);
+          // Do not tag packages that already exist. This will fail with a fatal error.
```
iclanton: Are we missing a check here?

chaseholland: No, the code moved but the comment didn't. d'oh. Removing the comment.

PR #1691, opened 2020-01-16
Extract debug certificate logic into new @microsoft/debug-certificate-manager

halfnibble:
The reason being for use in a new Webpack plugin. Coming soon.


common/reviews/api/debug-certificate-generator.api.md
```diff
@@ -0,0 +1,48 @@
...
+export interface ILogging {
+    // (undocumented)
+    log: (string: any) => void;
+    // (undocumented)
+    logError: (string: any) => void;
+    // (undocumented)
+    logVerbose: (string: any) => void;
+    // (undocumented)
+    logWarning: (string: any) => void;
+}
```
iclanton: Let's just use `Terminal` here.

common/reviews/api/debug-certificate-generator.api.md
```diff
@@ -0,0 +1,48 @@
+## API Report File for "@microsoft/debug-certificate-manager"
```
iclanton: Can you document the undocumented APIs?

common/reviews/api/debug-certificate-generator.api.md
```diff
@@ -0,0 +1,48 @@
...
+    logVerbose: (string: any) => void;
+    // (undocumented)
+    logWarning: (string: any) => void;
+}
+
+// @public (undocumented)
+export function untrustCertificate(parentTask: ILogging): boolean;
+
+
+// (No @packageDocumentation comment for this package)
```
iclanton: Can you add `@packageDocumentation`?

common/reviews/api/debug-certificate-generator.api.md
iclanton: Also - delete this file in favor of `debug-certificate-manger.api.md`.

core-build/gulp-core-build-serve/README.md
```diff
@@ -121,7 +121,7 @@ The certificate is generated and self-signed with a unique private key and an at
 it (Windows and macOS only). If the user does not agree to trust the certificate, provides invalid root
 credentials, or something else goes wrong, the `TrustCertTask` will fail and the `ServeTask` will serve
 with the default, non-trusted certificate. If trust succeeds, the certificate and the private key are
-dropped in the `.gcb-serve-data` directory in the user's home folder in `PEM` format. On platforms
+dropped in the `.rushstack-serve-data` directory in the user's home folder in `PEM` format. On platforms
```
iclanton: Let's rename this from `rushstack-serve-data` to something like just `.rushstack`.

core-build/gulp-core-build-serve/src/UntrustCertTask.ts
```diff
@@ -3,8 +3,8 @@
 
 import { GulpTask } from '@microsoft/gulp-core-build';
 import * as Gulp from 'gulp';
-import { CertificateStore } from './CertificateStore';
-import { untrustCertificate } from './certificates';
+import { CertificateStore } from '@microsoft/debug-certificate-manager';
+import { untrustCertificate } from '@microsoft/debug-certificate-manager';
```
iclanton: Put these in the same import.

libraries/debug-certificate-manager/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Installation
+
+`npm install @microsoft/debug-certificate-manager --save-dev`
+
+## Overview
+
+This library contains utilities for managing debug certificates in a development server environment. It provides functions to generate, self-sign, trust, and untrust .pem certificates for both Windows and Mac OS. It will also generate self-signed certificates on other OS's, but the user must manually trust and untrust them.
+
+[![npm version](https://badge.fury.io/js/%40microsoft%2Fdebug-certificate-manager.svg)](https://badge.fury.io/js/%40microsoft%2Fdebug-certificate-manager)
```
iclanton: This will be `@rushstack/debug-certificate-manager`

libraries/debug-certificate-manager/README.md
```diff
@@ -0,0 +1,42 @@
...
+[![Build Status](https://travis-ci.org/Microsoft/debug-certificate-manager.svg?branch=master)](https://travis-ci.org/Microsoft/debug-certificate-manager) [![Dependencies](https://david-dm.org/Microsoft/debug-certificate-manager.svg)](https://david-dm.org/Microsoft/debug-certificate-manager)
+
+
+## CertificateStore
+
+The CertificateStore class follows a singleton pattern and provides accessors and mutators for the certificate data stored in `.rushstack-serve-data`.
```
iclanton:
```suggestion
The `CertificateStore` class follows a singleton pattern and provides accessors and mutators for the certificate data stored in `.rushstack-serve-data`.
```

libraries/debug-certificate-manager/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Retrieve an instance with `CertificateStore.instance`.
```
iclanton: I'm not a huge fan of this pattern anymore. We should either just make the members of the class static, or instantiate the class. Personally I'd learn toward the latter.

libraries/debug-certificate-manager/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Overview
+
+This library contains utilities for managing debug certificates in a development server environment. It provides functions to generate, self-sign, trust, and untrust .pem certificates for both Windows and Mac OS. It will also generate self-signed certificates on other OS's, but the user must manually trust and untrust them.
+
+[![npm version](https://badge.fury.io/js/%40microsoft%2Fdebug-certificate-manager.svg)](https://badge.fury.io/js/%40microsoft%2Fdebug-certificate-manager)
+[![Build Status](https://travis-ci.org/Microsoft/debug-certificate-manager.svg?branch=master)](https://travis-ci.org/Microsoft/debug-certificate-manager) [![Dependencies](https://david-dm.org/Microsoft/debug-certificate-manager.svg)](https://david-dm.org/Microsoft/debug-certificate-manager)
+
+
+## CertificateStore
```
iclanton:
```suggestion
## `CertificateStore`
```

libraries/debug-certificate-manager/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+The CertificateStore class follows a singleton pattern and provides accessors and mutators for the certificate data stored in `.rushstack-serve-data`.
+
+Retrieve an instance with `CertificateStore.instance`.
+
+## ensureCertificate
```
iclanton:
```suggestion
## `ensureCertificate`
```

libraries/debug-certificate-manager/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Get the dev certificate from the store, or optionally, generate a new one and trust it if one does not exist in the store. Returns a certificate object following the `ICertificate` interface.
```
iclanton:
```suggestion
Get a dev certificate from the store, or optionally, or generate a new one and trust it if one does not exist in the store. Returns a certificate object following the `ICertificate` interface.
```

libraries/debug-certificate-manager/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```typescript
+export interface ICertificate {
+  pemCertificate: string | undefined;
+  pemKey: string | undefined;
+}
+```
+
+## untrustCertificate
```
iclanton:
```suggestion
## `untrustCertificate`
```

libraries/debug-certificate-manager/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@microsoft/debug-certificate-manager",
```
iclanton:
```suggestion
  "name": "@rushstack/debug-certificate-manager",
```

libraries/debug-certificate-manager/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "version": "0.1.0",
+  "description": "Cross-platform functionality to create debug ssl certificates.",
+  "main": "lib/index.js",
+  "typings": "dist/node-core-library.d.ts",
```
iclanton:
```suggestion
  "typings": "dist/debug-certificate-manager.d.ts",
```

libraries/debug-certificate-manager/src/CertificateStore.ts
```diff
@@ -81,10 +84,10 @@ export class CertificateStore {
       throw new Error('Unable to determine the current user\'s home directory');
     }
 
-    this._gcbServeDataPath = path.join(this._userProfilePath, '.gcb-serve-data');
-    FileSystem.ensureFolder(this._gcbServeDataPath);
+    this._serveDataPath = path.join(this._userProfilePath, '.rushstack-serve-data');
```
iclanton: This should maintain backwards compatibility with `.gcb-serve-data`.

It'd be cool if it looked in `.gcb-serve-data` and moved the data to the new folder.

libraries/debug-certificate-manager/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
iclanton: I'm not seeing a `tslint.json`.

common/reviews/api/debug-certificate-generator.api.md
halfnibble: Are you saying I should get a `Terminal` passed in, or that I should create my own `ConsoleTerminalProvider` for these functions. Or something else?

libraries/debug-certificate-manager/package.json
halfnibble: I was talking to APC about this. Unless we plan on moving over to the new group name in the next week or so, I think it would be best to leave it as @microsoft for now and let it be part of the mass update later. How do you feel about that?

common/reviews/api/debug-certificate-generator.api.md
iclanton: Let's pass in a `Terminal` instance.

libraries/debug-certificate-manager/package.json
iclanton: Let's just start out with the right scope so we don't have to potentially tombstone the @microsoft package.

common/reviews/api/debug-certificate-generator.api.md
halfnibble: The only example I could find of using Terminal with a GulpTask was in gulp-core-build-typescript. So I extracted the GCBTerminalProvider to an export in @microsoft/gulp-core-build and I'm using that to create the Terminal instance in @microsoft/gulp-core-build-serve. Let me know if I should do this differently.

core-build/gulp-core-build/src/utilities/GCBTerminalProvider.ts
```diff
@@ -1,12 +1,15 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-import { GulpTask } from '@microsoft/gulp-core-build';
+import { GulpTask } from '../tasks/GulpTask';
```
iclanton: Can you separate the local import into a different block?

libraries/debug-certificate-manager/src/certificates.ts
```diff
@@ -16,28 +11,41 @@ import { runSudoSync, ISudoSyncResult } from './sudoSync';
...
+   */
+  pemCertificate: string | undefined;
+
+  /**
+   * Private key used to sign the pem certificate
+   */
+  pemKey: string | undefined;
 }
 
 function _createDevelopmentCertificate(): ICertificate {
```
iclanton: Can you wrap all of these in a class?

core-build/gulp-core-build/src/utilities/GCBTerminalProvider.ts
halfnibble: Done.

libraries/debug-certificate-manager/src/certificates.ts
halfnibble: Let me know if this looks good. I called it, `CertificateManger`.

halfnibble:
While having `CertificateManager` convert previously generated certs from GCB seemed like a good idea, I've found it does not work in practice. At least not on Mac.

It might be the new path or the different "friendly name", but the certificates are reported as "invalid." Trusted but invalid.

The only way to make the certificate valid was to do both of the following:
1. Not copy the GCB cert over to .rushstack (generate a new cert).
2. Delete the "localhost" entry prior to trusting the newly generated cert.

Only doing one or the other (or neither) of the above results in an "Invalid_Cert" error in Chrome.

Therefore, instead of having this copy over the  old certificate, I think it should delete it and remove it from the KeyChain. @iclanton does that sound reasonable?

halfnibble:
Update: I think this is good to go now @iclanton 

Tested multiple scenarios. (It wasn't working before because the build command had not succeeded—but it does work to add 2 certs with the same commonName and different serial numbers).

libraries/debug-certificate-manager/src/CertificateStore.ts
halfnibble: This was found to not be backwards compatible, and therefore, we decided on generating a new certificate with a 1 character change in the serial number.

PR #1698, opened 2020-01-20
[rush] Make default branch and origin name configurable

Jabher:
I've been looking through sources and accidentally understood that moving this constanted variables to configuration will take me around 20 minutes, so why not.

This can be useful for a lot of people.

### DIFF
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -1045,7 +1045,7 @@ export class RushConfiguration {
   /**
    * The default fully-qualified git remote of the repository. This helps "rush change" find the right remote to compare against.
    */
-  public get repositoryDefaultFullyQualifiedRemote(): string {
+  public get repositoryDefaultFullyQualifiedRemoteBranch(): string {
     return `${this.repositoryDefaultRemote}/${this.repositoryDefaultBranch}`;
   }
 
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -83,6 +83,8 @@ export interface IRushRepositoryJson {
    * The remote url of the repository. This helps "rush change" find the right remote to compare against.
    */
   url: string;
+  defaultBranch: string;
+  defaultRemote: string;
 }
 
 /**
@@ -293,6 +295,8 @@ export class RushConfiguration {
 
   // Repository info
   private _repositoryUrl: string;
+  private _repositoryDefaultBranch: string;
+  private _repositoryDefaultRemote: string;
 
   private _pnpmOptions: PnpmOptionsConfiguration;
   private _yarnOptions: YarnOptionsConfiguration;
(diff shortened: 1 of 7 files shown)


common/changes/@microsoft/rush/add-git-properties_2020-01-20-16-47.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Add support of defaultBranch and defaultOrigin as repo configuration properties",
```
iclanton:
```suggestion
      "comment": "Make the default branch and default remote configurable.",
```

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -1022,6 +1028,27 @@ export class RushConfiguration {
...
+
+  /**
+   * The default git remote of the repository. This helps "rush change" find the right remote to compare against.
+   */
+  public get repositoryDefaultRemote(): string {
+    return this._repositoryDefaultRemote;
+  }
+
+  /**
+   * The default fully-qualified git remote of the repository. This helps "rush change" find the right remote to compare against.
```
```suggestion
   * The default fully-qualified git remote branch of the repository. This helps "rush change" find the right branch to compare against.
```

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -1022,6 +1028,27 @@ export class RushConfiguration {
...
+   */
+  public get repositoryDefaultFullyQualifiedRemote(): string {
```
iclanton: This should probably be `repositoryDefaultFullyQualifiedBranch` or `repositoryDefaultFullyQualifiedRemoteBranch`.

apps/rush-lib/src/utilities/VersionControl.ts
```diff
@@ -7,10 +7,7 @@ import {
   Executable,
   Path
 } from '@microsoft/node-core-library';
-
-const DEFAULT_BRANCH: string = 'master';
-const DEFAULT_REMOTE: string = 'origin';
-const DEFAULT_FULLY_QUALIFIED_BRANCH: string = `${DEFAULT_REMOTE}/${DEFAULT_BRANCH}`;
+import {RushConfiguration} from '..';
```
iclanton:
```suggestion
import { RushConfiguration } from '../RushConfiguration';
```

iclanton:
Can you also update the `rush init` `rush.json` template [here](https://github.com/microsoft/rushstack/blob/master/apps/rush-lib/assets/rush-init/rush.json#L207)?

apps/rush-lib/src/api/RushConfiguration.ts
Jabher: (this is relied to next change too)
remote is the stored string key for remote repository URL + settings relied to it; I do not know whether it can be named better. [Git docs](https://git-scm.com/docs/git-remote) does not give it other name rather than "remote"
> Manage the set of repositories ("remotes") whose branches you track.

apps/rush-lib/src/api/RushConfiguration.ts
Jabher: fixed

apps/rush-lib/src/api/RushConfiguration.ts
iclanton: This property is specifically referring to the remote branch, though. See docs here: https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches

Git refers to the <remote>/<branchname> as either "remote branch" or "remote tracking branch."

Jabher: ah, got it. my bad, I thought this is about an another line

PR #1716, opened 2020-01-25
[rush] Ensure the Rush version selector installs a working version of mkdirp.

iclanton:


iclanton:
Now that this (https://github.com/microsoft/rushstack/issues/1713#issuecomment-578418792) fix has been released, the fix in this PR is no longer necessary.

PR #1727, opened 2020-01-28
Update to rush 5.19.3

iclanton:


common/scripts/install-run.js
```diff
@@ -373,7 +373,12 @@ function installAndRun(packageName, packageVersion, packageBinName, packageBinAr
         cwd: process.cwd(),
         env: process.env
     });
-    return result.status;
+    if (result.status !== null) {
+        return result.status;
+    }
+    else {
```
halfnibble: Completely stylistic, but it looks like we use `} else {` in most place in this repo.

iclanton: This is compiled code. That's what the TS compiler produced.

halfnibble: Lol. I knew that... 😳

PR #1734, opened 2020-02-01
[localization-plugin] Add RESX support to the localization plugin.

iclanton:


webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -484,22 +402,22 @@ export class LocalizationPlugin implements Webpack.Plugin {
...
     // END configuration
 
     // START options.filesToIgnore
     { // eslint-disable-line no-lone-blocks
-      this._locJsonFilesToIgnore = new Set<string>();
-      for (const locJsonFilePath of this._options.filesToIgnore || []) {
-        const normalizedLocJsonFilePath: string = path.resolve(configuration.context!, locJsonFilePath);
-        this._locJsonFilesToIgnore.add(normalizedLocJsonFilePath);
+      this._filesToIgnore = new Set<string>();
+      for (const filePath of this._options.filesToIgnore || []) {
```
D4N14L: This seems less involved than above where you checked if the file was an absolute path, does this matter here? Will it always be relative to the configuration.context?

webpack/localization-plugin/src/utilities/ResxReader.ts
```diff
@@ -0,0 +1,270 @@
...
+        ResxReader._logWarningWithLocation(
+          options,
+          'Missing string comment in <data> element',
+          dataElement
+        );
+      }
+
+      return {
+        value,
+        comment: comment ||  ''
```
D4N14L: Seems to conflict with the above check for undefined comment

webpack/localization-plugin/src/LocalizationPlugin.ts
iclanton: Actually, `path.resolve` handles if the path is already absolute, so the other absolute check is unnecessary. Thanks for noticing that.

PR #1735, opened 2020-02-02
[set-webpack-public-path-plugin] Add functionality to use the webpack-produced asset name to find the script path.

iclanton:


webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -157,43 +159,51 @@ export class LocalizationPlugin implements Webpack.Plugin {
...
-                const localizedChunkAssets: { [locale: string]: string } = {};
-                for (const [locale, newAsset] of resultingAssets) {
-                  compilation.assets[newAsset.filename] = newAsset.asset;
-                  localizedChunkAssets[locale] = newAsset.filename;
-                }
+            for (const chunk of chunkGroup.chunks) {
+              // Clone the chunk files array because we're going to modify it
+              const chunkFiles: string[] = [...chunk.files];
+              const chunkFilesSet: Set<string> = new Set(chunkFiles);
+              for (const chunkFileName of chunkFiles) {
```
apostolisms: can this be optimized?

webpack/set-webpack-public-path-plugin/src/SetPublicPathPlugin.ts
```diff
@@ -127,48 +147,79 @@ export class SetPublicPathPlugin implements Webpack.Plugin {
...
+            });
+          } else {
+            return source;
+          }
+        });
+      });
+
+      compiler.hooks.emit.tap(PLUGIN_NAME, (compilation: Webpack.compilation.Compilation) => {
+        for (const chunkGroup of compilation.chunkGroups) {
+          for (const chunk of chunkGroup.chunks) {
```
apostolisms: can this be optimized?

webpack/localization-plugin/src/LocalizationPlugin.ts
iclanton: Not really. It'll still end up iterating through the same number of elements, regardless of how we do the iteration.

webpack/set-webpack-public-path-plugin/src/SetPublicPathPlugin.ts
iclanton: Same story - we still need to loop through each chunk-produced asset, so regardless of how we loop through them, it'll be the same number of elements.

PR #1736, opened 2020-02-03
[localization-plugin] Support wrapping strings in a "default" export.

iclanton:


webpack/localization-plugin/src/loaders/LocLoader.ts
```diff
@@ -1,20 +1,21 @@
...
 import { LocFileParser } from '../utilities/LocFileParser';
+import {
+  loaderFactory,
+  IBaseLoaderOptions
+} from './LoaderFactory';
 
-export interface ILocLoaderOptions {
+export interface ILocLoaderOptions extends IBaseLoaderOptions {
   pluginInstance: LocalizationPlugin;
+  exportAsDefault: boolean;
```
D4N14L: This appears to be duplicate now

iclanton: Good catch.

PR #1737, opened 2020-02-04
[localization-plugin] Generate loc file typings.

iclanton:


webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -84,11 +85,42 @@ export class LocalizationPlugin implements Webpack.Plugin {
...
 
+    let typingsPreprocessor: TypingsGenerator | undefined;
+    if (this._options.typingsOptions) {
+      typingsPreprocessor = new TypingsGenerator({
+        srcFolder: this._options.typingsOptions.sourceRoot || compiler.context,
+        generatedTsFolder: this._options.typingsOptions.generatedTsFolder,
+        exportAsDefault: this._options.exportAsDefault,
+        filesToIgnore: this._options.filesToIgnore
+      });
+    } else {
```
D4N14L: Nit: Just set to undefined right off the bat

PR #1744, opened 2020-02-12
Move RushConfiguration access in action to unblock Init

D4N14L:
`rush init` was broken by attempting to access RushConfiguration in the constructor of the publish action. I've moved the initialization into the run() method to initialize when the action is actually run.

### DIFF
--- a/common/changes/@microsoft/rush/danade-fixBrokenInit_2020-02-12-02-01.json
+++ b/common/changes/@microsoft/rush/danade-fixBrokenInit_2020-02-12-02-01.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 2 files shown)


common/changes/@microsoft/rush/danade-fixBrokenInit_2020-02-12-02-01.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix broken init command caused by early RushConfiguration access",
```
iclanton:
```suggestion
      "comment": "",
```

Issue #1745, opened 2020-02-12
[rush] Coloring eventHook fail


VincentSurelle:
## Please prefix the issue title with the project name i.e. [rush], [api-extractor] etc. ##

**Is this a feature or a bug?**

- [x] Feature
- [ ] Bug

**Please describe the actual behavior.**

I would like to colorize this kind of message :
```
Event hook “node common/scripts/post-rush-install.js” failed. Run “rush” with --debug to see detailed error information
```

It's an error than can have impact on some projects. I think keeping these messages white can lead to unexpected errors.

Red can be too much for some projects. I think yellowing the message to make it a warning is good enough.

I can obviously make a PR for this ;)

I just want to know if : 

- It's ok ?
- Red or Yellow ?

Thanks

iclanton:
>It's ok ?

Go ahead!

>Red or Yellow ?

I agree with you that red may be too much. Yellow sounds good.

VincentSurelle:
Merged in #1747

PR #1747, opened 2020-02-14
[rush-lib] Adds yellow output on event hook fail

VincentSurelle:
This PR adds a yellow coloring when an event hook fails.

It has been discussed on #1745

### DIFF
--- a/apps/rush-lib/src/logic/EventHooksManager.ts
+++ b/apps/rush-lib/src/logic/EventHooksManager.ts
@@ -45,8 +45,10 @@ export class EventHooksManager {
             }
           );
         } catch (error) {
-          console.error(`${os.EOL} Event hook "${script}" failed. Run "rush" with --debug` +
-            ` to see detailed error information.`);
+          console.error(
+            os.EOL + colors.yellow(`Event hook "${script}" failed. Run "rush" with --debug`) +
+            ` to see detailed error information.`
+          );
           if (isDebug) {
             console.error(os.EOL + error.message);
           }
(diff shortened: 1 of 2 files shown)


apps/rush-lib/src/logic/EventHooksManager.ts
```diff
@@ -45,8 +45,10 @@ export class EventHooksManager {
             }
           );
         } catch (error) {
-          console.error(`${os.EOL} Event hook "${script}" failed. Run "rush" with --debug` +
-            ` to see detailed error information.`);
+          console.error(
+            os.EOL + colors.yellow(`Event hook "${script}" failed. Run "rush" with --debug`) +
```
iclanton: Can you make the ` to see detailed error information.` part yellow too?

common/changes/@microsoft/rush/fix-event-hook-fail-color_2020-02-14-12-22.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Adds yellow output on event hook fail",
```
iclanton:
```suggestion
      "comment": "Make the event hook failure message print in yellow.",
```

apps/rush-lib/src/logic/EventHooksManager.ts
VincentSurelle: Done sorry, made a mistake when removing prettier auto-correct :/

PR #1758, opened 2020-02-26
[localization-plugin] Fix a few issues with localized asset path renaming.

iclanton:


webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -194,10 +194,18 @@ export class LocalizationPlugin implements Webpack.Plugin {
                 options.contentHashType === 'javascript' &&
                 assetPath.match(Constants.LOCALE_FILENAME_PLACEHOLDER_REGEX)
               ) {
-                return assetPath.replace(
-                  Constants.LOCALE_FILENAME_PLACEHOLDER_REGEX,
-                  `" + ${Constants.JSONP_PLACEHOLDER} + "`
+                // Does this look like an async chunk URL generator?
+                if (typeof options.chunk.id === 'string' && options.chunk.id.match(/^\" +/)) {
```
sachinjoseph: I am not very familiar with this, but are we interested in ids that look like regex `^" .*`?

Should this regex be a const?

iclanton: This is just looking for a particular case where we're being passed something that looks like a string concatenation snippet.

However, there is a mistake here
```suggestion
                if (typeof options.chunk.id === 'string' && options.chunk.id.match(/^\" \+/)) {
```

PR #1770, opened 2020-03-07
[rush] Patch pnpm last-install error handling

Js-Brecht:
There's a small bug in the work done in #1657 that I missed.  It's probably not something that would be encountered frequently, but it is there nonetheless.

If somebody switches from a global default path for pnpm's store, to a static path using the `local` option, or `RUSH_PNPM_STORE_PATH`, then the condition that checks for an error will return false, so it won't handle the error gracefully.  Since the store path DID change, though, it will fall through to PNPM, which will throw an exception.

The value for the global default is an empty string, which, of course, is falsy.  The error occurs here:
https://github.com/microsoft/rushstack/blob/15f796370e964876f7c28dba6680cf09244ac9bc/apps/rush-lib/src/api/LastInstallFlag.ts#L64-L69
If `oldState.storePath` evaluates to false, then it will not handle the error state.  That's erroneous, and that check is redundant, anyway.  That was just supposed to be a check to ensure that the `storePath` existed before, meaning the package manager hadn't been changed TO pnpm.  It would have been valid if it said `oldState.storePath !== undefined`... but that condition is handled by `oldState.packageManager === pkgManager` already.

### DIFF
--- a/apps/rush-lib/src/api/LastInstallFlag.ts
+++ b/apps/rush-lib/src/api/LastInstallFlag.ts
@@ -61,12 +61,10 @@ export class LastInstallFlag {
         const pkgManager: PackageManagerName = newState.packageManager;
         if (pkgManager === 'pnpm') {
           if (
-            ( // Only throw an error if the package manager hasn't changed from PNPM
-              oldState.packageManager === pkgManager
-            ) && ( // Throw if the store path changed
-              oldState.storePath &&
-              oldState.storePath !== newState.storePath
-            )
+            // Only throw an error if the package manager hasn't changed from PNPM
+            oldState.packageManager === pkgManager &&
+            // Throw if the store path changed
+            oldState.storePath !== newState.storePath
           ) {
             const oldStorePath: string = oldState.storePath || '<global>';
             const newStorePath: string = newState.storePath || '<global>';
--- a/common/changes/@microsoft/rush/patch-pnpm-store-options_2020-03-07-23-18.json
+++ b/common/changes/@microsoft/rush/patch-pnpm-store-options_2020-03-07-23-18.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "fix false-negative condition when pnpm store path has changed",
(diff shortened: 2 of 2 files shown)


Js-Brecht:
> Is that the expected behavior? The ones I marked with (?) I think are a little questionable. Should we throw in that case, versus just allowing PNPM to initialize a new store if one doesn't already exist? My primary concern with this feature is polluting a shared store with packages from another feed.

The reason it is throwing an exception here is to keep pnpm from throwing an exception.  **Any** time the store changes paths, PNPM will throw.  The decision to intercept that was made in the original issue thread.  Conversation is [here](https://github.com/microsoft/rushstack/issues/1414#issuecomment-554848122).

PNPM does initialize a new store if one doesn't exist wherever you've pointed it.  If you point it to one that exists, it will use what's already there.  All PNPM cares about is if the current `node_modules` is linked to the store it's currently configured to use.  If it's not, it tells you that you need to run `pnpm install --force` which wipes out `node_modules` and relinks everything from the current store.

Personally, I think that wiping out `node_modules` is sufficient, and Rush already does that if `last-install.flag !== the new state`; but I don't know all the ins and outs of Rush, so I just assumed that there is more going on, or that needs to happen, than I am aware of.

> Just out of curiosity, does PNPM isolate packages by the feed they originated from?

Well... to a degree.  It isolates them by registry domain, then by version.

It does not isolate packages that come from different feeds on the same domain...  kind of like Azure Artifacts do.  All packages that I get from different Azure feeds go into the same folder.

There might wind up being issues if you have multiple feeds, on the same domain, hosting the same package with the same version.  If they are the same version, then you will always get the package that is already on the disk, from whichever feed it was downloaded from first.  If they aren't the same content (for whatever reason), then there will be issues.  So, if it came from the same domain, and has the same version, then PNPM will not go ask for it again, unless the store is deleted/pruned.

D4N14L:
It's been a bit since this came up, but I've integrated the changes into this PR: https://github.com/microsoft/rushstack/pull/2023

D4N14L:
Closing this out as the change was integrated into #2023. Thanks for finding this and proposing a change!

PR #1772, opened 2020-03-10
[rush] Support custom environment variables for the underlying package manager install process, expose maxInstallAttempts as a parameter for rush install/update.

sachinjoseph:
Provide a way to set environment variables for the underlying package manager install command. Expose maxInstallAttempts as a parameter for rush install and update.

For example, this can be used to set `NODE_OPTIONS=--max-old-space-size=<number>` environment variable to work around package manager running out of memory during installation.

Expose maxInstallAttempts as a parameter for rush install/update: This parameter can be used to retry the installation a specified number of times, and this can greatly increase the chances of the install succeeding especially when the network performance is not great.

Sample usage of new config in rush.json:

```json
  "rushVersion": "5.19.3",
  "pnpmVersion": "4.11.5",
  "pnpmOptions": {
    "environmentVariables": {
      "NODE_OPTIONS": {
        "value": "--max-old-space-size=8192",
        "override": true
      }
    }
  },
```

### DIFF
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -241,7 +241,7 @@ export interface ICurrentVariantJson {
  *
  * @public
  */
-export class PackageManagerOptionsConfigurationBase implements IPackageManagerOptionsJsonBase {
+export abstract class PackageManagerOptionsConfigurationBase implements IPackageManagerOptionsJsonBase {
   /**
    * Enviroment variables for the package manager
    */
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -248,8 +248,8 @@ export abstract class PackageManagerOptionsConfigurationBase implements IPackage
   public readonly environmentVariables?: IConfigurationEnvironment
 
   /** @internal */
-  protected constructor(environmentVariables?: IConfigurationEnvironment) {
-    this.environmentVariables = environmentVariables;
+  protected constructor(json: IPackageManagerOptionsJsonBase) {
+    this.environmentVariables = json.environmentVariables;
   }
 }
 
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -128,7 +128,8 @@ export interface IConfigurationEnvironment {
 }
 
 /**
- * Represents an environment variable
+ * Represents the value of an environment variable, and if the value should be overridden if the variable is set
+ * in the parent environment.
  * @public
  */
 export interface IConfigurationEnvironmentVariable {
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -106,13 +106,32 @@ export interface IRushRepositoryJson {
 export type PnpmStoreOptions = 'local' | 'global';
 
 /**
- * Part of IRushConfigurationJson.
- * @internal
+ * Options for the package manager.
+ * @public
  */
 export interface IPackageManagerOptions {
+  /**
+   * Enviroment variables for the package manager
+   */
   environmentVariables?: {
-    [name: string]: string
-  };
+    [environmentVariableName: string]: IEnvironmentVariable;
+  }
+}
+
+/**
+ * Represents an environment variable
+ * @public
+ */
+export interface IEnvironmentVariable {
+  /**
+   * Value of the environment variable
+   */
+  value: string;
+
+  /**
+   * Set to true to override the environment variable even if it is set on the device
+   */
+  override?: boolean; //
 }
 
 /**
@@ -1039,6 +1058,9 @@ export class RushConfiguration {
     return this._projectFolderMaxDepth;
   }
 
+  /**
+   * {@inheritDoc IPackageManagerOptions}
+   */
   public get packageManagerOptions(): IPackageManagerOptions | undefined {
     return this._packageManagerOptions;
   }
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -174,7 +174,7 @@ export interface IPnpmOptionsJson extends IPackageManagerOptionsJsonBase {
  * @internal
  */
 export interface IYarnOptionsJson extends IPackageManagerOptionsJsonBase {
-    /**
+  /**
    * If true, then Rush will add the "--ignore-engines" option when invoking Yarn.
    * This allows "rush install" to succeed if there are dependencies with engines defined in
    * package.json which do not match the current environment.
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -140,7 +140,7 @@ export interface IConfigurationEnvironmentVariable {
   /**
    * Set to true to override the environment variable even if it is set on the device
    */
-  override?: boolean; //
+  override?: boolean;
 }
 
 /**
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -265,7 +265,7 @@ export abstract class PackageManagerOptionsConfigurationBase implements IPackage
 export class NpmOptionsConfiguration extends PackageManagerOptionsConfigurationBase {
   /** @internal */
   public constructor(json: INpmOptionsJson) {
-    super(json.environmentVariables);
+    super(json);
   }
 }
 
@@ -327,7 +327,7 @@ export class PnpmOptionsConfiguration extends PackageManagerOptionsConfiguration
 
   /** @internal */
   public constructor(json: IPnpmOptionsJson, commonTempFolder: string) {
-    super(json.environmentVariables);
+    super(json);
     this.pnpmStore = json.pnpmStore || 'local';
     if (EnvironmentConfiguration.pnpmStorePathOverride) {
       this.pnpmStorePath = EnvironmentConfiguration.pnpmStorePathOverride;
@@ -362,7 +362,7 @@ export class YarnOptionsConfiguration extends PackageManagerOptionsConfiguration
 
   /** @internal */
   public constructor(json: IYarnOptionsJson) {
-    super(json.environmentVariables);
+    super(json);
     this.ignoreEngines = !!json.ignoreEngines;
   }
 }
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -109,6 +109,16 @@ export type PnpmStoreOptions = 'local' | 'global';
  * Part of IRushConfigurationJson.
  * @internal
  */
+export interface IPackageManagerOptions {
+  environmentVariables?: {
+    [name: string]: string
+  };
+}
+
+/**
+ * Part of IRushConfigurationJson.
+ * @internal
+ */
 export interface IPnpmOptionsJson {
   /**
    * The store resolution method for PNPM to use
@@ -160,6 +170,7 @@ export interface IRushConfigurationJson {
   projects: IRushConfigurationProjectJson[];
   eventHooks?: IEventHooksJson;
   hotfixChangeEnabled?: boolean;
+  packageManagerOptions?: IPackageManagerOptions;
   pnpmOptions?: IPnpmOptionsJson;
   yarnOptions?: IYarnOptionsJson;
   ensureConsistentVersions?: boolean;
@@ -351,6 +362,8 @@ export class RushConfiguration {
   private _repositoryDefaultBranch: string;
   private _repositoryDefaultRemote: string;
 
+  private _packageManagerOptions?: IPackageManagerOptions;
+
   private _pnpmOptions: PnpmOptionsConfiguration;
   private _yarnOptions: YarnOptionsConfiguration;
 
@@ -449,6 +462,8 @@ export class RushConfiguration {
         + ` and ${packageManagerFields[1]} `);
     }
 
+    this._packageManagerOptions = rushConfigurationJson.packageManagerOptions;
+
     if (this._packageManager === 'npm') {
       this._packageManagerToolVersion = rushConfigurationJson.npmVersion!;
       this._packageManagerWrapper = new NpmPackageManager(this._packageManagerToolVersion);
@@ -463,10 +478,10 @@ export class RushConfiguration {
     this._shrinkwrapFilename = this._packageManagerWrapper.shrinkwrapFilename;
 
     this._tempShrinkwrapFilename = path.join(
-        this._commonTempFolder, this._shrinkwrapFilename
+      this._commonTempFolder, this._shrinkwrapFilename
     );
     this._packageManagerToolFilename = path.resolve(path.join(
-        this._commonTempFolder, `${this.packageManager}-local`, 'node_modules', '.bin', `${this.packageManager}`
+      this._commonTempFolder, `${this.packageManager}-local`, 'node_modules', '.bin', `${this.packageManager}`
     ));
 
     /// From "C:\repo\common\temp\pnpm-lock.yaml" --> "C:\repo\common\temp\pnpm-lock-preinstall.yaml"
@@ -475,9 +490,9 @@ export class RushConfiguration {
       parsedPath.name + '-preinstall' + parsedPath.ext);
 
     RushConfiguration._validateCommonRushConfigFolder(
-        this._commonRushConfigFolder,
-        this.packageManager,
-        this._shrinkwrapFilename
+      this._commonRushConfigFolder,
+      this.packageManager,
+      this._shrinkwrapFilename
     );
 
     this._projectFolderMinDepth = rushConfigurationJson.projectFolderMinDepth !== undefined
@@ -642,11 +657,11 @@ export class RushConfiguration {
       if (semver.major(Rush.version) !== semver.major(expectedRushVersion)
         || semver.minor(Rush.version) !== semver.minor(expectedRushVersion)) {
 
-          // If the major/minor are different, then make sure it's an older version.
-          if (semver.lt(Rush.version, expectedRushVersion)) {
-            throw new Error(`Unable to load ${rushJsonBaseName} because its RushVersion is`
-              + ` ${rushConfigurationJson.rushVersion}, whereas @microsoft/rush-lib is version ${Rush.version}.`
-              + ` Consider upgrading the library.`);
+        // If the major/minor are different, then make sure it's an older version.
+        if (semver.lt(Rush.version, expectedRushVersion)) {
+          throw new Error(`Unable to load ${rushJsonBaseName} because its RushVersion is`
+            + ` ${rushConfigurationJson.rushVersion}, whereas @microsoft/rush-lib is version ${Rush.version}.`
+            + ` Consider upgrading the library.`);
         }
       }
     }
@@ -741,9 +756,9 @@ export class RushConfiguration {
    * recognized config files.
    */
   private static _validateCommonRushConfigFolder(
-      commonRushConfigFolder: string,
-      packageManager: PackageManagerName,
-      shrinkwrapFilename: string
+    commonRushConfigFolder: string,
+    packageManager: PackageManagerName,
+    shrinkwrapFilename: string
   ): void {
     if (!FileSystem.exists(commonRushConfigFolder)) {
       console.log(`Creating folder: ${commonRushConfigFolder}`);
@@ -1022,6 +1037,10 @@ export class RushConfiguration {
    */
   public get projectFolderMaxDepth(): number {
     return this._projectFolderMaxDepth;
+  }
+
+  public get packageManagerOptions(): IPackageManagerOptions | undefined {
+    return this._packageManagerOptions;
   }
 
   /**
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -107,22 +107,31 @@ export type PnpmStoreOptions = 'local' | 'global';
 
 /**
  * Options for the package manager.
- * @public
+ * @internal
  */
-export interface IPackageManagerOptionsBase {
+export interface IPackageManagerOptionsJsonBase {
   /**
    * Enviroment variables for the package manager
    */
-  environmentVariables?: {
-    [environmentVariableName: string]: IEnvironmentVariable;
-  }
+  environmentVariables?: IConfigurationEnvironment
+}
+
+/**
+ * A collection of environment variables
+ * @internal
+ */
+export interface IConfigurationEnvironment {
+  /**
+   * Environment variables
+   */
+  [environmentVariableName: string]: IConfigurationEnvironmentVariable;
 }
 
 /**
  * Represents an environment variable
- * @public
+ * @internal
  */
-export interface IEnvironmentVariable {
+export interface IConfigurationEnvironmentVariable {
   /**
    * Value of the environment variable
    */
@@ -135,17 +144,17 @@ export interface IEnvironmentVariable {
 }
 
 /**
- * Options that are only used when the NPM package manager is selected.
- * @public
+ * Part of IRushConfigurationJson.
+ * @internal
  */
-export interface INpmOptions extends IPackageManagerOptionsBase {
+export interface INpmOptionsJson extends IPackageManagerOptionsJsonBase {
 }
 
 /**
  * Part of IRushConfigurationJson.
  * @internal
  */
-export interface IPnpmOptionsJson extends IPackageManagerOptionsBase {
+export interface IPnpmOptionsJson extends IPackageManagerOptionsJsonBase {
   /**
    * The store resolution method for PNPM to use
    */
@@ -163,7 +172,7 @@ export interface IPnpmOptionsJson extends IPackageManagerOptionsBase {
 /**
  * Part of IRushConfigurationJson.
  */
-export interface IYarnOptionsJson extends IPackageManagerOptionsBase {
+export interface IYarnOptionsJson extends IPackageManagerOptionsJsonBase {
   ignoreEngines?: boolean;
 }
 
@@ -196,7 +205,7 @@ export interface IRushConfigurationJson {
   projects: IRushConfigurationProjectJson[];
   eventHooks?: IEventHooksJson;
   hotfixChangeEnabled?: boolean;
-  npmOptions?: INpmOptions;
+  npmOptions?: INpmOptionsJson;
   pnpmOptions?: IPnpmOptionsJson;
   yarnOptions?: IYarnOptionsJson;
   ensureConsistentVersions?: boolean;
@@ -220,6 +229,39 @@ export interface ICurrentVariantJson {
 }
 
 /**
+ * Options that all package managers share.
+ *
+ * @public
+ */
+class PackageManagerConfigurationBase implements IPackageManagerOptionsJsonBase {
+  /**
+   * Enviroment variables for the package manager
+   */
+  public readonly environmentVariables?: IConfigurationEnvironment
+
+  /** @internal */
+  protected constructor(environmentVariables?: IConfigurationEnvironment) {
+    this.environmentVariables = environmentVariables;
+  }
+}
+
+/**
+ * Options that are only used when the NPM package manager is selected.
+ *
+ * @remarks
+ * It is valid to define these options in rush.json even if the NPM package manager
+ * is not being used.
+ *
+ * @public
+ */
+export class NpmOptionsConfiguration extends PackageManagerConfigurationBase {
+  /** @internal */
+  public constructor(json: INpmOptionsJson) {
+    super(json.environmentVariables);
+  }
+}
+
+/**
  * Options that are only used when the PNPM package manager is selected.
  *
  * @remarks
@@ -228,7 +270,7 @@ export interface ICurrentVariantJson {
  *
  * @public
  */
-export class PnpmOptionsConfiguration implements IPackageManagerOptionsBase {
+export class PnpmOptionsConfiguration extends PackageManagerConfigurationBase {
   /**
    * The method used to resolve the store used by PNPM.
    *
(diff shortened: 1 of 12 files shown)


apps/rush-lib/src/cli/actions/BaseInstallAction.ts
```diff
@@ -59,6 +60,11 @@ export abstract class BaseInstallAction extends BaseRushAction {
       description: 'Activates verbose logging for the package manager. You will probably want to pipe'
         + ' the output of Rush to a file when using this command.'
     });
+    this._maxInstallAttempts = this.defineIntegerParameter({
+      parameterLongName: '--max-install-attempts',
+      argumentName: 'NUMBER',
+      description: 'Attempts command the specified number of times.'
```
D4N14L: Maybe mention default value is 3?

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -160,6 +170,7 @@ export interface IRushConfigurationJson {
   projects: IRushConfigurationProjectJson[];
   eventHooks?: IEventHooksJson;
   hotfixChangeEnabled?: boolean;
+  packageManagerOptions?: IPackageManagerOptions;
```
iclanton: There are already `pnpmOptions` and `yarnOptions` objects. It seems like this should just be part of those.

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
iclanton:
```suggestion
      description: 'Overrides the default maximum number of install attempts. The default value is 3.'
```

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
```diff
@@ -96,6 +102,10 @@ export abstract class BaseInstallAction extends BaseRushAction {
       }
     }
 
+    if (this._maxInstallAttempts.value && this._maxInstallAttempts.value < 1) {
```
iclanton: We probably should also throw if they provide `0`

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -131,6 +134,7 @@ export class InstallManager {
   private _rushGlobalFolder: RushGlobalFolder;
   private _commonNodeModulesMarker: LastInstallFlag;
   private _commonTempFolderRecycler: AsyncRecycler;
+  private _maxInstallAttempts: number = 3;
```
iclanton: Intentionally increasing this?

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -50,8 +50,6 @@ import { CommonVersionsConfiguration } from '../api/CommonVersionsConfiguration'
 // The PosixModeBits are intended to be used with bitwise operations.
 /* eslint-disable no-bitwise */
 
-const MAX_INSTALL_ATTEMPTS: number = 2;
```
iclanton: Let's leave this here and rename it to `DEFAULT_MAX_INSTALL_ATTEMPTS`

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -897,6 +905,16 @@ export class InstallManager {
           // Example: "C:\MyRepo\common\temp\npm-local\node_modules\.bin\npm"
           const packageManagerFilename: string = this._rushConfiguration.packageManagerToolFilename;
 
+          const packageManagerEnv: NodeJS.ProcessEnv = process.env;
+
+          if (this._rushConfiguration.packageManagerOptions &&
+            this._rushConfiguration.packageManagerOptions.environmentVariables) {
```
iclanton:
```suggestion
          if (
            this._rushConfiguration.packageManagerOptions &&
            this._rushConfiguration.packageManagerOptions.environmentVariables
          ) {
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -897,6 +905,16 @@ export class InstallManager {
...
+
+            for (const envVar of Object.keys(this._rushConfiguration.packageManagerOptions.environmentVariables)) {
```
iclanton: Just use `for ( ... in ... )` instead of `Object.keys`

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -897,6 +905,16 @@ export class InstallManager {
           // Example: "C:\MyRepo\common\temp\npm-local\node_modules\.bin\npm"
           const packageManagerFilename: string = this._rushConfiguration.packageManagerToolFilename;
 
+          const packageManagerEnv: NodeJS.ProcessEnv = process.env;
```
iclanton: Is this the existing behavior?

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -897,6 +905,16 @@ export class InstallManager {
...
+
+          if (this._rushConfiguration.packageManagerOptions &&
+            this._rushConfiguration.packageManagerOptions.environmentVariables) {
+
+            for (const envVar of Object.keys(this._rushConfiguration.packageManagerOptions.environmentVariables)) {
+              packageManagerEnv[envVar] = this._rushConfiguration.packageManagerOptions.environmentVariables[envVar];
```
iclanton: What if the process's environment already has an env variable that's specified in the configuration? I'd think the local environment should take precedence. Maybe we should print a warning?

common/changes/@microsoft/rush/sachinjoseph-support-custom-environment-variables-for-install-manager_2020-03-10-18-12.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Support custom environment variables for the underlying package manager install process, expose maxInstallAttempts as a parameter for rush install/update.",
```
iclanton:
```suggestion
      "comment": "Support setting environment variables for package manager install processes in rush.json and expose --max-install-attempts as a parameter for rush install/update.",
```

common/reviews/api/rush-lib.api.md
```diff
@@ -295,6 +303,10 @@ export class RushConfiguration {
     readonly npmCacheFolder: string;
     readonly npmTmpFolder: string;
     readonly packageManager: PackageManagerName;
+    // Warning: (ae-incompatible-release-tags) The symbol "packageManagerOptions" is marked as @public, but its signature references "IPackageManagerOptions" which is marked as @internal
```
iclanton: This should be public.

common/reviews/api/rush-lib.api.md
```diff
@@ -295,6 +303,10 @@ export class RushConfiguration {
...
+    //
+    // (undocumented)
```
iclanton: Docs?

apps/rush-lib/src/api/RushConfiguration.ts
sachinjoseph: I thought this could be a place for settings that apply to the underlying package manager, irrespective of the specific choice made.

apps/rush-lib/src/logic/InstallManager.ts
sachinjoseph: yes, I want to increase it to 3. I will retain the constant and set it to 3 as per your suggestion

apps/rush-lib/src/logic/InstallManager.ts
sachinjoseph: Yes, we the process is launched with the "environment" in which rush was launched.

apps/rush-lib/src/logic/InstallManager.ts
sachinjoseph: I will add a warning. I wanted the rush config to override any setting on the machine by default.

I think you are right about the precedence. How about we support an optional flag `override` in rush config for each environment variable, with default value `false`? That would make it configureable and explicit.

iclanton: I like that.

apps/rush-lib/src/api/RushConfiguration.ts
iclanton: I think it's confusing to have two fields which, on the surface level, appear to do the same thing. I think we should make this the base type of the other two fields.

@octogonz - do you have an opinion here?

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
sachinjoseph: We throw if the value is less than `1`, that would cover `0`

apps/rush-lib/src/logic/InstallManager.ts
sachinjoseph: Done

apps/rush-lib/src/api/RushConfiguration.ts
sachinjoseph: Interesting that the build is passing without changelogs for `rush-lib`

iclanton: `rush-lib`'s changelog is covered by `rush`'s changelog because they're in the same publishing group.

apps/rush-lib/src/cli/actions/InstallAction.ts
```diff
@@ -34,7 +35,9 @@ export class InstallAction extends BaseInstallAction {
       recheckShrinkwrap: false,
       networkConcurrency: this._networkConcurrencyParameter.value,
       collectLogFile: this._debugPackageManagerParameter.value!,
-      variant: this._variant.value
+      variant: this._variant.value,
+      maxInstallAttempts: this._maxInstallAttempts.value ? this._maxInstallAttempts.value :
+        RushConstants.defaultMaxInstallAttempts
```
iclanton:
```suggestion
      maxInstallAttempts: this._maxInstallAttempts.value !== undefined
        ? this._maxInstallAttempts.value
        : RushConstants.defaultMaxInstallAttempts
```

apps/rush-lib/src/cli/actions/UpdateAction.ts
```diff
@@ -62,7 +63,9 @@ export class UpdateAction extends BaseInstallAction {
       recheckShrinkwrap: this._recheckParameter.value!,
       networkConcurrency: this._networkConcurrencyParameter.value,
       collectLogFile: this._debugPackageManagerParameter.value!,
-      variant: this._variant.value
+      variant: this._variant.value,
+      maxInstallAttempts: this._maxInstallAttempts.value ? this._maxInstallAttempts.value :
+        RushConstants.defaultMaxInstallAttempts
```
iclanton:
```suggestion
      maxInstallAttempts: this._maxInstallAttempts.value !== undefined
        ? this._maxInstallAttempts.value
        : RushConstants.defaultMaxInstallAttempts
```

apps/rush-lib/src/api/RushConfiguration.ts
sachinjoseph: >
>
> I think we should make this the base type of the other two fields.

Making this the base type in the rush.json would be a breaking change.

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -105,6 +105,35 @@ export interface IRushRepositoryJson {
...
+  environmentVariables?: {
+    [environmentVariableName: string]: IEnvironmentVariable;
+  }
+}
+
+/**
+ * Represents an environment variable
+ * @public
+ */
+export interface IEnvironmentVariable {
```
iclanton: Let's call this `IConfigurationEnvironmentVariable`

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -134,11 +134,18 @@ export interface IEnvironmentVariable {
   override?: boolean; //
 }
 
+/**
+ * Options that are only used when the NPM package manager is selected.
+ * @public
```
iclanton: Let's make this consistent with `IPnpmOptionsJson` and `IYarnOptionsJson`.

```suggestion
 * Part of IRushConfigurationJson.
 * @internal
```

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -221,7 +228,7 @@ export interface ICurrentVariantJson {
  *
  * @public
  */
-export class PnpmOptionsConfiguration {
+export class PnpmOptionsConfiguration implements IPackageManagerOptionsBase {
```
iclanton: It'd be simpler to just make a base class.

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -903,44 +903,37 @@ export class InstallManager {
...
-                  console.log(`  Value: ${envVarValueInRushConfig}`);
-                  packageManagerEnv[envVar] = envVarValueInRushConfig;
-                }
-              }
+          if (this._rushConfiguration.packageManager === 'npm') {
+            if (
+              this._rushConfiguration.npmOptions &&
+              this._rushConfiguration.npmOptions.environmentVariables
+            ) {
+              packageManagerEnv = InstallManager._mergeEnvironmentVariables(
```
iclanton: Instead of calling this three times, stick it in a temporary variable and call it once.

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1076,6 +1069,48 @@ export class InstallManager {
...
 
+  private static _mergeEnvironmentVariables(
+    baseEnv: NodeJS.ProcessEnv,
+    environmentVariables: { [environmentVariableName: string]: IEnvironmentVariable; } | undefined
+  ): NodeJS.ProcessEnv {
+    const packageManagerEnv: NodeJS.ProcessEnv = baseEnv;
+
+    if (environmentVariables) {
+      for (const envVar in environmentVariables) {
+        if ({}.hasOwnProperty.call(environmentVariables, envVar)) {
```
iclanton:
```suggestion
        if (environmentVariables.hasOwnProperty(envVar)) {
```

If that doesn't work, just remove the check and suppress the lint error. Our JSON parser doesn't populate `hasOwnProperty` for some reason.

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1076,6 +1069,48 @@ export class InstallManager {
...
+          let setEnvironmentVariable: boolean = true;
+          console.log(`\nProcessing definition for environment variable: ${envVar}`);
+
+          if (process.env[envVar]) {
+            setEnvironmentVariable = false;
+            console.log(colors.yellow(`WARNING: Environment variable already defined on the device:`));
```
iclanton:
```suggestion
            console.log(colors.yellow(`WARNING: Environment variable already defined:`));
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1076,6 +1069,48 @@ export class InstallManager {
...
+            console.log(`  Name: ${envVar}`);
+            console.log(`  Value set on the device: ${process.env[envVar]}`);
```
iclanton:
```suggestion
            console.log(`  Existing value: ${baseEnv[envVar]}`);
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1076,6 +1069,48 @@ export class InstallManager {
...
+  ): NodeJS.ProcessEnv {
+    const packageManagerEnv: NodeJS.ProcessEnv = baseEnv;
+
+    if (environmentVariables) {
+      for (const envVar in environmentVariables) {
+        if ({}.hasOwnProperty.call(environmentVariables, envVar)) {
+          let setEnvironmentVariable: boolean = true;
+          console.log(`\nProcessing definition for environment variable: ${envVar}`);
+
+          if (process.env[envVar]) {
```
iclanton:
```suggestion
          if (baseEnv[envVar]) {
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1076,6 +1069,48 @@ export class InstallManager {
...
+            setEnvironmentVariable = false;
+            console.log(colors.yellow(`WARNING: Environment variable already defined on the device:`));
+            console.log(`  Name: ${envVar}`);
+            console.log(`  Value set on the device: ${process.env[envVar]}`);
+            console.log(`  Value set in Rush config: ${environmentVariables[envVar].value}`);
```
iclanton:
```suggestion
            console.log(`  Value set in rush.json: ${environmentVariables[envVar].value}`);
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1076,6 +1069,48 @@ export class InstallManager {
...
+
+            if (environmentVariables[envVar].override &&
+              environmentVariables[envVar].override === true) {
+              setEnvironmentVariable = true;
+              console.log(colors.yellow(`WARNING: Overriding the environment variable with the value set in Rush config.`));
```
iclanton:
```suggestion
              console.log(colors.yellow(`WARNING: Overriding the environment variable with the value set in rush.json.`));
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1076,6 +1069,48 @@ export class InstallManager {
...
+            }
+            else {
+              console.log(colors.yellow(`WARNING: Not overriding the value of the environment variable.`));
+            }
+          }
+
+          if (setEnvironmentVariable === true) {
```
iclanton:
```suggestion
          if (setEnvironmentVariable) {
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1076,6 +1069,48 @@ export class InstallManager {
...
+            console.log(`Setting environment variable for package manager.`);
```
iclanton: This logging will be really verbose. Maybe we should only log the non-warnings if rush is invoked with `--debug`

apps/rush-lib/src/schemas/rush.schema.json
```diff
@@ -93,6 +94,22 @@
             "fewer-dependencies",
             "fast"
           ]
+        },
+        "environmentVariables": {
```
iclanton: There is a convention for "extending" objects in JSON schemata. Let's do that instead of duplicating this.

common/reviews/api/rush-lib.api.md
```diff
@@ -425,6 +432,9 @@ export class YarnOptionsConfiguration {
     //
     // @internal
     constructor(json: IYarnOptionsJson);
+    readonly environmentVariables?: {
+        [environmentVariableName: string]: IEnvironmentVariable;
```
iclanton: Let's rename this type to `IConfigurationEnvironmentVariable`

apps/rush-lib/src/logic/InstallManager.ts
sachinjoseph: The suggested change doesn't work. I will suppress the lint error.

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -105,11 +105,56 @@ export interface IRushRepositoryJson {
...
+export interface IConfigurationEnvironmentVariable {
+  /**
+   * Value of the environment variable
+   */
+  value: string;
+
+  /**
+   * Set to true to override the environment variable even if it is set on the device
+   */
+  override?: boolean; //
```
```suggestion
  override?: boolean;
```

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -182,6 +236,39 @@ export interface ICurrentVariantJson {
   variant: string | null; // Use `null` instead of `undefined` because `undefined` is not handled by JSON.
 }
 
+/**
+ * Options that all package managers share.
+ *
+ * @public
+ */
+export class PackageManagerOptionsConfigurationBase implements IPackageManagerOptionsJsonBase {
```
```suggestion
export abstract class PackageManagerOptionsConfigurationBase implements IPackageManagerOptionsJsonBase {
```

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -182,6 +236,39 @@ export interface ICurrentVariantJson {
...
+  /**
+   * Enviroment variables for the package manager
+   */
+  public readonly environmentVariables?: IConfigurationEnvironment
+
+  /** @internal */
+  protected constructor(environmentVariables?: IConfigurationEnvironment) {
+    this.environmentVariables = environmentVariables;
+  }
```
iclanton: Let's change this to take `IPackageManagerOptionsJsonBase`, as that's the analogous structure.

```suggestion
  protected constructor(json: IPackageManagerOptionsJsonBase) {
    this.environmentVariables = json.environmentVariables;
  }
```

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -126,8 +171,16 @@ export interface IPnpmOptionsJson {
 
 /**
  * Part of IRushConfigurationJson.
+ * @internal
  */
-export interface IYarnOptionsJson {
+export interface IYarnOptionsJson extends IPackageManagerOptionsJsonBase {
+    /**
```
```suggestion
  /**
```

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
```diff
@@ -59,6 +61,12 @@ export abstract class BaseInstallAction extends BaseRushAction {
       description: 'Activates verbose logging for the package manager. You will probably want to pipe'
         + ' the output of Rush to a file when using this command.'
     });
+    this._maxInstallAttempts = this.defineIntegerParameter({
```
iclanton: There is a `defaultValue` option here. Let's use that.

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
```diff
@@ -96,6 +104,10 @@ export abstract class BaseInstallAction extends BaseRushAction {
       }
     }
 
+    if (this._maxInstallAttempts.value && this._maxInstallAttempts.value < 1) {
+      throw new Error(`The value of "${this._maxInstallAttempts.longName}" must be positive.`);
```
iclanton:
```suggestion
      throw new Error(`The value of "${this._maxInstallAttempts.longName}" must be positive and nonzero.`);
```

apps/rush-lib/src/cli/actions/InstallAction.ts
```diff
@@ -34,7 +35,10 @@ export class InstallAction extends BaseInstallAction {
       recheckShrinkwrap: false,
       networkConcurrency: this._networkConcurrencyParameter.value,
       collectLogFile: this._debugPackageManagerParameter.value!,
-      variant: this._variant.value
+      variant: this._variant.value,
+      maxInstallAttempts: this._maxInstallAttempts.value !== undefined
```
iclanton: If we change `_maxInstallAttempts` to use `defaultValue`, I think we can remove these checks.

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1029,6 +1064,49 @@ export class InstallManager {
...
+  ): NodeJS.ProcessEnv {
+    const packageManagerEnv: NodeJS.ProcessEnv = baseEnv;
+
+    if (environmentVariables) {
+      // eslint-disable-next-line guard-for-in
+      for (const envVar in environmentVariables) {
+        let setEnvironmentVariable: boolean = true;
+        console.log(`\nProcessing definition for environment variable: ${envVar}`);
+
+        if (process.env[envVar]) {
```
iclanton:
```suggestion
        if (baseEnv.hasOwnProperty(envVar)) {
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1029,6 +1064,49 @@ export class InstallManager {
...
+          setEnvironmentVariable = false;
+          console.log(colors.yellow(`WARNING: Environment variable already defined:`));
+          console.log(`  Name: ${envVar}`);
+          console.log(`  Value set on the device: ${process.env[envVar]}`);
```
iclanton:
```suggestion
          console.log(`  Value set on the device: ${baseEnv[envVar]}`);
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1029,6 +1064,49 @@ export class InstallManager {
...
+          console.log(`  Value set in rush.json: ${environmentVariables[envVar].value}`);
+
+          if (environmentVariables[envVar].override &&
+            environmentVariables[envVar].override === true) {
```
iclanton:
```suggestion
          if (environmentVariables[envVar].override) {
```

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
sachinjoseph: positive implies non-zero, right?

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1029,6 +1064,49 @@ export class InstallManager {
...
+            setEnvironmentVariable = true;
+            console.log(colors.yellow(`WARNING: Overriding the environment variable with the value set in rush.json.`));
```
iclanton: Now that I'm thinking about this again, should we bother warning at all when we override the environment variable? Maybe we should just log that we're overriding an existing variable. The purpose of the `override` option is to let the repo owner control which ones should be overridden, so they should already be aware of what will happen.

apps/rush-lib/src/logic/InstallManager.ts
sachinjoseph: @iclanton What is a better way of phrasing "Value set on the device"? I mean, even if the variable is not defined on the device, rush can still be launched with with this variable defined.

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
iclanton: Zero is generally considered either both positive and negative or neither positive nor negative: https://math.stackexchange.com/questions/26705/is-zero-positive-or-negative

apps/rush-lib/src/logic/InstallManager.ts
sachinjoseph: Good catch!

apps/rush-lib/src/logic/InstallManager.ts
sachinjoseph: Makes sense.

apps/rush-lib/src/logic/InstallManager.ts
sachinjoseph: It is an optional parameter. Is it safe to drop the null check and strict comparison?

apps/rush-lib/src/logic/InstallManager.ts
iclanton: "Existing value"?

"device" doesn't really make sense in this context.

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1086,7 +1086,7 @@ export class InstallManager {
           if (environmentVariables[envVar].override &&
             environmentVariables[envVar].override === true) {
             setEnvironmentVariable = true;
-            console.log(colors.yellow(`WARNING: Overriding the environment variable with the value set in rush.json.`));
+            console.log(`Overriding the environment variable with the value set in rush.json.`);
```
iclanton: Maybe this should only be printed in `--debug` mode.

sachinjoseph: I think this should always be printed. I don't want rush to silently override a variable set on the command line only for the session or in a ci job step.

iclanton: Hmm okay. Let's see how chatty it ends up being.

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -105,11 +105,56 @@ export interface IRushRepositoryJson {
...
+ * @public
+ */
+export interface IConfigurationEnvironmentVariable {
+  /**
+   * Value of the environment variable
+   */
+  value: string;
+
+  /**
+   * Set to true to override the environment variable even if it is set on the device
```
```suggestion
   * Set to true to override the environment variable even if it is set in the parent environment
```

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -105,11 +105,56 @@ export interface IRushRepositoryJson {
...
+ */
+export interface IConfigurationEnvironment {
+  /**
+   * Environment variables
+   */
+  [environmentVariableName: string]: IConfigurationEnvironmentVariable;
+}
+
+/**
+ * Represents an environment variable
```
```suggestion
 * Represents the value of an environment variable, and if the value should be overridden if the variable is set
 * in the parent environment.
```

apps/rush-lib/src/logic/InstallManager.ts
```diff
@@ -1029,6 +1064,49 @@ export class InstallManager {
...
+
+    if (environmentVariables) {
+      // eslint-disable-next-line guard-for-in
+      for (const envVar in environmentVariables) {
+        let setEnvironmentVariable: boolean = true;
+        console.log(`\nProcessing definition for environment variable: ${envVar}`);
+
+        if (baseEnv.hasOwnProperty(envVar)) {
+          setEnvironmentVariable = false;
+          console.log(colors.yellow(`WARNING: Environment variable already defined:`));
```
iclanton: This shouldn't be a warning if they explicitly said they want to override it.

```suggestion
          console.log(`WARNING: Environment variable already defined:`);
```

apps/rush-lib/src/logic/InstallManager.ts
iclanton: Yeah - it defaults to false, so if it's undefined, this will evaluate to false.

apps/rush-lib/src/schemas/rush.schema.json
```diff
@@ -3,31 +3,44 @@
...
+        "additionalProperties": false
+      }
+    }
+  },
   "properties": {
     "$schema": {
       "description": "Part of the JSON Schema standard, this optional keyword declares the URL of the schema that the file conforms to. Editors may download the schema and use it to perform syntax highlighting.",
       "type": "string"
     },
-
```
iclanton: Can you put these blank lines back?

sachinjoseph: VS Code automatically removed them when I saved the file. Should we maintain these blank lines?

PR #1775, opened 2020-03-12
Genericize the TypingsGenerator utility from @rushstack/localization-plugin.

iclanton:


libraries/typings-generator/src/TypingsGenerator.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    if (!this._options.terminal) {
+      this._options.terminal = new Terminal(new ConsoleTerminalProvider({ verboseEnabled: true }));
+    }
+
+    this._options.fileExtensions = this._normalizeFileExtensions(this._options.fileExtensions);
+  }
+
+  public generateTypings(): void {
+    FileSystem.ensureEmptyFolder(this._options.generatedTsFolder);
```
apostolisms: this can be problematic if someone sets the generatedTsFolder to be `src` or  another directory that shouldn't be cleaned, either accidentally or on purpose.

iclanton: It's mentioned in the documentation. I think this is reasonable given that it's documented.

PR #1777, opened 2020-03-13
Add 'AdditionalCommands' to run subcommands during bulk Rush actions

D4N14L:


D4N14L:
Needs tests and `rush change`

apps/rush-lib/src/logic/taskRunner/ProjectTask.ts
```diff
@@ -58,24 +63,26 @@ export class ProjectTask implements ITaskDefinition {
...
     this._rushConfiguration = options.rushConfiguration;
-    this._commandToRun = options.commandToRun;
     this.isIncrementalBuildAllowed = options.isIncrementalBuildAllowed;
     this._packageChangeAnalyzer = options.packageChangeAnalyzer;
     this._packageDepsFilename = options.packageDepsFilename;
-}
+
+    this._commandsToRun = options.commandsToRun;
+    if (this._commandsToRun.length < 1) {
+      throw Error("No commands were defined for the task.");
```
sachinjoseph: _commandsToRun could also be empty if --skip parameter has all parameters listed as values, right?

sachinjoseph:
Should we clearly differentiate between commands and subcommands?

D4N14L:
After a longer set of discussions, we've decided to expand the application of this feature making this design out-of-date

PR #1800, opened 2020-03-27
[localization-plugin] Fix a few RESX parsing issues and improve translated strings resolution.

iclanton:


webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -439,15 +499,15 @@ export class LocalizationPlugin implements Webpack.Plugin {
...
     const filesMap: Map<string, Map<string, string>> = this._resolvedLocalizedStrings.get(localeName)!;
 
     const stringsMap: Map<string, string> = new Map<string, string>();
-    filesMap.set(locFilePath, stringsMap);
+    filesMap.set(localizedFilePath, stringsMap);
 
-    for (const stringName in locFileData) {
-      if (locFileData.hasOwnProperty(stringName)) {
-        const stringKey: string = `${locFilePath}?${stringName}`;
+    for (const stringName in localizedFileData) {
```
dmichon-msft: We can write this a bit more ergonomically using `Object.entries` which makes conversion to/from `Map` easy.
```ts
for (const [stringName, stringValue] of Object.entries(locFileData))
```

webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -655,4 +724,41 @@ export class LocalizationPlugin implements Webpack.Plugin {
...
+            chunkHasAnyLocModules = true;
+            break;
+          }
+        }
+      }
+
+      EntityMarker.markEntity(chunk, chunkHasAnyLocModules);
+    }
+
+    return EntityMarker.getMark(chunk)!;
```
dmichon-msft: You can avoid the need for the assertion by switching up the pattern a bit:
```ts
let chunkHasAnyLocModules: boolean | undefined = EntityMarker.getMark(chunk);
if (chunkHasAnyLocModules === undefined) {
    // Do what you currently do in this block
}

return chunkHasAnyLocModules;

File: webpack/localization-plugin/src/LocalizationPlugin.ts [hunk 3]
```diff
@@ -655,4 +724,41 @@ export class LocalizationPlugin implements Webpack.Plugin {
...
+  }
+
+  private _convertLocalizationFileToLocData(locFile: ILocalizationFile): ILocaleFileData {
+    const locFileData: ILocaleFileData = {};
+    for (const stringName in locFile) { // eslint-disable-line guard-for-in
```
dmichon-msft: `for (const [stringName, { value }] of Object.entries(locFile))` avoids the eslint issue

webpack/localization-plugin/README.md
```diff
@@ -122,6 +122,38 @@ translatedStrings: {
 }
 ```
 
+Alternatively, instead of directly specifying the translations, a path to a translated resource file can be
+specified. For example:
+
+```JavaScript
+translatedStrings: {
+  "en-us": {
+    "./src/strings1.loc.json": "./localization/en-us/strings1.loc.json"
```
dmichon-msft: Document the rules on where `.` resolves to?

webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -655,4 +724,41 @@ export class LocalizationPlugin implements Webpack.Plugin {
...
+      if (!chunkHasAnyLocModules) {
+        for (const module of chunk.getModules()) {
+          if (EntityMarker.getMark(module)) {
+            chunkHasAnyLocModules = true;
+            break;
+          }
+        }
+      }
+
+      // Check async chunks if this is a runtime chunk and we haven't directly found any localized modules
```
dmichon-msft: Surely the marker for this state should be different from the marker for the chunk itself needing localization? This state just necessitates a tweak to the runtime definition to reroute the loading.

webpack/localization-plugin/src/utilities/ResxReader.ts
```diff
@@ -1,31 +1,82 @@
...
   XmlElement
 } from 'xmldoc';
+
 import {
   ILocalizedString,
   ILocalizationFile
 } from '../interfaces';
-import { ILoggingFunctions } from './Logging';
 
 const STRING_NAME_RESX: RegExp = /^[A-z_][A-z0-9_]*$/;
```
dmichon-msft: Is there a reason we ban `$` in the names? It's a legal identifier character too.

webpack/localization-plugin/src/utilities/ResxReader.ts
```diff
@@ -245,26 +296,45 @@ export class ResxReader {
...
+  private static _logWithLocation(
+    loggingFn: (message: string) => void,
+    message: string,
+    filePath: string,
+    line?: number,
+    position?: number
+  ): void {
+    let location: string;
+    if (position !== undefined) {
+      location = `${filePath}(${line},${position})`;
```
dmichon-msft: Does this schema work with Ctrl+Click in VSCode? If not, would be a nice-to-have to switch to one that does.

build-tests/localization-plugin-test-03/webpack.config.js
```diff
@@ -2,12 +2,34 @@
...
+  let contextRelativePath = path.relative(__dirname, localizedResourcePath);
+  contextRelativePath = contextRelativePath.replace(/\\/g, '/'); // Convert Windows paths to Unix paths
+  if (!contextRelativePath.startsWith('.')) {
+    contextRelativePath = `./${contextRelativePath}`;
+  }
+
+  const result = {};
+  for (const localeName of localeNames) {
+    const expectedCombinedStringsPath = path.resolve(__dirname, 'localization', localeName, 'combinedStringsData.json');
+    if (FileSystem.exists(expectedCombinedStringsPath)) {
```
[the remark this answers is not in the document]
dmichon-msft: Admittedly this is test code, but the best way to break bad habits is to not do them anywhere

build-tests/localization-plugin-test-03/webpack.config.js
```diff
@@ -2,12 +2,34 @@
...
+  JsonFile,
+  FileSystem
+} = require('@rushstack/node-core-library');
 
 const { LocalizationPlugin } = require('@rushstack/localization-plugin');
 const { SetPublicPathPlugin } = require('@rushstack/set-webpack-public-path-plugin');
 const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 
+function resolveMissingString(localeNames, localizedResourcePath) {
```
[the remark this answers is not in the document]
iclanton: I'm going to rewrite this soon, and that'll come then.

webpack/localization-plugin/src/utilities/ResxReader.ts
iclanton: No good reason. I'll add that.

webpack/localization-plugin/README.md
iclanton: It's in the README. It's relative to the webpack compilation context.

webpack/localization-plugin/src/utilities/ResxReader.ts
iclanton: Yeah works fine in VSCode.

iclanton:
>I'm concerned about perf and the current cascading of localized expansion across chunks that aren't localized.

Chunks that contain no localized modules aren't localized. They're given a locale-agnostic name.

webpack/localization-plugin/src/LocalizationPlugin.ts
iclanton: You're referring to the case where an entrypoint doesn't directly contain any localized resources, but it references async chunks that do?

In that case, we do modify the runtime definition to reroute the loading, but the locale name for the async chunks needs to land somewhere in the entrypoint chunk so, therefore, the entrypoint ends up also becoming localized.

dmichon-msft: My concern is about the chunks between the entry point chunk and the localized chunks.

build-tests/localization-plugin-test-03/webpack.config.js
```diff
@@ -22,9 +19,11 @@ function resolveMissingString(localeNames, localizedResourcePath) {
   const result = {};
   for (const localeName of localeNames) {
     const expectedCombinedStringsPath = path.resolve(__dirname, 'localization', localeName, 'combinedStringsData.json');
-    if (FileSystem.exists(expectedCombinedStringsPath)) {
+    try {
       const loadedCombinedStringsPath = JsonFile.load(expectedCombinedStringsPath);
       result[localeName] = loadedCombinedStringsPath[contextRelativePath];
+    } catch (e) {
+      // If reading the file throws, it probably doesn't exist
```
dmichon-msft: It's safest to do:
```ts
if (e.code !== 'ENOENT' && e.code !== 'ENOTDIR') {
    // File exists, but reading failed.
    throw e;
}
```

webpack/localization-plugin/src/LocalizationPlugin.ts
iclanton: Only the entrypoint chunk(s) contain(s) the async chunk loading code, so only the entrypoint will end up localized.

build-tests/localization-plugin-test-03/webpack.config.js
iclanton: Kinda nasty to put that everywhere we read a file.

PR #1810, opened 2020-04-05
[localization-plugin] Fix a few issues with character escaping.

iclanton:


build-tests/localization-plugin-test-01/src/indexA.ts
```diff
@@ -3,3 +3,4 @@ import(/* webpackChunkName: 'chunk-without-strings' */ './chunks/chunkWithoutStr
   const chunk = new ChunkWithoutStringsClass();
   chunk.doStuff();
 });
+//foobar
```
iclanton, line 6: Remove

PR #1812, opened 2020-04-05
[localization-plugin] Add support for normalization of newlines in RESX files.

iclanton:


webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -690,6 +694,30 @@ export class LocalizationPlugin implements Webpack.Plugin {
         }
       }
       // END options.localizedData.pseudoLocales
+
+      // START options.
```
iclanton, line 698: Fix this comment

PR #1817, opened 2020-04-06
Update Rush files using "rush init" and turn on "preventManualShrinkwrapChanges."

iclanton:


iclanton:
Successful build: https://dev.azure.com/RushStack/GitHubProjects/_build/results?buildId=3565

PR #1826, opened 2020-04-11
[rush-lib] Use `command` instead of `which`

jkroepke:
Hi,

in some certain minimalistic build environments, `@microsoft/rush` can't installed because the command line tool `which` is not available. 

```
 Installing @microsoft/rush...
 Error: Unable to install package: Error: Unable to determine the path to the NPM tool: Error: Command failed: which npm
 /bin/sh: which: command not found
```

While we could fix this issue by our own by just installing `which` I would like to improve this cool project a little bit.

In shell scripting it's preferred to use `command -v` instead `which` because the first one is a POSIX shell built-in. It looks like `childProcess.execSync` already opens a shell internally, we could use just the built-in helpers instead an external small utility. 

Related resources:
* https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script

### DIFF
--- a/common/changes/@microsoft/rush/patch-1_2020-04-11-10-11.json
+++ b/common/changes/@microsoft/rush/patch-1_2020-04-11-10-11.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush",
+  "email": "mail@jkroepke.de"
+}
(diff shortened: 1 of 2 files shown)


common/changes/@microsoft/rush/patch-1_2020-04-11-10-11.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue where, on some minimal systems, Rush used a missing shell command to detect an application path.",
```

jkroepke: Thanks! I didn't find a CONTRIBUTION.md. I didn't know what should todo.

PR #1850, opened 2020-04-29
[rush] Pass --frozen-lockfile to rush install

sachinjoseph:
Also: Fixes https://github.com/microsoft/rushstack/issues/1300 (in a way😃). @octogonz FYI.

Disabled by default - to enable set the flag `usePnpmFrozenLockfileForRushInstall` in `experiments.json`.


sachinjoseph:
@octogonz Do you want to take a look?

PR #1883, opened 2020-05-22
[node-core-library] Introduce a FileSystemNotExistError error to better identify filesystem errors that occur when a file or folder doesn't exist

iclanton:


iclanton:
Taking this approach instead: https://github.com/microsoft/rushstack/pull/1884

Issue #1689, opened 2020-01-15
[GCB] Allow configuration of autoprefixer browser list (or drop IE10 support EOM)


KevinTCoughlin:
Currently, `gulp-core-build-sass` passes a hard-coded browser range including IE10 to `autoprefixer`:

https://github.com/microsoft/rushstack/blob/4cf4bdd0738e4cadbb198f6c0112a8eedc4aea54/core-build/gulp-core-build-sass/src/SassTask.ts#L90

Officially, IE10 reaches end of support on January 31, 2020:

>Internet Explorer (IE) 10 will reach end of support on January 31, 2020. From this point forward, IE 11 will be the only supported version of Internet Explorer on Windows Server 2012 and Windows Embedded 8 Standard. IE 11 is available via Microsoft Update Catalog, Windows Update and Windows Server Update Service (WSUS). Go here to learn more.
>https://support.microsoft.com/en-us/help/17454/lifecycle-faq-internet-explorer

IE10 requires a number of vendor specific prefixes, particularly with flex & grid. Dropping IE10 from the supported browser range will reduce emitted CSS size. This change would impact both first and third-party SharePoint developers.

We may consider making the browser list configurable if devs require that level of configuration.

I'm happy to raise the PR. The above change may need to be a `major` bump to `gulp-core-build-sass`.

cc: @iclanton @lijunle @patmill

KevinTCoughlin:
IIRC `load-themed-styles` is no longer compatible with IE < 11. In case that influences the above.

lijunle:
> IIRC load-themed-styles is no longer compatible with IE < 11. In case that influences the above.

When did `load-themed-styles` drop off the IE10 support? I don't see it is mentioned in its [`CHANGELOG`](https://github.com/microsoft/rushstack/blob/master/libraries/load-themed-styles/CHANGELOG.md)? 

> We may consider making the browser list configurable if devs require that level of configuration. I'm happy to raise the PR. The above change may need to be a major bump to gulp-core-build-sass.

Sounds reasonable for the layer of `gulp-core-build-sass`. Each end product has the choice to decide if support IE10. Also, reasonable to be a major bump and mention int the CHANGELOG.

One more thing from SharePoint aspect, we should have something like `browsers: '> 1% of SharePoint'`. It pushes the end user experience to the dev end.

KevinTCoughlin:
[...]

PR #1891, opened 2020-05-27
Allow configuration of autoprefixer plugin

KevinTCoughlin:
### Overview

Addresses #1689. IE 10 is EOL as of Jan 2020 ([announcement](https://support.microsoft.com/en-us/help/17454/lifecycle-faq-internet-explorer)). 

Currently, solutions built using `gulp-core-build-sass` have IE 10 compat hard-coded as the browserlist override. This results in a large number of vendor prefixes emitted in the resultant CSS.

### Example of vendor prefix delta when using flex-box:

Using https://autoprefixer.github.io/

**IE10 compat**

```css
div {
  display: -ms-flexbox;
  display: flex;
  -ms-flex-direction: column;
      flex-direction: column;
}
```

IE10 prefixes for flex-box are numerous due to historical reasons.

**IE11 compat**

```css
div {
  display: flex;
  flex-direction: column;
}
```
With this change, consumers can provide their own browserlist or rely on `autoprefixer`'s defaults depending on their application specific scenarios. This should improve application performance both by reducing property computations while constructing the CSSOM and reducing the amount of CSS sent over-the-wire.

### Dependency changelogs

- [postcss changelog](https://github.com/postcss/postcss/blob/master/CHANGELOG.md)
- [postcss-modules changelog](https://github.com/css-modules/postcss-modules/blob/master/CHANGELOG.md)
- [autoprefixer changelog](https://github.com/postcss/autoprefixer/releases)


core-build/gulp-core-build-sass/src/SassTask.ts
```diff
@@ -78,6 +78,11 @@ export interface ISassTaskConfig {
    * returnPromise and sourceMap will be ignored.
    */
   cleanCssOptions?: CleanCss.Options;
+
+  /**
+   * Allows the override of the options passed to autoprefixer.
+   */
+  autoprefixerOptions?: autoprefixer.Options;
```
iclanton, line 85: Can you pass what was already there as the defaults in the constructor?

KevinTCoughlin: Done in https://github.com/microsoft/rushstack/pull/1891/commits/07623c99ff9297c9d449d9cee3e886347f655aeb.

PR #1939, opened 2020-06-18
[Heft] Basic implementation of Heft build framework.

iclanton:


apps/heft/LICENSE
```diff
@@ -0,0 +1,24 @@
+@microsoft/heft
```
iclanton, line 1:
```suggestion
@rushstack/heft
```

octogonz:
Wahoo

apps/heft/LICENSE
```diff
@@ -0,0 +1,24 @@
+@rushstack/heft
```
octogonz, line 1: We shouldn't publish Heft without a README.md file. At the very least, I'd suggest to start with a placeholder file with a notice saying something simple like "This project is still under development and is being published to facilitate testing. For more details and support resources, please visit: https://rushjs.io"

common/reviews/api/heft.api.md
```diff
@@ -0,0 +1,170 @@
...
+export abstract class ActionHooksBase {
+    // (undocumented)
+    readonly afterLoadActionConfiguration: AsyncSeriesHook;
+    // (undocumented)
+    readonly loadActionConfiguration: AsyncSeriesHook;
+    readonly override: AsyncSeriesBailHook<IActionDataBase<ActionHooksBase>>;
+}
+
+// @public (undocumented)
+export type Build = IBuildActionData;
```
octogonz, line 27: Why are these types aliased?

apps/heft/src/cli/actions/DevDeployAction.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @public
+ */
+export class DevDeployHooks extends ActionHooksBase {}
+
+export class DevDeployAction extends HeftActionBase<IDevDeployActionData, DevDeployHooks> {
+  public constructor(options: IHeftActionBaseOptions) {
+    super(
+      {
+        actionName: 'dev-deploy',
+        summary: 'Deploy the current project, and optionally the whole repo, to a testing CDN.',
```
octogonz, line 21: How will the out-of-box implementation of this work? Is this operation generic enough across repos for it to be one of the default built-in actions?

apps/heft/src/cli/actions/DevDeployAction.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @public
+ */
+export class DevDeployHooks extends ActionHooksBase {}
+
+export class DevDeployAction extends HeftActionBase<IDevDeployActionData, DevDeployHooks> {
+  public constructor(options: IHeftActionBaseOptions) {
+    super(
+      {
+        actionName: 'dev-deploy',
```
octogonz, line 20: This is a local jargon that I haven't seen widely used elsewhere. The name is maybe not very intuitive. (What does "dev" mean? How does it different from a normal deployment? etc.)

apps/heft/build.js
```diff
@@ -0,0 +1,47 @@
...
+const glob = require('glob');
+const { FileSystem } = require('@rushstack/node-core-library');
+const { ToolPaths } = require('@microsoft/rush-stack-compiler-3.7');
+const { callNodeScript } = require('./callNodeScript');
+const child_process = require('child_process');
+
+const isProductionBuild = process.argv.indexOf('--production') !== -1;
+
+process.exitCode = 1;
+try {
```
octogonz, line 13: @iclanton This is so exciting!

I'll probably have some more feedback about the architecture, but I hesitate to voice opinions until there's enough functionality to see it building a somewhat realistic project.  Like, maybe we can aim to replace this `build.js` file for Heft itself.  

Another option is the TSDoc repo, which has some real-world [build.js files](https://github.com/microsoft/tsdoc/blob/master/tsdoc/build.js) that are pretty primitive and probably easy candidates.

apps/heft/LICENSE
iclanton: It's currently not set to publish, but I can add a simple README for now.

apps/heft/src/cli/actions/DevDeployAction.ts
iclanton: Right now it's been designed to work with Azure storage, with hooks for asset discovery and processing and account and storage key. It could easily be further genericized.

apps/heft/src/cli/actions/DevDeployAction.ts
iclanton: Let's leave this as-is for now and potentially rename to "deploy" in the future.

common/reviews/api/heft.api.md
iclanton: These are aliased for API simplicity in hooks. The idea is that you can say something like `compilation.hooks.build.tap('myplugin', (build: Build /* instead of IBuildActionData */) => { ... });`

iclanton: I'm open to refactoring this later.

apps/heft/build.js
iclanton: Agreed on all points. The architecture definitely needs some revision, but it's a good start.

PR #1962, opened 2020-06-25
[heft] Refactor the way action hooks are called to separate the action properties into their own object.

iclanton:


apps/heft/src/cli/actions/HeftActionBase.ts
```diff
@@ -17,20 +17,24 @@ import { PluginManager } from '../../pluginFramework/PluginManager';
...
  */
-export abstract class ActionHooksBase {
+export abstract class ActionHooksBase<TActionProperties extends object> {
   /**
    * This hook allows the action's execution to be completely overridden. Only the last-registered plugin
    * with an override hook provided applies.
    */
-  public readonly override: AsyncSeriesBailHook<IActionDataBase<ActionHooksBase>> = new AsyncSeriesBailHook([
-    'actionData'
+  public readonly override: AsyncSeriesBailHook<TActionProperties> = new AsyncSeriesBailHook([
```
octogonz, line 36: `override` is a keyword -- not a good identifier for an API

apps/heft/src/cli/actions/StartAction.ts
```diff
@@ -1,19 +1,24 @@
...
  * @public
  */
-export interface IStartActionData extends IActionDataBase<StartHooks> {}
+export interface IStartActionProperties {}
 
-export class StartAction extends HeftActionBase<IStartActionData, StartHooks> {
+/**
+ * @public
+ */
+export interface IStartActionContext extends IActionContext<StartHooks, IStartActionProperties> {}
```
octogonz, line 19: The generic `<StartHooks, IStartActionProperties>` has a couple disadvantages:

- This pattern cannot be extended further into a child interface that extends `IStartActionContext`
- The API documentation is abstracted.
- The generic parameters must be specified even when there are no changes to the hooks/properties
- If more inheritable items are added to the context, they will accumulate as generic parameters.

Compare with this:

```ts
export interface IStartActionContext extends IActionContext {
  hooks: StartHooks;  // (redefines base member)
  properties: IStartActionProperties;  // (redefines base member)
}
```

Might be more verbose to write, but for a person reading the API docs, the type of `hooks` and `properties` is much easier to understand.

apps/heft/src/cli/actions/HeftActionBase.ts
```diff
@@ -17,20 +17,24 @@ import { PluginManager } from '../../pluginFramework/PluginManager';
...
+  properties: TActionProperties;
 }
 
 /**
  * @public
  */
-export abstract class ActionHooksBase {
+export abstract class ActionHooksBase<TActionProperties extends object> {
   /**
    * This hook allows the action's execution to be completely overridden. Only the last-registered plugin
```
octogonz, line 33: The semantics here seem potentially muddy. Perhaps it should be an error if more than one plugin tries to do this?  Perhaps this is not an event handler at all, but some other kind of contract?

apps/heft/src/cli/actions/HeftActionBase.ts
```diff
@@ -17,20 +17,24 @@ import { PluginManager } from '../../pluginFramework/PluginManager';
...
    * with an override hook provided applies.
    */
-  public readonly override: AsyncSeriesBailHook<IActionDataBase<ActionHooksBase>> = new AsyncSeriesBailHook([
-    'actionData'
+  public readonly override: AsyncSeriesBailHook<TActionProperties> = new AsyncSeriesBailHook([
+    'actionContext'
   ]);
```
octogonz, line 38: The `AsyncSeriesBailHook<TActionProperties>` construct seems to produce an event handler callback signature like this:

```ts
call(arg1?: TActionProperties | undefined, arg2?: any, arg3?: any, ...args: any[]): any;
```

That might be the abstraction that tappable's pipeline invokes, but is there some way we could replace it in our own API with something stricter and intelligently named?

Something like this:
```ts
type OverrideHookCall<TActionProperties extends object> = (actionProperties: TActionProperties) => void;
```

Also, if this is an extensible event handler, the industry standard is more like this:
```ts
type EventHandler<TEventArgs> = (thingTheEventCameFrom: SomeBaseClass, eventArgs: TEventArgs) => void;

call: EventHandler<IOverrideHookEventArgs>;
```

And then `IOverrideHookEventArgs` is a documented and extensible.

octogonz: There's probably different ways to approach it. But the main feedback is that this isn't a great IntelliSense experience:

Capture

iclanton: Let's mark this as `@beta` for now and tackle this later as the API matures.

apps/heft/src/cli/actions/HeftActionBase.ts
iclanton: Let's mark this as `@beta` for now and tackle this later as the API matures.

apps/heft/src/cli/actions/StartAction.ts
iclanton: >This pattern cannot be extended further into a child interface that extends IStartActionContext

Anything that extends `IStartActionContext` can add properties

>The API documentation is abstracted.

I view this as a positive - the API documentation exists in one place.

>The generic parameters must be specified even when there are no changes to the hooks/properties

There's no default implementation for either, so I also view this as a positive.

>If more inheritable items are added to the context, they will accumulate as generic parameters.

That's a legitimate concern, so I'd say we should consider this change if/when we get to that point.

That alternate implementation is also far more likely to cause someone to make a mistake. Additionally, having the generics on the base types (and no generics on the child types) helps indicate that the base types used on the base interface properties aren't the ones any consumer of the API should be using.

PR #1969, opened 2020-06-26
Make rush-stack-compiler-3.1 a runtime dependency for gulp-core-build-typescript

D4N14L:
This is a bug when consuming gulp-core-build-typescript. Without this change, an error will be seen at buildtime stating that the dependency could not be found:
[typescript] Encountered 1 TypeScript error:
[typescript]   ERROR: ..\..\common\temp\node_modules\.pnpm\<packagerepo>\@microsoft\gulp-core-build-typescript\8.4.24\node_modules\@microsoft\gulp-core-build-typescript\lib\RSCTask.d.ts(3,37) - (TS2307) Cannot find module '@microsoft/rush-stack-compiler-3.1'.


D4N14L:
I attempted to use API-Extractor to generate the typings, since this import is only used for typings, however I ran into this bug: https://github.com/microsoft/rushstack/issues/1968

PR #1999, opened 2020-07-06
Make the next release of Rush a prerelease.

iclanton:


iclanton:
Abandoned in favor of using a different release branch.

PR #2006, opened 2020-07-07
[heft] Add support for ESLint.

iclanton:


apps/heft/src/plugins/TypeScriptPlugin/Eslint.ts
```diff
@@ -0,0 +1,111 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import * as path from 'path';
+import * as crypto from 'crypto';
+import * as TEslint from 'eslint';
```
octogonz, line 6: Can you somehow measure how much time is spent parsing TypeScript versus running rules? I believe ESLint collects metrics and exposes them  via some API. (?)

octogonz: This might help: https://github.com/eslint/eslint/blob/ec105b24f7e036ecdc4267f018529ac3765e29d5/lib/linter/timing.js#L42

apps/heft/src/plugins/TypeScriptPlugin/Eslint.ts
```diff
@@ -0,0 +1,111 @@
...
+    const eslintFailureLogMessages: (string | IColorableSequence)[][] = [];
+
+    for (const eslintFileResult of this._lintResult) {
+      const buildFolderRelativeFilePath: string = path.relative(
+        this._buildFolderPath,
+        eslintFileResult.filePath
+      );
+      for (const message of eslintFileResult.messages) {
+        // https://eslint.org/docs/developer-guide/nodejs-api#◆-lintmessage-type
+        const severity: string = message.severity === 1 ? 'WARNING' : 'ERROR';
```
octogonz, line 44: Capture

There are some formatting issues with the above output:
- The `@rushstack/eslint-config` package classifies every lint issue as an "error". We need to fix this, since nearly all lint issues are not serious and should be warnings. It's an easy fix that you could possibly do in this PR.
- Here Heft is remapping every lint issue to be a warning regardless of its severity, presumably to compensate for that.  If we keep this approach, then we cannot print `ERROR`.
- API Extractor's warnings are printed in red

The **important UI design principles** are:
- Log messages that call themselves "error" must be treated as errors (print to STDERR and return nonzero a exit code so Rush fails)
- Log messages that call themselves "warning" must be treated as warnings (print to STDERR and not change the exit code)
- The red color should never be used for anything that is not an actual error
- The yellow color should never be used for anything that is not an actual warning

octogonz: If we want to summarize the error/warning counts, they really should come at the very end, something like:
```
 ---- Bundle finished (800ms) ----
-------------------- Finished (2.55s) --------------------
Project: heft-test-01@1.0.0
ESLint reported 3 warnings <--- yellow
TSLint reported 3 errors and 4 warnings <--- red
API Extractor reported 2 errors <--- red
```
If a task (e.g. TypeScript) completed successfully, I wouldn't include it in the summary at all, because that is just taking up space without telling us anything we don't already know.  However, on 100% success we could show something like this:
```
-------------------- Finished (2.55s) --------------------
Project: heft-test-01@1.0.0
All tasks completed without warnings. <--- green
```

It is weird for "Finished" to be colored green when there are warnings. Maybe don't color it at all.

octogonz: On a separate topic, I'm still strongly in favor of [stream-collator](https://www.npmjs.com/package/@rushstack/stream-collator) which will eliminate all these `[api-extractor] ` prefixes cluttering up the output. Our UI design would instead show coherent chunks of reporting for individual tasks, sort of like you're doing with `Encountered 2 ESLint errors:`

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -0,0 +1,152 @@
...
+      for (const tslintFailure of this._lintResult.failures) {
+        const buildFolderRelativeFilename: string = path.relative(
+          this._buildFolderPath,
+          tslintFailure.getFileName()
+        );
+        const { line, character } = tslintFailure.getStartPosition().getLineAndCharacter();
+        const severity: string = tslintFailure.getRuleSeverity().toUpperCase();
+        this._terminal.writeWarningLine(
+          '  ',
+          Colors.yellow(`${severity}: ${buildFolderRelativeFilename}:${line + 1}:${character + 1}`),
```
octogonz, line 85: BTW I noticed that the API Extractor output is still using `filename.ts(123,234)` style formatting. This is easy to fix because `ExtractorMessage` exposes the individual parts:    `sourceFilePath`, `sourceFileLine`, `sourceFileColumn`, `formatMessageWithoutLocation()`

octogonz: Or, if Heft is blazing the trail for a new standard, we could change API Extractor's default formatting.

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -0,0 +1,152 @@
...
+          ' - ',
+          Colors.yellow(`(${tslintFailure.getRuleName()}) ${tslintFailure.getFailure()}`)
```
octogonz, line 87: FYI ESLint's `--format unix` output uses a `:` instead of ` - `:
```
src\index.ts:36:18: Interface name `MissingI` must match the RegExp: /^_?I[A-Z]/ [Error/@typescript-eslint/naming-convention]
```

I think I might prefer your hyphen, if the various editors recognize it.  I definitely do not like ESLint's way of using `Error/` to indicate an error.  Our `ERROR: ` prefix looks better.

apps/heft/src/plugins/TypeScriptPlugin/Eslint.ts
```diff
@@ -0,0 +1,111 @@
...
+  public reportFailures(): void {
+    const eslintFailureLogMessages: (string | IColorableSequence)[][] = [];
+
+    for (const eslintFileResult of this._lintResult) {
+      const buildFolderRelativeFilePath: string = path.relative(
+        this._buildFolderPath,
+        eslintFileResult.filePath
+      );
+      for (const message of eslintFileResult.messages) {
+        // https://eslint.org/docs/developer-guide/nodejs-api#◆-lintmessage-type
```
octogonz, line 43: Better to make a `const enum ESlintSeverity` and document that, rather than using the integers directly.

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
```
octogonz, line 1: 👍 for extracting TSLint into a separate file.

apps/heft/src/plugins/TypeScriptPlugin/Eslint.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  protected async initializeAsync(): Promise<void> {
+    this._eslint = new this._eslintPackage.ESLint({
+      cwd: this._buildFolderPath
+    });
+
+    this._eslintBaseConfiguration = await this._eslint.calculateConfigForFile(this._linterConfigFilePath);
+
+    this._eslintCli = new this._eslintPackage.CLIEngine({
```
octogonz, line 84: ESLint goes all nuts with `cosmiconfig` alternative filenames: `.eslintrc.js`, `.eslintrc.cjs`, `.eslintrc.yaml`, `.eslintrc.yml`, `.eslintrc.json`, `.eslintrc`, `package.json` (in that order of precedence).

Since Heft is opinionated, IMO it should reject all names other than `.eslintrc.js` for a couple reasons:
- In a monorepo, it's a confusing experience to have to consider 7 different filenames when searching for these files.
- In practice `.js` is the only generally useful file extension, because due to historical reasons basic config problems cannot be solved without scripting (e.g. [__dirname](https://github.com/microsoft/rushstack/blob/8b0a39576caf5a1cfb34321543fe118f17196067/libraries/stream-collator/.eslintrc.js#L6) in our setup)

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
iclanton: Here's a fix for that: https://github.com/microsoft/rushstack/pull/2009

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
iclanton: Yeah I didn't like ESLint's default formatting. VSCode seems fine with it, although I'm not sure if we want to test any specific editor detection?

apps/heft/src/plugins/TypeScriptPlugin/Eslint.ts
iclanton: stream-collator doesn't handle watch processes well. Perhaps we need two modes? An important part of a watch mode is seeing *something* getting logged, telling you that the process hasn't frozen.

I'm not completely sold on separating blocks of logs into contiguous streams as a good UI. In Rush it makes sense because it's separate projects that are effectively independent, so their parallel builds are irrelevant from a logging perspective. The same doesn't apply to a truly parallelized build of a single project.

I'm going to rework the error reporting in a separate PR to follow some of the ideas you've posted here, and then I think we should do subsequent PRs to rework the rest of the logging UI.

apps/heft/src/plugins/TypeScriptPlugin/Eslint.ts
iclanton: That module allows for rule duration logging. I've patched into that, but we'll need to do more research if we want to log more specific metrics.

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
octogonz: The other popular ones where I work are: Sublime, WebStorm, Atom

VIM and IntelliJ Ultimate are other possibilities.

Seems time consuming to install them all and try them, though...

PR #2046, opened 2020-07-22
[Module Minifier Plugin] Make @types/webpack peer dependency optional

dmichon-msft:
Move `@types/webpack` to optional peer dependency.
Fix base type of `IExtendedModule` to only be based on `webpack.compilation.Module`, not `webpack.Module`, since the latter is the webpack configuration "module" section, which is completely irrelevant.

### DIFF
--- a/webpack/module-minifier-plugin/package.json
+++ b/webpack/module-minifier-plugin/package.json
@@ -17,10 +17,13 @@
   },
   "peerDependencies": {
     "webpack": "^4.31.0",
-    "webpack-sources": "~1.4.3"
-  },
-  "optionalPeerDependencies": {
+    "webpack-sources": "~1.4.3",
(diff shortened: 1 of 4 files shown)


webpack/module-minifier-plugin/package.json
```diff
@@ -17,7 +17,9 @@
   },
   "peerDependencies": {
     "webpack": "^4.31.0",
-    "webpack-sources": "~1.4.3",
+    "webpack-sources": "~1.4.3"
+  },
+  "optionalPeerDependencies": {
```
iclanton, line 22: Looks like this is the correct way to specify an optional peerDependency: https://github.com/microsoft/rushstack/blob/master/webpack/localization-plugin/package.json#L20-L27

PR #2074, opened 2020-08-05
[rush] Ensure downstream workspace package version bumps also bump parent packages

D4N14L:


apps/rush-lib/src/logic/PublishUtilities.ts
```diff
@@ -642,25 +642,32 @@ export class PublishUtilities {
       dependencies[change.packageName] &&
       !PublishUtilities._isCyclicDependency(allPackages, parentPackageName, change.packageName)
     ) {
-      const requiredVersion: string = new DependencySpecifier(
+      const requiredVersion: DependencySpecifier = new DependencySpecifier(
         change.packageName,
         dependencies[change.packageName]
-      ).versionSpecifier;
+      );
+      const isAnyWorkspaceVersion: boolean =
```
iclanton, line 649: "AnyWorkspace" is a little confusing. It could be interpreted to mean "is it any variety of a workspace version," instead of "a workspace version that specifically selects anything."

Perhaps we should call this `isWorkspaceWildcardVersion`

apps/rush-lib/src/logic/PublishUtilities.ts
```diff
@@ -642,25 +642,32 @@ export class PublishUtilities {
...
         } else {
           // Either it already satisfies the new version, or doesn't.
           // If not, the downstream dep needs to be republished.
-          changeType = semver.satisfies(change.newVersion!, requiredVersion)
-            ? ChangeType.dependency
-            : ChangeType.patch;
+          // The downstream dep will also need to be republished if using `workspace:*` as this will publish
+          // as the exact version.
+          changeType =
+            semver.satisfies(change.newVersion!, requiredVersion.versionSpecifier) && !isAnyWorkspaceVersion
```
iclanton, line 668: Seems like this should be

```suggestion
            isAnyWorkspaceVersion || semver.satisfies(change.newVersion!, requiredVersion.versionSpecifier)
```

so a `workspace:*` version gets a "dependency"-type bump. Isn't that what `ChangeType.dependency` is for?

common/changes/@microsoft/rush/danade-FixBumpWorkspaceStar_2020-08-05-03-48.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Ensure downstream workspace package version bumps also bump parent packages",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue where a package version bump would not bump downstream packages with a `workspace:*` dependency specifier.",
```

apps/rush-lib/src/logic/PublishUtilities.ts
D4N14L: Agreed

apps/rush-lib/src/logic/PublishUtilities.ts
D4N14L: No, not from my understanding. A `ChangeType.dependency` bump is used to signify that a dependency was bumped, but that the current version range can satisfy the new dependency. That change type doesn't actually bump the version, but patch does.

PR #2095, opened 2020-08-12
[heft] Formalize the way extendable configuration files are loaded.

iclanton:


common/reviews/api/heft-configuration-loader.api.md
```diff
@@ -0,0 +1,68 @@
+## API Report File for "@rushstack/heft-configuration-loader"
```
octogonz, line 1: This name is pretty verbose, and sounds a bit like a Webpack loader plugin.  Should we shorten `@rushstack/heft-configuration-loader` to `@rushstack/heft-config`?

libraries/node-core-library/src/Resolve.ts
```diff
@@ -3,15 +3,23 @@
...
+/**
+ * @alpha
+ */
 export interface IResolvePackageOptions {
   doNotResolveSymlinks: boolean;
 }
 
-export class ResolveUtilities {
+/**
+ * @alpha
```
octogonz, line 20: Let's move this functionality into the `Import` API.  What we are resolving are NPM package imports, so they seem logically related enough to be the same API.

libraries/node-core-library/src/Resolve.ts
```diff
@@ -3,15 +3,23 @@
...
+ */
+export class Resolve {
```
octogonz, line 22: Let's move these `Resolve` members into the `Import` API.  What we are resolving are NPM package imports, so they seem logically related enough to be the same API.

libraries/node-core-library/src/Resolve.ts
```diff
@@ -3,15 +3,23 @@
 
 import { sync } from 'resolve';
 import * as path from 'path';
-import { FileSystem, PackageJsonLookup } from '@rushstack/node-core-library';
+
+import { PackageJsonLookup } from './PackageJsonLookup';
+import { FileSystem } from './FileSystem';
 
 const packageJsonLookup: PackageJsonLookup = new PackageJsonLookup();
```
octogonz, line 10: Should `PackageJsonLookup` have a default instance?  

The reason for not being a singleton is that if the filesystem changes after a lookup, the cache will return the old value for any subsequent lookups. This might be confusing for a task that is writing package.json files as outputs. 

But if a bunch of modules are all creating their own private caches, it defeats the point of caching.

libraries/node-core-library/src/Resolve.ts
```diff
@@ -87,7 +95,7 @@ export class ResolveUtilities {
       pathInsidePackage = packagePath.substr(lastSlashIndex + 1);
```
octogonz, line 95: [GitHub is deficient, so I have to copy+paste these lines in order to attach a comment.]
```ts
      return path.dirname(
        sync(packageName, {
          basedir: normalizedRootPath,
          packageFilter: (pkg: { main: string }): { main: string } => {
            pkg.main = 'package.json';
            return pkg;
          }
        })
      );
```

This code does something right:
- It patches the `"main"` field to point to `package.json`, so we can find the package folder even in cases where the entry point is in a subdirectory

But it has two other bugs:
- You forgot to add `preserveSymlinks: false`, so it applies obsolete Node.js lookups. The `resolve` package never fixed their default, so everyone has to remember to add this setting, otherwise it resolves incorrectly.
- By default `resolve` will also resolve Node.js built-in libraries. For example, when `package.json` refers to `http`, it means the NPM package of that name. Whereas `resolve` assumes you mean the Node.js `http` module. The workaround is to append a slash, [as we did here](https://github.com/microsoft/rushstack/blob/de6d949aa5cce8c9ed957b552c6e2fe3ba13dd0d/apps/rush-lib/src/logic/deploy/DeployManager.ts#L291)

Generally speaking, the `resolve.sync` API design is just super duper error prone. I've been wanting to banish all references to that API across our code base (of which there are many), and wrap it up behind a tidy `node-core-library` wrapper like `Import.resolveThing()`.  Maybe your PR can blaze that trail?

libraries/node-core-library/src/Resolve.ts
```diff
@@ -3,15 +3,23 @@
...
+import { PackageJsonLookup } from './PackageJsonLookup';
+import { FileSystem } from './FileSystem';
 
 const packageJsonLookup: PackageJsonLookup = new PackageJsonLookup();
 
+/**
+ * @alpha
+ */
 export interface IResolvePackageOptions {
   doNotResolveSymlinks: boolean;
```
octogonz, line 16: Why do we need this? When would it ever be useful?

libraries/node-core-library/src/Resolve.ts
```diff
@@ -3,15 +3,23 @@
...
 }
 
-export class ResolveUtilities {
+/**
+ * @alpha
+ */
+export class Resolve {
   /**
    * Gets the path of a given package name from the perspective of a given path
    */
```
octogonz, line 25: Suggested API design:

```ts

interface IImportResolveOptions {
  packageName: string;  // required for now
  baseFolderPath: string; // required for now

  // If true, if the package name matches a Node.js system module, then the return
  // value will be the package name without any path.  This will take precedence over
  // an installed NPM package of the same name.
  //
  // Example:
  // Import.resolveModulePath({ packageName: "fs", baseFolderPath: process.cwd() })
  //  --> "fs"
  includeSystemModules?: boolean;

  // If true, then packageName is allowed to refer to the package.json of the active project.
  // It will take precedence over any installed dependency with the same name.
  // Note that this requires an additional PackageJsonLookup calculation.
  //
  // Example:
  // Import.resolveModulePath({ packageName: "my-project", baseFolderPath: process.cwd(), 
  //   allowSelfReference: true })
  //  --> "path/to/my-project"
  allowSelfReference?: boolean;
}

interface IImportResolveModulePathOptions extends IImportResolveOptions {
}
interface IImportResolvePackageFolderOptions extends IImportResolveOptions {
}

class Import {
  // Example:
  // Import.resolveModulePath({ packageName: "colors", baseFolderPath: process.cwd() })
  //  --> "path/to/node_modules/colors/lib/index.js"
  public static resolveModulePath(options: IImportResolveModulePathOptions): string {
  }

  // Example:
  // Import.resolvePackageFolder({ packageName: "colors", baseFolderPath: process.cwd() })
  //   --> "path/to/node_modules/colors"
  public static resolvePackageFolder(options: IResolvePackageFolderOptions): string {
  }
}

```

octogonz: Bonus points if you can figure out how to model the [realPathSync](https://github.com/microsoft/rushstack/blob/de6d949aa5cce8c9ed957b552c6e2fe3ba13dd0d/apps/rush-lib/src/logic/deploy/DeployManager.ts#L302) hook that `rush deploy` needs. :-)

And also resolution of module+path expressions (e.g. `"@rushstack/eslint-config/patch/modern-module-resolution"`)

libraries/heft-configuration-loader/src/test/ConfigurationFileLoader.test.ts
```diff
@@ -0,0 +1,292 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import * as path from 'path';
+
+import { ConfigurationFileLoader, PathResolutionMethod, InheritanceType } from '../ConfigurationFileLoader';
```
octogonz, line 6: For end users, the error reporting experience is pretty important. It might be useful to have some unit tests for that.

For example, consider a situation like this:

**config/config1.json**
```
{
  "extends": "@rushstack/thing/includes/config2.json",
  "folderPaths": [ "./lib" ]
}
```

**node_modules/@rushstack/thing/includes/config2.json**
```
{
  "extends": "./config3.json"
}
```

**node_modules/@rushstack/thing/includes/config3.json**
```
{
  "filePaths": "value"
}
```

Various interesting errors are possible:
1. `"filePaths"` is supposed to be an array, but they specified a string
2. `"extends"` refers back to an earlier file, creating a circular reference
3. `config3.json` does not exist
4. `config3.json` contains invalid JSON and cannot be parsed at all
5. Everything loads okay, but after all the files are merged together, an error occurs because `"filePaths"` and `"folderPaths"` cannot be specified together. The schema allows either, but not both.

For each of these examples, we can ask:
- Does the error message clearly explain what the problem is?
- Can the end user easily determine which file they need to edit to fix the error?

We don't need to solve everything in this PR, but it would be useful to have some tests to track our progress.

octogonz:
General note: Once this is merged, we should revisit PR https://github.com/microsoft/rushstack/pull/2054 which proposes a standard way to specify default values in JSON.  Default values can probably eliminate the need for `.heft/clean.json` for most projects.

libraries/node-core-library/src/Resolve.ts
iclanton: That's a good point. There's already a `clearCache()` function on it, and we actually already do have a static instance used by `PackageJsonLookup.loadOwnPackageJson`. What would you want a default singleton to look like?

octogonz: `PackageJsonLookup.instance`?

With a comment something like this:
```ts
/**
 * A singleton instance of `PackageJsonLookup`, which is useful for short-lived processes
 * that can reasonably assume that the file system will not be modified after the cache
 * is populated.
 *
 * @remarks
 * For long-running processes that need to clear the cache at appropriate times,
 * it is recommended to create your own instance of `PackageJsonLookup` instead
 * of relying on this instance.
 */
```

libraries/node-core-library/src/Resolve.ts
iclanton: Removed.

common/changes/@rushstack/node-core-library/ianc-heft-config-files_2020-08-13-19-02.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/node-core-library",
+      "comment": "Include an experimental API for resolving packages and paths within packages.",
```
iclanton, line 5:
```suggestion
      "comment": "Include an Import.resolve API for resolving paths relative to other paths.",
```

libraries/node-core-library/src/Import.ts
```diff
@@ -1,13 +1,58 @@
...
+  resolvePath: string;
+
+  /**
+   * The path from which {@link IImportResolveOptions.resolvePath} should be resolved
+   */
+  baseFolderPath: string;
+
+  /**
+   * If true, if the package name matches a Node.js system module, then the return
+   * value will be the package name without any path.  This will take precedence over
```
octogonz, line 28: Use `@remarks` block to keep the summary from being too long.  The example will be more readable as a code block:

  /**
   * If true, if the package name matches a Node.js system module, then the return
   * value will be the package name without any path.  
   * 
   * @remarks
   * This will take precedence over an installed NPM package of the same name.
   *
   * Example:
   * ```ts
   * // Returns the string "fs" indicating the Node.js system module
   * Import.resolveModulePath({ 
   *   resolvePath: "fs", 
   *   basePath: process.cwd() 
   * })
   * ```
   */
  includeSystemModules?: boolean;

octogonz: Same for `allowSelfReference`

libraries/node-core-library/src/Import.ts
```diff
@@ -83,4 +128,105 @@ export class Import {
...
+        resolvedPackagePath = nodeJsPath.dirname(ownPackageJsonPath);
+      }
+    }
+
+    if (!resolvedPackagePath) {
+      try {
+        resolvedPackagePath = nodeJsPath.dirname(
+          Resolve.sync(
+            // Append a slash to the package name to ensure `resolve.sync` doesn't attempt to return a system package
+            `${packageName}/`,
```
octogonz, line 209: I don't see anything ever getting added to the `_builtInModules` set. (?)

According to the `resolve` package maintainer, appending a slash or not is the magic way to control whether it considers system modules or not.  He chose this design because it is how Node.js `require()` behaves.

However, I find that highly counterintuitive.  In 99% of cases, `Import.resolve()` will be resolving NPM packages, and it would be wrong to match a system module. 

@iclanton Thus I'm thinking the design can be like this:

- by default if `includeSystemModules=false`, we append a trailing slash ONLY if `resolvePath` does not already contain a slash
- if `includeSystemModules=true` then we do not append a slash.  Thus `resolvePath="fs"` would return the system path. I'm undecided about what to do with ``resolvePath="fs/"` in this case.  Faithfully emulate Node.js? Or assume it is a mistake an report an error?  Not sure what's better.

libraries/node-core-library/src/Import.ts
```diff
@@ -1,13 +1,58 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
+import * as nodeJsPath from 'path';
```
octogonz, line 4: It's weird not to call it `path`. If it conflicts with another name in this file, we should probably rename the other one.

libraries/node-core-library/src/Import.ts
```diff
@@ -1,13 +1,58 @@
...
+
+import { PackageJsonLookup } from './PackageJsonLookup';
+import { FileSystem } from './FileSystem';
+
+/**
+ * @public
+ */
+export interface IImportResolveOptions {
+  /**
+   * The path to resolve.
```
octogonz, line 17: Might be useful to give some examples of what is supported.

libraries/node-core-library/src/Import.ts
```diff
@@ -83,4 +128,105 @@ export class Import {
...
+        resolvedPackagePath = nodeJsPath.dirname(
+          Resolve.sync(
+            // Append a slash to the package name to ensure `resolve.sync` doesn't attempt to return a system package
+            `${packageName}/`,
+            {
+              basedir: normalizedRootPath,
+              preserveSymlinks: false,
+              packageFilter: (pkg: { main: string }): { main: string } => {
+                // In case the "main" property isn't defined, set it to something we know will exist
+                pkg.main = 'package.json';
```
octogonz, line 215: The above comment is incorrect. There are two different kinds of resolving:

- `f1("jquery") --> "./node_modules/jquery/lib/index.js"`
- `f2("jquery") --> "./node_modules/jquery"`

The trick for `f2()` is to hardwire `pkg.main` to return `package.json` and then call `path.dirname()` on the result.  For `f1()` we can use `resolve.sync()` in the normal way.

octogonz: BTW a missing `"main"` field is normally NOT a problem, because if unspecified then Node.js [will default to](https://nodejs.org/api/modules.html#modules_all_together) looking for `./index.js`, `./index.json`, `index.node`.

@BPapp-MS  FYI

libraries/node-core-library/src/Import.ts
```diff
@@ -83,4 +128,105 @@ export class Import {
...
+   * Resolves a path in a package, relative to another path.
+   */
+  public static resolve(options: IImportResolveOptions): string {
+    const { resolvePath } = options;
+
+    if (nodeJsPath.isAbsolute(resolvePath)) {
+      return resolvePath;
+    }
+
+    let normalizedRootPath: string = FileSystem.getRealPath(options.baseFolderPath);
```
octogonz, line 142: It seems like `require.sync()` already implements all these parsing features itself.  It almost exactly simulates Node.js `require()` behavior.

libraries/node-core-library/src/Import.ts
```diff
@@ -83,4 +128,105 @@ export class Import {
...
+    }
+
+    let packageName: string;
+    let pathInsidePackage: string | undefined;
+    if (slashAfterPackageNameIndex === -1) {
+      // This looks like a package name without a path
+      packageName = resolvePath;
+    } else {
+      packageName = resolvePath.substr(0, slashAfterPackageNameIndex);
+      pathInsidePackage = resolvePath.substr(slashAfterPackageNameIndex + 1);
```
octogonz, line 167: It seems like `require.sync()` already implements all these parsing features itself.  It almost exactly simulates Node.js `require()` behavior.

libraries/node-core-library/src/Import.ts
iclanton: This is what we called it for `FileSystem`, but I'll change it back.

libraries/node-core-library/src/Import.ts
iclanton: `_builtInModules` is set on [line 172](https://github.com/microsoft/rushstack/pull/2095/files/8fe1fb927b899284053cf3c555acd5d6cad7fcd9#diff-cf89596b0e435a3be751cbc2b37072baR171-R173)

`resolve` relies on a janky [hardcoded list](https://github.com/browserify/resolve/blob/master/lib/core.json) of "core" modules, while this solution gets that list from the runtime. I prefer my solution.

libraries/node-core-library/src/Import.ts
```diff
@@ -83,4 +173,109 @@ export class Import {
...
+        return ownPackage.packageRootPath;
+      }
+    }
+
+    try {
+      const resolvedPath: string = Resolve.sync(packageName, {
+        basedir: normalizedRootPath,
+        preserveSymlinks: false,
+        packageFilter: (pkg: { main: string }): { main: string } => {
+          // In case the "main" property isn't defined, set it to something we know will exist
```
octogonz, line 247:
```suggestion
          // Hardwire "main" to point to a file that is guaranteed to exist.
          // This helps resolve packages such as @types/node that have no entry point.
          // And then we can use path.dirname() below to locate the package folder,
          // even if the real entry point was in an subfolder with arbitrary nesting.
```

libraries/node-core-library/src/Import.ts
```diff
@@ -83,4 +173,109 @@ export class Import {
...
+
+      const packagePath: string = path.dirname(resolvedPath);
+      const packageJson: IPackageJson = PackageJsonLookup.instance.loadPackageJson(
+        path.join(packagePath, 'package.json')
+      );
+      if (packageJson.name === packageName) {
+        return packagePath;
+      } else {
+        throw new Error();
+      }
```
octogonz, line 261: Seems like we can delete this code. Since we hardwired `main` to be package.json, that file must exist, otherwise `resolve.sync()` would have failed.

octogonz: Also I can't think of any realistic case where `packageJson.name !== packageName` could happen.

iclanton: The

```
      if (packageJson.name === packageName) {
        return packagePath;
      } else {
        throw new Error();
      }
```

code ensures that what was passed to this function is a package name and not a module path.

libraries/node-core-library/src/Import.ts
```diff
@@ -83,4 +173,109 @@ export class Import {
     const importLazyLocal: (moduleName: string) => unknown = importLazy(require);
     return importLazyLocal(moduleName);
   }
+
+  /**
+   * Resolve a module, from the perspective of a specified folder. This is used to resolve a path that
+   * would be used in a `require(...)` or `import '...'` statement.
```
octogonz, line 179:
```suggestion
   * This resolves a module path using similar logic as the Node.js `require.resolve()` API,
   * but supporting extra features such as specifying the base folder.
   *
   * @remarks
   * A module path is a text string that might appear in a statement such as
   * `import { X } from "____";` or `const x = require("___");`.  The implementation is based
   * on the popular `resolve` NPM package.
   *
   * Suppose the `example` is an NPM package whose entry point is `lib/index.js`:
   * ```ts
   * // Returns "/path/to/project/node_modules/example/lib/index.js"
   * Import.resolveModule({ modulePath: 'example' });
   *
   * // Returns "/path/to/project/node_modules/example/lib/other.js"
   * Import.resolveModule({ modulePath: 'example/lib/other' });
   * ```
   * If you need to determine the containing package folder
   * (`/path/to/project/node_modules/example`), use {@link Import.resolvePackage} instead.
   *
   * @returns the absolute path of the resolved module.  If it resolves to a Node.js
   * system module, then its name is returned without any file path.
```

libraries/node-core-library/src/Import.ts
```diff
@@ -83,4 +173,109 @@ export class Import {
...
+          preserveSymlinks: false
+        }
+      );
+    } catch (e) {
+      throw new Error(`Cannot find module "${modulePath}" from "${options.baseFolderPath}".`);
+    }
+  }
+
+  /**
+   * Resolve the path to a package, from the perspective of a specified folder.
```
octogonz, line 224:
```suggestion
   * Performs module resolution to determine the folder where a package is installed.
   *
   * @remarks
   * Suppose `example` is an NPM package whose entry point is `lib/index.js`:
   * ```ts
   * // Returns "/path/to/project/node_modules/example"
   * Import.resolvePackage({ packageName: 'example' });
   * ```
   *
   * If you need to resolve a module path, use {@link Import.resolveModule} instead:
   * ```ts
   * // Returns "/path/to/project/node_modules/example/lib/index.js"
   * Import.resolveModule({ modulePath: 'example' });
   * ```
   *
   * @returns the absolute path of the package folder.
   * If {@link IImportResolveOptions.includeSystemModules} is specified
   * and a system module is found, then its name is returned without any file path.
```

libraries/node-core-library/src/Import.ts
```diff
@@ -1,13 +1,103 @@
...
+   *   resolvePath: "current-project",
+   *   basePath: process.cwd(),
+   *   allowSelfReference: true
+   * })
+   * ```
+   */
+  allowSelfReference?: boolean;
+}
+
+/**
```
octogonz, line 61:
```suggestion
/**
 * Options for {@link Import.resolveModule}
```

libraries/node-core-library/src/Import.ts
```diff
@@ -1,13 +1,103 @@
...
+ */
+export interface IImportResolveModuleOptions extends IImportResolveOptions {
+  /**
+   * The module identifier to resolve. For example "\@rushstack/node-core-library" or
+   * "\@rushstack/node-core-library/lib/index.js"
+   */
+  modulePath: string;
+}
+
+/**
```
octogonz, line 72:
```suggestion
/**
 * Options for {@link Import.resolvePackage}
```

libraries/node-core-library/src/Import.ts
```diff
@@ -1,13 +1,103 @@
...
+import * as path from 'path';
 import importLazy = require('import-lazy');
+import * as Resolve from 'resolve';
+import nodeModule = require('module');
+
+import { PackageJsonLookup } from './PackageJsonLookup';
+import { FileSystem } from './FileSystem';
+import { IPackageJson } from './IPackageJson';
+
+/**
```
octogonz, line 13:
```suggestion
/**
 * Common options shared by {@link IImportResolveModuleOptions} and {@link IImportResolvePackageOptions}
```

libraries/heft-config-file/src/ConfigurationFileLoader.ts
```diff
@@ -0,0 +1,429 @@
...
+    const annotation: IConfigurationFileFieldAnnotation<TParentProperty> | undefined =
+      options.parentObject[CONFIGURATION_FILE_FIELD_ANNOTATION];
+    if (annotation && annotation.originalValues.hasOwnProperty(options.propertyName)) {
+      return annotation.originalValues[options.propertyName] as TValue;
+    }
+
+    throw new Error(`No original value could be determined for property "${options.propertyName}"`);
+  }
+
+  private async _loadConfigurationFileAsyncInner(
```
octogonz, line 179:
```suggestion
  private async _loadConfigurationFileInnerAsync(
```

libraries/heft-config-file/src/ConfigurationFileLoader.ts
```diff
@@ -0,0 +1,429 @@
...
+ */
+export type IPropertyInheritanceTypes<TConfigurationFile> = {
+  [propertyName in keyof TConfigurationFile]?: InheritanceType;
+};
+
+/**
+ * @beta
+ */
+export interface IJsonPathsMetadata {
+  [jsonPath: string]: IJsonPathMetadata;
```
octogonz, line 82: This ABSOLUTELY needs documentation. :-P

libraries/heft-config-file/src/ConfigurationFileLoader.ts
```diff
@@ -0,0 +1,429 @@
...
+ */
+export interface IOriginalValueOptions<TParentProperty> {
+  parentObject: TParentProperty;
+  propertyName: keyof TParentProperty;
+}
+
+/**
+ * @beta
+ */
+export class ConfigurationFileLoader<TConfigurationFile> {
```
octogonz, line 111: I still suspect that `ConfigurationFile` might be a better name for this class:  If we wanted to modify and resave a config file, wouldn't this class be able to manage that?  Or would we need to introduce a separate class (`PackageJsonEditor`)?

libraries/heft-config-file/src/ConfigurationFileLoader.ts
```diff
@@ -0,0 +1,429 @@
...
+export enum PathResolutionMethod {
+  /**
+   * Resolve a path relative to the configuration file
+   */
+  resolvePathRelativeToConfigurationFile,
+
+  /**
+   * Resolve a path relative to the root of the project containing the configuration file
+   */
+  resolvePathRelativeToProjectRoot,
```
octogonz, line 39: FYI Jest and API Extractor both use tokens to make these nuances explicit.

Consider [this setting](https://api-extractor.com/pages/commands/config_file/#mainentrypointfilepath) for example:

Capture

It has a couple advantages:
- It allows the user to choose between multiple ways of resolving.
- It makes the resolution explicit, so nobody needs to memorize how each field gets resolved.

Given the complexity of Heft's config files, this might be worth implementing. It should go in a separate PR, though.

iclanton: Agreed.

libraries/node-core-library/src/Import.ts
octogonz: @iclanton It might be better to use the `PackageName` API for that, since it is a cheap syntax test that doesn't need to parse the file from disk. And it will give better errors.

PR #2106, opened 2020-08-14
[rush] Fix bad hash being written to pnpm-lock when version --bump is run

D4N14L:


apps/rush-lib/src/logic/VersionManager.ts
```diff
@@ -105,6 +105,11 @@ export class VersionManager {
       changeManager.updateChangelog(!!shouldCommit);
     }
 
+    // Refresh rush configuration
+    this._rushConfiguration = RushConfiguration.loadFromConfigurationFile(
```
sachinjoseph, line 109: Don't we reload the Rush config above?

D4N14L: We do, but the changes between then and here cause the package.json files to be modified even further. Reloading here ensures that we have the most up-to-date projects.

PR #2153, opened 2020-09-01
[heft] fix expands parameter resolution in getConfigHash

BPapp-MS:
makes it so that the getConfigHash function in Tslint.ts is compatible with the variety of values of the "extends" parameter on config files, including arrays of strings that can specify paths to either config files, or packages who define a config file as "main" in the package.json


apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -5,7 +5,7 @@ import * as path from 'path';
 import { Tslint as TTslint } from '@microsoft/rush-stack-compiler-3.7';
 import { Import } from '@rushstack/node-core-library';
 import * as crypto from 'crypto';
-import { Terminal, JsonFile } from '@rushstack/node-core-library';
```
dmichon-msft, line 8: Merge with the other import from `@rushstack/node-core-library`

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -36,28 +37,63 @@ export class Tslint extends LinterBase<TTslint.RuleFailure> {
...
   ): crypto.Hash {
     interface IMinimalConfig {
-      extends?: string;
+      extends?: string | string[];
     }
 
     terminal.writeVerboseLine(`Examining config file "${configFilePath}"`);
-
+    // if configFilePath is not a json file, assume that it is a package whose main file
+    // is a config file, per the the "extends" spec of tslint.json, found at
```
Claudiazhaoya, line 71: extra 'the'. minor typo :)

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -36,28 +36,63 @@ export class Tslint extends LinterBase<TTslint.RuleFailure> {
...
+    const extendsProperty: string | string[] | undefined = parsedConfig.extends;
+    let hash: crypto.Hash = previousHash || crypto.createHash('sha1');
+
+    if (extendsProperty instanceof Array) {
+      for (const extendFile of extendsProperty) {
+        const extendFilePath: string = Import.resolveModule({
+          modulePath: extendFile,
+          baseFolderPath: path.dirname(configFilePath)
+        });
+        hash = this.getConfigHash(extendFilePath, terminal, fileSystem, hash);
```
Claudiazhaoya, line 87: Probably use Tslint.getConfigHash here to align it with the other function calls?

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -36,28 +36,63 @@ export class Tslint extends LinterBase<TTslint.RuleFailure> {
...
+      }
+    } else if (extendsProperty) {
+      // note that if we get here, extendsProperty is a string
       const extendsFullPath: string = Import.resolveModule({
-        modulePath: parsedConfig.extends,
+        modulePath: extendsProperty,
         baseFolderPath: path.dirname(configFilePath)
       });
-      hash = Tslint.getConfigHash(extendsFullPath, terminal, fileSystem);
-    } else {
```
Claudiazhaoya, line 59: Why deleting this condition? What if there is no extend property in the config file?

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
BPapp-MS: fixed!

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
BPapp-MS: this condition is handled by the update() call in the return line

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -28,6 +27,7 @@ export class Tslint extends LinterBase<TTslint.RuleFailure> {
   private _enabledRules: TTslint.IRule[];
   private _ruleSeverityMap: Map<string, TTslint.RuleSeverity>;
   protected _lintResult: TTslint.LintResult;
+  private static _packageJsonLookup: PackageJsonLookup = new PackageJsonLookup();
```
iclanton, line 30: Use `PackageJsonLookup.instance`

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -36,28 +36,63 @@ export class Tslint extends LinterBase<TTslint.RuleFailure> {
     this._fileSystem = options.fileSystem;
   }
 
+  private static _resolvePackageMainFilePath(packagePath: string): string {
+    const packageJson: IPackageJson | undefined = this._packageJsonLookup.tryLoadPackageJsonFor(packagePath);
```
iclanton, line 40:
```suggestion
    const packageJson: IPackageJson | undefined = PackageJsonLookup.instance.tryLoadPackageJsonFor(packagePath);
```

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -36,28 +36,63 @@ export class Tslint extends LinterBase<TTslint.RuleFailure> {
...
+    if (!packageJson) {
+      throw new Error(
+        `Supplied package path ${packagePath} could not resolve a package.json while searching for main file`
+      );
+    } else if (!packageJson.main) {
+      throw new Error(`Package at path ${packagePath} specifies no main file`);
+    }
+    return path.resolve(packagePath, packageJson.main);
```
iclanton, line 48: This isn't necessarily correct. You may need to `.realPath` the `packagePath`

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -36,28 +36,63 @@ export class Tslint extends LinterBase<TTslint.RuleFailure> {
     this._fileSystem = options.fileSystem;
   }
 
+  private static _resolvePackageMainFilePath(packagePath: string): string {
```
iclanton, line 39: How is this appreciably different from `Import.resolveModule`?

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
```diff
@@ -36,28 +36,63 @@ export class Tslint extends LinterBase<TTslint.RuleFailure> {
...
+    return path.resolve(packagePath, packageJson.main);
+  }
+
+  /**
+   * getConfigHash returns the sha1 hash unique to the contents of the given config as
+   * well as all the configs that the given one extends.
+   *
+   * If suppied the previousHash parameter, the hash is updated with the contents of the
+   * file's extended configs and itself before being returned. Passing a digested hash to
+   * this parameter will result in an error.
```
iclanton, line 57:
```suggestion
   * Returns the sha1 hash of the contents of the config file at the provided path and the
   * the configs files that the referenced file extends.
   * 
   * @param previousHash - If supplied, the hash is updated with the contents of the
   * file's extended configs and itself before being returned. Passing a digested hash to
   * this parameter will result in an error.
```

common/changes/@rushstack/heft/benpapp-AddNodeArgsForTSConfig_2020-08-24-23-25.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft",
+      "comment": "Making it so that the getConfigHash function in Tslint.ts is compatible with the variety of values of the \"extends\" parameter on config files, including arrays of strings that can specify paths to either config files, or packages who define a config file as \"main\" in the package.json",
```
iclanton, line 5:
```suggestion
      "comment": "Fix issues with parsing of tslint.json config files, including adding support for an array provided to \"extends\" and proper Node module resolution to extended config files.",
```

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
BPapp-MS: IIRC the config file paths that were earlier returned by resolveModule were resulting in paths that were causing an error when we then tried to call fileSystem.readFile on them. It wasn't clear to me from the [documentation ](https://github.com/microsoft/rushstack/blob/1f3a8bd81c79e975ca55ec41f24011aae9839461/libraries/node-core-library/src/Import.ts#L181) of resolveModule why this was happening, I thought it was returning folder names but looking at the documentation again it could be ommitting file extensions?

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
BPapp-MS: Okay, do you know what type that property belongs to?

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
BPapp-MS: It seems to be working since the change to referencing Import, so I removed the ad hoc function

apps/heft/src/plugins/TypeScriptPlugin/Tslint.ts
BPapp-MS: Ended up removing this function

iclanton (approved):

PR #2159, opened 2020-09-02
[rush] Move MinimalRushConfiguration to rush-lib and expand properties

D4N14L:


apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -429,18 +429,218 @@ export type ResolutionStrategy = 'fewer-dependencies' | 'fast';
  * configuration file.
```
D4N14L, line 429: Update comment

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -429,18 +429,218 @@ export type ResolutionStrategy = 'fewer-dependencies' | 'fast';
...
+  public get rushConfigurationJson(): IRushConfigurationJson {
+    return this._rushConfigurationJson;
+  }
+
+  /**
+   * The version of rush specified by the rushVersion property of the rush.json configuration file. This
+   * should be a semver style version number like "4.0.0"
+   */
+  public get rushVersion(): string {
+    return this._rushConfigurationJson.rushVersion;
```
D4N14L, line 545: Note: this has changed so it no longer supports the "rushMinimumVersion" property of Rush <= 2.x

apostolisms (approved):

Zenix27 (approved):

sachinjoseph (changes requested):
@D4N14L We shouldn't move this to rush-lib. Loading rush-lib at runtime is quite heavy and I was hoping we could move away from loading rush-lib in Rush binary if the version requested is not the globally installed one.

Instead, can we move it to separate package that can then be referenced by both rush and rush-lib?

sachinjoseph:
> @D4N14L We shouldn't move this to rush-lib. Loading rush-lib at runtime is quite heavy and I was hoping we could move away from loading rush-lib in Rush binary if the version requested is not the globally installed one.
> 
> Instead, can we move it to separate package that can then be referenced by both rush and rush-lib?

Let's say the globally installed version of Rush is 5.32.2 and the repo requests 5.33.2. Now, executing any rush command will cause the global rush to load rush-lib@5.32.2, only to find out that it needs to actually load 5.33.2. I was hoping we could make changes to avoid loading 5.32.2 completely since the repo requests a different version.

octogonz (changes requested):
We discussed this offline yesterday. I'm not sure that exposing `MinimalRushConfiguration ` is the best approach.

PR #2171, opened 2020-09-07
[heft] feat: add --max-workers CLI option for jest

sammarks:
For CI environments, I couldn't find a way to change the `maxWorkers` configuration value for Jest using an environment variable. The best I've come up with so far is to change the `maxWorkers` config option, but that requires the max workers config to be the same everywhere.

Since a CI environment might be more / less powerful than a local development machine, or one person's machine might be more / less powerful than the other, I think it makes sense to forward that Jest CLI option through the Heft CLI.

That way, I can just update my build script inside CI to pass `--maxWorkers` as something else, and have the default value inside the configuration something sensible for local development environments.

I noticed the CLI options for testing could potentially apply to more than just Jest in the future (whenever other plugins are developed), so tried to keep the verbiage as generic as possible.

I also am not sure what the preference is for something like this with regard to testing. So if I'm missing some tests, let me know and I'll write those up.

Let me know if you have any feedback or if there's anything else I need to do!

### DIFF
--- a/apps/heft/src/cli/actions/TestAction.ts
+++ b/apps/heft/src/cli/actions/TestAction.ts
@@ -119,7 +119,8 @@ export class TestAction extends BuildAction {
         'Use this parameter to control maximum number of worker processes tests are allowed to use.' +
         ' This parameter is similar to the parameter noted in the Jest documentation, and can either be' +
         ' an integer representing the number of workers to spawn when running tests, or can be a string' +
-        ' representing a percentage of the available CPUs on the machine to utilize. Example values: "3", "25%"'
+        ' representing a percentage of the available CPUs on the machine to utilize. Example values: "3",' +
+        ' "25%%"' // The "%%" is required because argparse (used by ts-command-line) treats % as an escape character
     });
   }
 
--- a/apps/heft/src/cli/actions/TestAction.ts
+++ b/apps/heft/src/cli/actions/TestAction.ts
@@ -114,13 +114,12 @@ export class TestAction extends BuildAction {
 
     this._maxWorkers = this.defineStringParameter({
       parameterLongName: '--max-workers',
-      argumentName: 'STRING',
+      argumentName: 'COUNT_OR_PERCENTAGE',
       description:
-        'Change the default maximum workers for tests; this parameter conforms to the Jest' +
-        ' documentation, and can either be an integer representing the number of workers to spawn' +
-        ' when running tests, or can be a string representing a percentage of the available CPUs' +
-        ' on the machine to utilize.' +
-        ' This corresponds to the "--maxWorkers" parameter in Jest\'s documentation.'
+        'Use this parameter to control maximum number of worker processes tests are allowed to use.' +
+        ' This parameter is similar to the parameter noted in the Jest documentation, and can either be' +
+        ' an integer representing the number of workers to spawn when running tests, or can be a string' +
+        ' representing a percentage of the available CPUs on the machine to utilize. Example values: "3", "25%"'
     });
   }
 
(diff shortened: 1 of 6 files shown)


BPapp-MS (approved):

apps/heft/src/cli/actions/TestAction.ts
```diff
@@ -110,6 +111,17 @@ export class TestAction extends BuildAction {
         ' default reporter would have presented it. Include this output in your bug report.' +
         ' Do not use "--debug-heft-reporter" in production.'
     });
+
+    this._maxWorkers = this.defineStringParameter({
+      parameterLongName: '--max-workers',
+      argumentName: 'STRING',
```
D4N14L, line 117: MAX_WORKERS

D4N14L (approved):

apps/heft/src/cli/actions/TestAction.ts
iclanton: It should probably be `COUNT`, and this parameter should probably be an integer parameter.

```suggestion
      argumentName: 'COUNT',
```

common/changes/@rushstack/heft/max-workers-cli_2020-09-07-17-39.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft",
+      "comment": "add --maxWorkers option for running tests",
```
iclanton, line 5:
```suggestion
      "comment": "Add --max-workers option to the \"test\" action to control the maximum number of worker processes the test process can use.",
```

apps/heft/src/cli/actions/TestAction.ts
```diff
@@ -110,6 +111,17 @@ export class TestAction extends BuildAction {
...
+      description:
+        'Change the default maximum workers for tests; this parameter conforms to the Jest' +
+        ' documentation, and can either be an integer representing the number of workers to spawn' +
+        ' when running tests, or can be a string representing a percentage of the available CPUs' +
+        ' on the machine to utilize.' +
+        ' This corresponds to the "--maxWorkers" parameter in Jest\'s documentation.'
```
iclanton, line 123:
```suggestion
        'Use this parameter to control maximum number of worker processes tests are allowed to use.' +
        ' This parameter is similar to the parameter noted in the Jest documentation, and can either be' +
        ' an integer representing the number of workers to spawn when running tests, or can be a string' +
        ' representing a percentage of the available CPUs on the machine to utilize. Example values: "3", "25$"'
```

apps/heft/src/cli/actions/TestAction.ts
iclanton: Or perhaps

```suggestion
      argumentName: 'COUNT_OR_PERCENTAGE',
```

sammarks:
Just saw these comments; will implement the suggestions shortly.

Another thing to think about: should we also add a CLI flag to the `rush build` command as well? Typically in a CI environment I'm not calling the heft commands individually, rather the `rush build` command all at once.

Perhaps better for a follow-up PR? Could add into this one as well.

iclanton:
@sammarks - WRT the `rush build` command, you can add a custom parameter to the `build` action in your Rush repo. See docs here: https://rushjs.io/pages/maintainer/custom_commands/

sammarks:
Ah, yes! Forgot about that. That'll work just fine.

iclanton:
@sammarks - hope you don't mind, but I pushed a minor fix to your branch to escape the "%" character in the CLI help, and to update the test snapshots. If this looks good to you, it looks good to me and I'll go ahead and merge this in.

iclanton (approved):

sammarks:
No worries @iclanton, thanks for the fix and quick merge!

PR #2173, opened 2020-09-08
[heft] add --max-workers param to test action

BPapp-MS:
adds --max-workers parameter to test action which passes that value along to the --maxWorkers parameter of the Jest subprocess of the Jest plugin.  It would be nice to reimplement this someday with command line interfaces defined by plugins themselves, when support is added for that.

PR also includes one-line change which adds the --detectOpenHandles flag to the Jest subprocess when Heft is run in debug mode.


BPapp-MS:
Closing: redundant with[ this PR:](https://github.com/microsoft/rushstack/pull/2171) https://github.com/microsoft/rushstack/pull/2171

PR #2181, opened 2020-09-11
[localization-plugin] Handle webpack dev server hot update chunks.

halfnibble:
Webpack dev server will emit hot update chunks ending in 'hot-update.js', which conflict with the latest changes in the LocalizationPlugin. This attempts to address that.


apostolisms (approved):

D4N14L (approved):

iclanton:
This shouldn't be happening, but because it is, I'm cool with this fix.

Let's try to fix the underlying issue as well.

PR #2192, opened 2020-09-15
[heft]: Sass typings generator

halfnibble:
This PR is now ready for review. The SassTypingsPlugin generates .d.ts files in temp/sass-ts in the pre-compile step for TS type checking purposes. It works well with serve mode too.

There is an example project under build-tests/heft-sass-test.


halfnibble:
I forgot to mention, the problem is that the css and map files are being generated in the src/ directory, which makes no sense. The typings sass call outputs to lib/ and the webpack run outputs to dist/, so I have no idea why these files are being generated in the src/ folder.

libraries/typings-generator/package.json
```diff
@@ -27,7 +27,7 @@
     "@microsoft/rush-stack-compiler-3.5": "workspace:*",
     "@microsoft/node-library-build": "workspace:*",
     "@rushstack/eslint-config": "workspace:*",
-    "@rushstack/heft": "workspace:*",
+    "@rushstack/heft": "0.8.0",
```
D4N14L, line 30: Hmm. With all the changes in heft, we may not want to pin to a specific version. Maybe use a ^ version? Or "latest" if supported?

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  private async _transpileSassAsync(
+    fileContents: string,
+    filePath: string,
+    buildFolder: string
+  ): Promise<string> {
+    const result: Result = await LegacyAdapters.convertCallbackToPromise(render, {
+      data: fileContents,
+      importer: (url: string) => ({ file: this._patchSassUrl(url) }),
+      includePaths: [path.join(buildFolder, 'node_modules'), path.join(buildFolder, 'src')],
```
D4N14L, line 66: Just to clarify, is the `src` folder always going to contain the source files? Is that configurable?

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -0,0 +1,96 @@
...
+      indentedSyntax: path.extname(filePath).toLowerCase() === '.sass'
+    });
+
+    return result.css.toString();
+  }
+
+  private _patchSassUrl(url: string): string {
+    if (url[0] === '~') {
+      url = 'node_modules/' + url.substr(1);
+    } else if (url === 'stdin') {
```
D4N14L, line 76: When/where does this occur out of curiosity?

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -0,0 +1,96 @@
...
+  [className: string]: string;
+}
+
+/**
+ * Generates type files (.d.ts) for Sass/SCSS/CSS files.
+ *
+ * @public
+ */
+export class SassTypingsGenerator extends StringValuesTypingsGenerator {
+  private _classMap: IClassMap;
```
D4N14L, line 28: this can be made local to _getClassNamesFromCSSAsync it seems?

D4N14L (approved):

libraries/typings-generator/package.json
halfnibble: I agree with you, but this seems to be what other projects are using that have a circular dependency on Heft. Since it is only being used to build, and since build works, I think I will leave this as is for now. What we should do, however, is make a PR that updates all the projects currently set to "@rushstack/heft": "0.8.0". I think that should happen in a separate PR.

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
halfnibble: Good point. I have now made this configurable. In fact, I made most of these options configurable.

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
halfnibble: Good question! 

There is an example in `build-tests/heft-sass-test/src/styles.sass` that you can refer to.

In Sass, it is common to abbreviate 3rd party package imports with `~`, and something magically resolves it to "./node_modules/". For example, to import Sass mixins from the buttono project, it is typically done like so:
`@import '~buttono/buttono'`

For some reason, even though this is standard, it doesn't seem to resolve the import without this helper method.

halfnibble: (BTW. I stole this method from gulp-core-build-sass).

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
halfnibble: Good point! It is now local to the method.

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
D4N14L: #Sorry, I should clarify. When does the 'stdin' value get used? The tilde makes sense

halfnibble: Oh! I have absolutely no idea. I copied that from gulp-core-build-sass. 

I can't think of a scenario, so I should probably just remove it.

halfnibble: Ok. It works great without the stdin statement. So I've removed it.

apps/heft/src/schemas/sass.schema.json
```diff
@@ -0,0 +1,58 @@
...
+    },
+
+    "exportAsInterfaceName": {
+      "type": "string",
+      "description": "The name of the export interface."
+    },
+
+    "fileExtensions": {
+      "type": "array",
+      "description": "Files with these extensions will pass through the Sass transpiler..",
```
iclanton, line 42:
```suggestion
      "description": "Files with these extensions will pass through the Sass transpiler.",
```

apps/heft/src/plugins/SassTypingsPlugin/postcss-modules.d.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+// Add missing Definitely Typed typings
+declare module 'postcss-modules';
```
halfnibble, line 5: @iclanton It seemed particularly difficult to import a package lacking a DefinitelyTyped package. Am I doing something wrong with importing/using postcss-modules (as far as typings goes)?

apps/heft/src/schemas/sass.schema.json
halfnibble: Fixed.

apps/heft/src/plugins/SassTypingsPlugin/postcss-modules.d.ts
iclanton: We turned on a strict compiler check option that's probably causing this, but yeah, this is expected.

apps/rush/CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 # Change Log - @microsoft/rush
```
iclanton, line 1: Revert changes to this file.

apps/rush/package.json
```diff
@@ -1,6 +1,6 @@
 {
```
iclanton, line 1: Revert.

build-tests/heft-sass-test/README.md
```diff
@@ -0,0 +1,15 @@
+# heft-webpack-basic-test
```
iclanton, line 1: Update name.

build-tests/heft-webpack-basic-test/README.md
```diff
@@ -1,4 +1,4 @@
-# heft-webpack-basic-test
+# heft-sass-test
```
iclanton, line 1: Revert this.

common/config/rush/version-policies.json
```diff
@@ -90,7 +90,7 @@
   {
     "policyName": "rush",
     "definitionName": "lockStepVersion",
-    "version": "5.34.2",
+    "version": "5.34.3",
```
iclanton, line 93: Revert.

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * @public
+ */
+export interface ISassConfiguration {
+  /**
+   * Source code root directory.
+   * Defaults to "src/".
+   */
+  srcFolder?: string;
```
iclanton, line 19: I don't think it makes sense for this to be customizable.

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -0,0 +1,158 @@
...
+   * Determines if export values are wrapped in a default property, or not.
+   * Defaults to true.
+   */
+  exportAsDefault?: boolean;
+
+  /**
+   * The name of the export interface.
+   * Defaults to "IExportStyles".
+   */
+  exportAsInterfaceName?: string;
```
iclanton, line 37: Would anyone ever need to customize this?

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * Files with these extensions will pass through the Sass transpiler.
+   * Defaults to ["sass", "scss", "css"]
+   */
+  fileExtensions?: string[];
```
iclanton, line 43: This doesn't seem super useful.

iclanton:
Can you add a template file for `scss.json` here: https://github.com/microsoft/rushstack/tree/master/apps/heft/src/templates

iclanton (changes requested):

iclanton:
Can you update the [web rig](https://github.com/microsoft/rushstack/tree/master/rigs/heft-web-rig) to include `sass.json`, and include the typings output location in the base tsconfig in that project?

apps/heft/src/utilities/CoreConfigFiles.ts
```diff
@@ -202,6 +204,21 @@ export class CoreConfigFiles {
     return CoreConfigFiles._typeScriptConfigurationFileLoader;
   }
 
+  public static get sassConfigurationFileLoader(): ConfigurationFile<ISassConfigurationJson> {
+    const schemaPath: string = path.resolve(__dirname, '..', 'schemas', 'sass.schema.json');
+    CoreConfigFiles._sassConfigurationFileLoader = new ConfigurationFile<ISassConfigurationJson>({
+      projectRelativeFilePath: 'config/sass.json',
+      jsonSchemaPath: schemaPath,
+      jsonPathMetadata: {
+        '$.includePaths.*': {
```
iclanton, line 213: Seems like the `generatedTsFolder` should also be resolved.

common/changes/@rushstack/heft/halfnibble-sass-typings-generator_2020-09-17-04-55.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft",
+      "comment": "Add new Sass typings plugin.",
```
iclanton, line 5:
```suggestion
      "comment": "Add functionality to automatically generate typings for *.scss, *.sass, and *.css files.",
```

common/reviews/api/typings-generator.api.md
```diff
@@ -10,6 +10,8 @@ import { Terminal } from '@rushstack/node-core-library';
 export interface IStringValuesTypingsGeneratorOptions extends ITypingsGeneratorOptions<IStringValueTypings> {
     // (undocumented)
     exportAsDefault?: boolean;
+    // (undocumented)
+    exportAsInterfaceName?: string;
```
iclanton, line 14: Shouldn't this be called `exportAsDefaultInterfaceName`?

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
halfnibble: This one I kind of disagree with. The tsconfig.json file allows developers to make any folder the root of their source code. It could be `main` or `app/src`, etc. Therefore, Sass developers need the same flexibility.

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
halfnibble: No one would ever need to, so I can remove this customization.

apps/heft/src/utilities/CoreConfigFiles.ts
halfnibble: Done.

apps/rush/CHANGELOG.md
halfnibble: Done.

build-tests/heft-sass-test/README.md
halfnibble: Done.

build-tests/heft-webpack-basic-test/README.md
halfnibble: Oops. Done.

common/reviews/api/typings-generator.api.md
halfnibble: Yes. Fixed.

libraries/typings-generator/src/StringValuesTypingsGenerator.ts
```diff
@@ -25,6 +25,7 @@ export interface IStringValueTypings {
  */
 export interface IStringValuesTypingsGeneratorOptions extends ITypingsGeneratorOptions<IStringValueTypings> {
   exportAsDefault?: boolean;
+  exportAsDefaultInterfaceName?: string;
```
iclanton, line 28: Can you add docs for these two?

iclanton (approved):

PR #2221, opened 2020-09-22
[terminal] Fix dependency on @rushstack/node-core-library

dmichon-msft:
Multiple files have a runtime, not just dev time dependency on @rushstack/node-core-library. Fixing the dependency to be a runtime dependency so that it can be successfully installed in pnpm/rush.


octogonz (approved):
Thank you for fixing this!

PR #2252, opened 2020-10-02
[heft] Throw if jest config file doesn't exist

mmkal:
Fixes https://github.com/microsoft/rushstack/issues/2209

CC @iclanton 

I couldn't find any low-level unit tests for `JestPlugin`, so I created a test file. It doesn't feel quite right but this isn't a blocker for me so I won't be able to put much more time into it I'm afraid. Feel free to push to the branch/close if you think a different approach would be better.


apps/heft/src/plugins/JestPlugin/JestPlugin.ts
```diff
@@ -22,17 +22,21 @@ export class JestPlugin implements IHeftPlugin {
...
-          await this._runJestAsync(heftSession, heftConfiguration, test);
-        });
-      });
+    const expectedConfigPath: string = path.join(heftConfiguration.buildFolder, JEST_CONFIGURATION_LOCATION);
 
-      heftSession.hooks.clean.tap(PLUGIN_NAME, (clean: ICleanStageContext) => {
-        this._includeJestCacheWhenCleaning(heftConfiguration, clean);
-      });
+    if (!FileSystem.exists(expectedConfigPath)) {
+      throw new Error(`Expected to find jest config file at ${expectedConfigPath}`);
```
iclanton, line 28: This should go inside the `_runJestAsync` function, and it should emit an error through the `ScopedLogger`.

iclanton (changes requested):

iclanton (approved):
Looks good. Thanks!

mmkal:
@iclanton it looks like this found a few projects which are not running tests. Not sure if that's known/expected, or what you think the best course of action is - use a previous heft version for those projects?

iclanton:
I just changed the verb from `test` to `build` for those projects.

PR #2276, opened 2020-10-13
Add additional property in sass schema to be compatible with sass.json in odsp-common

Claudiazhaoya:
The current sass.json schema doesn't have the property called 'useCSSModule', which is declared and used in sass.json in odsp-common. To get it compatible with odsp-common, I add this property in heft sass schema.

### DIFF
--- a/apps/heft/src/schemas/sass.schema.json
+++ b/apps/heft/src/schemas/sass.schema.json
@@ -39,6 +39,11 @@
         "type": "string",
         "pattern": "[^\\\\]"
       }
+    },
+
+    "useCSSModules": {
+      "type": "boolean",
+      "description": "Determines if CSS Modules are used, or not"
     }
   }
 }
--- a/common/changes/@rushstack/heft/zhas-add-property-in-sass-schema_2020-10-12-23-50.json
+++ b/common/changes/@rushstack/heft/zhas-add-property-in-sass-schema_2020-10-12-23-50.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft",
+      "comment": "Add additional property in sass schema",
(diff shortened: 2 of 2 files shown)


iclanton:
Adding this property to the schema won't do anything. Right now `sass.json` is only used for generating TypeScript declarations.

@halfnibble - how do you think we should handle turning CSS modules on and off? Does a property in `sass.json` make sense? Or should this be expected to be controlled by the webpack config, and if the webpack config is autogenerated by another plugin, this property end up in another config file?

halfnibble:
It looks like we have a coincidental naming collision in config files. I didn't realize our internal repo used the same `config/sass.json` config filename (must be a good name 😉).

The SassTypingsPlugin is really just a plugin for TypeScript. Because of this, we could consider moving what is currently Heft's `sass.json` to a part of the TypeScript config file. 

<ins>My first question is, are we:</ins>
A) Planning a full feature Heft plugin to handle the entire Sass transpilation, autoprefixing, etc. process.
or
B) Planning to let existing Webpack plugins handle everything Sass related except type checking.

I assumed we were going with option B. (although option A. might be cool further down the road). If Webpack plugins will continue to handle the majority of Sass processing, then the `useCSSModules` config would need to be accessible to the custom webpack config generator. 

<ins>This gives me an interesting idea...</ins>
It would be cool to have a project and/or monorepo wide custom config file that Heft parses and makes available to custom plugins. Then odsp-common could specify a schema for a file like `custom-config.json` that includes the `useCSSModules` property. And other open source tools and templates (e.g. `create-react-app`) could specify custom configs related to setting up their build system.

iclanton:
We've decided to take a different approach here (renaming the `sass.json` file in our internal repo), so I'm closing this PR out.

PR #2279, opened 2020-10-13
[Rush] Add .changeignore file support

### LINKED ISSUE #2263: [rush] exclude certain files from `rush change` diff calculation

- [x] Feature
- [ ] Bug

If a project edits a test file, and no production code, `rush change` demands that a changefile be generated even though there are no (meaningful) updates required.

Allow configuring files to ignore from the diff calculation. One option could be `.npmignore`, or to use the package.json `files` property. This may need to follow source maps to avoid false negatives for packages that don't include "source" typescript, though. It may be less complex to pass in directly, e.g. `rush change ... --ignore '**/*.test.ts' --ignore '**/*.md'`, or to put something like `"ignoreChanges": ["**/*.test.ts", "**/*.md"]` in rush.json.

* **Tool:** rush
* **Tool Version:** 5.34.3

mmkal:
Fixes #2263 

`rush change` will now look for a `.changeignore` file in each project with changed files. If it finds one, it will filter out files similarly to `.gitignore`, `.npmignore`, `.eslintignore` etc.

For example, to make sure test changes don't result in unwanted publishes, you could add `**/*.test.ts` to each project's `.changeignore` file.


apps/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -312,11 +317,11 @@ export class ChangeAction extends BaseRushAction {
   }
 
   private _getChangedPackageNames(): string[] {
-    const changedFolders: (string | undefined)[] | undefined = VersionControl.getChangedFolders(
+    const changedFiles: string[] = VersionControl.getChangedFiles(
```
mmkal, line 320: Switched to `getChangedFiles` since ignore patterns might apply to files or folders (e.g. `__tests__` or `**/*.test.ts`).

I'm not sure why the return type was `(string | undefined)[] | undefined` given `VersionControl.getChangedFolders` and `VersionControl.getChangedFiles` _always_ return an array of strings (it's just empty if there are no changes). So I updated it to `string[]`

apps/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -312,11 +317,11 @@ export class ChangeAction extends BaseRushAction {
...
       this._targetBranch,
       this._noFetchParameter.value
     );
-    if (!changedFolders) {
+    if (changedFiles.length === 0) {
```
mmkal, line 324: I think this was a logic error because of the nullable return type for `changeFolders` mentioned above. I suspect this optimisation was never hitting since `changedFolders` would always be truthy, even with no changes (`![] === false`). This should now skip the (potentially) expensive change-checking if no git changes are detected.

apps/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -354,14 +359,17 @@ export class ChangeAction extends BaseRushAction {
...
-      if (folder && Path.isUnderOrEqual(folder, projectFolder)) {
-        return true;
-      }
+  private _hasProjectChanged(changedFiles: string[], projectFolder: string): boolean {
+    const relevantFiles: string[] = changedFiles.filter((file) => Path.isUnderOrEqual(file, projectFolder));
+
+    const changeIgnoreFile: string = path.join(projectFolder, CHANGE_IGNORE_FILE_NAME);
+
+    const ignoreConfig: Ignore = ignore();
+    if (relevantFiles.length > 0 && FileSystem.exists(changeIgnoreFile)) {
```
mmkal, line 368: The `> 0` check is an optimisation: it avoids accessing the filesystem when there are no changes for this project (for most changes in large repos, this will be the case for almost every project). In that case, the `.some(...)` call below instantly returns false anyway.

mmkal (commented):
CC @iclanton - again I couldn't find where tests for this kind of thing live. I don't want to invent testing conventions so would be great to get some help/guidance re how you think this should be tested. It's pretty straightforward and the interesting parts are delegated to the `ignore` package, but still would be good to make sure it doesn't have some boolean error somewhere.

apps/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -311,26 +316,39 @@ export class ChangeAction extends BaseRushAction {
...
 
+    const getChangeIgnorePatterns: (fsPath: string) => string[] = memoize((fsPath: string): string[] => {
+      if ((repoRootFolder && Path.isUnder(repoRootFolder, fsPath)) || !FileSystem.exists(fsPath)) {
+        return [];
+      }
+      const parentIgnoreStrings: string[] = getChangeIgnorePatterns(path.dirname(fsPath));
+
+      const changeIgnoreFilePath: string = path.join(fsPath, RushConstants.changeignoreFileName);
+
+      return FileSystem.exists(changeIgnoreFilePath)
```
dmichon-msft, line 339:
```suggestion
      try {
        return [...parentIgnoreStrings, FileSystem.readFile(changeIgnoreFilePath);
      } catch (error) {
        if (FileSystem.isNotExistsError(error)) {
          return parentIgnoreStrings;
        }
        throw error;
      }
```

Uses half as many file system operations, since `FileSystem.exists()` is internally just a try/catch around an `fs.stat` call.

apps/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -311,26 +316,39 @@ export class ChangeAction extends BaseRushAction {
...
+        ? [...parentIgnoreStrings, FileSystem.readFile(changeIgnoreFilePath)]
+        : parentIgnoreStrings;
+    });
+
     this.rushConfiguration.projects
```
dmichon-msft, line 344: I recommend inverting the lookup:
```ts
const changedFilesByProject: Map<RushConfigurationProject, string[]> = new Map();

for (const file of changedFiles) {
  const relevantProject: RushConfigurationProject | undefined = this.rushConfiguration.findProjectByPosixRelativePath(file);
  if (!relevantProject) {
    continue;
  }
  let projectFiles: string[] | undefined = changedFilesByProject.get(relevantProject);
  if (!projectFiles) {
    if (!project.shouldPublish || project.versionPolicy && project.versionPolicy.exemptFromRushChange) {
      continue;
    }
    projectFiles = [];
    changedFilesByProject.set(relevantProject, projectFiles);
  }

  // Potentially test the file here
  projectFiles.push(file);
}

for (const [project, relevantFiles] of changedFilesByProject) {
  // Do your ignorer config processing and filtering here, and find out what is left.
  // Might consider recording which files prompted the change for posterity, so that we can have a --verbose option or similar for rush change that logs which files required the change file.
}
```

apps/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -311,26 +316,39 @@ export class ChangeAction extends BaseRushAction {
...
+    if (changedFiles.length === 0) {
       return [];
     }
     const changedPackageNames: Set<string> = new Set<string>();
 
     const repoRootFolder: string | undefined = VersionControl.getRepositoryRootPath();
     const projectHostMap: Map<string, string> = this._generateHostMap();
 
+    const getChangeIgnorePatterns: (fsPath: string) => string[] = memoize((fsPath: string): string[] => {
+      if ((repoRootFolder && Path.isUnder(repoRootFolder, fsPath)) || !FileSystem.exists(fsPath)) {
```
dmichon-msft, line 332: Be advised that, as currently implemented, this function is hideously expensive.

dmichon-msft, line 332: This will skip checking the entire tree for a deleted folder, which means that if said folder was supposed to be ignored for change detection, it will no longer be. I recommend skipping the check.

apps/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -311,26 +316,39 @@ export class ChangeAction extends BaseRushAction {
...
+        return [];
+      }
+      const parentIgnoreStrings: string[] = getChangeIgnorePatterns(path.dirname(fsPath));
```
dmichon-msft, line 335: Aren't ignore patterns defined relative to the file that defines them?

apps/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -353,14 +371,20 @@ export class ChangeAction extends BaseRushAction {
...
+  private _hasProjectChanged(
+    changedFiles: string[],
+    projectFolder: string,
+    getChangeIgnorePatterns: (fsPath: string) => string[]
+  ): boolean {
+    const relevantFiles: string[] = changedFiles.filter((file) => Path.isUnderOrEqual(file, projectFolder));
+
+    const ignoreConfig: Ignore = ignore();
+
+    for (const ignorePattern of flatMap(relevantFiles.map(path.dirname), getChangeIgnorePatterns)) {
```
dmichon-msft, line 383: It's almost certainly worth deduplicating intermediary values to improve on performance.
```
const foldersToConsider: Set<string> = new Set(relevantFiles.map(path.dirname));
for (const folder of foldersToConsider) {
  const parent: string = path.dirname(folder);
  if (/* is parent in project */) {
    foldersToConsider.add(parent);
  }

  // Can lookup the local ignore file and add the patterns here, including relevant prefixing.
}

dmichon-msft (commented):
I have performance concerns with the change as currently implemented. `Path.isUnderOrEqual` and `Path.isUnder` are currently hideously expensive, and this change causes the same (total changed files) * (number of projects) performance relationship that PackageChangeAnalyzer was experiencing until I refactored it to use the new `findProjectByPosixRelativePath` function to linearize the relationship.

It might be fine in a small repo, but it'll have very visible scaling influences for large PRs (e.g. bulk file moves) in a larger monorepo.

PR #2282, opened 2020-10-14
[Heft] Allow files to be ignored for typings generation in sass.json

iclanton:


apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -35,7 +35,12 @@ export interface ISassConfiguration {
    * The paths should be relative to the project root.
    * Defaults to ["node_modules", "src"]
    */
-  includePaths?: string[];
+  importPaths?: string[];
+
+  /**
+   * A list of file paths relative to the "src" folder that should be excluded from typings generation.
+   */
```
Claudiazhaoya, line 42: Probably better to use excludePaths to align with importPaths?

Claudiazhaoya (approved):

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
iclanton: This has different behavior than `importPaths`. I wanted to changes the name of that one to make it clearly different. Maybe `importPaths` isn't the right name.

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -109,14 +114,14 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
...
-    includePaths: string[] | undefined
+    importPaths: string[] | undefined
   ): Promise<string> {
     const result: Result = await LegacyAdapters.convertCallbackToPromise(render, {
       data: fileContents,
       file: filePath,
       importer: (url: string) => ({ file: this._patchSassUrl(url) }),
-      includePaths: includePaths
-        ? includePaths
+      includePaths: importPaths
```
D4N14L, line 123: I agree with the change to label it "importPaths", but will this cause confusion for devs used to providing "includePaths" to `node-sass`? Just thinking out loud if the breaking change is the right call here.

D4N14L (approved):

PR #2298, opened 2020-10-18
[rush] Provide override for package.json script executed by bulk commands

ifeanyiecheruo:
Proof of concept solution for #2283 (Support customizing CI builds from Dev builds)

Adds a script configuration option for bulk commands that specifies
the package.json script to execute (falls-back to the bulk action name)

If an array of names is specified the first name that matches a scripts
entry in package.json will be executed (falls-back to the bulk action name).

This change also
- updates the default configuration for the builtin build command to have
  a script entry of "rush-build"
- updates the default configuration of the built in rebuild command to have
  a script entry of ["rush-build", "build"]

This enables a scenario where...
- a package can specify what happens for rush build
- a package can specify what happens rushx build (clean, build)
Without the two behaviors conflicting

```json
"script": {
  "rush-build": "heft test --clean",
}
```

### DIFF
--- a/apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
+++ b/apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
@@ -31,15 +31,14 @@ import { IRushConfigurationProjectJson } from '../../api/RushConfigurationProjec
  */
 export interface IBulkScriptActionOptions extends IBaseScriptActionOptions {
   enableParallelism: boolean;
+  /**
+   * Optional package.json script to run otherwise use `actionName` as the script to run..
+   */
+  script?: string | string[];
   ignoreMissingScript: boolean;
   ignoreDependencyOrder: boolean;
   incremental: boolean;
   allowWarningsInSuccessfulBuild: boolean;
-
-  /**
-   * Optional command to run. Otherwise, use the `actionName` as the command to run.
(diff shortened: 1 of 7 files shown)


apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -118,7 +124,7 @@ export class BulkScriptAction extends BaseScriptAction {
       isIncrementalBuildAllowed: this._isIncrementalBuildAllowed,
       ignoreMissingScript: this._ignoreMissingScript,
       ignoreDependencyOrder: this._ignoreDependencyOrder,
-      packageDepsFilename: Utilities.getPackageDepsFilenameForCommand(this._commandToRun)
+      packageDepsFilename: Utilities.getPackageDepsFilenameForCommand(this._commandToRun[0])
```
dmichon-msft, line 127: Ideally this should use the name of the command that actually executed, in case there is a collision between multiple Rush commands.

dmichon-msft (commented):
I like this in general, just a minor note about the package deps file name for cacheability.

halimb:
@iclanton this PR is referenced in the official documentation: https://rushjs.io/pages/advanced/watch_mode/#the-watchforchanges-setting-experimental

PR #2305, opened 2020-10-21
[heft] Add typings options

halfnibble:
This PR makes 2 changes.

1. Supports `undefined` as a value returned by `parseFileAndGenerateTypingsAsync`. When `undefined` is returned, no typings file is generated. 
2. Adds the `"fileExtensions"` configuration option to the SassTypingsGenerator. This is so repos that only want to process Sass Modules may do so (e.g. by configuring extensions as `["module.sass", "module.scss"]`


apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -110,6 +119,10 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
     });
   }
 
+  private _isSassPartial(filePath: string): boolean {
+    return path.basename(filePath)[0] === '_';
```
D4N14L, line 123: Maybe throw a comment in here that this is always expected to be the case, to clarify why we do this to future devs seeing the code

libraries/typings-generator/src/TypingsGenerator.ts
```diff
@@ -124,9 +124,16 @@ export class TypingsGenerator {
...
     try {
       const fileContents: string = await FileSystem.readFileAsync(locFilePath);
-      const typingsData: string = await this._options.parseAndGenerateTypings(fileContents, locFilePath);
+      const typingsData: string | undefined = await this._options.parseAndGenerateTypings(
+        fileContents,
+        locFilePath
+      );
       const generatedTsFilePath: string = this._getTypingsFilePath(locFilePath);
 
+      if (typingsData === undefined) {
```
D4N14L, line 133: Comment what cases this would return undefined and expected behaviour

D4N14L (approved):

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -30,6 +30,12 @@ export interface ISassConfiguration {
    */
   exportAsDefault?: boolean;
 
+  /**
+   * Files with these extensions will pass through the Sass transpiler for typings generation.
+   * Defaults to ["sass", "scss", "css"]
```
iclanton, line 35: For consistency with [this](https://github.com/microsoft/rushstack/blob/master/apps/heft/src/schemas/typescript.schema.json#L68-L74), let's require that these start with a `.`

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
```diff
@@ -110,6 +119,10 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
     });
   }
 
+  private _isSassPartial(filePath: string): boolean {
```
iclanton, line 122: Can you include a doc comment describing what a sass partial is?

apps/heft/src/schemas/sass.schema.json
```diff
@@ -32,6 +32,15 @@
       "description": "Determines if export values are wrapped in a default property, or not."
     },
 
+    "fileExtensions": {
+      "type": "array",
+      "description": "Files with these extensions will pass through the Sass transpiler for typings generation.",
+      "items": {
+        "type": "string",
+        "pattern": "[^\\\\]"
```
iclanton, line 40: Let's update this pattern to something like

```suggestion
        "pattern": "^\\.[A-z0-9-._]*[A-z0-9-]+$"
```

This says: 
- Must start with a `.`
- Other characters must be alphanumeric, `-`, `_`, or `.` (except for the last character)

That doesn't cover 100% of legal file extensions, but it should be pretty close.

It'd also be good to update [this expression](https://github.com/microsoft/rushstack/blob/master/apps/heft/src/schemas/typescript.schema.json#L73). Might be a good idea to extract this to a resource in a shared schema file.

apps/heft/src/templates/sass.json
```diff
@@ -23,6 +23,15 @@
    */
   // "exportAsDefault": true,
 
+  /**
+   * Files with these extensions will pass through the Sass transpiler for typings generation.
+   */
+  // "fileExtensions": [
+  //   "sass",
+  //   "scss",
+  //   "css
```
iclanton, line 32:
```suggestion
  //   ".sass",
  //   ".scss",
  //   ".css
```

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
iclanton: Can you also make sure that works with `typings-generator`. Maybe we even require that in `typings-generator`?

iclanton (changes requested):

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
halfnibble: @iclanton Hello Ian, I think I've addressed all your feedback.

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
halfnibble: @D4N14L Hello Daniel, I think I've addressed all your feedback.

apps/heft/src/plugins/SassTypingsPlugin/SassTypingsGenerator.ts
halfnibble: The TypingsGenerator class appears to have logic to test for the `.` and add it if it's missing. 
Everything seems to work fine after adding the `.`

D4N14L (approved):

apps/heft/src/schemas/sass.schema.json
```diff
@@ -32,6 +32,15 @@
       "description": "Determines if export values are wrapped in a default property, or not."
     },
 
+    "fileExtensions": {
+      "type": "array",
+      "description": "Files with these extensions will pass through the Sass transpiler for typings generation.",
+      "items": {
+        "type": "string",
+        "pattern": "^\\.[A-z0-9-._]*[A-z0-9-]+$"
```
iclanton, line 40:
```suggestion
        "pattern": "^\\.[A-z0-9-_.]*[A-z0-9-_]+$"
```

apps/heft/src/schemas/typescript.schema.json
```diff
@@ -70,7 +70,7 @@
           "description": "File extensions that should be copied from the src folder to the destination folder(s)",
           "items": {
             "type": "string",
-            "pattern": "\\..+"
+            "pattern": "^\\.[A-z0-9-._]*[A-z0-9-]+$"
```
iclanton, line 73:
```suggestion
            "pattern": "^\\.[A-z0-9-_.]*[A-z0-9-_]+$"
```

iclanton (approved):

PR #2318, opened 2020-10-28
Use source-map-loader to map back to .ts files in webpack tutorial

ifeanyiecheruo:
One possible solution for issue #2316

Configures the heft-webpack-basic-tutorial project to use [source-map-loader](https://www.npmjs.com/package/source-map-loader) to resolve the transitive closure of mappings from the .js files in ./dist to the .ts files in ./src

Note this falls apart if the project has a dependency on another rush project.
source-map-loader has a strong assumption that its webpack config is in a common ancestor folder of all the traversed source maps leading to the wrong relative paths being generated for pretty much all source files discovered.

using `devtool: 'eval-source-map'` helps unblock the above issue by embedding the source map in generated bundles so that they can always be found but still leaves some odd quirks. e.g browsing source files in Chrome reveals the poorly generated paths

### DIFF
--- a/common/config/rush/browser-approved-packages.json
+++ b/common/config/rush/browser-approved-packages.json
@@ -3,19 +3,11 @@
   "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/approved-packages.schema.json",
   "packages": [
     {
-      "name": "@rushstack/heft-web-rig",
-      "allowedCategories": [ "libraries", "tests" ]
-    },
-    {
       "name": "react",
       "allowedCategories": [ "tests" ]
     },
     {
       "name": "react-dom",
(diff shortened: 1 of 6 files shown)


common/config/rush/browser-approved-packages.json
```diff
@@ -13,6 +13,10 @@
     {
       "name": "react-dom",
       "allowedCategories": [ "tests" ]
+    },
+    {
+      "name": "source-map-loader",
+      "allowedCategories": [ "tests" ]
```
iclanton, line 19: This should go in `nonbrowser-approved-packages.json`

iclanton (approved):

PR #2319, opened 2020-10-28
Update Webpack to 4.44.2

iclanton:


webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -198,10 +198,7 @@ export class LocalizationPlugin implements Webpack.Plugin {
 
     if (isWebpackDevServer) {
       if (typingsPreprocessor) {
-        compiler.hooks.afterEnvironment.tapPromise(
-          PLUGIN_NAME,
-          async () => await typingsPreprocessor!.runWatcherAsync()
-        );
+        compiler.hooks.afterEnvironment.tap(PLUGIN_NAME, () => typingsPreprocessor!.runWatcherAsync());
```
halfnibble, line 201: Is this no longer async, or do we just not need to wait for the promise resolution?

halfnibble (approved):

webpack/localization-plugin/src/LocalizationPlugin.ts
iclanton: `afterEnvironment` is apparently sync (not sure how this worked before honestly), so here we're just firing-and-forgetting `runWatcherAsync`.

webpack/localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -238,7 +235,7 @@ export class LocalizationPlugin implements Webpack.Plugin {
                   assetPath.match(Constants.LOCALE_FILENAME_TOKEN_REGEX)
                 ) {
                   // Does this look like an async chunk URL generator?
-                  if (typeof options.chunk.id === 'string' && options.chunk.id.match(/^\" \+/)) {
+                  if (typeof options.chunk.id === 'string' && (options.chunk.id as string).match(/^\" \+/)) {
```
dmichon-msft, line 238: We shouldn't need this cast. Maybe need to store `options.chunk.id` in a local `const` so the resolver can guarantee it didn't change?

dmichon-msft (approved):
Just one minor note, otherwise looks good

webpack/localization-plugin/src/LocalizationPlugin.ts
iclanton: The typings here are wrong. It's marked as a `number` in the typings, so we need to explicitly cast this here because otherwise its inferred type is `never`.

PR #2335, opened 2020-11-03
[rush-lib] Fix bug where version process is using a wrong `git.addChanges` signature

nikuph:
When calling `rush version --bump` it fails with:
```
* EXECUTING: git add . (common/config/rush/version-policies.json)

ERROR: spawnSync /bin/sh ENOTDIR
```

This is caused by mixing up `workingDirectory` and `pathspec`.

### DIFF
--- a/apps/rush-lib/src/cli/actions/VersionAction.ts
+++ b/apps/rush-lib/src/cli/actions/VersionAction.ts
@@ -227,7 +227,7 @@ export class VersionAction extends BaseRushAction {
     });
 
     if (changeLogUpdated) {
-      git.addChanges('.', this.rushConfiguration.changesFolder);
+      git.addChanges(this.rushConfiguration.changesFolder, '.');
       git.addChanges(':/**/CHANGELOG.json');
       git.addChanges(':/**/CHANGELOG.md');
       git.commit('Deleting change files and updating change logs for package updates.');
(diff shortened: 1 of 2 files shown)


apps/rush-lib/src/cli/actions/VersionAction.ts
```diff
@@ -227,7 +227,7 @@ export class VersionAction extends BaseRushAction {
     });
 
     if (changeLogUpdated) {
-      git.addChanges('.', this.rushConfiguration.changesFolder);
+      git.addChanges(this.rushConfiguration.changesFolder, '.');
```
D4N14L, line 230: This is the default, so you can just cut the last param off here

D4N14L: Actually, this one is still correct. We are trying to add every change in the folder. Could you revert this change?

apps/rush-lib/src/cli/actions/VersionAction.ts
```diff
@@ -239,7 +239,7 @@ export class VersionAction extends BaseRushAction {
     });
 
     if (packageJsonUpdated) {
-      git.addChanges('.', this.rushConfiguration.versionPolicyConfigurationFilePath);
+      git.addChanges(this.rushConfiguration.versionPolicyConfigurationFilePath, '.');
```
D4N14L, line 242: This is the default, so you can just cut the last param off here

D4N14L (changes requested):

apps/rush-lib/src/cli/actions/VersionAction.ts
nikuph: Sorry, too tired 🙈

Mixed it up here and thought it's the same issue as below.

apps/rush-lib/src/cli/actions/VersionAction.ts
nikuph: Removed the last param 👍

D4N14L (approved):
Looks good! Thanks for your contribution!

iclanton:
This has been released in Rush v5.35.2.

PR #2340, opened 2020-11-06
[heft] Improve error handling and make --debug print stacks of errors that occur in heft's internal initialization.

iclanton:


apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -141,38 +158,41 @@ export class HeftToolsCommandLineParser extends CommandLineParser {
   }
 
   public async execute(args?: string[]): Promise<boolean> {
-    this._terminalProvider.verboseEnabled = this.isDebug;
+    // Defensively set the exit code to 1 so if the tool crashes for whatever reason, we'll have a nonzero exit code.
+    process.exitCode = 1;
```
D4N14L, line 162: Should this get set back to 0 before exiting?

iclanton: It'll get set back to 0 at the end of `onExecute`. See here: https://github.com/microsoft/rushstack/pull/2340/files#diff-6b764ccba8a9718f67ad50c0858e8d8cbc5098442c98a48acdefaea13dbaf131R221

D4N14L (approved):

PR #2361, opened 2020-11-17
[rush] Fix an issue where local peer dependencies were not considered when determining build order

kiltyj:
## Summary
With Rush version 5.30.0 and later, we started seeing build failures in our repository, which includes packages that list other packages in our repository as peer dependencies. The build failures were related to packages building out of dependency order due to these local peer dependency relationships not being considered. This change adds peer dependencies to the list of local dependencies that are considered for determining package build order.

## Details
This change fixes our build, and seems like a safe and correct fix to me, but I acknowledge that I could be missing a valid reason for excluding local peer dependencies. However, given that our builds were working prior to 5.30.0 this seems like it might have been an inadvertent change.

## How it was tested
I manually invoked "rush rebuild" with a fresh version of our repository containing multiple local peer relationships to confirm that our build failures were no longer occurring.

### DIFF
--- a/apps/rush-lib/src/api/RushConfigurationProject.ts
+++ b/apps/rush-lib/src/api/RushConfigurationProject.ts
@@ -222,7 +222,8 @@ export class RushConfigurationProject {
       this._localDependencyProjects = [
         ...this._getLocalDependencyProjects(this.packageJson.dependencies),
         ...this._getLocalDependencyProjects(this.packageJson.devDependencies),
-        ...this._getLocalDependencyProjects(this.packageJson.optionalDependencies)
+        ...this._getLocalDependencyProjects(this.packageJson.optionalDependencies),
+        ...this._getLocalDependencyProjects(this.packageJson.peerDependencies)
       ];
     }
     return this._localDependencyProjects;
--- a/common/changes/@microsoft/rush/fix-local-peer-dep-build-order_2020-11-17-22-35.json
+++ b/common/changes/@microsoft/rush/fix-local-peer-dep-build-order_2020-11-17-22-35.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix an issue where local peer dependencies were not considered when determining build order",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush",
+  "email": "kiltyj@users.noreply.github.com"
+}


iclanton:
Let's say you have a build graph that look like this:

```
(A)<-🍐-(C)
 |     /
 |    /
 |   /
 |  /
 v v
 (B)
```

And let's say project `(A)` has a peerDependency (🍐) on project `(C)`. In this repo, project `(A)`'s peerDependency needs to be satisfied with project `(B)` taking a dependency on project `(C)`. If you build this repo, `(C)` will be built before `(B)`, so anything that `(A)` imports from `(C)` at runtime should be satisfied by the time `(B)` is built.

If `(A)` needs access to `(C)` at build time (for example, if `(A)` uses TypeScript typings information from `(C)`), `(A)` should take a devDependency in addition to its peerDependency on `(C)`.

kiltyj:
Thanks for the response, @iclanton. I think I see now. Adding these peer dependencies as dev dependencies does fix our build without the change in this PR.

Trying to come up with cases where a peer dependency should not be a dev dependency, the only case I can think of is if (using your example build graph) project `(A)` expects project `(C)` to mutate some existing object, e.g. a global like `window`, or some object passed into project `(A)`'s API by project `(B)`. In that case, though (assuming no typings, etc. are needed) the dependency is only at runtime, so there shouldn't be an issue building without the dev dependency declared. 

Generally speaking, it seems like if a peer dependency needs to be imported, it should also be a dev dependency. Does this seem correct?

iclanton:
>Generally speaking, it seems like if a peer dependency needs to be imported, it should also be a dev dependency. Does this seem correct?

That's the approach we've taken, and it's worked well for us so far. As an example, take a look at the [`package.json` file for the `@rushstack/localization-plugin` project](https://github.com/microsoft/rushstack/blob/master/webpack/localization-plugin/package.json) in this repo: notice that the dependency on `@rushstack/set-webpack-public-path-plugin` is listed under both `peerDependencies` and `devDependencies`. This is actually an interesting case in that this is an optional peer dependency, so we build `@rushstack/localization-plugin` with access to TypeScript typing information from `@rushstack/set-webpack-public-path-plugin`, but at runtime we try to resolve it and don't fail if it isn't provided.

D4N14L:
> Generally speaking, it seems like if a peer dependency needs to be imported, it should also be a dev dependency. Does this seem correct?

For the most part, this is true. The only time it's not is if, for example, `(A)` needs the dependency at runtime, since there's nothing that _strictly guarantees_ that `(B)` will provide this dependency (peer dependencies are just a constraint imposed by the package manager, and they can choose to enforce this constraint or not, ex. npm v1 and v2 used to automatically install these peer deps if not satisfied, but since then will [simply warn the dev](https://nodejs.org/es/blog/npm/peer-dependencies/)).

Though in this case, I can't really see a use-case for the peer dependency other than specifying somewhat broad compatibility constraints for APIs being exported by the package, as the package will be consuming the version specified in dependencies regardless of what version `(B)` is using.

kiltyj:
Makes sense, @iclanton & @D4N14L. Thanks so much for taking the time to respond! Closing this and correcting our dev dependencies.

Issue #2360, opened 2020-11-17
Add this commit message to rush.json-> gitPolicy


jaboko:
https://github.com/microsoft/rushstack/blob/decfa7ba4a31b65e407f0cdd52ebda50a0b02c0a/apps/rush-lib/src/cli/actions/VersionAction.ts#L232

I think you should put this message in the rush.json parameter like versionBumpCommitMessage to allow you to add [skip ci] to the message.

Thanks!

iclanton:
That seems pretty reasonable. Would you be interested in opening a PR to make this change?

jaboko:
> That seems pretty reasonable. Would you be interested in opening a PR to make this change?

Yes, I will do it soon

PR #2369, opened 2020-11-22
[rush-lib] Added 'rush version' commit message to rush.json-> gitPolicy

jaboko:
## Summary
Issue #2360. Added variable commit message `changeLogUpdateCommitMessage` in rush.json -> gitPolicy for `rush version` like `versionBumpCommitMessage` 

## Details
At the moment, if I call the `rush version` in Azure Devops Pipeline, the message without [skip ci] will immediately commit and push to the Git. This causes re-build and may lead to build cycle if no additional check is done.

I have added the ability to edit this message by analogy with the `versionBumpCommitMessage`

**_Question_**
Do I need to add the parameter in `apps\rush-lib\assets\rush-init\rush.json`? Because when I add the parameter, I get a json validation error.

### DIFF
--- a/common/changes/@microsoft/rush/issue-2360-add-rush-version-commit-message_2020-11-22-22-01.json
+++ b/common/changes/@microsoft/rush/issue-2360-add-rush-version-commit-message_2020-11-22-22-01.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Add 'rush version' commit message to rush.json-> gitPolicy",
+      "comment": "Add the ability to customize the commit message used when \"rush version\" is run.",
       "type": "none"
     }
   ],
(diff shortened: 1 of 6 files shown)


common/changes/@microsoft/rush/issue-2360-add-rush-version-commit-message_2020-11-22-22-01.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Add 'rush version' commit message to rush.json-> gitPolicy",
```
iclanton, line 5:
```suggestion
      "comment": "Add the ability to customize the commit message used when \"rush version\" is run.",
```

iclanton:
>Do I need to add the parameter in apps\rush-lib\assets\rush-init\rush.json? Because when I add the parameter, I get a json validation error.

Yeah, you should add the parameter there. What's the error that you get?

jaboko:
> > Do I need to add the parameter in apps\rush-lib\assets\rush-init\rush.json? Because when I add the parameter, I get a json validation error.
> 
> Yeah, you should add the parameter there. What's the error that you get?

There was a json validation error. This is because this parameter is not in the schematic.
I added a parameter to rush-init/rush.json.

iclanton (approved):

PR #2379, opened 2020-12-04
[heft] Ensure rootDir is consistently specified in tsconfig. 

halfnibble:
## Summary

Ensure the "rootDir" setting, which sets the root source directory and can impact output folder structure, is consistently specified in all tsconfig.json files, including build rig profiles. 

In places where "rootDirs" virtual directories are necessary for typings folders, ensure the source directory is specified both in "rootDirs" and "rootDir".

## Details

Basically, through trial and error, I determined that not setting the "rootDir" property can effect the output folder structure if the root source directory does not contain any TypeScript files. It appears that "rootDir" and "rootDirs" are actually not closely related and "src" needs to be in "rootDir" at a minimum. The "src" directory must also be in "rootDirs" if that setting is added to specify one or more virtual directories (e.g. for "loc-json-ts" and "sass-ts"). 

## How it was tested

On a work project. Plus it builds.

### DIFF
--- a/common/changes/@microsoft/rush-stack-compiler-2.4/halfnibble-ensure-rootdir_2020-12-04-23-00.json
+++ b/common/changes/@microsoft/rush-stack-compiler-2.4/halfnibble-ensure-rootdir_2020-12-04-23-00.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush-stack-compiler-2.4",
+      "comment": "Ensure rootDir is consistently specified. ",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@microsoft/rush-stack-compiler-2.4",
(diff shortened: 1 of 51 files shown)


common/changes/@microsoft/rush-stack-compiler-2.4/halfnibble-ensure-rootdir_2020-12-04-23-00.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush-stack-compiler-2.4",
+      "comment": "Ensure rootDir is consistently specified. ",
```
iclanton, line 5:
```suggestion
      "comment": "Ensure rootDir is consistently specified.",
```

iclanton (approved):

PR #2383, opened 2020-12-07
[set-webpack-public-path-plugin] Remove uglify from set-webpack-public-path-plugin

iclanton:


webpack/set-webpack-public-path-plugin/src/codeGenerator.ts
```diff
@@ -141,31 +138,18 @@ export function getSetPublicPathCode(
...
-  } else {
-    const minifyOutput: uglify.MinifyOutput = uglify.minify(joinedScript, {
-      compress: {
-        dead_code: true
-      }
-    });
-
-    return `${EOL}${minifyOutput.code}`;
-  }
+  // Minified version of this code:
```
D4N14L, line 141: If you think this will ever need to be modified, maybe throw some steps in to follow to get to the minified code? Otherwise not a big deal

D4N14L (approved):

webpack/set-webpack-public-path-plugin/src/codeGenerator.ts
iclanton: Minifying code is pretty straightforward. I'm not really worried about that.

PR #2411, opened 2020-12-30
[rush] Make setting the credential env variable imply writing is allowed.

iclanton:
This PR renames the `RUSH_BUILD_CACHE_CREDENTIAL` env variable to `RUSH_BUILD_CACHE_WRITE_CREDENTIAL`, and make setting it allow cache write.

The motivation here is that a CI machine can set this env variable to fill the cache that users have read-only access to.


apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
```diff
@@ -68,7 +73,7 @@ export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase
         this._storageContainerName
       ];
 
-      if (this.isCacheWriteAllowed) {
+      if (this._isCacheWriteAllowedByConfiguration) {
```
D4N14L, line 76: This does not take the cred into consideration

apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
```diff
@@ -197,15 +202,15 @@ export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase
       if (sasString) {
         const connectionString: string = this._getConnectionString(sasString);
         blobServiceClient = BlobServiceClient.fromConnectionString(connectionString);
-      } else if (!this.isCacheWriteAllowed) {
+      } else if (!this._isCacheWriteAllowedByConfiguration) {
```
D4N14L, line 205: Is it possible to have a read-only cred? My guess is not, if we're assuming the existence of a cred implies write access. But if it does, then how do we differentiate? If we have a read-only cred and set "write allowed" to false, would we be able to use the cred to connect and read?

iclanton: You can, but something has to hydrate the cache, so I'd think the scenario we push users toward would be CI setting a write credential in an env variable.

If, for some reason, you want a read credential in CI, you can run the `rush update-cloud-credentials` command.

octogonz (approved):

PR #2414, opened 2021-01-06
[typings-generator] Keep typings folder during run watch

halfnibble:
## Summary

Instead of ensuring an empty folder when generating typings in watch mode, just ensure the folder exists.

## Details

This is to prevent typings being erased after preCompile and while TSC is compiling in the compile phase. 

## How it was tested

Very carefully.

### DIFF
--- a/common/changes/@rushstack/typings-generator/halfnibble-keep-typings-dir-watch_2021-01-06-00-13.json
+++ b/common/changes/@rushstack/typings-generator/halfnibble-keep-typings-dir-watch_2021-01-06-00-13.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/typings-generator",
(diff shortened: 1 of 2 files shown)


iclanton (approved):

common/changes/@rushstack/typings-generator/halfnibble-keep-typings-dir-watch_2021-01-06-00-13.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/typings-generator",
+      "comment": "Do not empty typings folder when running in watch mode.",
+      "type": "patch"
```
iclanton, line 6: Let's make this a minor bump as this is technically a breaking change.

PR #2415, opened 2021-01-06
[rush] Allow the cache entry name to be configured.

iclanton:
This PR introduces a `cacheEntryNamePattern` to `common/config/rush/build-cache.json` to allow the cache entry IDs to be configured.

This field allows for patterns containing alphanumeric characters, slashes, underscores, and hyphens and tokens `[hash]` (which is replaced with the project state hash), `[projectName]` (which is replaced with the project's package name), and `[projectName:normalized]` (which is replaced with the project's package name, but with `/` characters replaced with `+` characters).

### DIFF
--- a/apps/rush-lib/src/logic/buildCache/CacheEntryId.ts
+++ b/apps/rush-lib/src/logic/buildCache/CacheEntryId.ts
@@ -13,6 +13,9 @@ export type GetCacheEntryIdFunction = (options: IGenerateCacheEntryIdOptions) =>
 const HASH_TOKEN_NAME: string = 'hash';
 const PROJECT_NAME_TOKEN_NAME: string = 'projectName';
 
+// This regex matches substrings that look like [token]
+const TOKEN_REGEX: RegExp = /\[[^\]]*\]/g;
+
 export class CacheEntryId {
   private constructor() {}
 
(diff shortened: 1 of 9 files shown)


apps/rush-lib/src/schemas/build-cache.schema.json
```diff
@@ -27,7 +39,9 @@
             "cacheProvider": {
               "type": "string",
               "enum": ["local-only"]
-            }
+            },
+
+            "cacheEntryNamePattern": { "$ref": "#/definitions/anything" }
```
D4N14L, line 44: Confused... why is this the `anything` definition and not a string as above?

apps/rush-lib/src/logic/buildCache/CacheEntryId.ts
```diff
@@ -0,0 +1,144 @@
...
+const HASH_TOKEN_NAME: string = 'hash';
+const PROJECT_NAME_TOKEN_NAME: string = 'projectName';
+
+export class CacheEntryId {
+  private constructor() {}
+
+  public static parsePattern(pattern?: string): GetCacheEntryIdFunction {
+    if (!pattern) {
+      return ({ projectStateHash }) => projectStateHash;
+    } else {
```
D4N14L, line 22: While I want to say that this feels like a lot of effort to parse a few known tokens... The state machine does set it up to be easily expanded in the future. My only other thought would be a very tight regex with capture groups for the `[blah]` tokens, replacing the capture groups, and then validating the re-composed string as a whole as you already do.

D4N14L (approved):

apps/rush-lib/src/schemas/build-cache.schema.json
iclanton: Simplicity. We've done this in other places too. For example: https://github.com/microsoft/rushstack/blob/master/apps/rush-lib/src/schemas/command-line.schema.json#L90-L100

Because each of these has a `"additionalProperties": false`, we need to include dummy entries for all of the shared fields.

apps/rush-lib/src/logic/buildCache/CacheEntryId.ts
iclanton: Yeah that is simpler.

Issue #2345, opened 2020-11-09
[rush] Version Policies


matteobruni:
## Please prefix the issue title with the project name i.e. [rush], [api-extractor] etc. ##

**Is this a feature or a bug?**

- [ ] Feature
- [x] Bug

**Please describe the actual behavior.**

I have this version policies file

```json
[
  {
    "definitionName": "individualVersion",
    "policyName": "1.19-prerelease",
    "version": "1.18.11",
    "nextBump": "prerelease"
  },
  {
    "definitionName": "individualVersion",
    "policyName": "2.2-prerelease",
    "version": "2.1.11",
    "nextBump": "prerelease"
  },
  {
    "definitionName": "individualVersion",
    "policyName": "1.2-prerelease",
    "version": "1.1.11",
    "nextBump": "prerelease"
  },
  {
    "definitionName": "individualVersion",
    "policyName": "1.4-prerelease",
    "version": "1.3.11",
    "nextBump": "prerelease"
  }
]
```

But when I run `rush version` it returns me this error

```shell
ERROR: JSON validation failed:
[omissis]/tsparticles/common/config/rush/version-policies.json
Error: #/3
Data does not match any schemas from 'oneOf'
Error: #/3/definitionName (The name of version policy definition)
No enum match for: individualVersion
Error: #/3 (Lockstep version policy)
Additional properties not allowed: nextBump,version
Error: #/2
Data does not match any schemas from 'oneOf'
Error: #/2/definitionName (The name of version policy definition)
No enum match for: individualVersion
Error: #/2 (Lockstep version policy)
Additional properties not allowed: nextBump,version
Error: #/1
Data does not match any schemas from 'oneOf'
Error: #/1/definitionName (The name of version policy definition)
No enum match for: individualVersion
Error: #/1 (Lockstep version policy)
Additional properties not allowed: nextBump,version
Error: #/0
Data does not match any schemas from 'oneOf'
Error: #/0/definitionName (The name of version policy definition)
No enum match for: individualVersion
Error: #/0 (Lockstep version policy)
Additional properties not allowed: nextBump,version
```

**What is the expected behavior?**

Update all the corresponding packages to the new version, but probably I didn't understand how since the documentation is just a JSON file.

**If this is a bug, please provide the tool version, Node.js version, and OS.**

* **Tool:** rush + pnpm
* **Tool Version:** 5.35.2 + 5.11.1
* **Node Version:** 14.9.0
  * **Is this a LTS version?** No
[...]

PR #2422, opened 2021-01-08
[rush] Revise command line selection operations

### LINKED ISSUE #2354: [rush] Add a "rush build" CLI parameter for building a project's dependencies but NOT the project
## Summary

Quote from the Zulp discussion:

> Hi. What I want to achieve is that there is a flag that enables to build / run some command for all the dependencies of the module, but not the module itself.
> I know there is --from which will run command for all modules that are dependent from some package. And --to that will run command for all modules on which module is dependent.
> But the problem with it is that both of those commands are including the mentioned module.
> What's my use-case? I want to build all dependencies of the particular app, but I don't want to build the app itself, as I first want to run lint, tests, and other commands, and only after everything is good, I want to run build for the app itself. For now, I'm working around that using versions policies that I use as a sort of tag to map library packages as a library and build them all (even tho my app may not use those deps) using --to-version-policy flag.
> 
> Maybe the introduction of some tags will be a nice thing?

Link to the discussion: https://rushstack.zulipchat.com/#narrow/stream/262513-general/topic/Build.20deps.20but.20app

Proposed solution by @octogonz was to add `--from-but-not`  or `--dependencies-only` flags.

I prefer `--dependencies-only` and  `--dependents-only` flags as they just sound better and kinda easier to understand than `--from` and `--to` that are pretty hard to get used to TBH :) `--for-dependencies-of` and `--for-dependents-of` or  smth similar is a bit more verbose maybe, but IMO easier to get :) 

Thanks! Hopefully this will be added soon :)

dmichon-msft:
## Summary

Adds support for ~~`--from-except` and~~ `--to-except` to invoke a command on ~~all dependents on a specific project, or~~ all dependencies of a specified project.

Adds support for `--only PROJECT` to invoke a command on `PROJECT` without adding its dependents or dependencies to the selection.

Adds support for `--affected-by PROJECT` and `--affected-by-except PROJECT` to provide the functionality formerly provided by `--from` (i.e. execute the command on PROJECT and all dependents thereof, without reprocessing the dependencies).

~~Updates the interaction between `--from` and `--to` in `TaskSelector` to intersect the results of the former with the latter.
This ensures that `rush build --from A --to B` builds all projects that depend on `A` (including `A`) and are dependencies of `B` (including `B`). This is intended to support watch scenarios where some files have changed in `A` and the developer is interested in the effects on `B`.~~

Adds new properties to `RushConfigurationProject` that contain the dependencies and dependents as `ReadonlySet<RushConfigurationProject>` for ease of graph analysis.
Factors out directed graph manipulation functions for compositing. Adds unit tests for the same.
Switches to using generators for collection manipulation where appropriate.
~~Adds `--with-dependencies` flag to include all dependencies of the selected projects.~~
Fixes #2354
Alters behavior of `--from` to include all dependencies per #2349

## Details

Name of parameters subject to revision, pending discussion in #2354.
Optimized the calculation of the dependency graph to use a loop instead of recursion.

Moves responsibility for evaluating `--from`, `--to`, etc. out of TaskSelector and into the calling command, with the aid of selection helper functions. This change is to allow `--watch` to specify exactly the set of projects it wishes to execute.

## How it was tested

Invoked a variety of `rush build` scenarios with a combination of `--from`, `--to`, `--to-except`, `--affected-by`, `--affected-by-except`, `--only` on the `@rushstack` repo.
Selection helpers have unit test coverage.

Edit: fixed issue reference
Edit 2: reverted behavioral changes to `--to` and `--from`
Edit 3: added `--with-dependencies` flag
Edit 4: removed `--with-dependencies` flag and made it the default behavior of `--from`. Removed `--from-except` since including dependencies makes the option nonsense.


iclanton:
@dmichon-msft #1234 doesn't seem like the right reference.

dmichon-msft:
> @dmichon-msft #1234 doesn't seem like the right reference.

Fixed reference to 2354

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -182,36 +207,66 @@ export class BulkScriptAction extends BaseScriptAction {
...
-    this._toFlag = this.defineStringListParameter({
+    this._toProject = this.defineStringListParameter({
       parameterLongName: '--to',
       parameterShortName: '-t',
+      argumentName: 'PROJECT2',
+      description:
+        'Restricts command execution to the specified project and all dependencies of the specified project. ' +
+        '"." can be used as shorthand to specify the project in the current working directory. ' +
+        'Specifying "--to" or "--to-except" multiple times will take the union of matching projects. ' +
+        'Specifying "--from" as well will further restrict execution to the intersection of both filters.',
```
octogonz, line 218: > 'Specifying "--from" as well will further restrict execution to the intersection of both filters.',

Suppose we have this graph:

```
A --[depends-on]--> B  --[depends-on]--> C
```

I would expect this command to build all projects:
```shell
$ rush build --to B --to-except B --from B --from-except B
```

However the description above seems to imply that the set will be "restricted" to some sort of intersection.  (?)

Would appending additional CLI parameters /ever/ reduce the set of projects to be built? If so that seems counterintuitive.

octogonz:
Thanks a ton for making this PR!  I really need this feature.  :-)

apostolisms (approved):

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
dmichon-msft: @octogonz , with the modification, the command:
```
$ rush build --to B --from B
```
will build `B` and only `B` regardless of the shape of the graph.

Similarly,
```
$ rush build --to B --from A
```
matches 0 projects because:
`--to B` matches `B, C`
`--from A` matches `A`
And the intersection of these two is the empty set.

The issue this is addressing is that currently if you have the following graph:
```
A --[depends-on]--> B  --[depends-on]--> C --[depends-on]--> D
```
And you want to rebuild only `B` and `C`, there is no way to ask for that to happen.
With the change, you can do this with:
```
$ rush build --from C --to B
```
i.e. build everything from `C` to `B`

I suppose my question would be, right now, does anyone combine `--from` and `--to` in build commands? I never do because the combination is completely useless.

octogonz: > I suppose my question would be, right now, does anyone combine `--from` and `--to` in build commands? I never do because the combination is completely useless.

I use this all the time.  For example, suppose that I made a change to API Extractor:
```
# 1. Build all the dependencies needed by API Documenter
# 2. Build API Documenter itself
# 3. Build only projects that depend on API Documenter to make sure no tests were broken
$ rush build --to api-documenter --from api-documenter
```

This select 19 projects, which is much faster than building the full 106 projects in this repo.

The behavior you described is a significant design change, that would cause the above command to do something not useful.

octogonz: I'm also curious how your design would handle more complex graphs.  For example, consider:
```shell
$ rush build --from A --to H
```


With the current design, Rush would build the union of everything needed to build H, plus everything that depends on A.

It sounds like your design would instead compute an intersection like this:



Is that meaningful?  What does it represent exactly?

octogonz: FYI PR #2349 considered it a bug that Rush does not build B in this example:

```shell
$ rush build --from A
```



E may fail to compile because B as skipped.  So PR #2349 fixes this problem by ensuring that B will get included.

Perhaps there are two competing mentalities here:

1. *"Give me a quick way to skip compiling irrelevant projects, but do not expect me to remember how projects are chained together or whether they need to be rebuilt"* VERSUS
2. *"I save time by optimistically assuming most projects are up to date; if I'm wrong then I'll get a weird build error, but no problem, I can just do a full rebuild"*

octogonz (changes requested):
Based on the PR discussion, this PR is making a nontrivial design change that conflicts with https://github.com/microsoft/rushstack/pull/2349

We need to sort that out before merging.

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
octogonz: Here's another interesting case:
```shell
# Build only these 3 projects and their dependencies
$ rush build --to D --to E --to F
```
Will this still work with your new design?

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -182,36 +207,66 @@ export class BulkScriptAction extends BaseScriptAction {
...
+      completions: this._getProjectNames.bind(this)
+    });
```
octogonz, line 220: > And you want to rebuild only B and C, there is no way to ask for that to happen.

BTW we could instead solve that with an explicit parameter like `--include` :

```shell
# Build B and C only
$ rush build --include B --include C
```

You could also imagine an `--exclude` parameter:
```shell
# Build A and all its dependents and their dependencies (PR 2349 semantics), but skip B
$ rush build --from A --exclude B
```

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
dmichon-msft: I'm revising the PR to separate out adding support for `--from-except` and `--to-except` and will revisit adding a way to express "rebuild the minimal set of projects required to reflect changes to A in B" as a follow-on. As long as the machinery exists under the hood, the watch mode machinery doesn't really care if there's a way to ask for that set from the command line.

dmichon-msft: Closing for now since the change to behavior of `--from` and `--to` has been reverted. Will revisit in a later issue, since it clearly needs more discussion.

apps/rush-lib/src/logic/TaskSelector.ts
```diff
@@ -60,143 +59,71 @@ export class TaskSelector {
...
-    }
+    const expandedTo: Set<RushConfigurationProject> | false =
+      hasTo && expandAllDependencies<RushConfigurationProject>(expandToProjects(toProjects));
 
-    // Register all downstream dependents
-    for (const dependentProject of dependents.values()) {
-      this._registerTask(dependentProject);
+    if (expandedFrom && expandedTo) {
+      // Select all projects that depend on those in fromProjects OR that are dependencies of those in toProjects
+      // To include all dependencies of --from, per PR#2349, use expandAllDependencies(expandedFrom)
```
dmichon-msft, line 81: The behavior desired in #2349 can be facilitated by use of either `expandAllDependencies(expandedFrom)` or applying a final round of `expandAllDependencies()` to the result.

I will still need a follow up means to call this class and request only the critical path between projects `X` (inclusive) and `Y` (exclusive) as that was my primary motivation for implementing this feature.

Specifically, the scenario is that a developer is watching (via `webpack-dev-server` or `jest` or whatever) project `Y`, and is making changes to some other project, e.g. `X` and wants to see the effects of their changes as quickly as possible in project `Y`.

wbern:
Thanks for the work done here! Running into these issues myself now with --from flag.

Any chance this could be done for the install command as well?

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
dmichon-msft: Supporting the described "build only X and Y" scenario has been postponed to a later change that will focus on the concept of a "project set" or "working set" or "tag" or whatever term we actually decide to use.

apps/rush-lib/src/logic/TaskSelector.ts
dmichon-msft: Added `--with-dependencies` per discussion.

dmichon-msft:
@octogonz , give this another look with the reverted behavior and new version of the internals?

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -182,37 +203,76 @@ export class BulkScriptAction extends BaseScriptAction {
...
-        'Run command in the specified project and all projects that directly or indirectly depend on the ' +
-        'specified project. "." can be used as shorthand to specify the project in the current working directory.',
-      completions: this._getProjectNames.bind(this)
+        'Run command on the selection instead of all projects. ' +
+        'Adds all projects with the specified version policy, and all projects that depend on them, to the current selection.'
     });
+
+    this._withDependencies = this.defineFlagParameter({
+      parameterLongName: '--with-dependencies',
+      parameterShortName: '-w',
```
dmichon-msft, line 270: It occurs to me that we may want to reserve `-w` for `--watch`, unless we would prefer for watching to be a completely different command.

wbern: Agreed!

wbern:
With regards to my comment in #2345 

>@octogonz Honestly I think --from should include everything that's needed. As a colleague of mine said, to and from makes an hourglass, or at least so we thought at first. 🙂

Would it be drastic to let --from be "--with-dependencies", and have a flag that says "--without-dependencies"? I want to merge this really badly, so it pains me to bring this up, but a good design is important too. 😄

octogonz:
> Would it be drastic to let --from be "--with-dependencies", and have a flag that says "--without-dependencies"? I want to merge this really badly, so it pains me to bring this up, but a good design is important too. 😄

Yes, I was actually thinking the same thing.  Do the safe internally consistent thing by default, and then the extra CLI parameter would be for people know what they are doing.  We could even argue that it's not exactly a breaking design change, to broaden the set of projects.  (PR #2349 had assumed that was the intended behavior, for example.)

We can think a little more about whether `--without-dependencies` is the best name. But with this plan, PR #2435 seems good to go.

dmichon-msft:
@wbern , with latest iteration, `--from` makes a truncated cone, and `--to X --from X` is redundant, since it does the same thing as `--from X`.

Adding a `--without-dependencies` or such can be left as a later exercise, since I solved my issue for `--watch` by having `TaskSelector` accept the exact set of projects to execute the task on.

@octogonz , latest iteration should now resolve the behavior conflict.

apps/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -227,25 +228,55 @@ export class RushConfigurationProject {
...
+          yield* self._getLocalDependencyProjects(self.packageJson.devDependencies);
+          yield* self._getLocalDependencyProjects(self.packageJson.optionalDependencies);
+        })()
+      );
     }
     return this._localDependencyProjects;
   }
 
+  /**
+   * The set of projects withint he rush configuration which directly depend on this project.
```
octogonz, line 270:
```suggestion
   * The set of projects within the Rush configuration which directly depend on this project.
```

apps/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -227,25 +228,55 @@ export class RushConfigurationProject {
...
+   * Excludes those that declare this project as a cyclicDependencyProject
+   */
+  public get localDependentProjectSet(): ReadonlySet<RushConfigurationProject> {
```
octogonz, line 273: Is this important to expose as a `@public` API?

I guess what I'm wondering is *"What would I use this API for?"*  If there's a meaningful use case, we could explain it in the `@remarks` section.

common/reviews/api/rush-lib.api.md
```diff
@@ -413,10 +413,15 @@ export class RushConfigurationProject {
     // @internal
     constructor(projectJson: IRushConfigurationProjectJson, rushConfiguration: RushConfiguration, tempProjectName: string);
     get cyclicDependencyProjects(): Set<string>;
+    // @deprecated
     get downstreamDependencyProjects(): string[];
+    get downstreamDependencyProjectSet(): Set<string>;
```
octogonz, line 418: Should it be a `ReadonlySet`?

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -182,37 +224,94 @@ export class BulkScriptAction extends BaseScriptAction {
...
+      description:
+        'Run command on the selection instead of all projects. ' +
+        'Add the specified project (and only the specified project) to the current selection. ' +
+        '"." can be used as shorthand to specify the project in the current working directory. ' +
+        'Additional use of any selection commands will further expand the selection.',
+      completions: this._getProjectNames.bind(this)
+    });
+
+    this._affectedByProject = this.defineStringListParameter({
+      parameterLongName: '--affected-by',
```
octogonz, line 275: I think our final spec called it `--impacted-by` with `-i`.

Here's the comment from Teams for reference:

```
rush build --to project
rush build -t project

rush build --to-except project
rush build -T project

rush build --from project
rush build -f project

rush build --only project
rush build -o project

rush build --impacted-by project
rush build -i project

rush build --impacted-by-except project
rush build -I project

rush build --changed-projects-only
rush build -c
```

common/reviews/api/rush-lib.api.md
dmichon-msft: Can't, because `RushConfiguration` reads and mutates it.

apps/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: I can switch to `@internal`. It's the companion of `localDependencyProjectSet`, which is part of the public API. The less useful form, `downstreamProjects` is already public, though maybe that should be revised?

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -24,8 +24,9 @@ import { Utilities } from '../../utilities/Utilities';
 import { RushConstants } from '../../logic/RushConstants';
 import { EnvironmentVariableNames } from '../../api/EnvironmentConfiguration';
 import { LastLinkFlag, LastLinkFlagFactory } from '../../api/LastLinkFlag';
-import { IRushConfigurationProjectJson } from '../../api/RushConfigurationProject';
+import { IRushConfigurationProjectJson, RushConfigurationProject } from '../../api/RushConfigurationProject';
 import { BuildCacheConfiguration } from '../../api/BuildCacheConfiguration';
+import * as Selection from '../../logic/Selection';
```
octogonz, line 29: Make `Selection` a class.  

Using `import * as Selection` has a number of downsides:
- When I search our code for `Selection.union`, I may miss some places where developers inconsistently chose not to use `import * as Selection`.
- If the class gets renamed, people can forget to update some of the imports, and the compiler won't catch that
- As a general pattern `import * as Selection` only works for internal APIs; API Extractor won't know how to document it as a namespace because it is compositionally constructed

apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
```diff
@@ -531,7 +531,11 @@ export class WorkspaceInstallManager extends BaseInstallManager {
     }
 
     const localDependencyProjectNames: Set<string> = new Set<string>(
-      project.localDependencyProjects.map((x) => x.packageName)
+      (function* (deps: Iterable<RushConfigurationProject>): Iterable<string> {
+        for (const dep of deps) {
+          yield dep.packageName;
+        }
+      })(project.localDependencyProjectSet)
```
octogonz, line 538: `project.localDependencyProjects.map((x) => x.packageName)` really is more intuitive IMO

apps/rush-lib/src/logic/taskRunner/TaskRunner.ts
```diff
@@ -156,10 +156,28 @@ export class TaskRunner {
...
+              return 1;
+            }
+            return 0;
+          })
+          .map((x) => {
+            if (x.criticalPathLength !== undefined) {
+              return `  ${x.name.padEnd(maxNameLength, ' ')} (Depth: ${x.criticalPathLength})`;
+            }
+            return `  ${x.name}`;
+          })
```
octogonz, line 180: What is being changed here?  (What was wrong with the old way?)

apps/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -227,25 +228,55 @@ export class RushConfigurationProject {
...
-        ...this._getLocalDependencyProjects(this.packageJson.devDependencies),
-        ...this._getLocalDependencyProjects(this.packageJson.optionalDependencies)
-      ];
+      const self: RushConfigurationProject = this;
+      this._localDependencyProjects = new Set(
+        (function* () {
+          yield* self._getLocalDependencyProjects(self.packageJson.dependencies);
+          yield* self._getLocalDependencyProjects(self.packageJson.devDependencies);
+          yield* self._getLocalDependencyProjects(self.packageJson.optionalDependencies);
+        })()
```
iclanton, line 263: Just concatenate the arrays with spread operators. This is too complicated.

apps/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -377,15 +411,28 @@ export class RushConfigurationProject {
...
+   */
+  private *_getLocalDependentProjects(): Iterable<RushConfigurationProject> {
+    for (const projectName of this.downstreamDependencyProjectSet) {
+      const localProject: RushConfigurationProject | undefined = this._rushConfiguration.getProjectByName(
+        projectName
+      );
+      if (localProject && localProject.localDependencyProjectSet.has(this)) {
+        yield localProject;
+      }
+    }
```
iclanton, line 436: Just return an array.

apps/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -358,10 +389,13 @@ export class RushConfigurationProject {
     return isMain;
   }
 
-  private _getLocalDependencyProjects(
+  /**
+   * Compute the local rush projects that this project immediately depends on,
+   * according to the specific dependency group from package.json
+   */
+  private *_getLocalDependencyProjects(
```
iclanton, line 396: Just return an array here.

apps/rush-lib/src/cli/actions/BaseRushAction.ts
```diff
@@ -131,13 +131,11 @@ export abstract class BaseRushAction extends BaseConfiglessRushAction {
     return super.onExecute();
   }
 
-  protected mergeProjectsWithVersionPolicy(
-    projectsParameters: CommandLineStringListParameter,
-    versionPoliciesParameters: CommandLineStringListParameter
-  ): RushConfigurationProject[] {
+  protected *evaluateProjects(
```
iclanton, line 134: This is a pretty vague name.

octogonz:
Fixes https://github.com/microsoft/rushstack/issues/1447

apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
dmichon-msft: Well, `[...project.localDependencyProjects].map(x => x.packageName)` is the closest equivalent, but I can do it, it's not a perf bottleneck.

iclanton:
All of the generating functions feel more complicated than necessary.

apps/rush-lib/src/logic/taskRunner/TaskRunner.ts
dmichon-msft: This added the `depth: ` parameter to give some semblance of the tree rather than just a list of projects.

apps/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: But the generator is actually less code, unlike the other scenarios you called out

apps/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: Generators being too complicated is a self-fulfilling prophecy. If no one ever uses them for things they do efficiently, no one will ever get used to them. Much like the "if you assume all your users can't understand URLs, they will never learn to understand URLs" issue of modern browsers.

apps/rush-lib/src/cli/actions/BaseRushAction.ts
dmichon-msft: Revising to `evaluateProjectParameter`

apps/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: Updated.

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
dmichon-msft: Updated.

apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
dmichon-msft: Switched to using the array and map. Wonder if it might be worth extending `Set` with `.map()` for ergonomics.

apps/rush-lib/src/logic/taskRunner/TaskRunner.ts
dmichon-msft: Reverted the verbosity.

apps/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: Switched to `Selection.union()` here because I already have a helper function that does this.

apps/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: Switched to returning a `Set<RushConfigurationProject>` since the caller directly does so anyway.

apps/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: Switched to a `Set<RushConfigurationProject>` like for its counterpart.

apps/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -226,24 +228,55 @@ export class RushConfigurationProject {
...
     }
-    return this._localDependencyProjects;
+    return this._dependencyProjects;
+  }
+
+  /**
+   * The set of projects within the Rush configuration which declare this project as a dependency.
+   * Excludes those that declare this project as a `cyclicDependencyProject`.
+   *
+   * The counterpart to `localDependencyProjectSet`.
```
dmichon-msft, line 273: Missed a name

octogonz:
Here's an initial draft of the docs for @dmichon-msft's feature.

dmichon-msft:
> Here's an initial draft of the docs for @dmichon-msft's feature.

@octogonz , is there supposed to be a link here?

octogonz:
@dmichon-msft looks like I forgot to paste it: :-)

https://rushjs.io/pages/developer/selecting_subsets/

apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
dmichon-msft: Switched to using a (fake) class.

apps/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: This is now `consumingProjects`, and left public so that consumers of rush-lib can walk the tree in both directions if desired.

wbern:
Looks really good!

dmichon-msft:
@octogonz , I believe I have addressed all of your comments at this point.

octogonz (approved):

octogonz:
Released with Rush 5.38.0

PR #2431, opened 2021-01-12
[heft] Improve a couple minor things.

halfnibble:
## Summary

Two small changes to help Heft developers.

## Details

1. Export the Async utility methods so other Heft plugin developers can make use of them for plugins that support watchMode and multiple parallel async operations.

2. Update the default staticAssetsToCopy to this:
```
    "fileExtensions": [
      ".json",
      // Fonts
      ".eot",
      ".ttf",
      ".woff",
      // Images
      ".dds",
      ".gif",
      ".ico",
      ".jpeg",
      ".jpg",
      ".png",
      ".svg",
      ".webp",
      // Styles
      ".css",
      ".sass",
      ".scss"
    ]
```
## How it was tested

In our internal repo.

### DIFF
--- a/apps/heft/src/utilities/Async.ts
+++ b/apps/heft/src/utilities/Async.ts
@@ -3,7 +3,13 @@
 
 import { ScopedLogger } from '../pluginFramework/logging/ScopedLogger';
 
+/**
+ * Helpful async utility methods
+ */
 export class Async {
+  /**
+   * Utility method to limit the number of parallel async operations with a promise queue.
+   */
   public static async forEachLimitAsync<TEntry>(
     array: TEntry[],
     parallelismLimit: number,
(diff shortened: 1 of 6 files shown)


apps/heft/src/utilities/Async.ts
```diff
@@ -3,7 +3,13 @@
 
 import { ScopedLogger } from '../pluginFramework/logging/ScopedLogger';
 
+/**
+ * Helpful async utility methods
+ */
 export class Async {
```
iclanton, line 9: Perhaps this should be better formalized and moved to `node-core-library`? @octogonz?

halfnibble: @octogonz and @iclanton  Before I forget, I can redo this PR to just have the staticAssetsToCopy change.

But I really think we should have a `heft-plugin-utils` project on here with methods that are useful when creating plugins for heft. I know we've copied the Async static methods from here into 2 other internal repos, and I think it would be best if we found a way to consolidate that code into one place.

PR #2437, opened 2021-01-14
[rush-lib] - [2436] - Add new event hooks for pre and post rush add

### LINKED ISSUE #2436: [rush]
## Summary
`preRushInstall` and `postRushInstall` hooks are not called when using `rush add -p somepackage`, but the logs in terminal when running this command do state `Running "rush update"`. These hooks do run when running `rush update` or `rush install` directly.

## Repro steps
- Add a `preRushInstall` and/or `postRushInstall` event hook to the rush.json
- Run `rush add -p lodash` in a package
- Note how `Running "rush update"` is logged

**Expected result:**: the same event hooks are called as when running `rush update`
**Actual result:** neither `preRushInstall` or `postRushInstall` are called

## Details
When reading the difference in implementations of the [`BaseInstallAction.ts`](https://github.com/microsoft/rushstack/blob/7b1767ba71a85f39fc1cfe2cf1266077430249f5/apps/rush-lib/src/cli/actions/BaseInstallAction.ts) (extended by the Update Action) versus the [`AddAction.ts`](https://github.com/microsoft/rushstack/blob/7b1767ba71a85f39fc1cfe2cf1266077430249f5/apps/rush-lib/src/cli/actions/AddAction.ts); I see the former registers the events with the `eventHooksManager` where the latter does not.

The UpdateAction also does not extend the BaseInstallAction. The documentation says that `rush add -p`:

> Adds a specified package as a dependency of the current project ... and then runs "rush update" ...

Instead, the `doRushAdd` appears to not actually perform the equivalent of the `rush update` action, instead calling the `installManager.doInstall()` directly https://github.com/microsoft/rushstack/blob/83001f419e191098e5137e047f5cbe478dc7de1d/apps/rush-lib/src/logic/PackageJsonUpdater.ts#L236 but without the hooks that are registered in BaseInstallAction.

I'm happy to write up a fix for this so the relevant lifecycle is called, if that is agreed to be expected behavior? But do also understand that doing that, and suddenly starting to call lifecycle hooks at times they weren't previously called, may also cause it's own issue, so perhaps would make sense to have new event hooks for preRushAdd, postRushAdd, in which case current behavior is not changed in the interim.

Either way, would be happy to contribute a PR as deemed appropriate.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | Y
| `rushVersion` from rush.json? | 5.34.3 (but review of observed code in master)
| `useWorkspaces` from rush.json? | N
| Operating system? | Mac
| Would you consider contributing a PR? | Yes
| Node.js version (`node -v`)? | 14.15.4

cblagg:
## Summary
Fixes #2436 - This introduces new event hooks for pre and post `rush add` to resolve the issue noted in [2436] in a backwards compatible manner.

## Details
The issue [#2436] notes how the pre and post install scripts are not called when performing `rush add` although the console output informs that a `rush update` is being performed, and the documentation correlates to this action being performed as part of this lifecycle.

However, to avoid a potentially breaking change (starting calling lifecycle scripts when they previously weren't), this resolves the issue by adding new event hooks for this action, so configuration can be added where the expected/desired behavior is needed, by referencing the same script(s) to be called.

## How it was tested
- To be expanded and tested if decided as agreed upon approach


ovidiubute:
@cblagg I could really use this feature, do you want to continue your PR?

PR #2483, opened 2021-02-11
[rush-lib] Disable build cache feature in `--watch` after initial build

dmichon-msft:
## Summary

Disables the build cache feature during after the initial build with `--watch` finishes. This is for the following reasons:
- Prevent Rush from deleting and recreating files by rehydrating from cache, which tends to break downstream file watchers, such as `webpack --watch`
- Avoid the CPU overhead of writing to the local build cache
- Avoid the disk usage of creating new local build cache entries every time a change occurs

## Details

Works by removing awareness of the build cache from the build process after the first build finishes.
Will reduce performance if the developer reverts to a prior build state during a `--watch` process, since this will now require a build instead of a cache hit.

## How it was tested
`rush build --watch` on the `rushstack` repo, changing files in `package-deps-hash` and reverting said changes, observing that the rebuild no longer uses the cache, even when it reverts to a previously known build state. The initial build still uses the cache, however.

### DIFF
--- a/apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
+++ b/apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
@@ -234,6 +234,9 @@ export class BulkScriptAction extends BaseScriptAction {
       projectsToWatch
     });
 
+    let buildCacheConfiguration: BuildCacheConfiguration | undefined =
+      options.taskSelectorOptions.buildCacheConfiguration;
+
     // Loop until Ctrl+C
     // eslint-disable-next-line no-constant-condition
     while (true) {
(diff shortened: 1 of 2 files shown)


apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -220,7 +220,10 @@ export class BulkScriptAction extends BaseScriptAction {
    */
   private async _runWatch(options: IExecuteInternalOptions): Promise<void> {
     const {
-      taskSelectorOptions: { selection: projectsToWatch },
+      taskSelectorOptions: {
+        buildCacheConfiguration: initialBuildCacheConfiguration,
+        selection: projectsToWatch
```
iclanton, line 225: I thought `selection` was going to be renamed in the `IExecuteInternalOptions` type?

dmichon-msft: This is `ITaskSelectorConstructor`. The rename was in `IProjectWatcherOptions`. This value is the set of projects to invoke the command on for single execution.

iclanton (approved):

PR #2489, opened 2021-02-15
[rush] Don't try to upload a cache entry to Azure Storage if it already exists.

iclanton:
We have a CI job that populates our Azure Storage cache. It occasionally fails if two jobs run at the same time and both try to upload the same blob cache entry to Azure.

This change checks if the blob exists right before trying to upload. If the blob exists, we'll consider the upload "successful" because it would have been filled from cache if it had existed when the project's build started.

### DIFF
--- a/apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
+++ b/apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
@@ -119,12 +119,19 @@ export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase 
 
     const blobClient: BlobClient = await this._getBlobClientForCacheIdAsync(cacheId);
     const blockBlobClient: BlockBlobClient = blobClient.getBlockBlobClient();
-    try {
-      await blockBlobClient.upload(entryStream, entryStream.length);
+
+    const blobAlreadyExists: boolean = await blockBlobClient.exists();
+    if (blobAlreadyExists) {
+      terminal.writeVerboseLine('Build cache entry blob already exists.');
       return true;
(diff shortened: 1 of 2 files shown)


halfnibble (approved):

D4N14L (approved):

apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
```diff
@@ -119,12 +119,19 @@ export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase
 
     const blobClient: BlobClient = await this._getBlobClientForCacheIdAsync(cacheId);
     const blockBlobClient: BlockBlobClient = blobClient.getBlockBlobClient();
-    try {
-      await blockBlobClient.upload(entryStream, entryStream.length);
+
+    const blobAlreadyExists: boolean = await blockBlobClient.exists();
```
dmichon-msft, line 123: Should we try securing the lease to the blob instead? Seems like that should handle concurrency better.
We should also find out what the error code is if you try to write to an immutable blob and treat that as success.

iclanton: Leasing the blob seems more complicated than it's worth. I can try to handle the simultaneous modification error.

dmichon-msft (approved):

apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
iclanton: I'll do that in a followup

PR #2528, opened 2021-03-02
[rush] Prefer frozen lockfile during `rush update` to minimize lockfile churn

### LINKED ISSUE #2525: [rush] Use `--prefer-frozen-lockfile` in `rush update` (without `--full`)
## Summary

When running `rush update`, I expect it to only change the lockfile if necessary to satisfy new constraints from `rush.json`, `package.json`, `pnpmfile.js`, or `common-versions.json`.

However, the current command runs `pnpm install --no-prefer-frozen-lockfile`, which is a bit more permissive of lockfile updates.

## Repro steps

Pull latest `main` branch of a Rush monorepo, run `rush update`

  **Expected result:** 
No changes detected, lockfile remains the same and does equivalent of `rush install`

  **Actual result:** 
Updates made to pnpm-lock.yaml and repo-state.json

## Details
Change `--no-prefer-frozen-lockfile` to `--prefer-frozen-lockfile`, at least for pnpm >= 5 on workspaces, or just behind an experiment/pnpm config setting.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.40.3 |
| `rushVersion` from rush.json? | 5.40.3 |
| Operating system? | Linux |
| Node.js version (`node -v`)? | 10.24.0 |

dmichon-msft:
## Summary

Fixes #2525

## Details
Switches pnpm to use `--prefer-frozen-lockfile` for `rush update` when `useWorkspaces` is `true`

## How it was tested
Verified locally on rushstack repo by running `rush update` that no change occurs if the lockfile is up to date, and that modifications occur if it is not.

### DIFF
--- a/apps/rush-lib/src/logic/base/BaseInstallManager.ts
+++ b/apps/rush-lib/src/logic/base/BaseInstallManager.ts
@@ -502,6 +502,9 @@ export abstract class BaseInstallManager {
         } else {
           args.push('--frozen-shrinkwrap');
         }
+      } else if (this._rushConfiguration.pnpmOptions.useWorkspaces) {
+        // In workspaces, we want to avoid unnecessary lockfile churn
+        args.push('--prefer-frozen-lockfile');
       } else {
         // Ensure that Rush's tarball dependencies get synchronized properly with the pnpm-lock.yaml file.
         // See this GitHub issue: https://github.com/pnpm/pnpm/issues/1342
--- a/apps/rush-lib/src/logic/base/BaseInstallManager.ts
+++ b/apps/rush-lib/src/logic/base/BaseInstallManager.ts
@@ -493,16 +493,15 @@ export abstract class BaseInstallManager {
         args.push('--no-lock');
       }
 
-      if (
-        this._rushConfiguration.experimentsConfiguration.configuration.usePnpmFrozenLockfileForRushInstall &&
-        !this._options.allowShrinkwrapUpdates
-      ) {
+      const { configuration: experiments } = this._rushConfiguration.experimentsConfiguration;
+
+      if (experiments.usePnpmFrozenLockfileForRushInstall && !this._options.allowShrinkwrapUpdates) {
         if (semver.gte(this._rushConfiguration.packageManagerToolVersion, '3.0.0')) {
           args.push('--frozen-lockfile');
         } else {
           args.push('--frozen-shrinkwrap');
         }
-      } else if (this._rushConfiguration.pnpmOptions.preferFrozenLockfileForUpdate) {
+      } else if (experiments.usePnpmPreferFrozenLockfileForRushUpdate) {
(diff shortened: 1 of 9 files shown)


D4N14L (approved):

apps/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -502,6 +502,9 @@ export abstract class BaseInstallManager {
         } else {
           args.push('--frozen-shrinkwrap');
```
D4N14L, line 503: Should we update this terminology?

apps/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -502,6 +502,9 @@ export abstract class BaseInstallManager {
...
         }
+      } else if (this._rushConfiguration.pnpmOptions.preferFrozenLockfileForUpdate) {
+        // In workspaces, we want to avoid unnecessary lockfile churn
```
D4N14L, line 506: You may also want to condition this option on workspaces support, though opting in does imply that they know what they're doing

sachinjoseph (approved):

apps/rush-lib/src/schemas/rush.schema.json
```diff
@@ -103,6 +103,10 @@
         "environmentVariables": {
           "$ref": "#/definitions/environmentVariables"
         },
+        "preferFrozenLockfileForUpdate": {
+          "description": "If treu, the for \"rush update\", pnpm will use the --prefer-frozen-lockfile flag to minimize shrinkwrap changes.",
```
D4N14L, line 107: treu -> true

apps/rush-lib/src/logic/base/BaseInstallManager.ts
sachinjoseph: If we condition it on workspaces, then the flag name would also need to imply that - something like preferFrozenLockfileForUpdateInWorkspaces. But like you said, opting in does imply that they know what they're doing, so IMO, this should not be conditioned on workspaces support.

sachinjoseph (approved):

apps/rush-lib/src/logic/base/BaseInstallManager.ts
dmichon-msft: Not related.

iclanton:
Would we want this to eventually become the default behavior? If so, should this be an experiment instead of a `rush.json` option?

D4N14L:
> Would we want this to eventually become the default behavior? If so, should this be an experiment instead of a `rush.json` option?

Yes, I think this would be the case (especially since we are planning on workspaces becoming the default as well)

dmichon-msft:
@iclanton, @D4N14L, switch to an experiment from a pnpmOptions value.

iclanton (approved):

PR #2538, opened 2021-03-07
[rush] Allow merge conflicts in repo-state.json to be automatically resolved.

iclanton:
## Summary

When the `preventManualShrinkwrapChanges` option in `rush.json` is turned on, the `repo-state.json` file contains a hash of the shrinkwrap file. Whenever a merge conflict is encountered in the shrinkwrap file, one is almost guaranteed to occur in `repo-state.json`. PNPM gracefully handles merge conflicts in the `pnpm-lock.yaml` file, and this PR adds that same graceful handling to `repo-state.json` when the user runs `rush update`.

## Details

This change detects a merge conflict marker in `repo-state.json` and marks the file as "not valid," but doesn't throw an exception if `rush update` is being run.

## How it was tested

I intentionally caused a merge conflict in `repo-state.json` and ran `rush update` (with the changes in this PR) on that repo. I also tested that `rush install` and `rush update` still work as expected with these changes when there isn't a conflict.

### DIFF
--- a/apps/rush-lib/src/logic/RepoStateFile.ts
+++ b/apps/rush-lib/src/logic/RepoStateFile.ts
@@ -42,15 +42,18 @@ export class RepoStateFile {
   private _variant: string | undefined;
   private _pnpmShrinkwrapHash: string | undefined;
   private _preferredVersionsHash: string | undefined;
+  private _isValid: boolean;
   private _modified: boolean = false;
 
   private constructor(
     repoStateJson: IRepoStateJson | undefined,
+    isValid: boolean,
     filePath: string,
     variant: string | undefined
   ) {
     this._repoStateFilePath = filePath;
     this._variant = variant;
+    this._isValid = isValid;
 
     if (repoStateJson) {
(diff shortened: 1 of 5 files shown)


octogonz (approved):

apps/rush-lib/src/logic/RepoStateFile.ts
```diff
@@ -87,16 +97,46 @@ export class RepoStateFile {
...
+        for (
+          let nextNewlineIndex: number = 0;
+          nextNewlineIndex > -1;
+          nextNewlineIndex = fileContents.indexOf('\n', nextNewlineIndex + 1)
+        ) {
+          if (fileContents.substr(nextNewlineIndex + 1, 7) === '<<<<<<<') {
+            foundMergeConflictMarker = true;
+            repoStateJson = {
+              preferredVersionsHash: 'INVALID',
+              pnpmShrinkwrapHash: 'INVALID'
```
D4N14L, line 127: Does this mean that if any new properties are added they'll need to be included here? Though it does make it more clear what's going wrong.

D4N14L (approved):

PR #2556, opened 2021-03-16
[rush] Fix `rush install` when omitImportersFromPreventManualShrinkwrapChanges is active

dmichon-msft:
## Summary

The code path for the preventManualShrinkwrapChanges validation policy during "rush install" was missed when the experiment was originally added, so running "rush update" would not alter the lockfile or the repo-state.json file, but a subsequent "rush install" from a clean state would fail.

## Details

Refactors how the parameter is passed to ensure that it is consistent between the validator and the repo-state.json generation.

## How it was tested

Ran "rush update" using the built version in another repo.
Ran "rush purge" in said repo.
Ran "rush install" in said repo, verified success.
Manually modified the "importers" section in the lockfile and a corresponding package.json
Ran "rush purge" in said repo.
Ran "rush install" in said repo, verified success.

### DIFF
--- a/apps/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
+++ b/apps/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
@@ -205,17 +205,11 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
   public readonly shrinkwrapFilename: string;
 
   private readonly _shrinkwrapJson: IPnpmShrinkwrapYaml;
-  private readonly _hashSerializeOptions: IPnpmShrinkWrapFileSerializeOptions;
 
-  private constructor(
-    shrinkwrapJson: IPnpmShrinkwrapYaml,
-    shrinkwrapFilename: string,
-    hashSerializeOptions: IPnpmShrinkWrapFileSerializeOptions
-  ) {
+  private constructor(shrinkwrapJson: IPnpmShrinkwrapYaml, shrinkwrapFilename: string) {
     super();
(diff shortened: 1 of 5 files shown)


apps/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -292,7 +285,11 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
           throw new AlreadyReportedError();
         }
 
-        if (this.getShrinkwrapHash() !== policyOptions.repoState.pnpmShrinkwrapHash) {
+        const hashSerializeOptions: IPnpmShrinkWrapFileSerializeOptions = {
+          omitImporters: policyOptions.validateOnlyExternalPackageLayout
```
iclanton, line 289: It feels a little odd that this is set two different ways in the two places the `getShrinkwrapHash` function is called. Do we not have access to the `policyOptions` object in the `refreshState` function?

dmichon-msft: There isn't really a common node between these two code paths. If I pass it into RepoStateFile during construction, that's a public API change, which so far this experiment has avoided doing.

dmichon-msft: Found a solution. Passing `IExperimentsJson` as an optional 3rd parameter to the validate() function, and now all the code paths in `PnpmShrinkwrapFile` just use the `IExperimentsJson` for consistency (`getShrinkwrapHash()` and `validate()`).

iclanton (approved):

PR #2574, opened 2021-03-27
[heft] Add `--trace` parameter to Heft to emit timeline trace in Chrome Trace Events format

dmichon-msft:
## Summary
Adds support for `heft --trace <command>` to record a timeline view in Chrome Trace Events format of all tap invocations.

## Details
Logs all tap invocations, listing the name of the plugin and of the hook. The stack view is a bit misleading due to the viewer assuming that durations that fully intersect other durations are sub-operations.

image

## How it was tested
Running locally to compile rush-lib, and loading the output file into Edge Dev Tools.

### DIFF
--- a/apps/heft/src/plugins/TimelineTracePlugin.ts
+++ b/apps/heft/src/plugins/TimelineTracePlugin.ts
@@ -1,0 +1,188 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import { FileWriter } from '@rushstack/node-core-library';
+
+import { HeftConfiguration } from '../configuration/HeftConfiguration';
+import { IHeftPlugin } from '../pluginFramework/IHeftPlugin';
+import { HeftSession } from '../pluginFramework/HeftSession';
+import { IHeftLifecycle } from '../pluginFramework/HeftLifecycle';
+import { HookInterceptor, SyncHook, Tap, TapOptions } from 'tapable';
+import { IBuildStageContext } from '../stages/BuildStage';
+import { ITestStageContext } from '../stages/TestStage';
+import { ICleanStageContext } from '../stages/CleanStage';
+
+const PLUGIN_NAME: 'TimelineTracePlugin' = 'TimelineTracePlugin';
+
+/**
+ * This plugin generates a timeline trace of the Heft execution in the Chrome Trace Events format.
+ */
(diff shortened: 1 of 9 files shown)


apps/heft/src/plugins/TimelineTracePlugin.ts
```diff
@@ -0,0 +1,188 @@
...
+const PLUGIN_NAME: 'TimelineTracePlugin' = 'TimelineTracePlugin';
+
+/**
+ * This plugin generates a timeline trace of the Heft execution in the Chrome Trace Events format.
+ */
+export class TimelineTracePlugin implements IHeftPlugin {
+  public readonly pluginName: 'TimelineTracePlugin' = PLUGIN_NAME;
+
+  public apply(heftSession: HeftSession, heftConfiguration: HeftConfiguration): void {
+    const traceFileName: string = `${heftConfiguration.buildFolder}/heft.trace.${Date.now()}.json`;
```
iclanton, line 24: Same as my [comment](https://github.com/microsoft/rushstack/pull/2569/files#r605366585) on https://github.com/microsoft/rushstack/pull/2569. Seems like this should come from the user.

apps/heft/src/stages/BuildStage.ts
```diff
@@ -307,7 +307,8 @@ export class BuildStage extends StageBase<BuildStageHooks, IBuildStageProperties
     buildStage,
     watchMode
   }: IRunSubstageWithLoggingOptions): Promise<void> {
-    if (buildStage.hooks.run.isUsed()) {
+    // isUsed() also checks interceptors
+    if (buildStage.hooks.run.taps.length) {
```
iclanton, line 311: Let's make this more explicit

```suggestion
    if (buildStage.hooks.run.taps.length > 0) {
```

apps/heft/src/stages/StageBase.ts
```diff
@@ -80,7 +80,8 @@ export abstract class StageBase<
   }
 
   public async executeAsync(): Promise<void> {
-    if (this.stageHooks.overrideStage.isUsed()) {
+    // isUsed() also checks interceptors
+    if (this.stageHooks.overrideStage.taps.length) {
```
iclanton, line 84:
```suggestion
    if (this.stageHooks.overrideStage.taps.length > 0) {
```

dmichon-msft:
This was based on the legacy version of heft. Closing.

iclanton:
This is for the old Heft design.

PR #2581, opened 2021-03-31
Verify changefile schema when running `rush change --verify`

D4N14L:
## Summary
Fixes an issue where non-valid changefiles would slip past `rush change --verify` due to verify only checking the changefiles satisfy the changes made in the project.

## Details
To fix, I created a schema to loosely validate the structure of the changefile and any fields that are required to have specific values. I validate against this schema while parsing out the changefiles during `rush change --verify`.

## How it was tested
Made a change in a repo and manually modified the `type` field to be invalid. Confirmed that the validation logic threw.

### DIFF
--- a/apps/rush-lib/src/logic/ChangeFiles.ts
+++ b/apps/rush-lib/src/logic/ChangeFiles.ts
@@ -1,8 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
+import * as path from 'path';
 import { EOL } from 'os';
-import { JsonFile, Import } from '@rushstack/node-core-library';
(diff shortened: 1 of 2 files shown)


common/changes/@microsoft/rush/user-danade-VerifyChangeType_2021-03-31-18-55.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "\"Validate changefiles against a schema when running 'rush change --verify'\"",
```
iclanton, line 5:
```suggestion
      "comment": "Validate changefiles against a schema when running 'rush change --verify'",
```

iclanton (approved):

PR #2592, opened 2021-04-07
[api-extractor-model] Fix typo in api-extractor-model README

Rugvip:
Just a small README typo fix! :)

### DIFF
--- a/apps/api-extractor-model/README.md
+++ b/apps/api-extractor-model/README.md
@@ -43,7 +43,7 @@ might look like this:
         - ApiMethodSignature
         - ApiPropertySignature
       - ApiNamespace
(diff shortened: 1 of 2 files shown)


Rugvip:
Do you want a changelog entry for this or is it fine to go out with the next change?

iclanton:
@Rugvip - can you create a changelog entry? You can leave the message blank to make it just go out with the next release.

iclanton (approved):

Rugvip:
Sorry, forgot about this one. Tried installing the local tooling to create the change but the dependency install step froze and killed the internet on my machine, so put off going down that troubleshooting rabbit hole 😅

PR #2612, opened 2021-04-15
Fix incremental TypeScript compilation in Heft

dmichon-msft:
## Summary
When adding support for older versions of TypeScript, diagnostics and emit were accidentally redirected to being the responsibility of the inner program (which does not support incremental) instead of the BuilderProgram.

## Details
Fixes to prefer the BuilderProgram if defined. Also touches up other parts of the code and makes setting `noEmitOnError: false` explicit in the base rigs.

Also ensures that emit diagnostics from the incremental compilation are included in the diagnostic output.

## How it was tested
Validated that building `@rushstack/rush-lib` multiple times in a row with no changes yields check, transform, print, and emit times of 0 ms.

### DIFF
--- a/common/changes/@rushstack/heft-node-rig/incremental-ts_2021-04-15-01-30.json
+++ b/common/changes/@rushstack/heft-node-rig/incremental-ts_2021-04-15-01-30.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-node-rig",
+      "comment": "Explicitly set noEmitOnError: false",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/heft-node-rig",
+  "email": "dmichon-msft@users.noreply.github.com"
+}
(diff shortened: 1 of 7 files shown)


common/changes/@rushstack/heft-node-rig/incremental-ts_2021-04-15-01-30.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-node-rig",
+      "comment": "Explicitly set noEmitOnError: false",
```
iclanton, line 5:
```suggestion
      "comment": "Explicitly set the noEmitOnError TypeScript compiler option to false in the base tsconfig.",
```

common/changes/@rushstack/heft-web-rig/incremental-ts_2021-04-15-01-30.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-web-rig",
+      "comment": "Explicitly set noEmitOnError: false",
```
iclanton, line 5:
```suggestion
      "comment": "Explicitly set the noEmitOnError TypeScript compiler option to false in the base tsconfig.",
```

iclanton (approved):

PR #2624, opened 2021-04-20
[rush-lib] Reorganize unit tests for PackageChangeAnalyzer

elliot-nelson:
## Summary

In preparation for some future changes to `PackageChangeAnalyzer`, I've done some (very light) refactoring in the code and rewritten the unit tests to cover most of the functionality in the class.  The goal is to make it easier to prove that future changes don't break the existing API.

## Details

 - For callers of `PackageChangeAnalyzer`, the only change in this PR is that _constructing_ a PackageChangeAnalyzer no longer immediately queries the state of git (file system).  The methods `getPackageDeps` and `getProjectStateHash` are now "lazy-load" instead of "eager-load".  This change makes the class easier to reason about and write unit tests for.

 - Unit tests now cover the two primary responsibilities of this class (one is to produce the list of dependencies relevant for each project, the other to take those dependencies and turn them into a final hash value).

 - Removed the sort-of-injector pattern in favor of a traditional jest mock on the test subject and cleaned up `any` usage.

### DIFF
--- a/apps/rush-lib/src/logic/PackageChangeAnalyzer.ts
+++ b/apps/rush-lib/src/logic/PackageChangeAnalyzer.ts
@@ -15,9 +15,6 @@ import { RushConfigurationProject } from '../api/RushConfigurationProject';
 import { RushConstants } from './RushConstants';
 
 export class PackageChangeAnalyzer {
-  // Allow this function to be overwritten during unit tests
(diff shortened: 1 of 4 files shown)


iclanton (approved):

iclanton:
Lol at your test path names.

Issue #2607, opened 2021-04-13
[rush] When passing --json install-run-rush should not output extra information


M3kH:
When trying to use the commands that offers `--json` as option for the output we don't need extra output in the `stdin` but they can be in the `stderr`.

Example:
`node common/scripts/install-run-rush.js list --json`

## Summary
I was trying to parse the output of `rush list --json` with `jq`.
The `install-run-rush.js` outputs has additional information that didn't let me parse the output.
`install-run-rush.js` use `console.log` rather then `console.error` and so uses `stdin` rather then `stderr`.

## Standard questions

Please answer these questions to help us investigate your issue more quickly:

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.42.2 |
| `rushVersion` from rush.json? | 5.42.2 |
| `useWorkspaces` from rush.json? | true |
| Operating system? | Windows, Mac, Linux |
| Would you consider contributing a PR? | Yes |
| Node.js version (`node -v`)? |  12.20.1|

amxfonseca:
Every time the command is invoked the following heading is printed. This is a no go for most commands that need to parse the output of the command

```
The rush.json configuration requests Rush version 5.42.2

Invoking "rush version"
```

M3kH:
@iclanton if you don't mind I will like to do a pull request so that we could use as food for thoughts.

iclanton:
@M3kH Go for it!

std4453:
Just stumbled into this, been parsing `rush list --json` and works fine locally but breaks in CI. Is there anyone that can review and approve these PRs?

PR #2625, opened 2021-04-20
Move CLI information output to stderr for install-run-rush

M3kH:
## Summary
Trying to solve #2607 by forward informational CLI output on `stderr`

## Details
Rather than using and `stdin` for informational output we want to use `sterr`, this is also the right usage for the stderr.

## How it was tested
N/A


iclanton:
This isn't a solution to #2607. These are informational messages that should be printed to stdout. When the `--json` flag is passed, those messages should be suppressed.

Can we discuss in-person tomorrow? I think I have a good solution, but I'd like to discuss it first.

M3kH:
> This isn't a solution to #2607. These are informational messages that should be printed to stdout. When the `--json` flag is passed, those messages should be suppressed.
> 
> 
> 
> Can we discuss in-person tomorrow? I think I have a good solution, but I'd like to discuss it first.

@iclanton I don't think you want to discuss with me right? 😅 otherwise sure I can make time.

iclanton:
@M3kH - lol sorry, I started typing that response, got distracted, and then typed something I meant to type to someone else.

You want to fix this issue by completely suppressing the banner messages, not just print them to a different stream. The easiest way to do that may be to add a feature to `ts-command-line` where a flag parameter can be marked as causing the tool to produce machine-readable output. Then you can plumb its inclusion to a property or function in the action class and the CommandLineParser class directing the tool not to print anything but the machine-readable output.

dmichon-msft:
#2777 is looking at an alternate approach to the problem, namely it is turning off the logging from install-run-rush.js altogether unless `--debug` or `-d` is passed on the command line, and only printing the output of the underlying command. This combines with an ability to disable Rush's own additional boilerplate logging with a configuration parameter in the action schema.

M3kH:
@dmichon-msft indeed I think that is a better solution.
I will still consider not using `stdout` for debugging purposes, which is a practice of many `CLI` tools.

PR #2627, opened 2021-04-20
[rush-lib] Fix issue on Windows with paths that contains spaces

manrueda:
## Summary

Currently all Rush commands that involve the `install-run` scripts fail on Windows when the project's absolute path has a space (" ") on it.

This is caused by a limitation of Node.js's `spawn` method that cannot execute a binary using a path with spaces on it. The limitation is known, and the recommended path is to execute the binary on shell mode and wrap the path in double quotes to allow Windows shell (CMD) properly resolve the path.

## Details

My changes involve the detection of the failing case (space on path + Windows OS) and only in that case use the shell mode + quotes wrapping on the path.

In this way we ensure that only the case with issue will change and use shell mode, Linux and MacOS does not work if the path gets wrapped in quotes.

## How it was tested

I build the project and copy the resulting `install-run.js` content into a project that had the issue and confirm that it works.

Also test it on MacOS to confirm that no side-effects were introduced.

### DIFF
--- a/apps/rush-lib/src/scripts/install-run.ts
+++ b/apps/rush-lib/src/scripts/install-run.ts
@@ -452,9 +452,16 @@ export function installAndRun(
   const originalEnvPath: string = process.env.PATH || '';
   let result: childProcess.SpawnSyncReturns<Buffer>;
   try {
+    // Node.js on Windows can not spawn a file when the path has a space on it
+    // unless the path gets wrapped in a cmd friendly way and shell mode is used
+    const shouldUseShell: boolean = binPath.includes(' ') && os.platform() === 'win32';
+    const platformBinPath: string = shouldUseShell ? `"${binPath}"` : binPath;
+
     process.env.PATH = [binFolderPath, originalEnvPath].join(path.delimiter);
-    result = childProcess.spawnSync(binPath, packageBinArgs, {
+    result = childProcess.spawnSync(platformBinPath, packageBinArgs, {
       stdio: 'inherit',
+      windowsVerbatimArguments: false,
+      shell: shouldUseShell,
       cwd: process.cwd(),
       env: process.env
     });
--- a/common/changes/@microsoft/rush/fix-windows-support-install-scripts_2021-04-20-23-21.json
+++ b/common/changes/@microsoft/rush/fix-windows-support-install-scripts_2021-04-20-23-21.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix an issue where Rush fails to run on Windows when the repository absolute path contains a space",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush",
+  "email": "manrueda@users.noreply.github.com"
+}


apps/rush-lib/src/scripts/install-run.ts
```diff
@@ -452,9 +452,16 @@ export function installAndRun(
...
+    // Node.js on Windows can not spawn a file when the path has a space on it
+    // unless the path gets wrapped in a cmd friendly way and shell mode is used
+    const shouldUseShell: boolean = binPath.includes(' ') && os.platform() === 'win32';
+    const scapedBinPath: string = shouldUseShell ? `"${binPath}"` : binPath;
+
     process.env.PATH = [binFolderPath, originalEnvPath].join(path.delimiter);
-    result = childProcess.spawnSync(binPath, packageBinArgs, {
+    result = childProcess.spawnSync(scapedBinPath, packageBinArgs, {
       stdio: 'inherit',
+      windowsVerbatimArguments: false,
```
D4N14L, line 463: Isn't this the default value already?

apps/rush-lib/src/scripts/install-run.ts
```diff
@@ -452,9 +452,16 @@ export function installAndRun(
   const originalEnvPath: string = process.env.PATH || '';
   let result: childProcess.SpawnSyncReturns<Buffer>;
   try {
+    // Node.js on Windows can not spawn a file when the path has a space on it
+    // unless the path gets wrapped in a cmd friendly way and shell mode is used
+    const shouldUseShell: boolean = binPath.includes(' ') && os.platform() === 'win32';
+    const scapedBinPath: string = shouldUseShell ? `"${binPath}"` : binPath;
```
D4N14L, line 458: Is 'scaped' a verb? Did you mean 'escaped'? Is wrapping in quotes really considered 'escaping'? What even are words?

D4N14L (approved):

iclanton (changes requested):
Can you address @D4N14L's comments?

apps/rush-lib/src/scripts/install-run.ts
manrueda: It is but only when `shell` is false. When `shell` is true, `windowsVerbatimArguments` changes to true too. In this way it's always false.

apps/rush-lib/src/scripts/install-run.ts
manrueda: I can rename it, any preference for the name?

D4N14L: Maybe `platformBinPath` to emphasize that it's modified due to platform specific reasons?

manrueda: Updated with the proposed name

D4N14L (approved):

manrueda:
Hi @iclanton, noticed that @D4N14L already gave the good look to the change. When could I expect this change to be release? I have a couple of developers blocked because of this issue (we have spaces on the user folder name because of corporate policies)

Thanks!

manrueda:
Hi again, sorry to insist... but can I get a timeline to when this will get released or if it needs some other changes?
@D4N14L @iclanton @octogonz 

Thanks!

common/changes/@microsoft/rush/fix-windows-support-install-scripts_2021-04-20-23-21.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix an issue where all Rush script fail to run on Windows when the repository absolute path contains a space",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue where Rush fails to run on Windows when the repository absolute path contains a space",
```

iclanton (approved):

manrueda:
@iclanton updated the change file content with your recommendation

PR #2630, opened 2021-04-21
[rush] Fix `rush install` after workspace projects have been deleted upstream

dmichon-msft:
## Summary
Ensures that `rush install` does not throw an error if the last local install contained projects that have since been deleted upstream, and the user pulls latest changes and runs `rush install`

## Details
Skips reading the existing common/temp/pnpm-workspace.yaml file when constructing what the content is expected to be. Diffing at write time is still applied.

## How it was tested
Used the rush version to reinstall in rushstack.

### DIFF
--- a/apps/rush-lib/src/logic/pnpm/PnpmWorkspaceFile.ts
+++ b/apps/rush-lib/src/logic/pnpm/PnpmWorkspaceFile.ts
@@ -60,9 +60,11 @@ export class PnpmWorkspaceFile extends BaseWorkspaceFile {
 
   /** @override */
   protected serialize(): string {
+    // Ensure stable sort order when serializing
+    Sort.sortSet(this._workspacePackages);
+
     const workspaceYaml: IPnpmWorkspaceYaml = {
-      // Ensure stable sort order when serializing
-      packages: Sort.sort(Array.from(this._workspacePackages))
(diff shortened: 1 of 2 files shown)


apps/rush-lib/src/logic/pnpm/PnpmWorkspaceFile.ts
```diff
@@ -42,15 +41,9 @@ export class PnpmWorkspaceFile extends BaseWorkspaceFile {
...
-      // Populate with the existing file, or an empty list if the file doesn't exist
-      workspaceYaml = FileSystem.exists(workspaceYamlFilename)
-        ? yamlModule.safeLoad(FileSystem.readFile(workspaceYamlFilename).toString())
-        : { packages: [] };
-    } catch (error) {
-      throw new Error(`Error reading "${workspaceYamlFilename}":${os.EOL}  ${error.message}`);
-    }
+    // Ignore any existing file since this file is generated and we need to handle deleting packages
+    // If we need to support manual customization, that should be an additional parameter for "base file"
+    const workspaceYaml: IPnpmWorkspaceYaml = { packages: [] };
```
D4N14L, line 46: May as well remove this entirely and just init an empty set below

D4N14L (approved):

PR #2633, opened 2021-04-21
[rush] Fix Rush to use new .pnpmfile.cjs filename in PNPM 6

### LINKED ISSUE #2629: [rush] pnpm version 6 renames `pnpmfile.js` to `.pnpmfile.cjs` but rush doesn't
## Summary

Since pnpm version 6, the convention for pnpm hooks filename has changed from `pnpmfile.js` to `.pnpmfile.cjs` but rush doesn't output such file inside `common/temp/`, so the hooks never get executed.

## Repro steps

0. Create a Rush project, using `useWorkspaces: true` and `pnpm` version `6.0.2`.
0. Add a hook with a message inside `common/rush/pnpmfile.js`, for example:
  ```js
    context.log('Hook for: ' + packageJson.name)
    return packageJson
}
  ```
2. Execute `rush update --full --recheck`

**Expected result:** The console outputs the `Hook from: XXX` messages

**Actual result:** The console doesn't output our messages (hooks are not executed)

## Details

This issue should be easy to fix:
> If pnpm version is >= 6.0.0 then the automatically generated file should be named `.pnpmfile.cjs` instead of `pnpmfile.js`
In order to support multiple pnpm versions with the same file, it could also output an adapter file as `.pnpmfile.cjs`:
```js
// .pnpmfile.cjs
module.exports = require('./pnpmfile')
```

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.36.2 |
| `rushVersion` from rush.json? | 5.36.2 |
| Operating system? | windows 10 |
| Node.js version (`node -v`)? | 14.16.1 |

D4N14L:
## Summary

In preparing to use PNPM 6, it was determined that the new `.pnpmfile.cjs` filename is required when copying in the pnpmfile or creating our own shim pnpmfile in workspaces. This change updates Rush to copy `.pnpmfile.cjs` for PNPM 6 and above, and continues to use `pnpmfile.js` for below. Fixes #2629 

## How it was tested

Tested by performing install and validating that `pnpm-lock.yaml` changes are included and correct in PNPM 6-enabled monorepo.

### DIFF
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -636,6 +636,7 @@ export class RushConfiguration {
     RushConfiguration._validateCommonRushConfigFolder(
       this._commonRushConfigFolder,
       this.packageManager,
+      this._packageManagerToolVersion,
       this._shrinkwrapFilename,
       this._experimentsConfiguration
     );
@@ -929,6 +930,15 @@ export class RushConfiguration {
   }
 
   /**
+   * Returns the pnpmfile filename for the version of PNPM supplied.
+   */
+  private static _getPnpmfileFilename(packageManagerToolVersion: string): string {
+    return semver.gte(packageManagerToolVersion, '6.0.0')
+      ? RushConstants.pnpmfileV6Filename
+      : RushConstants.pnpmfileV1Filename;
+  }
+
+  /**
    * If someone adds a config file in the "common/rush/config" folder, it would be a bad
    * experience for Rush to silently ignore their file simply because they misspelled the
    * filename, or maybe it's an old format that's no longer supported.  The
@@ -938,6 +948,7 @@ export class RushConfiguration {
   private static _validateCommonRushConfigFolder(
     commonRushConfigFolder: string,
     packageManager: PackageManagerName,
+    packageManagerToolVersion: string,
     shrinkwrapFilename: string,
     experiments: ExperimentsConfiguration
   ): void {
@@ -982,7 +993,7 @@ export class RushConfiguration {
 
       // If the package manager is pnpm, then also add the pnpm file to the known set.
       if (packageManager === 'pnpm') {
-        knownSet.add(RushConstants.pnpmfileFilename.toUpperCase());
+        knownSet.add(RushConfiguration._getPnpmfileFilename(packageManagerToolVersion).toUpperCase());
       }
 
       // Is the filename something we know?  If not, report an error.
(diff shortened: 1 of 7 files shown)


dmichon-msft (approved):

apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -928,6 +929,15 @@ export class RushConfiguration {
     return tempNamesByProject;
   }
 
+  /**
+   * Returns the pnpmfile filename for the version of PNPM supplied.
+   */
+  private static _getPnpmfileFilename(packageManagerToolVersion: string): string {
```
iclanton, line 935: Seems like this should be in a utility somewhere instead of in `RushConfiguration`

apps/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -392,7 +393,7 @@ export abstract class BaseInstallManager {
       const committedPnpmFilePath: string = this._rushConfiguration.getPnpmfilePath(this._options.variant);
       const tempPnpmFilePath: string = path.join(
         this._rushConfiguration.commonTempFolder,
-        RushConstants.pnpmfileFilename
+        path.basename(committedPnpmFilePath)
```
iclanton, line 396: Would it make more sense to refactor `getPnpmfilePath` in `RushConfiguration` to return the filename instead of the whole path so we don't need to peel the path back apart here?

apps/rush-lib/src/logic/installManager/RushInstallManager.ts
```diff
@@ -78,7 +78,8 @@ export class RushInstallManager extends BaseInstallManager {
    * @override
    */
   public async prepareCommonTempAsync(
-    shrinkwrapFile: BaseShrinkwrapFile | undefined
+    shrinkwrapFile: BaseShrinkwrapFile | undefined,
+    variant: string | undefined
```
iclanton, line 82: Unused?

common/changes/@microsoft/rush/user-danade-FixPnpmfile_2021-04-21-22-03.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix support for PNPM 6 and use .pnpmfile.cjs",
```
iclanton, line 5:
```suggestion
      "comment": "Fix support for pnpmfile in PNPM 6.",
```

apps/rush-lib/src/api/RushConfiguration.ts
D4N14L: The idea of keeping it here is that in general, you should be obtaining the path to the file from `RushConfiguration`, similar to how we get the shrinkwrap path, since you will also need the PNPM version which already comes from `RushConfiguration`.

apps/rush-lib/src/logic/installManager/RushInstallManager.ts
D4N14L: Yeah, not necessary

iclanton (approved):

PR #2640, opened 2021-04-23
[rush-lib] Improve user error messages for common credential issues

elliot-nelson:
## Summary

There are some corner cases while using the rush build cache feature that result in unhelpful error messages.  Improve error messages displayed to the user so they have more immediate direction.  (This PR focuses on the Azure Storage Blob provider.)

The `RestError` that can be thrown when reading the cache is now formatted with the status code and error code, in addition to error-specific detail, for example:

```
// Old message
Error getting cache entry from Azure Storage: RestError

// New message
Error getting cache entry from Azure Storage: RestError 409 PublicAccessNotPermitted

You need to configure Azure Storage SAS credentials to access the build cache.
Update the credentials by running "rush update-cloud-credentials",
or provide a SAS in the RUSH_BUILD_CACHE_WRITE_CREDENTIAL environment variable.
```

## Details

 - If you turn on cache writing, but your credentials are invalid (as in, they cannot read OR write from the cache), then the check for blob existence after the build blows up silently, causing the build to fail (this is extra confusing because the only warning printed is the warning about your bad credentials at the top of the build, which makes it look like having bad credentials is _supposed_ to fail your build, which isn't the case).  Wrapping the blob check and printing the error prevents this potential build failure.

 - If you have valid credentials, but those credentials don't have read permission, you get an unhelpful `RestError` message.  Generally this issue is because your user doesn't have permission to read data blobs from your account storage, so let's point the user in that direction.  (In addition to printing out the actual error code and name.)

 - If you have configured your Azure Storage account to require credentials for read access, but you don't have any credentials, the storage provider will default to attempting to read the cache with no credentials.  Provide the user the standard "set up your credentials" speech in this case.  (In addition to printing out the actual error code and name.)

 - If you have configured the build cache and have credentials but those credentials are invalid (a bad copy/paste into the environment variable, or maybe even you configured the cache with S3 and then switched it to Azure, or the credentials used to be valid and were invalidated manually), give the user the standard "set up your credentials" speech in this case.  (In addition to printing out the actual error code and name.)

## How it was tested

Manually tested situations by:

(1) Set up cloud credentials, then manually corrupted the saved cloud credentials in `~/.rush-user` and ran a build.

(2) Turned on cache write, set up cloud credentials, then manually corrupted the credentials and ran a build.

(3) Set up cloud credentials, removed storage blob data reader permissions in Azure Portal, then ran a build.

(4) Turned off cache write, connected to private storage account, cleared cloud credentials, then ran a build.

### DIFF
--- a/apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
+++ b/apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
@@ -100,7 +100,47 @@ export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase 
         return undefined;
       }
     } catch (e) {
-      terminal.writeWarningLine(`Error getting cache entry from Azure Storage: ${e}`);
+      const errorMessage: string =
+        'Error getting cache entry from Azure Storage: ' +
+        [e.name, e.message, e.response?.status, e.response?.parsedHeaders?.errorCode]
+          .filter((piece: string | undefined) => piece)
+          .join(' ');
+
+      if (e.response?.parsedHeaders?.errorCode === 'PublicAccessNotPermitted') {
+        // This error means we tried to read the cache with no credentials, but credentials are required.
+        // We'll assume that the configuration of the cache is correct and the user has to take action.
+        terminal.writeWarningLine(
+          `${errorMessage}\n\n` +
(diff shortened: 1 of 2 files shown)


iclanton (approved):

iclanton:
Awesome. This is a huge improvement. Thanks @elliot-nelson!

PR #2646, opened 2021-04-26
Upgrade webpack 5

scamden:
## Summary

Upgrade Webpack 5

## Details

The latest webpack fixes a regression where modules couldn't be resolved when they contained a # in the path. This was throwing a warning when using the new webpack persistent cache and causing rush build to fail.

## How it was tested

Ran build with no failures

Side Note: Would it be possible to read the webpack module from the rig if it exists instead of forcing a version?

### DIFF
--- a/common/changes/@rushstack/heft-webpack5-plugin/pr-upgrade-webpack-5_2021-04-26-17-07.json
+++ b/common/changes/@rushstack/heft-webpack5-plugin/pr-upgrade-webpack-5_2021-04-26-17-07.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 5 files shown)


iclanton (approved):

iclanton:
>Would it be possible to read the webpack module from the rig if it exists instead of forcing a version?

Technically yes, but you'd still need to include either the webpack4 or webpack5 plugin since the API changed in Webpack 5. I'd think at that point it'd just be easier for project maintainers to only have to include one dependency.

scamden:
The nice thing though would be to be able to get changes to webpack like this PR without having to submit a PR haha. Could we fall back? Use the plugin's version when we don't find one in the rig similar to jest or typescript? Maybe just error if the major version is incorrect?

PR #2649, opened 2021-04-27
[Rush] Use simpler and more accurate check before skipping installs in workspaces

D4N14L:
## Summary

A couple issues with the install state check were encountered recently in Rush workspaces:
- When only deletions are performed in package.json, the check would errantly allow for skipping install because we would only validate that the packages in package.json are present (and compatible) in the lockfile
- Merge issues in git repos that allow non-binary merge that became apparent after fixing a bug that would incidentally force merge conflicts (#2610)

## Details

The method used to determine whether or not the shrinkwrap is up-to-date was made to provide the same functionality as PNPM, which can be found here: [https://github.com/pnpm/pnpm/blob/main/packages/lockfile-utils/src/satisfiesPackageManifest.ts](https://github.com/pnpm/pnpm/blob/main/packages/lockfile-utils/src/satisfiesPackageManifest.ts)

This will allow for much increased accuracy on whether or not the shrinkwrap needs to be updated, reducing false positives and dev headaches with trying to force Rush to update. While we do provide a method to override this check (`--recheck`), this is not the most obvious. I briefly discussed with @iclanton if we should even be bothering with the "up-to-date" check in the case of updates, but this functionality is still useful and time-saving here as long as we assume that the check is _correct_.

## How it was tested

Tested in a large monorepo that contained a bad merge conflict that was passing through the existing "up-to-date" check when running `rush install` even though it was out-of-date. Once the changes were validated to match what PNPM reports, I ran `rush update` to update the shrinkwrap, and again ran `rush install` to validate that the check passing as expected.

### DIFF
--- a/apps/rush-lib/src/logic/base/BaseShrinkwrapFile.ts
+++ b/apps/rush-lib/src/logic/base/BaseShrinkwrapFile.ts
@@ -10,7 +10,7 @@ import { IShrinkwrapFilePolicyValidatorOptions } from '../policy/ShrinkwrapFileP
 import { PackageManagerOptionsConfigurationBase } from '../../api/RushConfiguration';
 import { PackageNameParsers } from '../../api/PackageNameParsers';
 import { IExperimentsJson } from '../../api/ExperimentsConfiguration';
-import { PackageJsonEditor } from '../../api/PackageJsonEditor';
+import { RushConfigurationProject } from '../../api/RushConfigurationProject';
 
 /**
  * This class is a parser for both npm's npm-shrinkwrap.json and pnpm's pnpm-lock.yaml file formats.
@@ -126,7 +126,7 @@ export abstract class BaseShrinkwrapFile {
    *
    * @virtual
    */
-  public abstract isWorkspaceProjectModified(workspaceKey: string, packageJson: PackageJsonEditor): boolean;
+  public abstract isWorkspaceProjectModified(project: RushConfigurationProject): boolean;
 
   /** @virtual */
   protected abstract serialize(): string;
--- a/apps/rush-lib/src/logic/base/BaseShrinkwrapFile.ts
+++ b/apps/rush-lib/src/logic/base/BaseShrinkwrapFile.ts
@@ -10,6 +10,7 @@ import { IShrinkwrapFilePolicyValidatorOptions } from '../policy/ShrinkwrapFileP
 import { PackageManagerOptionsConfigurationBase } from '../../api/RushConfiguration';
 import { PackageNameParsers } from '../../api/PackageNameParsers';
 import { IExperimentsJson } from '../../api/ExperimentsConfiguration';
+import { PackageJsonEditor } from '../../api/PackageJsonEditor';
 
 /**
  * This class is a parser for both npm's npm-shrinkwrap.json and pnpm's pnpm-lock.yaml file formats.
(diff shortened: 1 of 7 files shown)


apps/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -505,75 +506,82 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
-      ...(projectImporter.optionalDependencies || {}),
-      ...(projectImporter.dependencies || {}),
-      ...(projectImporter.devDependencies || {})
-    };
-    if (!allDependencies.hasOwnProperty(packageName)) {
-      return undefined;
+    // First, get the unique package names and map them to package versions.
+    const dependencyVersions: Map<string, PackageJsonDependency> = new Map();
+    for (const packageDependency of [...packageJson.dependencyList, ...packageJson.devDependencyList]) {
+      // We will also filter out peer dependencies since these are not included by the shrinkwrap.
```
dmichon-msft, line 526: Probably better to state that they are not installed at development time.

apps/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -505,75 +506,82 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
+    if (dependencyVersions.size !== Object.keys(importer.specifiers).length) {
+      return true;
     }
 
-    if (!packageDescription.dependencies.hasOwnProperty(dependencyName)) {
-      return undefined;
+    // Finally, validate that all values in the importer are also present in the dependency list.
+    for (const [importerPackageName, importerVersionSpecifier] of Object.entries(importer.specifiers)) {
+      const foundDependency: PackageJsonDependency | undefined = dependencyVersions.get(importerPackageName);
+      if (!foundDependency || foundDependency.version !== importerVersionSpecifier) {
```
dmichon-msft, line 579: We might want to run the package.json through pnpmfile.js before this test. Otherwise we give people confusing results if their pnpmfile overrides a local package's dependencies (i.e. pnpm is happy, rush is not)

D4N14L: We already check to see if pnpmfile has been modified and prevent skipping installs if the file is updated. We also officially document that running `rush update --full` after updating pnpmfile.js is recommended in the template pnpmfile that gets dropped after running `rush init`, as well as on the rushjs.io website: https://rushjs.io/pages/configs/pnpmfile_js/

I'm going to leave this as-is, since the worst that can happen is someone runs into this case and has to run `rush update` now that the packages are changed on-disk. PNPM will still block install with an issue saying that the shrinkwrap is out of date, and update will still be able to be run.

apps/rush-lib/src/logic/base/BaseShrinkwrapFile.ts
```diff
@@ -148,11 +120,13 @@ export abstract class BaseShrinkwrapFile {
...
-    dependencySpecifier: DependencySpecifier,
-    workspaceKey: string
-  ): DependencySpecifier | undefined;
+  /**
+   * Returns whether or not the workspace specified by the shrinkwrap matches the state of
+   * a given package.json. Returns true if any dependencies are not aligned with the shrinkwrap.
+   *
+   * @virtual
+   */
+  public abstract isWorkspaceProjectModified(workspaceKey: string, packageJson: PackageJsonEditor): boolean;
```
iclanton, line 129: What happens if the repo isn't using workspaces?

iclanton (approved):

apps/rush-lib/src/logic/base/BaseShrinkwrapFile.ts
D4N14L: If it isn't using workspaces, it uses the old logic. It's not ideal, of course, but the changes I made won't be usable in non-workspace world due to the lack of `importers` in the pnpm-lock.yaml ... we'd be probably better off with the existing logic over there.

apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
```diff
@@ -253,6 +236,12 @@ export class WorkspaceInstallManager extends BaseInstallManager {
           )
         );
       }
+
+      // Now validate that the shrinkwrap file matches what is in the package.json
+      if (shrinkwrapFile && shrinkwrapFile.isWorkspaceProjectModified(rushProject)) {
```
iclanton, line 241:
```suggestion
      if (shrinkwrapFile?.isWorkspaceProjectModified(rushProject)) {
```

apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
```diff
@@ -253,6 +236,12 @@ export class WorkspaceInstallManager extends BaseInstallManager {
...
+        shrinkwrapWarnings.push(`Modified package.json found for "${rushProject.packageName}"`);
```
iclanton, line 242:
```suggestion
        shrinkwrapWarnings.push(`Modified dependencies for "${rushProject.packageName}" do not match the current shinkwrap.`);
```

apps/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -514,15 +515,20 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
   }
 
   /** @override */
-  public isWorkspaceProjectModified(workspaceKey: string, packageJson: PackageJsonEditor): boolean {
+  public isWorkspaceProjectModified(project: RushConfigurationProject): boolean {
+    const workspaceKey: string = this.getWorkspaceKeyByPath(
```
iclanton, line 519: Can `getWorkspaceKeyByPath` be private?

iclanton (approved):

apps/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
D4N14L: Punting to future change

PR #2655, opened 2021-04-29
[rush-lib] Fix a regression in the S3 cloud build cache provider

elliot-nelson:
## Summary

As reported in the zulipchat: a regression was introduced in https://github.com/microsoft/rushstack/pull/2614 that broke the signature generation for S3 requests (the header name signature delimiter was `,` instead of `;`).  Update the code and snapshots to correct.

(As the PR template suggests... `Pushing additional commits with "small" fixes often invalidates testing.` 😢 )

## Details

N/A (this fix is for the S3 option of the experimental cloud build cache feature).

## How it was tested

- Manually confirmed that _before_ #2614, could get/set from cache correctly; _afterwards_, received AWS Error 400.
- Manually confirmed that after this PR, could get/set from cache correctly again.


iclanton (approved):

iclanton:
This has been published in Rush 5.45.6.

PR #2663, opened 2021-04-29
[Rush] Refactor per-project shrinkwrap to remove implementation details from the install manager

D4N14L:
## Summary

Refactors the per-project shrinkwrap into it's own class that can be used with an implementation for each package manager. Also moves orphaned project detection into the shrinkwrap.

## Details

Package-manager specific implementation details were being placed in a base file extended by other package managers. This refactor is an effort to remove this implementation detail from the base level files in an attempt to better generalize Rush to package managers. This will allow features developed for other package managers (ex. yarn workspaces) to be implemented more easily.

## How it was tested

- [x]  Tested against a large workspaces repo and compared the `shrinkwrap-deps.json` file generated for projects and found no diff before and after.

- [x] Tested against a legacy rush repo and compared the `shrinkwrap-deps.json` file generated for projects and found no diff before and after.


apps/rush-lib/src/logic/PackageChangeAnalyzer.ts
```diff
@@ -105,22 +105,20 @@ export class PackageChangeAnalyzer {
       const projectDependencyManifestPaths: string[] = [];
 
       for (const project of this._rushConfiguration.projects) {
-        const dependencyManifestFilePath: string = PnpmProjectDependencyManifest.getFilePathForProject(
-          project
-        );
-        const relativeDependencyManifestFilePath: string = Path.convertToSlashes(
-          path.relative(this._rushConfiguration.rushJsonFolder, dependencyManifestFilePath)
+        const projectShrinkwrapFilePath: string = BaseProjectShrinkwrapFile.getFilePathForProject(project);
```
iclanton, line 108: Couldn't you just have this return the relative path?

apps/rush-lib/src/logic/UnlinkManager.ts
```diff
@@ -61,12 +61,10 @@ export class UnlinkManager {
...
-      const projectDependencyManifestFilePath: string = PnpmProjectDependencyManifest.getFilePathForProject(
-        rushProject
-      );
-      if (FileSystem.exists(projectDependencyManifestFilePath)) {
-        console.log(`Deleting ${projectDependencyManifestFilePath}`);
-        FileSystem.deleteFile(projectDependencyManifestFilePath);
+      const projectShrinkwrapFilePath: string = BaseProjectShrinkwrapFile.getFilePathForProject(rushProject);
+      if (FileSystem.exists(projectShrinkwrapFilePath)) {
+        console.log(`Deleting ${projectShrinkwrapFilePath}`);
+        FileSystem.deleteFile(projectShrinkwrapFilePath);
```
iclanton, line 67: Couldn't you use `deleteIfExistsAsync`?

apps/rush-lib/src/logic/base/BaseProjectShrinkwrapFile.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  /**
+   * Generate and write the project shrinkwrap file to <project>/.rush/temp/shrinkwrap-deps.json.
+   *
+   * @virtual
+   */
+  public abstract updateProjectShrinkwrapAsync(): Promise<void>;
+
+  public get projectShrinkwrapFilename(): string {
```
iclanton, line 49:
```suggestion
  public get projectShrinkwrapFilePath(): string {
```

apps/rush-lib/src/logic/base/BaseProjectShrinkwrapFile.ts
```diff
@@ -0,0 +1,60 @@
...
+    return this._projectShrinkwrapFilename;
+  }
+
+  protected get project(): RushConfigurationProject {
+    return this._project;
+  }
+
+  protected get shrinkwrapFile(): BaseShrinkwrapFile {
+    return this._shrinkwrapFile;
+  }
```
iclanton, line 59: I've started writing these as just `public readonly project: RushConfigurationProject`.

It doesn't *enforce* that it's readonly, but TS will complain if anyone tries to assign to it.

apps/rush-lib/src/logic/base/BaseProjectShrinkwrapFile.ts
```diff
@@ -0,0 +1,60 @@
...
+  public static getFilePathForProject(project: RushConfigurationProject): string {
+    return path.join(project.projectRushTempFolder, RushConstants.projectDependencyManifestFilename);
+  }
+
+  /**
+   * If the <project>/.rush/temp/shrinkwrap-deps.json file exists, delete it. Otherwise, do nothing.
+   */
+  public deleteIfExistsAsync(): Promise<void> {
+    return FileSystem.deleteFileAsync(this._projectShrinkwrapFilename, { throwIfNotExists: false });
+  }
```
iclanton, line 40:
```suggestion
  public async deleteIfExistsAsync(): Promise<void> {
    await FileSystem.deleteFileAsync(this._projectShrinkwrapFilename, { throwIfNotExists: false });
  }
```

apps/rush-lib/src/logic/base/BaseShrinkwrapFile.ts
```diff
@@ -105,20 +106,38 @@ export abstract class BaseShrinkwrapFile {
...
+  public abstract getProjectShrinkwrap(
+    project: RushConfigurationProject
+  ): BaseProjectShrinkwrapFile | undefined;
+
+  /**
+   * Returns whether or not the current state of the shrinkwrap file is compatible with workspace installs.
+   *
+   * @virtual
+   **/
+  public abstract isWorkspaceCompatible(): boolean;
```
iclanton, line 140: Can this also just be an abstract readonly property?

apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
```diff
@@ -420,18 +421,23 @@ export class WorkspaceInstallManager extends BaseInstallManager {
...
+
+    const projectShrinkwrapPromises: Promise<void>[] = this.rushConfiguration.projects.map((x) => {
+      const projectShrinkwrapFile:
+        | BaseProjectShrinkwrapFile
+        | undefined = tempShrinkwrapFile.getProjectShrinkwrap(x);
+      return projectShrinkwrapFile?.updateProjectShrinkwrapAsync() ?? Promise.resolve();
+    });
+    console.log(`have promises (${new Date().getTime()})`);
+    await Promise.all(projectShrinkwrapPromises);
+    console.log(`done writes (${new Date().getTime()})`);
```
iclanton, line 440: Clean up console.logs?

apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
```diff
@@ -420,18 +421,23 @@ export class WorkspaceInstallManager extends BaseInstallManager {
...
+      this.rushConfiguration.pnpmOptions,
+      this.rushConfiguration.tempShrinkwrapFilename
+    )!;
+
+    const projectShrinkwrapPromises: Promise<void>[] = this.rushConfiguration.projects.map((x) => {
+      const projectShrinkwrapFile:
+        | BaseProjectShrinkwrapFile
+        | undefined = tempShrinkwrapFile.getProjectShrinkwrap(x);
+      return projectShrinkwrapFile?.updateProjectShrinkwrapAsync() ?? Promise.resolve();
+    });
```
iclanton, line 437:
```suggestion
    const projectShrinkwrapPromises: Promise<void>[] = this.rushConfiguration.projects.map(async (x) => {
      const projectShrinkwrapFile:
        | BaseProjectShrinkwrapFile
        | undefined = tempShrinkwrapFile.getProjectShrinkwrap(x);
      await projectShrinkwrapFile?.updateProjectShrinkwrapAsync();
    });
```

apps/rush-lib/src/logic/npm/NpmShrinkwrapFile.ts
```diff
@@ -120,13 +121,13 @@ export class NpmShrinkwrapFile extends BaseShrinkwrapFile {
   }
 
   /** @override */
-  public getWorkspaceKeys(): ReadonlyArray<string> {
-    throw new InternalError('Not implemented');
+  public getProjectShrinkwrap(project: RushConfigurationProject): BaseProjectShrinkwrapFile | undefined {
+    return undefined;
```
iclanton, line 125: If we don't have a per-project shrinkwrap, the incremental and cached builds should look at the repo-wide shrinkwrap. Can we confirm that we do this?

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
```diff
@@ -0,0 +1,238 @@
...
+export class PnpmProjectShrinkwrapFile extends BaseProjectShrinkwrapFile {
+  /**
+   * Generate and write the project shrinkwrap file to <project>/.rush/temp/shrinkwrap-deps.json.
+   * @returns True if the project shrinkwrap was created or updated, false otherwise.
+   */
+  public async updateProjectShrinkwrapAsync(): Promise<void> {
+    // If the feature is not enabled, clean up the shrinkwrap and return
+    if (
+      this.project.rushConfiguration.experimentsConfiguration.configuration
+        .legacyIncrementalBuildDependencyDetection
```
iclanton, line 28: Weren't you going to get rid of this experiment?

apps/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -493,34 +495,56 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
+
+  /** @override */
+  public getImporterKeyByPath(workspaceRoot: string, projectFolder: string): string {
     return Path.convertToSlashes(path.relative(workspaceRoot, projectFolder));
   }
 
-  public getWorkspaceImporter(importerPath: string): IPnpmShrinkwrapImporterYaml | undefined {
-    return BaseShrinkwrapFile.tryGetValue(this._shrinkwrapJson.importers, importerPath);
+  public getImporter(importerKey: string): IPnpmShrinkwrapImporterYaml | undefined {
+    return BaseShrinkwrapFile.tryGetValue(this._shrinkwrapJson.importers, importerKey);
```
iclanton, line 534: Might be faster to populate `Map` objects for parts of `_shrinkwrapJson`.

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
```diff
@@ -0,0 +1,238 @@
...
+  protected generateWorkspaceProjectShrinkwrapMap(): Map<string, string> | undefined {
+    // Obtain the workspace importer from the shrinkwrap, which lists resolved dependencies
+    const importerKey: string = this.shrinkwrapFile.getImporterKeyByPath(
+      this.project.rushConfiguration.commonTempFolder,
+      this.project.projectFolder
+    );
+    const importer: IPnpmShrinkwrapImporterYaml | undefined = this.shrinkwrapFile.getImporter(importerKey);
+    if (!importer) {
+      // It's not in here. This is possible when perfoming filtered installs
+      return undefined;
```
iclanton, line 49: Should this check if this is a project that wasn't installed? And/or, should we never try to write the per-project shrinkwrap for projects that aren't in the installation set?

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
```diff
@@ -0,0 +1,238 @@
...
+    let integrity: string | undefined = shrinkwrapEntry?.resolution?.integrity;
+    if (!integrity) {
+      // git dependency specifiers do not have an integrity entry. Instead, they specify the tarball field.
+      // So instead, we will hash the contents of the dependency entry and use that as the integrity hash.
+      // Ex:
+      // github.com/chfritz/node-xmlrpc/948db2fbd0260e5d56ed5ba58df0f5b6599bbe38:
+      //   ...
+      //   resolution:
+      //     tarball: 'https://codeload.github.com/chfritz/node-xmlrpc/tar.gz/948db2fbd0260e5d56ed5ba58df0f5b6599bbe38'
+      const sha256Digest: string = crypto
```
iclanton, line 131: You should see how often this gets called.

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    // Add the optional dependencies of the dependency, and don't blow up if they don't exist
+    for (const [name, version] of Object.entries(shrinkwrapEntry.optionalDependencies || {})) {
+      this._addDependencyRecursive(
+        projectShrinkwrapMap,
+        name,
+        version,
+        shrinkwrapEntry,
+        (throwIfShrinkwrapEntryMissing = false)
```
iclanton, line 161:
```suggestion
        /* throwIfShrinkwrapEntryMissing */ false
```

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
```diff
@@ -0,0 +1,238 @@
...
+      );
+    }
+
+    // When using workspaces, hoisting of peer dependencies to a singular top-level project is not possible.
+    // Therefore, all packages that are consumed should be specified in the dependency tree. Given this, there
+    // is no need to look for peer dependencies, since it is simply a constraint to be validated by the
+    // package manager.
+    if (!this.shrinkwrapFile.isWorkspaceCompatible()) {
```
iclanton, line 169: This should probably just get passed into this function instead of calling this over and over.

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
```diff
@@ -0,0 +1,238 @@
...
+      // Check to see if the peer dependency is satisfied with the current shrinkwrap
+      // entry. If not, check the parent shrinkwrap entry. Finally, if neither have
+      // the specified dependency, check that the parent mentions the dependency in
+      // it's own peer dependencies. If it is, we can rely on the package manager and
+      // make the assumption that we've already found it further up the stack.
+      if (
+        (shrinkwrapEntry.dependencies && shrinkwrapEntry.dependencies.hasOwnProperty(peerDependencyName)) ||
+        (parentShrinkwrapEntry &&
+          parentShrinkwrapEntry.dependencies &&
+          parentShrinkwrapEntry.dependencies.hasOwnProperty(peerDependencyName)) ||
```
iclanton, line 194:
```suggestion
        (parentShrinkwrapEntry?.dependencies?.hasOwnProperty(peerDependencyName)) ||
```

apps/rush-lib/src/logic/base/BaseProjectShrinkwrapFile.ts
D4N14L: Did that for some, kept shrinkwrapFile this way to be overridable so that we can just do a typecast of the property in the packagemanager-specific lockfile.

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
D4N14L: Very rarely, thankfully. Only for packages w/o integrity, which only happens in rare situations (like github-sourced packages)

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
D4N14L: Made this a field so now it's no overhead :)

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
D4N14L: Talked offline, but basically we should be getting out of the business of validating the lockfile and throwing when it doesn't match what we expect, mostly because we keep getting bitten by this. By returning `undefined` here, it gives us the ability to quietly fail, and then we delete the shrinkwrap for projects not in the installation set in order to clean up the current state.

apps/rush-lib/src/logic/UnlinkManager.ts
D4N14L: We don't have a shrinkwrap at this point, which is required to create the `ProjectShrinkwrap`. Since we only need the path here and the shrinkwrap would just get thrown away, I'm going to leave this as-is. This will go away when the linking phase is removed with legacy Rush installs as well, so it will be short-lived.

octogonz (approved):

Issue #2669, opened 2021-05-02
[rush] Build cache onboarding for developer is rough (if not publicly readable)


elliot-nelson:
## Summary

If you are rolling out a new _cloud build cache_ for your monorepo and it does _not_ allow public read access, the onboarding story is a little rough for local developers.  Maybe we can improve the output to make it still visible, but not so obnoxious.

## Details

Enable a cloud build cache, then run `rush build` for the first time on a machine that has no cloud credentials yet.  You'll see spam like this for every project in the monorepo:

```console
==[ @###/###-######-######## ]==================================[ 41 of 51 ]==
Error getting cache entry from Azure Storage: RestError 409 PublicAccessNotPermitted

You need to configure Azure Storage SAS credentials to access the build cache.
Update the credentials by running "rush update-cloud-credentials", 
or provide a SAS in the RUSH_BUILD_CACHE_WRITE_CREDENTIAL environment variable.

@###/###-#######-######## was skipped.

==[ @###/######-####### ]========================================[ 42 of 51 ]==
Error getting cache entry from Azure Storage: RestError 409 PublicAccessNotPermitted

You need to configure Azure Storage SAS credentials to access the build cache.
Update the credentials by running "rush update-cloud-credentials", 
or provide a SAS in the RUSH_BUILD_CACHE_WRITE_CREDENTIAL environment variable.

@###/######-####### completed successfully in 1 minute 26.2 seconds.

(etc. etc. etc.)
```

We want the user to know that they are missing out on the cloud build cache and should follow the instructions to configure it, but maybe we could notify them _once_ and then remind them at the end?  Or use some other mechanism to let them know that they have a new feature to take advantage of.

## Standard questions

Please answer these questions to help us investigate your issue more quickly:

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.45.5  |
| `rushVersion` from rush.json? | 5.45.5  |
| `useWorkspaces` from rush.json? | true  |
| Operating system? | Mac  |
| Would you consider contributing a PR? | Yes |
| Node.js version (`node -v`)? | 12.17.0 |

elliot-nelson:
Closing.  (The build cache has gone through several iterations since I filed this, and these specific details aren't relevant anymore.)

PR #2680, opened 2021-05-07
[rush-lib] Certain build cache warnings are emitted only once

elliot-nelson:
## Summary

Implements #2669.

There are certain build cache warnings that don't (and shouldn't) fail the build, but aren't project-specific, which means that they will likely be hit by every project that attempts to hit the cloud cache.  These warnings are now emitted the first time they are hit, and then suppressed (written to verbose) after that.

## Details

Because `BuildCacheProvider` instances are created when the build cache configuration is first initialized, and reused by each `ProjectBuilder -> ProjectBuildCache` task, it's pretty easy in this instance to track whether a warning has been emitted across the potentially parallel tasks.

These particular warnings might be even better suited to a `Utilities.printMessageInBox` effect, but we don't want to overuse that and (right now) it isn't designed to print `writeWarning`.

In addition to only printing the warnings once, these warnings now hide the previously displayed _error message_ (like "409 PublicAccessNotPermitted"), and replace it with a note that the user is missing out on a faster build.  Open to wording suggestions here!  In this scenario, what's happening is that the build administration team is attempting to roll out the new build cache, and developers are all finding out about it at different times (some folks will have the latest changes minutes later, others may take a week to catch up).  Hopefully this wording change helps get folks onboarded _without_ pinging build administrators about the "error" they are seeing.

## How it was tested

Manually tested the different warning situations, confirmed output now results in only ONE copy of the warning on the first encountered project (and then again at the bottom of the build).

```console
==[ @###/######-############-### ]===============================[ 17 of 51 ]==
@###/######-############-### was restored from the build cache.

==[ @###/####-##### ]============================================[ 18 of 51 ]==
You could be missing out on a faster build!

You need to configure Azure Storage SAS credentials to access the build cache.
Update the credentials by running "rush update-cloud-credentials", 
or provide a SAS in the RUSH_BUILD_CACHE_WRITE_CREDENTIAL environment variable.

==[ @###/######-############-### ]===============================[ 19 of 51 ]==
@###/######-############-### completed successfully in 20.48 seconds.
```

### DIFF
--- a/apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
+++ b/apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
@@ -43,6 +43,12 @@ export interface IAzureStorageBuildCacheProviderOptions {
 
 const SAS_TTL_MILLISECONDS: number = 7 * 24 * 60 * 60 * 1000; // Seven days
 
+enum CacheWarningType {
+  CacheCredentialsRequired = 'CacheCredentialsRequired',
+  CacheCredentialsInvalid = 'CacheCredentialsInvalid',
+  CacheCredentialsNotAuthorized = 'CacheCredentialsNotAuthorized'
+}
+
 export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase {
   private readonly _storageAccountName: string;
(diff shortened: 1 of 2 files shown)


elliot-nelson:
Closing for now - I think there's a different path we might follow for on boarding local cache users.

PR #2690, opened 2021-05-11
[heft] Fix the relative path in the "sources" parameter in .js.map files

iclanton:
## Summary

Currently there is an issue with the `"sources"` parameter in `.js.map` files emitted to `lib` folders by Heft. This PR fixes the relative paths emitted to that parameter.

## Details

Currently, the output to `/build-tests/heft-node-everything-test/lib/index.js.map` looks like this:

```JSON
{
  "version": 3,
  "file": "index.js",
  "sourceRoot": "",
  "sources": ["../../../src/index.ts"],
  "names": [],
  "mappings": ";AAAA,4FAA4F;AAC5F,2DAA2D;;;AAE3D;;GAEG;AACH,MAAa,SAAS;CAAG,CAAC,kCAAkC;AAA5D,8BAAyB",
  "sourcesContent": [
    "// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.\r\n// See LICENSE in the project root for license information.\r\n\r\n/**\r\n * @public\r\n */\r\nexport class TestClass {} // tslint:disable-line:export-name\r\n"
  ]
}
```

Notice that the path (`../../../src/index.ts`) in the `"sources"` property does not point to the correct original TS source file, relative to the `lib/index.js.map` file. Instead, it's relative to the `.heft/build-cache/lib/index.js.map` file. This can cause problems for tools like Webpack's `source-map-loader`.

This PR makes the TypeScript unaware that it's emitting files to something other than the `outDir` specified in `tsconfig.json`, so it produces correct relative paths in sourcemaps.

## How it was tested

Built the `heft-node-everything-test` project locally and confirmed that this issue no longer repros.

### DIFF
--- a/apps/heft/src/plugins/TypeScriptPlugin/EmitFilesPatch.ts
+++ b/apps/heft/src/plugins/TypeScriptPlugin/EmitFilesPatch.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
+import * as path from 'path';
 import { InternalError } from '@rushstack/node-core-library';
 import type * as TTypescript from 'typescript';
 import {
@@ -39,8 +40,9 @@ export interface ICachedEmitModuleKind {
 export class EmitFilesPatch {
   private static _patchedTs: ExtendedTypeScript | undefined = undefined;
 
-  // eslint-disable-next-line
-  private static _baseEmitFiles: any | undefined = undefined;
+  private static _baseEmitFiles: any | undefined = undefined; // eslint-disable-line
+  private static _originalOutDir: string | undefined = undefined;
+  private static _redirectedOutDir: string | undefined = undefined;
 
   public static install(
     ts: ExtendedTypeScript,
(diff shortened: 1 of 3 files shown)


apps/heft/src/plugins/TypeScriptPlugin/EmitFilesPatch.ts
```diff
@@ -191,6 +197,30 @@ export class EmitFilesPatch {
...
+    if (!EmitFilesPatch.isInstalled) {
+      throw new InternalError(
+        'EmitFilesPatch.getRedirectedFilePath() cannot be used unless the patch is installed'
+      );
+    }
+
+    // Redirect from "path/to/lib" --> "path/to/.heft/build-cache/lib"
+    let redirectedFilePath: string = filePath;
+    if (EmitFilesPatch._redirectedOutDir !== undefined) {
+      if (Path.isUnderOrEqual(filePath, EmitFilesPatch._originalOutDir!)) {
```
dmichon-msft, line 210: This function is extraordinarily expensive, can we fix?

apps/heft/src/plugins/TypeScriptPlugin/EmitFilesPatch.ts
```diff
@@ -191,6 +197,30 @@ export class EmitFilesPatch {
...
+        redirectedFilePath = path.resolve(
+          EmitFilesPatch._redirectedOutDir,
+          path.relative(EmitFilesPatch._originalOutDir!, filePath)
+        );
+      } else {
+        // The compiler is writing some other output, for example:
```
dmichon-msft, line 216: Is this the only output that isn't written to the output directory? Seems likely. In which case, can we just check against that one file?

iclanton: Currently, I believe it is. But that may change in the future.

dmichon-msft (approved):

PR #2693, opened 2021-05-12
[rush] Remove support for PNPM < 5.0.0

D4N14L:
## Summary

Remove support for PNPM < 5.0.0. There were a few legacy features that were worked around with newer versions of PNPM. Since PNPM is now at v6, we should be good to remove support for many special-cased situations required for older versions of PNPM.

## Details

- Sets minimum PNPM version to 5.0.0
- All places where the packageManagerToolVersion was referenced for PNPM are now removed, with the exception of `PnpmLinkManager._getPathToLocalInstallation(...)` which is used for legacy Rush installs, and `PnpmPackageManager` constructor which is used to determine the pnpmfile name (changed in PNPM 6)
- `resolutionStrategy` was completely removed, with the exception of the schema which was done to provide documentation for the feature deprecation and to avoid breaking changes due to schema validation failures. Should be removed in next major Rush bump
- Tests updated to target newer versions of PNPM

## How it was tested

Tested by running against simple repo with old PNPM version, confirmed that Rush commands will not run due to older version of PNPM.


common/changes/@microsoft/rush/user-danade-RemovePnpm4_2021-05-12-19-04.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Remove support for PNPM < 5.0.0 and deprecate \"resolutionStrategy\" option",
```
iclanton, line 5:
```suggestion
      "comment": "Remove support for PNPM < 5.0.0 and remove the \"resolutionStrategy\" option",
```

iclanton (approved):

PR #2696, opened 2021-05-13
[node-core-library] Add `ignoreUndefinedValues` option to JsonFile serialization APIs

dmichon-msft:
## Summary
Adds a configuration option `ignoreUndefinedValues` to the JsonFile serialization APIs to make their behavior consistent with `JSON.stringify()` for the common case of setting values to be discarded as `undefined`. This is generally preferred because the `delete` operator is notoriously bad for performance.

## Details
The underlying APIs already drop keys with `undefined` values, the JsonFile class added an extra validation layer in front of serialization that turned this standard behavior into an error. Passing `ignoreUndefinedValues: true` in the options simply turns off this extra validation and defers to `JSON.stringify()` or `jju.stringify()`.

## How it was tested
Added unit test cases for keys with `undefined` as their value and the `ignoreUndefinedValues: true` option.

Edited: `dropUndefinedValues` -> `ignoreUndefinedValues`

### DIFF
--- a/common/changes/@rushstack/node-core-library/json-file-undefined_2021-05-13-21-25.json
+++ b/common/changes/@rushstack/node-core-library/json-file-undefined_2021-05-13-21-25.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/node-core-library",
+      "comment": "Add `dropUndefinedValues` option to JSONFile to discard keys with undefined values during serialization, i.e. the standard behavior of JSON.stringify() and other JSON serializers.",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@rushstack/node-core-library",
+  "email": "dmichon-msft@users.noreply.github.com"
+}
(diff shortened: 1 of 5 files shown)


common/reviews/api/node-core-library.api.md
```diff
@@ -424,6 +424,7 @@ export interface IJsonFileSaveOptions extends IJsonFileStringifyOptions {
 
 // @public
 export interface IJsonFileStringifyOptions {
+    dropUndefinedValues?: boolean;
```
D4N14L, line 427:
```suggestion
    ignoreUndefinedValues?: boolean;
```

D4N14L (approved):

common/changes/@rushstack/node-core-library/json-file-undefined_2021-05-13-21-25.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/node-core-library",
+      "comment": "Add `dropUndefinedValues` option to JSONFile to discard keys with undefined values during serialization, i.e. the standard behavior of JSON.stringify() and other JSON serializers.",
```
iclanton, line 5:
```suggestion
      "comment": "Add `dropUndefinedValues` option to JsonFile to discard keys with undefined values during serialization, this is the standard behavior of `JSON.stringify()` and other JSON serializers.",
```

iclanton:
This looks good to me, but I like @D4N14L's suggested property name better.

iclanton (approved):

PR #2711, opened 2021-05-25
Get Jest to run on non-TypeScript project

idahogurl:


iclanton:
This may be able to be covered by @D4N14L's PR that is moving Jest into its own project.

I think the best option would be to add an property to the `build.properties` object called something along the lines of `isTypescriptProject: boolean | undefined`.

iclanton:
#2727 is in, so you should be able to use that property now.

D4N14L:
I've now integrated this check into my PR here: #2721

PR #2719, opened 2021-05-27
Add Jest data sourced from Typescript into the build properties

D4N14L:
## Summary

Add Jest data to the `IBuildProperties`, such that the `JestPlugin` can be split out from the parent `@rushstack/heft` package and into it's own.

## Details

In attempting to split the `JestPlugin` out into it's own package, we will need to set some build properties that can be passed through the `HeftSession` into the plugin. Since the split-out `JestPlugin` will need to build using the `heft-node-rig`, which itself will _also_ depend on `JestPlugin` (to include in the default rig), we need to make both `heft` and `heft-node-rig` be cyclic dependencies of the new package. This change is being made in advance of those changes (along with a publish run of `heft`) in order to prepare to split JestPlugin out.

## How it was tested

Build and validation in PR. Change is simply adding (not moving) data to a new place that is not yet consumed elsewhere.

### DIFF
--- a/apps/heft/src/stages/BuildStage.ts
+++ b/apps/heft/src/stages/BuildStage.ts
@@ -125,6 +125,7 @@ export class BuildStageHooks extends StageHooksBase<IBuildStageProperties> {
  * @public
  */
 export interface IBuildStageProperties {
+  // Input
   production: boolean;
   lite: boolean;
   locale?: string;
@@ -132,6 +133,10 @@ export interface IBuildStageProperties {
   watchMode: boolean;
   serveMode: boolean;
   webpackStats?: unknown;
+
+  // Output
+  emitFolderNameForTests?: string;
+  emitExtensionForTests?: '.js' | '.cjs' | '.mjs';
 }
 
 /**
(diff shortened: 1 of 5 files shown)


apps/heft/src/stages/BuildStage.ts
```diff
@@ -125,13 +125,18 @@ export class BuildStageHooks extends StageHooksBase<IBuildStageProperties> {
...
   lite: boolean;
   locale?: string;
   maxOldSpaceSize?: string;
   watchMode: boolean;
   serveMode: boolean;
   webpackStats?: unknown;
+
+  // Output
+  emitFolderNameForTests?: string;
+  emitExtensionForTests?: '.js' | '.cjs' | '.mjs';
```
iclanton, line 139:
```suggestion

  // Output
  /**
   * @beta
   */
  emitFolderNameForTests?: string;
  /**
   * @beta
   */
  emitExtensionForTests?: '.js' | '.cjs' | '.mjs';
```

iclanton (approved):

PR #2727, opened 2021-06-04
[heft] Populate `IBuildStageProperties.isTypeScriptProject` in the `TypeScriptPlugin`

D4N14L:
## Summary

Add `IBuildStageProperties.isTypeScriptProject` output property and populate in the `TypeScriptPlugin`. This will be used in #2721 to eventually f.ix (avoiding auto-resolve) #2711 (thanks @idahogurl !!) in a method consistent with @iclanton 's comment here: https://github.com/microsoft/rushstack/pull/2711#issuecomment-854080989

### DIFF
--- a/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptPlugin.ts
+++ b/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptPlugin.ts
@@ -298,6 +298,12 @@ export class TypeScriptPlugin implements IHeftPlugin {
     };
 
     // Set some properties used by the Jest plugin
+    JestTypeScriptDataFile.saveForProject(heftConfiguration.buildFolder, {
+      emitFolderNameForTests: typeScriptConfiguration.emitFolderNameForTests || 'lib',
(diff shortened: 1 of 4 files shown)


apps/heft/src/plugins/TypeScriptPlugin/TypeScriptPlugin.ts
```diff
@@ -237,6 +237,8 @@ export class TypeScriptPlugin implements IHeftPlugin {
     if (tsconfigPaths.length === 0) {
       // If there are no TSConfigs, we have nothing to do
       return;
+    } else {
+      buildProperties.isTypeScriptProject = true;
```
iclanton, line 241:
```suggestion
      buildProperties.isTypeScriptProject = false;
      return;
    } else {
      buildProperties.isTypeScriptProject = true;
```

I see this property as having three values: `undefined` means "too early to tell", and then `true` and `false` mean "is/is not TypeScript" respectively.

iclanton (approved):

PR #2738, opened 2021-06-10
Disable allowUnreachableCode in tsconfig-base in the Heft rigs.

iclanton:
## Summary

This PR disallows [unreachable code](https://www.typescriptlang.org/tsconfig#allowUnreachableCode) in the Heft rigs.

## How it was tested

Built the repo.

### DIFF
--- a/apps/rush-lib/src/logic/LinkManagerFactory.ts
+++ b/apps/rush-lib/src/logic/LinkManagerFactory.ts
@@ -16,8 +16,8 @@ export class LinkManagerFactory {
       case 'yarn':
         // Yarn uses the same node_modules structure as NPM
         return new NpmLinkManager(rushConfiguration);
+      default:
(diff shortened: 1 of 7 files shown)


D4N14L (approved):

octogonz:
@iclanton is this a warning or an error?

During development, it seems rather draconian to hard fail if someone adds a `return` statement in the middle of a function.

The TypeScript compiler has a general lack of distinction between "errors" versus "warnings".  But Heft's plugin can has a partial workaround like this:

https://github.com/microsoft/rushstack/blob/b577067ae4fd88014a01ab0ee9b537c0e9ba104a/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts#L704-L713

iclanton:
This is a warning:

image

PR #2753, opened 2021-06-16
Update to tapable version 2.2.0

iclanton:
## Summary

Updates tapable to version `~2.2.0`.

## How it was tested

Built this repo.

### DIFF
--- a/apps/heft/src/pluginFramework/InternalHeftSession.ts
+++ b/apps/heft/src/pluginFramework/InternalHeftSession.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-import { SyncHook } from 'tapable';
+import { SyncHook, FullTap } from 'tapable';
 
 import { IHeftPlugin } from './IHeftPlugin';
 import { HeftSession, RegisterAction } from './HeftSession';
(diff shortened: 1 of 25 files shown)


apps/heft/src/utilities/TapableUtilities.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import { Hook, FullTap } from 'tapable';
+
+export class TapableUtilities {
+  public static getTaps<THook extends Hook<unknown, unknown>>(hook: THook): FullTap[] {
+    const result: FullTap[] = [];
+
+    hook.intercept({
```
dmichon-msft, line 10: Something to be aware of is that calling `.intercept()` automatically marks a `Hook` as used. I had to work around this in the timeline profiler PR by switching to a different check at call sites.

apps/heft/src/pluginFramework/InternalHeftSession.ts
```diff
@@ -62,16 +63,19 @@ export class InternalHeftSession {
...
-      if (!accessor) {
-        const accessingPlugins: Set<string> = new Set<string>(pluginHook.taps.map((x) => x.name));
-        throw new Error(
-          `Plugin "${plugin.pluginName}" does not provide an accessor property, so it does not provide ` +
-            `access to other plugins. Plugins requesting access to "${plugin.pluginName}: ` +
-            Array.from(accessingPlugins).join(', ')
-        );
-      } else {
-        pluginHook.call(accessor);
+    if (pluginHook) {
```
dmichon-msft, line 66: Should we include the `isUsed()` check right here?

PR #2805, opened 2021-07-10
[api-extractor] fix: avoid using "default" keyword as var name

### LINKED ISSUE #2804: [api-extractor] dts rollup generates invalid "default" var name
## Summary

api-extractor generates invalid dts rollup result, which contains "default" var name.

## Repro steps

I found the error when switching [my own `api-extractor` fork](https://github.com/adventure-yunfei/web-build-tools/pull/4) to the standard release version.

After some investigation, I found three cases:

1. with import type syntax:
    ```ts
    export declare interface Type1 {
        prop: import('foo').default;
    }
    ```
2. with (weird) named import:
    ```ts
    import { default as FooType2 } from 'foo';
    export declare interface Type2 {
        prop: FooType2;
    }
    ```
3. with re-export default:
    ```ts
    // index.d.ts
    import { FooType3 } from './re-export';
    export declare interface Type3 {
        prop: FooType3;
    }

    // re-export.d.ts
    export { default as FooType3 } from 'foo';
    ```

  **Expected result:** 
Avoid using "default" keyword as var name, e.g.:

```ts
import default_2 from 'foo';

export declare interface Type1 {
    prop: default_2;
}

export { }
```

  **Actual result:** 

api-extractor will generate invalid ts syntax `import { default } from 'foo';`.

e.g. for case 1, generating dts rollup result:

```ts
import { default } from 'foo';

export declare interface Type1 {
    prop: default;
}

export { }
```
## Details

`default` import should be handled by `AstImportKind.DefaultImport`, but there're more cases that `default` export is imported by other ways. (like lastest merged https://github.com/microsoft/rushstack/pull/1916).

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/api-extractor` version? | 7.18.1 |
| API Extractor scenario? | rollups (.d.ts) |
| TypeScript compiler version? | 4.3.5 |
| Node.js version (`node -v`)? | 14.17.2 |

adventure-yunfei:
## Summary

Fixes #2804

## Details

For case:
```ts
export declare interface Type1 {
    prop: import('foo').default;
}
```
Generating dts rollup result:
```ts
import { default as default_2 } from 'foo';

export declare interface Type1 {
    prop: default_2;
}

export { }
```

> Ideally should generate `import default_2 from 'foo';`, but it needs to fix every AstImport type that may contain "default" import. Too complicated.

## How it was tested

Tested cases in #2804.


iclanton:
Can you run `rush change` to describe your fix?

iclanton:
Looks like your build has a warning:

```
Warning: src/namedDefaultImport/index.ts:4:1 - (ae-missing-release-tag) "DefaultImportTypes" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
```

octogonz (approved):

PR #2811, opened 2021-07-13
[rush] Fix the --no-fetch parameter

iclanton:
## Summary

https://github.com/microsoft/rushstack/pull/2795 caused a regression where the `rush change` `--no-fetch` parameter causes a `git fetch`, and absence of that parameter doesn't fetch.

## How it was tested

Tested `rush change` and `rush change -v` locally with and without the `--no-fetch` parameter.


D4N14L:
😬

D4N14L (approved):

PR #2814, opened 2021-07-15
[rush-lib] Fix incremental build state calculation during filtered installs

D4N14L:
## Summary
Filtered installs were broken due to attempting to load config files in dependencies while calculating incremental build state.

## Details
Calculating incremental build state is done by analyzing the world in the repo during initialization, and then proceeding only with the required data. This is done (partially) in an attempt to preserve perf when calculating this state. However, recent changes in #2643 caused this calculation to fail for filtered installs with the error message:
```
Error calculating incremental build state. Instead running full rebuild. Error: Cannot find module '<package A>' from '<path to some project>'
```

This is because it introduced loading of a config file in a project that has not been included in the filtered install. This change makes it so that the config file is only ever loaded when encountering the project during a build, thus making it fail only when attempting to build a package that wasn't installed.

## How it was tested

Unit tests + own repo which had issues with this.

### DIFF
--- a/apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
+++ b/apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
@@ -34,6 +34,7 @@ export class ProjectChangeAnalyzer {
    * undefined === data isn't available (i.e. - git isn't present)
    */
   private _data: Map<string, Map<string, string>> | undefined | UNINITIALIZED = UNINITIALIZED;
+  private _filteredData: Map<string, Map<string, string>> = new Map<string, Map<string, string>>();
   private _projectStateCache: Map<string, string> = new Map<string, string>();
   private _rushConfiguration: RushConfiguration;
   private readonly _git: Git;
@@ -55,6 +56,12 @@ export class ProjectChangeAnalyzer {
     projectName: string,
     terminal: Terminal
   ): Promise<Map<string, string> | undefined> {
+    // Check the cache for any existing data
+    const existingData: Map<string, string> | undefined = this._filteredData.get(projectName);
+    if (existingData) {
+      return existingData;
+    }
+
     if (this._data === UNINITIALIZED) {
       this._data = await this._getDataAsync(terminal);
     }
@@ -70,22 +77,26 @@ export class ProjectChangeAnalyzer {
     }
 
     const unfilteredProjectData: Map<string, string> = this._data.get(projectName)!;
-    const filteredProjectData: Map<string, string> = new Map(unfilteredProjectData);
+    let filteredProjectData: Map<string, string> | undefined;
 
     const ignoreMatcher: Ignore | undefined = await this._getIgnoreMatcherForProjectAsync(project, terminal);
     if (ignoreMatcher) {
       // At this point, `filePath` is guaranteed to start with `projectRelativeFolder`, so
       // we can safely slice off the first N characters to get the file path relative to the
       // root of the project.
-      for (const [filePath] of unfilteredProjectData) {
+      filteredProjectData = new Map<string, string>();
+      for (const [filePath, fileHash] of unfilteredProjectData) {
         const relativePath: string = filePath.slice(project.projectRelativeFolder.length + 1);
-        if (ignoreMatcher.ignores(relativePath)) {
-          // Remove from the filtered data as we encounter the ignored files
-          filteredProjectData.delete(filePath);
+        if (!ignoreMatcher.ignores(relativePath)) {
+          // Add the file path to the filtered data if it is not ignored
+          filteredProjectData.set(filePath, fileHash);
         }
       }
+    } else {
+      filteredProjectData = unfilteredProjectData;
     }
 
+    this._filteredData.set(projectName, filteredProjectData);
     return filteredProjectData;
   }
 
@@ -189,7 +200,7 @@ export class ProjectChangeAnalyzer {
           owningProjectHashDeps = new Map<string, string>();
           projectHashDeps.set(owningProject.packageName, owningProjectHashDeps);
         }
-        owningProjectHashDeps!.set(filePath, fileHash);
+        owningProjectHashDeps.set(filePath, fileHash);
       }
     }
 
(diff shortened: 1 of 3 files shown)


apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -61,14 +61,32 @@ export class ProjectChangeAnalyzer {
...
+
+    const unfilteredProjectData: Map<string, string> = this._data.get(projectName)!;
+    const filteredProjectData: Map<string, string> = new Map(unfilteredProjectData);
+
+    const ignoreMatcher: Ignore | undefined = await this._getIgnoreMatcherForProjectAsync(project, terminal);
+    if (ignoreMatcher) {
+      // At this point, `filePath` is guaranteed to start with `projectRelativeFolder`, so
+      // we can safely slice off the first N characters to get the file path relative to the
+      // root of the project.
+      for (const [filePath] of unfilteredProjectData) {
```
iclanton, line 80: Is using `unfilteredProjectData.keys()` more efficient?

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -61,14 +61,32 @@ export class ProjectChangeAnalyzer {
...
+        const relativePath: string = filePath.slice(project.projectRelativeFolder.length + 1);
+        if (ignoreMatcher.ignores(relativePath)) {
+          // Remove from the filtered data as we encounter the ignored files
+          filteredProjectData.delete(filePath);
```
iclanton, line 84: It may be faster to iterate over `unfilteredProjectData` and only populate `filteredProjectData` with the un-ignored entries than to initialize `filteredProjectData` with the entries that will end up getting removed.

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -61,14 +61,32 @@ export class ProjectChangeAnalyzer {
...
+        }
       }
     }
+
+    return filteredProjectData;
```
iclanton, line 89: We should cache this result.

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -178,14 +182,14 @@ export class ProjectChangeAnalyzer {
...
-        if (!ignoreMatcher || !ignoreMatcher.ignores(relativePath)) {
-          projectHashDeps.get(owningProject.packageName)!.set(filePath, fileHash);
+        let owningProjectHashDeps: Map<string, string> | undefined = projectHashDeps.get(
+          owningProject.packageName
+        );
+        if (!owningProjectHashDeps) {
+          owningProjectHashDeps = new Map<string, string>();
+          projectHashDeps.set(owningProject.packageName, owningProjectHashDeps);
         }
+        owningProjectHashDeps!.set(filePath, fileHash);
```
iclanton, line 192: Why is the non-null assertion required here?

apps/rush-lib/src/logic/test/ProjectChangeAnalyzer.test.ts
```diff
@@ -33,6 +33,9 @@ describe(ProjectChangeAnalyzer.name, () => {
       },
       findProjectForPosixRelativePath(path: string): object | undefined {
         return projects.find((project) => path.startsWith(project.projectRelativeFolder));
+      },
+      getProjectByName(name: string): object | undefined {
+        return projects.find((project) => project.packageName === name);
```
iclanton, line 38: Seems like it would be more efficient to just have a lazily-initialized map of projects by name.

D4N14L: I'm gonna leave as-is since it's just unit-test code and there's existing examples using this pattern.

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
D4N14L: Good catch. Was a leftover

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
D4N14L: Good call

apps/rush-lib/src/logic/test/ProjectChangeAnalyzer.test.ts
iclanton: Looks like it's being used here: https://github.com/microsoft/rushstack/pull/2814/files#diff-a97c75b62243d6b88ead5b45905024ed5352dbf21bd23460511267ad924548c6R67

apps/rush-lib/src/logic/test/ProjectChangeAnalyzer.test.ts
```diff
@@ -33,6 +33,9 @@ describe(ProjectChangeAnalyzer.name, () => {
       },
       findProjectForPosixRelativePath(path: string): object | undefined {
         return projects.find((project) => path.startsWith(project.projectRelativeFolder));
+      },
+      getProjectByName(name: string): object | undefined {
```
iclanton, line 37: The return type here should be `RushConfigurationProject | undefined`, shouldn't it?

iclanton: Same with the function above.

iclanton: Ohh nevermind. Sorry, I didn't realize this was in a Test.

apps/rush-lib/src/logic/test/ProjectChangeAnalyzer.test.ts
iclanton: Ohh sorry, I thought this was in RushConfiguration. Nevermind.

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
D4N14L: Ehhhhhhh I went back and fourth on this one. I was erring on the side of "the add/subtract operation is the expensive part and assume that there's a minimum of files ignored", but then that's also assuming initialization of a fresh map from an existing one is faster than creating a new one and adding the elements individually. Which is probably generous. I'll change it

iclanton (approved):

PR #2816, opened 2021-07-19
[heft] Extract sass plugin to separate @rushstack/heft-sass-plugin package

jonasb:
## Summary

Extracts the default sass plugin to a separate package to make heft leaner and avoid issues related to its dependencies.

`@rushstack/heft-sass-plugin` is a new package with the same functionality as the old default sass plugin.

Breaking change since sass is no longer the default. However, if `@rushstack/heft-web-rig` is used there should be no break since the new plugin is added to that rig.

Related issue #2548.

## Details

* Refactored the functionality of the old plugin. Apart from being in a separate plugin, I believe the functionality is identical.
* When building there were some unrelated changes to `pnpm-lock.yaml`, so added those changes as separate commit to keep things clearer
* Updated UPGRADING documentation, but the final version needs to be updated
* Renamed `sass.schema.json` to `heft-sass-plugin.schema.json` to align with the jest plugin. Unsure if that's considered a breaking change?

## How it was tested

* Relying on existing tests, i.e. `rush build`
* Ensured that `heft-sass-test` works by changing `.sass` file during `npm start` and changes were reflected in browser automatically
* Added empty file `build-tests/heft-sass-test/config/sass.json` and removed `@rushstack/heft-sass-plugin` from `build-tests/heft-sass-test/config/heft.json` to get `project-validation` warning during build


D4N14L:
@halfnibble for awareness.

common/config/rush/browser-approved-packages.json
```diff
@@ -2,6 +2,10 @@
 {
   "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/approved-packages.schema.json",
   "packages": [
+    {
```
D4N14L, line 5: Not a `browser-approved-package`. Can you move this to the non-browser file?

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
```diff
@@ -1,12 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-import * as path from 'path';
+import { LegacyAdapters } from '@rushstack/node-core-library';
```
D4N14L, line 4: We tend to order using the following priority (alphabetically within each):
- system dependencies
- external dependencies (declared in `package.json`, obtained from NPM)
- monorepo dependencies (declared in `package.json`, obtained from monorepo workspace dependency)
- internal module dependencies (included in package)

Though I don't think that this is documented anywhere (or even guidance.... @octogonz ?)

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
```diff
@@ -61,14 +72,30 @@ export class SassTypingsPlugin implements IHeftPlugin {
...
+      await SassTypingsPlugin._getSassConfigurationLoader(
+        buildFolder
+      ).tryLoadConfigurationFileForProjectAsync(logger.terminal, buildFolder, heftConfiguration.rigConfig);
 
     return {
       ...sassConfigurationJson
     };
   }
+
+  private static _getSassConfigurationLoader(buildFolder: string): ConfigurationFile<ISassConfigurationJson> {
```
D4N14L, line 84: Unused param

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
```diff
@@ -51,7 +58,11 @@ export class SassTypingsPlugin implements IHeftPlugin {
 
     await sassTypingsGenerator.generateTypingsAsync();
     if (isWatchMode) {
-      Async.runWatcherWithErrorHandling(async () => await sassTypingsGenerator.runWatcherAsync(), logger);
```
D4N14L, line 54: This method was added by @iclanton to cleanup floating promises. Do you think it's safe to go back to a simple try-catch?

apps/heft/tsconfig.json
```diff
@@ -5,5 +5,5 @@
     "typeRoots": ["./custom-typings", "./node_modules/@types/"],
```
D4N14L, line 5: Can also remove from here

apps/heft/UPGRADING.md
```diff
@@ -1,5 +1,25 @@
 # Upgrade notes for @rushstack/heft
 
+### Heft 0.XX.0
```
jonasb, line 3: Should I specify a future version here?

common/config/rush/browser-approved-packages.json
jonasb: Sure, makes sense. Done.

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
jonasb: Ok, I've reordered the imports now.

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
jonasb: Nice catch, removed.

apps/heft/tsconfig.json
jonasb: Done. Also removed the `include` below.

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
jonasb: `try { await foo() } catch {}` would be safe, however it is a change in behavior. I missed that `runWatcherWithErrorHandling()` is fire and forget and my code would wait on the promise to complete. So I've make a copy of `Async.runWatcherWithErrorHandling()` and reverted to the same code. Thanks for pointing it out!

jonasb:
Any feedback of the renaming of `sass.schema.json` to `heft-sass-plugin.schema.json`? I did it to conform to the jest plugin, but I guess it only makes a difference if the published url `https://developer.microsoft.com/json-schemas/heft/sass.schema.json` is changes?

halfnibble:
> Any feedback of the renaming of `sass.schema.json` to `heft-sass-plugin.schema.json`? I did it to conform to the jest plugin, but I guess it only makes a difference if the published url `https://developer.microsoft.com/json-schemas/heft/sass.schema.json` is changes?

This name change makes sense to me!

halfnibble:
@jonasb This is a change we need! If you have not already, can you please verify this works with the heft-sass-test project?
https://github.com/microsoft/rushstack/tree/master/build-tests/heft-sass-test

You should be able to build the monorepo and run `npm start` from within heft-sass-test and a new browser window should popup with a little UI code. Then you should be able to change the styles and it should recompile and reload the window.

jonasb:
I've tested heft-sass-test now and it worked fine! Thanks @halfnibble for the suggestion!

apps/heft/UPGRADING.md
halfnibble: @iclanton This looks good to me. Do you want to specify the version to place here? This PR is configured for a major version upgrade.

halfnibble (approved):

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
D4N14L: Doesn't seem like they were re-ordered? Should look largely like the original, with the `import * as path from 'path'` at the top and the `@rushstack` imports at the bottom. If this is happening because of the prettier-on-commit functionality of the repo then you can ignore this. But it would be odd if so, since we use this format pretty widely across the repo.

apps/heft/UPGRADING.md
D4N14L: We don't want to do a major version just yet, so it would likely be a minor version bump to the next minor version (which should be `0.35.0` I believe)

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
jonasb: Sorry, only fixed `SassTypingsPlugin.ts` earlier. This file should be fixed now as well.

apps/heft/UPGRADING.md
jonasb: Changed to 0.35.0 now. Also changed the type of my change from `major` to `minor`

D4N14L (approved):

D4N14L (changes requested):
Actually, one more request. Could you add an entry to the `ProjectValidatorPlugin` for the `sass.json` config file (similar to the `jest.config.json` check) to warn for the missing plug-in? Since we're removing the default functionality https://github.com/microsoft/rushstack/blob/master/apps/heft/src/plugins/ProjectValidatorPlugin.ts

jonasb:
Good idea! Output when running without plugin but with a `config/sass.json` file.

```
Project build folder is "/Users/jonas/dev/explore/rushstack/build-tests/heft-sass-test"
Starting build
 ---- Pre-compile started ---- 
 ---- Pre-compile finished (4ms) ---- 
 ---- Compile started ---- 
[project-validation] Warning: The configuration file "/Users/jonas/dev/explore/rushstack/build-tests/heft-sass-test/config/sass.json" exists in your project, but the associated Heft plugin is not enabled. To fix this, you can add "@rushstack/heft-sass-plugin" to your package.json "devDependencies" and use "config/heft.json" to load it. For details, see Heft's UPGRADING.md notes and this article: https://rushstack.io/pages/heft_tasks/sass-typings/
```

D4N14L (approved):
Looks good! Since the updated `ProjectValidationPlugin` will require docs-side changes, we'll probably want to wait for @octogonz to bless and merge this PR.

common/changes/@rushstack/heft-sass-plugin/master_2021-07-19-17-25.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-sass-plugin",
+      "comment": "Extract default Sass plugin to separate package",
+      "type": "patch"
```
iclanton, line 6:
```suggestion
      "type": "minor"
```

common/changes/@rushstack/heft/master_2021-07-19-17-25.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft",
+      "comment": "Extract default Sass plugin to separate @rushstack/heft-sass-plugin package",
```
iclanton, line 5:
```suggestion
      "comment": "(BREAKING CHANGE) Extract default Sass plugin to separate @rushstack/heft-sass-plugin package",
```

heft-plugins/heft-sass-plugin/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@rushstack/heft-sass-plugin",
+  "version": "0.1.0",
```
iclanton, line 3:
```suggestion
  "version": "0.0.0",
```

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
```diff
@@ -1,21 +1,29 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-import { IHeftPlugin } from '../../pluginFramework/IHeftPlugin';
-import { HeftSession } from '../../pluginFramework/HeftSession';
-import { HeftConfiguration } from '../../configuration/HeftConfiguration';
-import { IBuildStageContext, IPreCompileSubstage } from '../../stages/BuildStage';
+import * as path from 'path';
+import {
```
iclanton, line 5:
```suggestion
import type {
```

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
```diff
@@ -71,4 +79,22 @@ export class SassTypingsPlugin implements IHeftPlugin {
       ...sassConfigurationJson
     };
   }
+
+  private static _getSassConfigurationLoader(): ConfigurationFile<ISassConfigurationJson> {
+    return new ConfigurationFile<ISassConfigurationJson>({
```
iclanton, line 84: We should memoize this

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
```diff
@@ -1,21 +1,29 @@
...
 import { ISassConfiguration, SassTypingsGenerator } from './SassTypingsGenerator';
-import { CoreConfigFiles } from '../../utilities/CoreConfigFiles';
-import { ScopedLogger } from '../../pluginFramework/logging/ScopedLogger';
-import { Async } from '../../utilities/Async';
+import { Async } from './utilities/Async';
 
 export interface ISassConfigurationJson extends ISassConfiguration {}
 
 const PLUGIN_NAME: string = 'SassTypingsPlugin';
+const PLUGIN_SCHEMA_PATH: string = path.resolve(__dirname, 'schemas', 'heft-sass-plugin.schema.json');
```
iclanton, line 21:
```suggestion
const PLUGIN_SCHEMA_PATH: string = `${__dirname}/schemas/heft-sass-plugin.schema.json`;
```

This is faster than using `path`.

iclanton (approved):
Looks good overall. I'll confirm with @octogonz that the package is created.

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
jonasb: Done

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
jonasb: I'm a bit cautious about this due to os compatibility. And as far as I can see `path.join()` and `path.resolve()` are used consistently throughout the repo.

iclanton: Yeah we're trying to move away from that. It's more expensive than you'd think. Slashes work fine on Windows.

iclanton (approved):

heft-plugins/heft-sass-plugin/src/SassTypingsPlugin.ts
jonasb: Ok, good to know!

jonasb:
Thanks for all the reviews and thanks for merging @iclanton !

PR #2819, opened 2021-07-20
[api-extractor] Fix api-extractor generator to only output lower cases of directives

Claudiazhaoya:
## Summary

This PR aims to fix the output of report by api-extractors to only have lower cased directives.

### DIFF
--- a/apps/api-extractor/src/generators/ApiReportGenerator.ts
+++ b/apps/api-extractor/src/generators/ApiReportGenerator.ts
@@ -60,10 +60,10 @@ export class ApiReportGenerator {
     // Emit the triple slash directives
     for (const typeDirectiveReference of Array.from(collector.dtsTypeReferenceDirectives).sort()) {
       // https://github.com/microsoft/TypeScript/blob/611ebc7aadd7a44a4c0447698bfda9222a78cb66/src/compiler/declarationEmitter.ts#L162
-      writer.writeLine(`/// <reference types="${typeDirectiveReference}" />`);
+      writer.writeLine(`/// <reference types="${typeDirectiveReference.toLowerCase()}" />`);
     }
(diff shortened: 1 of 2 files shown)


apps/api-extractor/src/generators/ApiReportGenerator.ts
```diff
@@ -60,10 +60,10 @@ export class ApiReportGenerator {
     // Emit the triple slash directives
     for (const typeDirectiveReference of Array.from(collector.dtsTypeReferenceDirectives).sort()) {
       // https://github.com/microsoft/TypeScript/blob/611ebc7aadd7a44a4c0447698bfda9222a78cb66/src/compiler/declarationEmitter.ts#L162
-      writer.writeLine(`/// <reference types="${typeDirectiveReference}" />`);
+      writer.writeLine(`/// <reference types="${typeDirectiveReference.toLowerCase()}" />`);
```
D4N14L, line 63: Add comments on both as to why the lower casing is needed

D4N14L: Are we sure this is the correct fix? Looking online (ex. https://stackoverflow.com/questions/37617325/typing-references-in-typescript-are-case-sensitive) it seems that this field is case-insensitive on Windows, but case-sensitive at runtime on Linux, so ideally we would want to include the exact correct casing here. We could end up actually breaking more people than we fix if this is in-fact the case.

iclanton: Here's some code in Heft that corrects the casing of a folder path: https://github.com/microsoft/rushstack/blob/a13865bef9a20dab28c044be3504c7326bfe94b1/apps/heft/src/configuration/HeftConfiguration.ts#L170-L175

We could do something similar here.

Claudiazhaoya:
The issue has been solved with some other circumvent solution

PR #2823, opened 2021-07-22
[module-minifier-plugin] Fix comment file generation, WorkerPoolMinifier

dmichon-msft:
## Summary
Fixes faulty code in the license file generator that was causing a crash.

## Details
Comments were being stored as objects but expected to be strings, and was not caught during the build due to an `any` on the typings of the `factoryMeta` fields. Was not caught during the build tests because none of them had preserved comments in the source text.
Updated the code to unpack the objects back into the raw string form.

Added license comments to the localization-plugin-test-02 build test suite to validate that comment extraction works as expected. Updated the same test to leverage `WorkerPoolMinifier` instead of `SynchronousMinifier`.

Fixed incorrect configuration of the `WorkerPool` instance inside of `WorkerPoolMinifier` that resulted in use of it hanging the process.

## How it was tested
Ran the new localization-plugin-test-02 build and verified the generated license.txt file in the dist-prod output.

### DIFF
--- a/common/changes/@rushstack/module-minifier-plugin/fix-minifier-comment-extraction_2021-07-22-21-47.json
+++ b/common/changes/@rushstack/module-minifier-plugin/fix-minifier-comment-extraction_2021-07-22-21-47.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/module-minifier-plugin",
+      "comment": "Fix comment file generation logic. Add to build test.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/module-minifier-plugin",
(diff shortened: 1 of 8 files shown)


common/changes/@rushstack/module-minifier-plugin/fix-minifier-comment-extraction_2021-07-22-21-47.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/module-minifier-plugin",
+      "comment": "Fix comment file generation logic. Add to build test.",
```
iclanton, line 5:
```suggestion
      "comment": "Fix comment file generation logic.",
```

iclanton (approved):

PR #2838, opened 2021-08-02
[rush] Improve startup performance with read cache

dmichon-msft:
## Summary
Introduces read caching into `RigConfig` and `RushProjectConfiguration`, to reduce redundant work during startup.
Stops instantiating an `Ignorer` instance if the project does not have a `rush-project.json` (either locally or from a rig).

## Details
Instantiation of `Ignorer` was superfluous when no rules were present, and saves performance by being skipped.
Caching of `RigConfig` and `RushProjectConfiguration` removes duplicate reads between the ProjectChangeAnalyzer initialization and subsequent reads for the build cache feature.

## How it was tested
Comparing `rush tab-complete` and `rush list` between the repo command and the locally built version shows about a 50% savings in the `@rushstack` repo.


iclanton:
When are projects' `RushProjectConfiguration`s getting instantiated multiple times in the same Rush run?

dmichon-msft:
> When are projects' `RushProjectConfiguration`s getting instantiated multiple times in the same Rush run?

Once by ProjectChangeAnalyzer, a second time for the build cache checks.

iclanton:
Can you create changelogs?

apps/rush-lib/src/api/RushProjectConfiguration.ts
```diff
@@ -126,6 +126,9 @@ export class RushProjectConfiguration {
       }
     });
 
+  private static readonly _configCache: Map<RushConfigurationProject, RushProjectConfiguration | false> =
```
iclanton, line 129: The `RushProjectConfiguration | false` here is kinda confusing.

build-tests/install-test-workspace/workspace/common/pnpm-lock.yaml
```diff
@@ -5,27 +5,27 @@ importers:
   typescript-newest-test:
```
iclanton, line 5: Revert?

libraries/rig-package/src/RigConfig.ts
```diff
@@ -199,80 +206,107 @@ export class RigConfig {
...
-            projectFolderPath: options.projectFolderPath,
-
-            rigFound: false,
-            filePath: '',
-            rigPackageName: '',
-            rigProfile: ''
-          });
-        }
+    const fromCache: RigConfig | false | undefined =
+      !options.bypassCache && !overrideRigJsonObject && RigConfig._configCache.get(projectFolderPath);
```
iclanton, line 212:
```suggestion
    const fromCache: RigConfig | undefined = options.bypassCache || overrideRigJsonObject
      ? undefined
       : RigConfig._configCache.get(projectFolderPath);
```

libraries/rig-package/src/RigConfig.ts
```diff
@@ -199,80 +206,107 @@ export class RigConfig {
...
 
-      rigFound: true,
-      filePath: rigConfigFilePath,
-      rigPackageName: json.rigPackageName,
-      rigProfile: json.rigProfile || 'default'
-    });
+        rigFound: true,
+        filePath: rigConfigFilePath,
+        rigPackageName: json!.rigPackageName,
+        rigProfile: json!.rigProfile || 'default'
```
iclanton, line 238: The constructor is private, so we can probably make `'default'` the default if this property is undefined.

iclanton (approved):

build-tests/install-test-workspace/workspace/common/pnpm-lock.yaml
dmichon-msft: This file updates every time I run `rush build`

apps/rush-lib/src/api/RushProjectConfiguration.ts
dmichon-msft: Do you have a preferred method for explicitly storing that the cached config value is "does not exist"? I prefer to use primitives, but could use a Symbol or something instead.

PR #2839, opened 2021-08-03
[rush] Move ConfigurationFile logging to new "debug" severity

elliot-nelson:
## Summary

This is a potential fix for https://github.com/microsoft/rushstack/issues/2790.

In short:

 - Add an additional log severity, `debug`, below `verbose`.
 - Like verbose, the _default_ behavior (in a typical usage like `ConsoleTerminalProvider`) is to suppress `debug`-level messages, unless `debugEnabled` is set to true.
 - All of the logging in `heft-config-file` has been moved to `debug` level instead of `verbose` level.
 - The meaning of top-level rush flag `--debug` has expanded -- not only does `--debug` enable full stack traces on exception, it also displays additional diagnostic information (`heft-config-file` detail being the first example).

## Details

I think any changes to Terminal need to consider two use cases: what does this mean for the "generic" usage of Terminals/TerminalProviders, and what does it mean specifically for rush/rush-lib.

Generically, I think this change is very reasonable; the default `Terminal(ConsoleTerminalProvider)` behavior would be very familiar to anyone who's used a logging system before.  (Basically, 5 logging levels, where `Error / Warning / Log` are above water by default, and `Verbose / Debug` are below water by default.)  Verbose and Debug are separate flags, rather than an organized number scheme, but anyone can use a Terminal in their app and easily enforce that debug is a superset of verbose, if desired.

For rush specifically, these channels have more meaning: the `verbose` severity is used to funnel all of the user build script output during a build, which means it's not well suited to any other information.  The new `debug` channel is an attempt to provide a channel where rush itself can provide diagnostic info, and it is turned off by default.

## How it was tested

Updated unit test snapshots showing that ConfigurationFile output now goes to the `debug` severity.  Tested building with my local rush copy to confirm that the "rig loading" spam is now gone by default, and can be brought back by entering `rush --debug build` (this is the fix for #2790).

## Known issues / problems / concerns

I've made the bare minimum changes to the stream handling in `ProjectBuilder`, so as of this PR, if debug _is_ enabled, it ends up being treated the same as verbose, meaning:

```console
# user build stdout goes to the .log file in each folder
rush build

# now user build stdout goes to .log and stdout
rush build -v

# user build stdout PLUS detailed diagnostics goes to .log and stdout
rush --debug build -v

# user build stdout PLUS detailed diagnostics goes only to .log file
rush --debug build
```

As a user, that last behavior seems a little odd.  A possible solution would be to treat the `debug` channel as if it was the `warning` channel whenever `--debug` is enabled (or, if writing to stderr for diagnostics is undesired, we just need to make more detailed changes to the stream handling and splitting).


common/changes/@rushstack/heft-config-file/experimental-terminal_2021-08-03-19-33.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-config-file",
+      "comment": "Detailed logging moved from verbose to debug severity.",
```
iclanton, line 5:
```suggestion
      "comment": "Moved detailed logging from verbose to debug severity.",
```

common/changes/@rushstack/node-core-library/experimental-terminal_2021-08-03-19-33.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/node-core-library",
+      "comment": "Added new message severity \"debug\", below verbose.",
```
iclanton, line 5:
```suggestion
      "comment": "Added new Terminal message severity \"debug\", below verbose.",
```

common/reviews/api/node-core-library.api.md
```diff
@@ -718,9 +720,10 @@ export class Sort {
 export class StringBufferTerminalProvider implements ITerminalProvider {
     constructor(supportsColor?: boolean);
     get eolCharacter(): string;
+    getDebugOutput(options?: IStringBufferOutputOptions): string;
     getErrorOutput(options?: IStringBufferOutputOptions): string;
     getOutput(options?: IStringBufferOutputOptions): string;
-    getVerbose(options?: IStringBufferOutputOptions): string;
+    getVerboseOutput(options?: IStringBufferOutputOptions): string;
```
iclanton, line 726: This is technically a breaking change, but this is generally only used for tests, so it's probably fine.

common/reviews/api/node-core-library.api.md
```diff
@@ -748,12 +753,14 @@ export class Terminal {
...
 }
 
-// @beta (undocumented)
+// @beta
 export enum TerminalProviderSeverity {
     // (undocumented)
-    error = 2,
+    debug = 4,
+    // (undocumented)
+    error = 0,
```
iclanton, line 761: This is also technically a breaking change.

libraries/node-core-library/src/Terminal/ITerminalProvider.ts
```diff
@@ -2,13 +2,29 @@
...
+ * Terminal providers may choose to suppress certain messages based on their severity,
+ * or to route some messages to other providers or not based on severity.
+ *
+ *   Severity  | Purpose
+ *   --------- | -------
+ *   error     | Build errors and fatal issues in rush
+ *   warning   | Not necessarily fatal, but indicate a problem the user should fix
+ *   log       | Informational messages from the rush system
+ *   verbose   | Additional information that may not always be necessary
+ *   debug     | Highest detail level, best used for troubleshooting information
```
iclanton, line 18:
```suggestion
 *   error     | Build errors and fatal issues
 *   warning   | Not necessarily fatal, but indicate a problem the user should fix
 *   log       | Informational messages
 *   verbose   | Additional information that may not always be necessary
 *   debug     | Highest detail level, best used for troubleshooting information
```

`node-core-library` isn't specifically tied to Rush.

libraries/node-core-library/src/Terminal/ITerminalProvider.ts
```diff
@@ -2,13 +2,29 @@
...
 export enum TerminalProviderSeverity {
-  log,
-  warning,
   error,
-  verbose
+  warning,
+  log,
+  verbose,
+  debug
 }
```
iclanton, line 28: I definitely see why you reordered these, but it causes a breaking API change.

libraries/node-core-library/src/Terminal/test/__snapshots__/Terminal.test.ts.snap
```diff
@@ -2,6 +2,7 @@
```
iclanton, line 2: Can you add a test exercising debug?

iclanton (approved):

libraries/node-core-library/src/Terminal/ITerminalProvider.ts
elliot-nelson: The order change here and the `getVerbose()` rename are only nice-to-haves.  I did end up marking the the rush change as MAJOR I believe, but now in hindsight, it probably makes more sense to revert those pieces and then this whole change could be minor.

iclanton: Let's avoid the major bump for now. There are a few other breaking changes we've been talking about, so we can make those then.

elliot-nelson:
Latest push:

 - Added unit tests for the `writeDebug` and `writeDebugLine` methods.
 - Reverted the "log severity" line changes and the "getVerboseOutput" changes (these changes were breaking changes, so without them this PR can drop to a minor change.
 - Updated rush change major to minor.

common/changes/@rushstack/heft-config-file/experimental-terminal_2021-08-03-19-33.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-config-file",
+      "comment": "Moved detailed logging from verbose to debug severity.",
```
iclanton, line 5:
```suggestion
      "comment": "Move detailed logging from verbose to debug severity.",
```

common/changes/@rushstack/node-core-library/experimental-terminal_2021-08-03-19-33.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/node-core-library",
+      "comment": "Added new Terminal message severity \"debug\", below verbose.",
```
iclanton, line 5:
```suggestion
      "comment": "Add new Terminal message severity \"debug\", below verbose.",
```

iclanton (approved):

PR #2843, opened 2021-08-05
[rush-lib] Add flags for parameters to rush telemetry

BPapp-MS:
## Summary

Users are not at the moment able to record telemetry on the parameters used by rush install/update, including whether the project selection parameters such as --to or --from are getting use. This PR adds many parameters to the output file.

## Details

Parameters are read in for their values, stringified and added to the object passed to Telemetry.log(). The selectionParameters property is tested for initialization by the subclass and if it exists, the value of its .getTelemetry() function is added to extraData, just as it is in the implementation of the build and rebuild telemetry, which is implemented in BulkScriptAction.ts

## How it was tested

I've set configured another project to use this branch and I've observed the results of the telemetry from several calls of rush install, update, build, and rebuild, with many different combinations of parameters.

### DIFF
--- a/apps/rush-lib/src/cli/actions/BaseInstallAction.ts
+++ b/apps/rush-lib/src/cli/actions/BaseInstallAction.ts
@@ -40,6 +40,10 @@ export abstract class BaseInstallAction extends BaseRushAction {
   protected _debugPackageManagerParameter!: CommandLineFlagParameter;
   protected _maxInstallAttempts!: CommandLineIntegerParameter;
   protected _ignoreHooksParameter!: CommandLineFlagParameter;
+  /*
+   * Subclasses can initialize the _selectionParameters property in order for
+   * the parameters to be written to the telemetry file
+   */
   protected _selectionParameters?: SelectionParameterSet;
 
   protected onDefineParameters(): void {
@@ -189,9 +193,9 @@ export abstract class BaseInstallAction extends BaseRushAction {
         bypassPolicy: (!!this._bypassPolicyParameter.value).toString(),
         noLink: (!!this._noLinkParameter.value).toString(),
         networkConcurrency: this._networkConcurrencyParameter.value
-          ? this._networkConcurrencyParameter.toString()
+          ? this._networkConcurrencyParameter.value.toString()
           : 'unspecified',
-        debugPackageManager: (!!this._debugPackageManagerParameter).toString(),
+        debugPackageManager: (!!this._debugPackageManagerParameter.value).toString(),
         maxInstallAttempts: this._maxInstallAttempts.value
           ? this._maxInstallAttempts.value.toString()
           : 'unspecified',
--- a/apps/rush-lib/src/cli/actions/BaseInstallAction.ts
+++ b/apps/rush-lib/src/cli/actions/BaseInstallAction.ts
@@ -193,7 +193,7 @@ export abstract class BaseInstallAction extends BaseRushAction {
         debug: installManagerOptions.debug.toString(),
         full: installManagerOptions.fullUpgrade.toString(),
         ...this.getParameterStringMap(),
-        ...this._selectionParameters?.getTelemetry() };
+        ...this._selectionParameters?.getTelemetry()
       };
       this.parser.telemetry.log({
         name: 'install',
--- a/apps/rush-lib/src/cli/actions/BaseInstallAction.ts
+++ b/apps/rush-lib/src/cli/actions/BaseInstallAction.ts
@@ -8,10 +8,7 @@ import { Import } from '@rushstack/node-core-library';
 import {
   CommandLineFlagParameter,
   CommandLineIntegerParameter,
-  CommandLineStringParameter,
-  CommandLineParameterKind,
-  CommandLineIntegerListParameter,
-  CommandLineStringListParameter
+  CommandLineStringParameter
 } from '@rushstack/ts-command-line';
 
 import { BaseRushAction } from './BaseRushAction';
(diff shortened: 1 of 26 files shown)


common/changes/@microsoft/rush/benpapp-AddFlagsToRushTelemetry_2021-08-05-20-52.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Adds properties to the extraData section of the telemetry file for parameter usage in the install commands",
```
iclanton, line 5:
```suggestion
      "comment": "Add properties to the extraData section of the telemetry file for parameter usage in the install commands",
```

apps/rush-lib/src/cli/actions/InstallAction.ts
```diff
@@ -7,6 +7,8 @@ import { RushCommandLineParser } from '../RushCommandLineParser';
 import { SelectionParameterSet } from '../SelectionParameterSet';
 
 export class InstallAction extends BaseInstallAction {
+  // must match name of _selectionParameters in BaseInstallAction for telemetry to work
+  // worthy of the override parameter in TypeScript > 4.3
   protected _selectionParameters!: SelectionParameterSet;
```
iclanton, line 12: Why does this need to be redeclared?

iclanton: Is it because this one is required and the base one isn't?

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
```diff
@@ -181,15 +187,30 @@ export abstract class BaseInstallAction extends BaseRushAction {
...
+        networkConcurrency: this._networkConcurrencyParameter.value
+          ? this._networkConcurrencyParameter.value.toString()
+          : 'unspecified',
+        debugPackageManager: (!!this._debugPackageManagerParameter.value).toString(),
+        maxInstallAttempts: this._maxInstallAttempts.value
+          ? this._maxInstallAttempts.value.toString()
+          : 'unspecified',
+        ignoreHooks: (!!this._ignoreHooksParameter.value).toString(),
+        debug: installManagerOptions.debug.toString(),
+        full: installManagerOptions.fullUpgrade.toString()
```
iclanton, line 204: Can we get this from the set of all parameters (like via the `this.parameters` object) instead of enumerating them explicitly?

apps/rush-lib/src/cli/actions/InstallAction.ts
BPapp-MS: The reality is that it only needs to be initialized which that line didn't enforce so I removed this declaration, which would at least cause an error if the name were to change

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
```diff
@@ -181,15 +190,42 @@ export abstract class BaseInstallAction extends BaseRushAction {
...
+          case CommandLineParameterKind.Choice:
+          case CommandLineParameterKind.String:
+          case CommandLineParameterKind.Integer:
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            extraData[parameter.longName] = JSON.stringify((parameter as any).value);
+            break;
+          case CommandLineParameterKind.StringList:
+          case CommandLineParameterKind.IntegerList:
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const arrayValue: ReadonlyArray<any> | undefined = (
```
iclanton, line 212:
```suggestion
            const arrayValue: ReadonlyArray<string | number> | undefined = (
```

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
```diff
@@ -181,15 +190,42 @@ export abstract class BaseInstallAction extends BaseRushAction {
...
+      };
+
+      for (const parameter of this.parameters) {
+        switch (parameter.kind) {
+          case CommandLineParameterKind.Flag:
+          case CommandLineParameterKind.Choice:
+          case CommandLineParameterKind.String:
+          case CommandLineParameterKind.Integer:
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            extraData[parameter.longName] = JSON.stringify((parameter as any).value);
```
iclanton, line 207: Can you just list out the specific types here?

apps/rush-lib/src/cli/actions/BaseInstallAction.ts
```diff
@@ -181,15 +187,23 @@ export abstract class BaseInstallAction extends BaseRushAction {
...
+        clean: (!!this._purgeParameter.value).toString(),
+        debug: installManagerOptions.debug.toString(),
+        full: installManagerOptions.fullUpgrade.toString()
+      };
+
+      extraData = { ...extraData, ...this.getParameterStringMap() };
+
+      if (this._selectionParameters) {
+        extraData = { ...extraData, ...this._selectionParameters.getTelemetry() };
+      }
```
iclanton, line 201:
```suggestion
      const extraData: Record<string, string> = {
        mode: this.actionName,
        clean: (!!this._purgeParameter.value).toString(),
        debug: installManagerOptions.debug.toString(),
        full: installManagerOptions.fullUpgrade.toString(),
        ...this.getParameterStringMap(),
        ...this._selectionParameters?.getTelemetry() };
      };
```

iclanton (approved):

PR #2853, opened 2021-08-12
[heft] Restore generation of tsbuildinfo path to compensate for improper resolution semantics

dmichon-msft:
## Summary
Using the `tsBuildInfoFile` in tsconfig-base.json in rigs was resolving incorrectly. This PR restores it to being auto-generated inside of the Heft TypeScriptPlugin.

## Details
Uses the same algorithm as previously, but routed into the `temp/` folder instead of `.heft/build-cache`.

## How it was tested
Local `rush build`, verified that projects have a build info file in the expected location.

### DIFF
--- a/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
+++ b/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
+import * as crypto from 'crypto';
 import * as path from 'path';
 import * as semver from 'semver';
 import {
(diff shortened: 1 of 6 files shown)


apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
```diff
@@ -121,13 +122,33 @@ export class TypeScriptBuilder extends SubprocessRunnerBase<ITypeScriptBuilderCo
...
   private _tsReadJsonCache: Map<string, object> = new Map<string, object>();
   private _cachedFileSystem: TypeScriptCachedFileSystem = new TypeScriptCachedFileSystem();
 
   public get filename(): string {
     return __filename;
   }
 
+  private get _tsCacheFilePath(): string {
+    if (!this.__tsCacheFilePath) {
+      const configHash: crypto.Hash = Tslint.getConfigHash(
```
iclanton, line 135: Seems like this should be moved out of the linter code and either into TypeScriptBuilder or into utilities.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
```diff
@@ -121,13 +122,33 @@ export class TypeScriptBuilder extends SubprocessRunnerBase<ITypeScriptBuilderCo
   private _typescriptTerminal!: Terminal;
   private _emitCompletedCallbackManager: EmitCompletedCallbackManager;
 
+  private __tsCacheFilePath!: string;
```
iclanton, line 125:
```suggestion
  private __tsCacheFilePath: string | undefined;
```

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
```diff
@@ -121,13 +122,33 @@ export class TypeScriptBuilder extends SubprocessRunnerBase<ITypeScriptBuilderCo
...
+        this._typescriptTerminal,
+        this._cachedFileSystem
+      );
+      configHash.update(JSON.stringify(this._configuration.additionalModuleKindsToEmit || {}));
+      const serializedConfigHash: string = configHash.digest('hex');
+
+      this.__tsCacheFilePath = path.posix.join(
+        this._configuration.buildMetadataFolder,
+        `ts_${serializedConfigHash}.json`
+      );
```
iclanton, line 146: Should just be able to say

```suggestion
      this.__tsCacheFilePath = `${this._configuration.buildMetadataFolder}/ts_${serializedConfigHash}.json`;
```

right?

iclanton (approved):

PR #2863, opened 2021-08-17
[module-minifier-plugin] Fix compatibility issue with mini-css-extract-plugin

dmichon-msft:
## Summary
Configures the ModuleMinifierPlugin to ignore non-JavaScript modules/assets as appropriate.

## Details
If the asset being rendered does not use the JavaScript module template, leaves the rendered modules in place instead of replacing the content with a marker.

## How it was tested
Used built version against a separate project with mini-css-extract-plugin in use.

### DIFF
--- a/webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
+++ b/webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
@@ -226,7 +226,8 @@ export class ModuleMinifierPlugin implements webpack.Plugin {
          */
         function dehydrateAsset(modules: Source, chunk: webpack.compilation.Chunk): Source {
           for (const mod of chunk.modulesIterable) {
-            if (mod.id === null || !submittedModules.has(mod.id)) {
+            // If the id is null, it won't be part of the chunk
+            if (mod.id !== null && !submittedModules.has(mod.id)) {
               console.error(
                 `Chunk ${chunk.id} failed to render module ${mod.id} for ${(mod as IExtendedModule).resource}`
               );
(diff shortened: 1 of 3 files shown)


webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
```diff
@@ -467,7 +451,7 @@ export class ModuleMinifierPlugin implements webpack.Plugin {
                     }
                   );
                 } else {
-                  // Skip minification for all other assets, though the modules still are
+                  // Skip minification for all other assets, though the modules still might be
```
D4N14L, line 454: This comment seems incomplete

webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
```diff
@@ -509,9 +493,31 @@ export class ModuleMinifierPlugin implements webpack.Plugin {
           }
         );
 
-        for (const template of [compilation.chunkTemplate, compilation.mainTemplate]) {
-          (template as unknown as IExtendedChunkTemplate).hooks.modules.tap(TAP_AFTER, dehydrateAsset);
-        }
+        (compilation.chunkTemplate as unknown as IExtendedChunkTemplate).hooks.modules.tap(
```
D4N14L, line 496: Will these two ever be tapped differently from eachother? Would it make sense to create a common func to share the logic?

D4N14L (approved):

webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
dmichon-msft: This was done because the signatures don't match. The `moduleTemplate` parameter is 3rd in `chunkTemplate`, 4th in `mainTemplate`.

D4N14L: Yeah, that makes sense for being why you changed it. I'm talking about the contents of the tap. You could still call out into the same func for both though.

PR #2870, opened 2021-08-21
[heft-jest-plugin] Upgrade to Jest 26

jonasb:
## Summary

Upgrade Jest dependency to 26

## Details

The driver for updating Jest for me is to get the [improved fake timers](https://jestjs.io/blog/2020/05/05/jest-26). AFAICT this works well. I've tried to upgrade to Jest 27 as well but have run into some issues, so creating this PR for now.

If someone wants to have a look at Jest 27 the WIP branch is here: https://github.com/jonasb/rushstack/tree/jest-27

## How it was tested

* `rush build`

### DIFF
--- a/apps/api-documenter/package.json
+++ b/apps/api-documenter/package.json
@@ -33,6 +33,6 @@
     "@types/js-yaml": "3.12.1",
     "@types/node": "10.17.13",
     "@types/resolve": "1.17.1",
-    "jest": "~25.4.0"
+    "jest": "~26.6.2"
   }
 }
--- a/apps/rush-lib/package.json
+++ b/apps/rush-lib/package.json
@@ -76,7 +76,7 @@
     "@types/strict-uri-encode": "2.0.0",
     "@types/tar": "4.0.3",
     "@types/z-schema": "3.16.31",
-    "jest": "~25.4.0",
+    "jest": "~26.6.2",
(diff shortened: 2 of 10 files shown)


dmichon-msft:
When testing locally I encountered substantial performance degredation when upgrading from Jest 25 -> 26 (or 27), on the order of 40%. Something to check for and see if we can mitigate.

elliot-nelson:
FWIW, on my MacBook, `rush-lib` unit tests run in ~30 seconds on Jest 25.4 and ~28 seconds on Jest 26.6.  That's averaging a few passes in a row for each, but of course isn't particularly scientific.

It seems like a bummer to delay an upgrade of a core lib like Jest, but 40% is a concerning number.  I wonder if that comes up for certain types of unit tests or expectations that aren't present in rush-lib.

dmichon-msft:
Hm, probably need to check my numbers again. I went from 25 all the way to 27 and had noticed some regression, that got worse when I rolled back to 26, but it may be heavily correlated with specific details of how the jest project is configured. I've been doing my best to eliminate babel-jest (and transformations in general) in projects I manage, since that tends to be the largest expense of the load time.

iclanton:
Supplanted by https://github.com/microsoft/rushstack/pull/3076

PR #2873, opened 2021-08-26
[heft-sass-plugin] Support emitting the compiled CSS to disk

dmichon-msft:
## Summary
Adds an option `cssOuputFolder` to sass.json, which, if specified, will cause the Heft SASS plugin to emit the compiled CSS into said folder, preserving folder structure. If the source file name does not exactly end in `.css`, the `.css` extension will be appended to keep resolution simple (e.g. `.scss.css`).

The intended use is for consuming projects to not need to recompile the SASS of packages they import, while still being able to control how it is packaged for the final application. SASS is an implementation detail, but making the CSS available improves flexibility (e.g. to use css-in-js or to extract it into a concatenated stylesheet with the mini-css-extract-plugin)

## Details
Modifies the `typings-generator` infrastructure to understand the concept of producing multiple output files for a give input.
Modifies the `heft-sass-plugin` to have a config option `cssOutputFolder` that resolves relative to the project build folder. After the SASS compiler runs, the output CSS (before module CSS parsing) will be emitted to disk at this location, if specified.

## How it was tested
Used the `heft-sass-test` project, which has been modified to emit compiled SASS into `lib/`, no longer directly copy the `.sass`, `.scss`, or `.css` files from the source tree, and skip the `sass-loader` in the webpack config.
Dependency watching for the sass files is handled within the sass plugin.

### DIFF
--- a/build-tests/heft-sass-test/config/sass.json
+++ b/build-tests/heft-sass-test/config/sass.json
@@ -1,0 +1,3 @@
+{
+  "cssOutputFolders": ["lib"]
+}
--- a/build-tests/heft-sass-test/config/typescript.json
+++ b/build-tests/heft-sass-test/config/typescript.json
@@ -52,7 +52,7 @@
     /**
      * File extensions that should be copied from the src folder to the destination folder(s).
      */
-    "fileExtensions": [".sass", ".scss", ".css"]
+    "fileExtensions": []
 
     /**
      * Glob patterns that should be explicitly included.
--- a/build-tests/heft-sass-test/src/_stylesImport.sass
+++ b/build-tests/heft-sass-test/src/_stylesImport.sass
@@ -1,6 +1,5 @@
 /**
- * This file gets copied to the "lib" folder because its extension is registered in copy-static-assets.json
- * Then Webpack uses sass-loader to transpile, css-loader to embed, and finally style-loader to apply it to the DOM.
+ * This file is a SASS partial and therefore has no direct output file, but gets embedded into other files.
  */
 
 // This will be imported by styles.sass
--- a/common/changes/@rushstack/heft-sass-plugin/sass-plugin-emit_2021-08-26-00-29.json
+++ b/common/changes/@rushstack/heft-sass-plugin/sass-plugin-emit_2021-08-26-00-29.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-sass-plugin",
+      "comment": "Add a \"cssOuputFolders\" option to \"config/sass.json\" to emit compiled CSS.",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@rushstack/heft-sass-plugin",
+  "email": "dmichon-msft@users.noreply.github.com"
+}
--- a/heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
+++ b/heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
@@ -2,10 +2,10 @@
 // See LICENSE in the project root for license information.
 
 import * as path from 'path';
-import { render, Result } from 'node-sass';
+import { render, Result, SassError } from 'node-sass';
 import postcss from 'postcss';
 import cssModules from 'postcss-modules';
-import { LegacyAdapters } from '@rushstack/node-core-library';
+import { FileSystem } from '@rushstack/node-core-library';
 import { IStringValueTypings, StringValuesTypingsGenerator } from '@rushstack/typings-generator';
 
 /**
@@ -23,6 +23,11 @@ export interface ISassConfiguration {
    * Defaults to "temp/sass-ts/".
    */
   generatedTsFolder?: string;
+
+  /**
+   * Output directories for compiled CSS
+   */
+  cssOutputFolders?: string[] | undefined;
 
   /**
    * Determines if export values are wrapped in a default property, or not.
@@ -87,6 +92,17 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
       sassConfiguration.exportAsDefault === undefined ? true : sassConfiguration.exportAsDefault;
     const exportAsDefaultInterfaceName: string = 'IExportStyles';
     const fileExtensions: string[] = sassConfiguration.fileExtensions || ['.sass', '.scss', '.css'];
+    const { cssOutputFolders } = sassConfiguration;
+
+    const getCssPaths: ((relativePath: string) => string[]) | undefined = cssOutputFolders
+      ? (relativePath: string): string[] => {
+          return cssOutputFolders.map(
+            (folder: string) =>
+              `${folder}/${relativePath.endsWith('.css') ? relativePath : `${relativePath}.css`}`
+          );
+        }
+      : undefined;
+
     super({
       srcFolder,
       generatedTsFolder,
@@ -95,24 +111,55 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
       fileExtensions,
       filesToIgnore: sassConfiguration.excludeFiles,
 
+      getAdditionalOutputFiles: getCssPaths,
+
       // Generate typings function
-      parseAndGenerateTypings: async (fileContents: string, filePath: string) => {
+      parseAndGenerateTypings: async (fileContents: string, filePath: string, relativePath: string) => {
         if (this._isSassPartial(filePath)) {
           // Do not generate typings for Sass partials.
           return;
         }
+
         const css: string = await this._transpileSassAsync(
           fileContents,
           filePath,
           buildFolder,
           sassConfiguration.importIncludePaths
         );
-        const classNames: string[] = await this._getClassNamesFromCSSAsync(css, filePath);
-        const sortedClassNames: string[] = classNames.sort((a, b) => a.localeCompare(b));
-        const sassTypings: IStringValueTypings = { typings: [] };
-        for (const exportName of sortedClassNames) {
-          sassTypings.typings.push({ exportName });
+
+        let classMap: IClassMap = {};
+        const cssModulesClassMapPlugin: postcss.Plugin<TCssModules> = cssModules({
+          getJSON: (cssFileName: string, json: IClassMap) => {
+            // This callback will be invoked durint the promise evaluation of the postcss process() function.
+            classMap = json;
+          },
+          // Avoid unnecessary name hashing.
+          generateScopedName: (name: string) => name
+        });
+
+        await postcss([cssModulesClassMapPlugin]).process(css, { from: filePath });
+
+        if (getCssPaths) {
+          await Promise.all(
+            getCssPaths(relativePath).map(async (cssFile: string) => {
+              // The typings generator processes files serially and the number of output folders is expected to be small,
+              // thus throttling here is not currently a concern.
+              await FileSystem.writeFileAsync(cssFile, css, {
+                ensureFolderExists: true
+              });
+            })
+          );
         }
+
+        const sortedClassNames: string[] = Object.keys(classMap).sort();
+
+        const sassTypings: IStringValueTypings = {
+          typings: sortedClassNames.map((exportName: string) => {
+            return {
+              exportName
+            };
+          })
+        };
 
         return sassTypings;
       }
@@ -133,20 +180,33 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
     buildFolder: string,
     importIncludePaths: string[] | undefined
   ): Promise<string> {
-    const result: Result = await LegacyAdapters.convertCallbackToPromise(render, {
-      data: fileContents,
-      file: filePath,
-      importer: (url: string) => ({ file: this._patchSassUrl(url) }),
-      includePaths: importIncludePaths
-        ? importIncludePaths
-        : [path.join(buildFolder, 'node_modules'), path.join(buildFolder, 'src')],
-      indentedSyntax: path.extname(filePath).toLowerCase() === '.sass'
-    });
+    const result: Result = await new Promise(
+      (resolve: (result: Result) => void, reject: (err: Error) => void) => {
+        render(
+          {
+            data: fileContents,
+            file: filePath,
+            importer: (url: string) => ({ file: this._patchSassUrl(url) }),
+            includePaths: importIncludePaths
+              ? importIncludePaths
+              : [path.join(buildFolder, 'node_modules'), path.join(buildFolder, 'src')],
+            indentedSyntax: path.extname(filePath).toLowerCase() === '.sass'
+          },
+          (err: SassError, result: Result) => {
+            if (err) {
+              // Extract location information and format into the error message until we have a concept
+              // of location-aware diagnostics in Heft.
+              return reject(new Error(`${err.file}(${err.column},${err.line}): ${err.message}`));
+            }
+            resolve(result);
+          }
+        );
+      }
+    );
 
     // Register any @import files as dependencies.
-    const target: string = result.stats.entry;
     for (const dependency of result.stats.includedFiles) {
-      this.registerDependency(target, dependency);
+      this.registerDependency(filePath, dependency);
     }
 
     return result.css.toString();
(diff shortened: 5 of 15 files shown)


build-tests/heft-sass-test/src/_stylesImport.sass
```diff
@@ -1,6 +1,5 @@
 /**
- * This file gets copied to the "lib" folder because its extension is registered in copy-static-assets.json
- * Then Webpack uses sass-loader to transpile, css-loader to embed, and finally style-loader to apply it to the DOM.
+ * This file is a SASS partial, so has no output file, but gets embedded into other files.
```
halfnibble, line 2: Suggested wording:
```suggestion
 * This file is a SASS partial and has no corresponding output file, but gets embedded into other files.
```

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
```diff
@@ -133,45 +181,43 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
...
-    const result: Result = await LegacyAdapters.convertCallbackToPromise(render, {
-      data: fileContents,
-      file: filePath,
-      importer: (url: string) => ({ file: this._patchSassUrl(url) }),
-      includePaths: importIncludePaths
-        ? importIncludePaths
-        : [path.join(buildFolder, 'node_modules'), path.join(buildFolder, 'src')],
-      indentedSyntax: path.extname(filePath).toLowerCase() === '.sass'
-    });
+    const result: Result = await new Promise(
```
halfnibble, line 184: This style is perfectly fine with me, but I just want to make sure we don't have a particular style guideline for callback apis. @iclanton or @D4N14L do we recommend using LegacyAdapters.convertCallbackToPromise wrapped in try/catch, or is creating a new promise acceptable or perhaps preferred?

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
```diff
@@ -133,45 +181,43 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
...
+          }
+        );
+      }
+    );
 
     // Register any @import files as dependencies.
-    const target: string = result.stats.entry;
     for (const dependency of result.stats.includedFiles) {
-      this.registerDependency(target, dependency);
+      this.registerDependency(filePath, dependency);
```
halfnibble, line 210: This appears like a subtle and correct change, however, I'm not sure it is best. 
E.g. If we have a set of Sass files with @include statements such that:
A.sass -> B.sass -> C.sass
and C.sass is changed, do we want to start at the top (A.sass, which I believe is result.stats.entry) and work our way down, or do we want to start at B.sass (the next parent filePath) and compile and then trigger A.sass (which will work down the dependency to B.sass anyway).

I'm not sure this will generate extra work, but I think it is worth investigating.

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
```diff
@@ -87,6 +92,20 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
       sassConfiguration.exportAsDefault === undefined ? true : sassConfiguration.exportAsDefault;
     const exportAsDefaultInterfaceName: string = 'IExportStyles';
     const fileExtensions: string[] = sassConfiguration.fileExtensions || ['.sass', '.scss', '.css'];
+    const { cssOutputFolder } = sassConfiguration;
+
+    const cssFolder: string | undefined = cssOutputFolder
+      ? path.resolve(buildFolder, cssOutputFolder)
```
halfnibble, line 98: Should we use `path.join` just in case the output folder has not been generated yet?

heft-plugins/heft-sass-plugin/src/schemas/heft-sass-plugin.schema.json
```diff
@@ -32,6 +32,11 @@
       "description": "Determines if export values are wrapped in a default property, or not."
     },
 
+    "cssOutputFolder": {
+      "type": "string",
+      "description": "If specified, will output .css and .scss.json files."
```
halfnibble, line 37: Are we outputting .scss.json files? In the heft-sass-test project, I'm seeing `.scss.css` and `.sass.css` files that correspond to the .scss and .sass style files in that project.

build-tests/heft-sass-test/config/sass.json
```diff
@@ -0,0 +1,3 @@
+{
+  "cssOutputFolder": "lib"
```
halfnibble, line 2: Do you think this should be an array? As long as we only intend to run the bundler in one output directory, then having a single output folder is all we need. And I can't think of any good reason to also output to other directories. But I also can't help but wonder if we should support more than one output folder. 🤔

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
D4N14L: path.resolve doesn't care whether it exists or not, it just resolves the path to remove relative pathing, ex. `/a/../b` -> `/b`

halfnibble: Ah, interesting. I don't know why I thought it actually resolved to a real disk path, but thanks for clarifying and helping me correct a misunderstanding. :)

halfnibble: Wait a minute, does this mean that a cssOutputFolder of `'lib'` would not resolve correctly? I feel like the docs here on node.js could be better.
https://nodejs.org/api/path.html#pathresolvepaths

> The given sequence of paths is processed from right to left, with each subsequent path prepended until an absolute path is constructed. For instance, given the sequence of path segments: /foo, /bar, baz, calling path.resolve('/foo', '/bar', 'baz') would return /bar/baz because 'baz' is not an absolute path but '/bar' + '/' + 'baz' is.

D4N14L: Yeah, the docs aren't quite clear... In this case, the absolute path would be the `buildFolder` and the `cssOutputFolder` would be appended, since it's not an absolute path.

halfnibble: Ah ha ha. I see how I misread the docs and what it actually means now. Thanks. I'll mark this resolved.

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
D4N14L: Hmm, in all honesty I kinda like using `Promise` directly when the signature is more complicated, it seems to clarify to me what is going on. However, it's also very doable to use the `LegacyAdapters` method and catch the resulting error. The original error should bubble up anyway, so this seems like a stylistic preference.

libraries/typings-generator/src/TypingsGenerator.ts
```diff
@@ -80,89 +87,141 @@ export class TypingsGenerator {
...
+        cwd: this._options.srcFolder
+      });
+
+      const queue: Set<string> = new Set();
+      let timeout: NodeJS.Timeout | undefined = undefined;
+      const flushQueue: () => void = () => {
+        const toProcess: string[] = Array.from(queue);
+        queue.clear();
+        timeout = undefined;
+        this._reprocessFiles(toProcess).catch(reject);
```
D4N14L, line 124: Since this is kicking off an unawaited `Promise` on every change, and `this._reprocessFiles(...)` touches a handful of fields (`this._relativePaths`, `this._consumersOfFile`, and `this._dependenciesOfFile`), should we wait for the previous `this._reprocessFiles(...)` call to complete to avoid weird race-condition issues? It seems like you're using a 100ms debounce to dedupe file changes, which would help... I imagine my case would be rather rare, and would only occur if two batches large enough to take a long time to process _and_ processes the same file in both during processing that they collide.

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
dmichon-msft: I did this because I wanted to be clearer about that the error type was a SassError object, while Promise makes no guarantees about the shape of its rejected error type.

dmichon-msft: The SassError contains the extra file, column, and line information which is very helpful to developers when surfacing the error.

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
dmichon-msft: `result.stats.entry` and `filePath` are expected to refer to the same file, except that since the values will be compared via exact string comparisons it is better to use the value passed into the function as the key in the table, in order to ensure that we find the file in the dependency table.

heft-plugins/heft-sass-plugin/src/schemas/heft-sass-plugin.schema.json
dmichon-msft: Description is outdated. Will update.

libraries/typings-generator/src/TypingsGenerator.ts
dmichon-msft: Good catch, do need to ensure the function is not reentrant.

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
halfnibble: Sounds good to me!

build-tests/heft-sass-test/config/sass.json
dmichon-msft: Technically we should probably output to the same set of directories that typescript is emitting to, but we don't have a general solution for that in place.

D4N14L (approved):

halfnibble (approved):

build-tests/heft-sass-test/config/typescript.json
```diff
@@ -52,7 +52,7 @@
     /**
      * File extensions that should be copied from the src folder to the destination folder(s).
      */
-    "fileExtensions": [".sass", ".scss", ".css"]
+    "fileExtensions": []
```
iclanton, line 55: It should probably still copy `.css` files.

common/changes/@rushstack/heft-sass-plugin/sass-plugin-emit_2021-08-26-00-29.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-sass-plugin",
+      "comment": "Support the \"cssOuputFolder\" option to emit compiled CSS.",
```
iclanton, line 5:
```suggestion
      "comment": "Add a \"cssOuputFolder\" option to \"config/sass.json\" to emit compiled CSS.",
```

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
```diff
@@ -87,6 +92,16 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
       sassConfiguration.exportAsDefault === undefined ? true : sassConfiguration.exportAsDefault;
     const exportAsDefaultInterfaceName: string = 'IExportStyles';
     const fileExtensions: string[] = sassConfiguration.fileExtensions || ['.sass', '.scss', '.css'];
+    const { cssOutputFolders } = sassConfiguration;
+
+    const getCssPaths: ((relativePath: string) => string[]) | undefined = cssOutputFolders
+      ? (relativePath: string): string[] => {
+          return cssOutputFolders.map((folder: string) =>
+            path.resolve(folder, relativePath.endsWith('.css') ? relativePath : `${relativePath}.css`)
```
iclanton, line 100:
```suggestion
            `${folder}/${relativePath.endsWith('.css') ? relativePath : `${relativePath}.css`}`
```

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
```diff
@@ -95,25 +110,53 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
...
-        for (const exportName of sortedClassNames) {
-          sassTypings.typings.push({ exportName });
+
+        let classMap: IClassMap = {};
+        const cssModulesClassMapPlugin: postcss.Plugin<TCssModules> = cssModules({
+          getJSON: (cssFileName: string, json: IClassMap) => {
+            classMap = json;
+          },
+          // Avoid unnecessary name hashing.
+          generateScopedName: (name: string) => name
```
iclanton, line 135: These are synchronous callbacks?

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
```diff
@@ -95,25 +110,53 @@ export class SassTypingsGenerator extends StringValuesTypingsGenerator {
...
+        await postcss([cssModulesClassMapPlugin]).process(css, { from: filePath });
+
+        if (getCssPaths) {
+          await Promise.all(
+            getCssPaths(relativePath).map((cssFile: string) =>
+              FileSystem.writeFileAsync(cssFile, css, {
+                ensureFolderExists: true
+              })
+            )
+          );
```
iclanton, line 147:
```suggestion
          await Promise.all(
            getCssPaths(relativePath).map(async (cssFile: string) =>
              await FileSystem.writeFileAsync(cssFile, css, {
                ensureFolderExists: true
              })
            )
          );
```

So this shows up in error stacks.

Should this use `Async.forEachAsync`?

heft-plugins/heft-sass-plugin/src/schemas/heft-sass-plugin.schema.json
```diff
@@ -46,7 +55,7 @@
       "description": "A list of paths used when resolving Sass imports.",
       "items": {
         "type": "string",
-        "pattern": "[^\\\\]"
+        "pattern": "[^\\\\\\/]"
```
iclanton, line 58: Disallowing all folder separators?

libraries/typings-generator/src/TypingsGenerator.ts
```diff
@@ -80,89 +87,166 @@ export class TypingsGenerator {
...
+
+        setTimeout(debouncedFlush, 100);
+      };
+
+      watcher.on('add', onChange);
+      watcher.on('change', onChange);
+      watcher.on('unlink', async (relativePath) => {
+        await Promise.all(
+          this.getOutputFilePaths(relativePath).map((outputFile: string) => {
+            return FileSystem.deleteFileAsync(outputFile);
```
iclanton, line 166:
```suggestion
          this.getOutputFilePaths(relativePath).map(async (outputFile: string) => {
            await FileSystem.deleteFileAsync(outputFile);
```

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
dmichon-msft: The TypingsGenerator currently runs on all files serially, so at most this will have the concurrency of the number of requested output folders.

heft-plugins/heft-sass-plugin/src/schemas/heft-sass-plugin.schema.json
dmichon-msft: Intent was to forbid absolute paths. Revising.

heft-plugins/heft-sass-plugin/src/SassTypingsGenerator.ts
dmichon-msft: Guaranteed to be invoked by the time the awaited promise completes.

build-tests/heft-sass-test/config/typescript.json
dmichon-msft: CSS files go through the same pipeline as the scss files, so copying is taken care of by the sass plugin.

iclanton (approved):

PR #2875, opened 2021-08-27
[heft] Ensure proper casing when file path is in different system environments

Claudiazhaoya:
## Summary

This PR is to fix the casing issues that path causes in different environments.  

## Details

The context of this PR is I encounter a issue as follows in a Windows build environment where the drive letter is lowercased in the CWD:
Encountered 1 warnings:
##[error]  [webpack] (undefined) (webpack)/buildin/global.js
##[error]There are multiple modules with names that only differ in casing.
##[error]This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.

This error cannot be reproduced locally because my local env is different than the cloud env. 

This is how changing chdir to the string "src" fix the problem for me. It doesn't have to be 'src', any must-exist folder can be here. The point is I need to do chdir twice to make the drive letter have the correct case.
PS D:\rushstack> node -e "console.log(process.cwd())"
D:\rushstack
PS D:\rushstack> node -e "process.chdir('d:\\rushstack'), console.log(process.cwd())"
D:\rushstack
PS D:\rushstack> node -e "process.chdir('D:\\'), process.chdir('d:\\rushstack'), console.log(process.cwd())"
d:\rushstack


elliot-nelson:
I think this change could use some more context:

 - Your steps for reproducing the issue in the PR body, what problem exactly are they reproducing? How does this affect the behavior of rush in your cloud build?
 - How does chdir to the string "src" fix the problem for you?  This doesn't seem related to the case-sensitivity problem -- perhaps it's specific to your environment / monorepo folder structure?

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -229,6 +229,7 @@ export class HeftToolsCommandLineParser extends CommandLineParser {
     if (currentCwd !== buildFolder) {
       // Update the CWD to the project's build root. Some tools, like Jest, use process.cwd()
       this._terminal.writeVerboseLine(`CWD is "${currentCwd}". Normalizing to project build folder.`);
+      process.chdir('src');
```
iclanton, line 232: Why does changing the CWD to something else and then back to the buildfolder fix the issue?

What if the `src` folder doesn't exist?

common/changes/@rushstack/heft/user-zhas-fix-cloud-build-case-error_2021-08-27-22-45.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft",
+      "comment": "Fix the cloud build error caused by process.chdir",
```
iclanton, line 5:
```suggestion
      "comment": "Ensure `process.cwd()` is set to the project root with correct file path casing.",
```

iclanton (changes requested):

apps/heft/src/cli/HeftToolsCommandLineParser.ts
Claudiazhaoya: We verified process.chdir(process.cwd().toLowerCase()) doesn't do anything.
Not sure the potential reason, but it looks like if the only thing different is the casing, we'll have to bounce to a different directory first. 
Something like:
PS D:\rushstack> node -e "console.log(process.cwd())"
D:\rushstack
PS D:\rushstack> node -e "process.chdir('d:\rushstack'), console.log(process.cwd())"
D:\rushstack
PS D:\rushstack> node -e "process.chdir('D:\'), process.chdir('d:\rushstack'), console.log(process.cwd())"
d:\rushstack

Claudiazhaoya: I'm looking for a folder that must exist in a project. Does such folder exist?

iclanton: Perhaps we can ensure the `.heft` folder has been created at this point and use that? That's the only one we can actually guarantee exists because it's the only one we can definitively control.

Claudiazhaoya: Can we ensure that or we haven't? If not, we need to ensure that before doing chdir

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -229,6 +229,9 @@ export class HeftToolsCommandLineParser extends CommandLineParser {
     if (currentCwd !== buildFolder) {
       // Update the CWD to the project's build root. Some tools, like Jest, use process.cwd()
       this._terminal.writeVerboseLine(`CWD is "${currentCwd}". Normalizing to project build folder.`);
+      if (FileSystem.existsAsync('.heft')) {
+        process.chdir('.heft');
+      }
```
iclanton, line 234: This won't work the way you expect. `FileSystem.existsAsync` will always return truthy because it returns a promise, even if that promise resolves to `false`. Once that gets fixed, this won't correct the path if a) the current folder isn't already the project root (i.e. - someone is invoking Heft from `/my-project/src`) or if it's the first time Heft has been run on a fresh clone and the `.heft` folder doesn't exist.

The `.heft` folder's absolute path is also already exposed as `this._heftConfiguration.projectHeftDataFolder`.

iclanton: My recommendation would be to make this function `async` and change this to:

```suggestion
      await FileSystem.ensureFolderAsync(this._heftConfiguration.projectHeftDataFolder);
      process.chdir(this._heftConfiguration.projectHeftDataFolder);
```

iclanton: As an aside - is there a performance hit for changing the cwd twice? What does `chdir` do under the hood?

Claudiazhaoya: @dmichon-msft Any idea how this behavior would change the performance?

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -222,13 +222,14 @@ export class HeftToolsCommandLineParser extends CommandLineParser {
...
 
-  private _normalizeCwd(): void {
+  private async _normalizeCwd(): Promise<void> {
     const buildFolder: string = this._heftConfiguration.buildFolder;
     this._terminal.writeLine(`Project build folder is "${buildFolder}"`);
     const currentCwd: string = process.cwd();
     if (currentCwd !== buildFolder) {
       // Update the CWD to the project's build root. Some tools, like Jest, use process.cwd()
       this._terminal.writeVerboseLine(`CWD is "${currentCwd}". Normalizing to project build folder.`);
+      await FileSystem.ensureFolderAsync(this._heftConfiguration.projectHeftDataFolder);
```
iclanton, line 232: You're missing the `process.chdir(this._heftConfiguration.projectHeftDataFolder);` part.

Also we should make sure this folder doesn't immediately get deleted again if they're running `heft clean` or `heft build --clean`.

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -222,13 +221,15 @@ export class HeftToolsCommandLineParser extends CommandLineParser {
...
-    const buildFolder: string = this._heftConfiguration.buildFolder;
+  private async _normalizeCwdAsync(): Promise<void> {
+    let buildFolder: string = this._heftConfiguration.buildFolder;
     this._terminal.writeLine(`Project build folder is "${buildFolder}"`);
     const currentCwd: string = process.cwd();
     if (currentCwd !== buildFolder) {
       // Update the CWD to the project's build root. Some tools, like Jest, use process.cwd()
       this._terminal.writeVerboseLine(`CWD is "${currentCwd}". Normalizing to project build folder.`);
+      // The reason why we need to redirect to another directory first is to solve the sensitive casing issues in cloud buiild
+      process.chdir(path.resolve('..', __dirname));
```
dmichon-msft, line 232:
```suggestion
      // If `process.cwd()` and `buildFolder` differ only by casing on Windows, the chdir operation will not fix the casing, which is the entire purpose of the exercise.
      // As such, chdir to a different directory first. That directory needs to exist, so use the parent of the current directory. This will not work if the current folder is the drive root, but that is a rather exotic case.
      process.chdir(path.dirname(buildFolder));
```

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -222,13 +221,15 @@ export class HeftToolsCommandLineParser extends CommandLineParser {
     process.exitCode = 0;
   }
 
-  private _normalizeCwd(): void {
-    const buildFolder: string = this._heftConfiguration.buildFolder;
+  private async _normalizeCwdAsync(): Promise<void> {
```
dmichon-msft, line 224:
```suggestion
  private _normalizeCwd(): void {
```
No longer has any async operations.

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -164,7 +163,7 @@ export class HeftToolsCommandLineParser extends CommandLineParser {
     this._terminalProvider.verboseEnabled = this.isDebug;
 
     try {
-      this._normalizeCwd();
+      await this._normalizeCwdAsync();
```
dmichon-msft, line 166:
```suggestion
      this._normalizeCwd();
```
No longer needs to be async

apps/heft/src/cli/HeftToolsCommandLineParser.ts
dmichon-msft: This needs to be resolved relative to the project. `__dirname` may be nonsensical if the developer is using Yarn.

dmichon-msft (approved):

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -229,6 +228,10 @@ export class HeftToolsCommandLineParser extends CommandLineParser {
     if (currentCwd !== buildFolder) {
       // Update the CWD to the project's build root. Some tools, like Jest, use process.cwd()
       this._terminal.writeVerboseLine(`CWD is "${currentCwd}". Normalizing to project build folder.`);
+      // If `process.cwd()` and `buildFolder` differ only by casing on Windows, the chdir operation will not fix the casing, which is the entire purpose of the exercise.
```
sachinjoseph, line 231: Use `trueCasePathSync` like it's used [here](https://github.com/microsoft/rushstack/blob/master/apps/rush-lib/src/api/RushConfiguration.ts)?

D4N14L: This was discussed offline, but unfortunately it can't be used here since, while it will change the path to have the correct drive casing, the subsequent `process.chdir` to the correctly cased path is a no-op since the path is the same (even though it has different casing).

apps/heft/src/cli/HeftToolsCommandLineParser.ts
iclanton: We use `__dirname` all over the place when resolving schemas and no one has complained about that on Yarn. Changing to `__dirname` may actually be the best option here.

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -229,6 +228,10 @@ export class HeftToolsCommandLineParser extends CommandLineParser {
...
+      // As such, chdir to a different directory first. That directory needs to exist, so use the parent of the current directory.
+      // This will not work if the current folder is the drive root, but that is a rather exotic case.
+      process.chdir(path.resolve(__dirname, '..'));
```
dmichon-msft, line 234: If we are trusting that `__dirname` maps to a physical folder, then the safest option is:
```suggestion
      process.chdir(__dirname);
```

dmichon-msft (commented):
One minor quibble still.

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
-
+import * as path from 'path';
```
iclanton, line 3:
```suggestion

```

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -1,6 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
-
 import {
```
iclanton, line 3:
```suggestion

import {
```

dmichon-msft (approved):

apps/heft/src/cli/HeftToolsCommandLineParser.ts
```diff
@@ -16,7 +16,6 @@ import {
 } from '@rushstack/node-core-library';
 import { ArgumentParser } from 'argparse';
 import { SyncHook } from 'tapable';
-
 import { MetricsCollector } from '../metrics/MetricsCollector';
```
iclanton, line 19:
```suggestion

import { MetricsCollector } from '../metrics/MetricsCollector';
```

iclanton (approved):

PR #2877, opened 2021-08-29
[rush] Add semver defaults to be used by `rush add`

### LINKED ISSUE #1468: [rush] ignores config options from common/config/rush/.npmrc

- [ ] Feature
- [x] Bug

I've added `save-exact=true` to `common/config/rush/.npmrc` file, but that setting is being ignored.

The docs state:

```
# Rush uses this file to configure the package registry, regardless of whether the
# package manager is PNPM, NPM, or Yarn.  Prior to invoking the package manager,
```

So if it is only using `.npmrc` for registry info, then maybe it is a misnomer to name the file `.npmrc` as that makes it appear as if it will be able to configure other aspects of NPM.

Btw, I am aware there is `--exact` flag for `rush add`, but it is difficult to codify that in the team environment. That is why I usually use a local `.npmrc` file.

1. Add `save-exact=true` to `common/config/rush/.npmrc`
2. Run `rush add $package_name`
3. Open `package.json` and observe that the version is not exact.

The version to be exact.

* **Tool:** rush
* **Tool Version:** 5.11.2
* **Node Version:** 10.15.3
* **OS:** macOS

Js-Brecht:
## Summary

Adds `dependencyVersionDefaults` to `common-versions.json`, to be consumed by `rush add` for determining a user-defined default semver style.

## Details

It's sometimes desired to specify different default semver styles for various kinds of packages, or simply to define a global fallback.

To accomplish this, support for `dependencyVersionDefaults` has been added to `CommonVersionsConfiguration` (and related schema/configs), and it exposes the `getDefaultSemverForPackage` method to calculate the default semver for a package.

`PackageJsonUpdater` now exposes `SemverStyle.Auto`; when no semver style is declared explicitly on the command line, `SemverStyle.Auto` will be used. With that, `PackageJsonUpdater` knows that it should ask `CommonVersionsConfiguration` for the desired semver for the package.

The global fall back for an auto semver selection is `tilde`.  This accounts for when no `dependenyVersionDefaults` are defined, or when none are matched.

Since there is no `--tilde` cli option on `rush add`, this `dependencyVersionDefaults` could potentially disallow using tilde semvers entirely.  Because of this, I've added the `--tilde` flag to `AddAction`.

---

One thing I wasn't entirely sure about was the desired method for sharing utilities.  I wanted to use the same `matchesWithStar` that `DeployManager` uses, so I moved it from `logic/deploy/Util.ts` to `utilities/Utilities.ts`

---

Fixes #1468 

## How it was tested

* Unit tests for `CommonVersionsConfiguration.getDefaultSemverStyle()`
* Manually testing it against one of my rush monorepos

#### Test cases

* Version selection for package existing in repo matches existing
* All semver cli options for `rush add` work to explicitly set the desired semver style
  * `--caret`
  * `--tilde`
  * `--exact`
  * "passthrough"
* Specific package name patterns match only that package
* Package name patterns with wildcard match correctly
* Fallback pattern (`*`) provides general default for semver style
* Packages that do not match any dependency version default package name patterns fall back to `tilde`
* Undefined or empty `dependencyVersionDefaults` results in `tilde` semver style


iclanton:

Js-Brecht:
@iclanton thanks for letting me know.  I'll make sure to do that

PR #2880, opened 2021-08-30
Benpapp/add params to heft telemetry

BPapp-MS:
## Summary

Similar to the recent PR adding commandline parameters to the output of the rush telemetry file, this PR adds the same structure into the metrics collector data for the purpose of being read by a heft plugin.

## Details

The change is pretty straightforward, it re-uses the function for getting the stringified parameters and values that I used for the telemetry in BaseInstallAction and BulkScriptAction

## How it was tested

Testing by installing this version of heft on a local repository and logging the telemetry results pulled from the metrics collector

### DIFF
--- a/apps/heft/src/metrics/MetricsCollector.ts
+++ b/apps/heft/src/metrics/MetricsCollector.ts
@@ -108,8 +108,8 @@ export class MetricsCollector {
    */
   public record(
     command: string,
-    parameterMap: Record<string, string>,
-    performanceData?: Partial<IPerformanceData>
+    performanceData?: Partial<IPerformanceData>,
+    parameters?: Record<string, string>
   ): void {
     if (this._startTimeMs === undefined) {
       throw new InternalError('MetricsCollector has not been initialized with setStartTime() yet');
@@ -136,7 +136,7 @@ export class MetricsCollector {
       machineCores: os.cpus().length,
       machineProcessor: os.cpus()[0].model,
       machineTotalMemoryMB: os.totalmem(),
-      commandParameters: parameterMap
+      commandParameters: parameters || {}
     };
 
     this.hooks.recordMetric.call('inner_loop_heft', metricsData);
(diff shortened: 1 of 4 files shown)


common/reviews/api/heft.api.md
```diff
@@ -274,6 +274,7 @@ export interface IHeftSessionHooks {
 // @public (undocumented)
 export interface IMetricsData {
     command: string;
+    commandParameters?: Record<string, string>;
```
iclanton, line 277: Kinda seems like this should always be defined, and if no parameters are used, it should be `{}`.

iclanton (approved):

common/reviews/api/heft.api.md
BPapp-MS: At the time I think I was trying to be backwards compatible if dependent projects were calling the record function, but you're right I think this is appropriate to define always

iclanton: You can have the parameter in the `record` function be optional and not make this optional.

common/reviews/api/heft.api.md
```diff
@@ -362,7 +363,7 @@ export class _MetricsCollector {
     flushAsync(): Promise<void>;
     // (undocumented)
     readonly hooks: MetricsCollectorHooks;
-    record(command: string, performanceData?: Partial<_IPerformanceData>): void;
+    record(command: string, parameterMap?: Record<string, string>, performanceData?: Partial<_IPerformanceData>): void;
```
iclanton, line 366: To make this not a breaking API change, the new parameter should go on the end.

apps/heft/src/metrics/MetricsCollector.ts
```diff
@@ -98,9 +103,14 @@ export class MetricsCollector {
...
    *
    * @param command - Describe the user command, e.g. `start` or `build`
-   * @param params - Optional parameters
+   * @param parameterMap - Optional map of parameters to their values
+   * @param performanceData - Optional performance data
    */
-  public record(command: string, performanceData?: Partial<IPerformanceData>): void {
+  public record(
+    command: string,
+    parameterMap?: Record<string, string>,
```
iclanton, line 111: The `parameterMap` name is a little misleading since this takes a `Record`, not a `Map`

PR #2887, opened 2021-09-03
[heft] Fix building with Typescript 4.4 due to changed internal interface

jonasb:
## Summary

When building a project with Typescript 4.4 the typescript Heft plugin would fail with "Error: directoryExists is not a function". This PR fixes it by adding a missing directoryExists argument to the call to the Typescript internal `matchFiles` method.

Fixes https://github.com/microsoft/rushstack/issues/2878.

## How it was tested

* Upgraded typescript-newest-test to Typescript 4.4.2. `install-test-workspace` was failing before this change
* Upgraded heft-node-everything-test to Typescript 4.4.2 locally. Had to remove `tslint-microsoft-contrib` from that project since it wasn't compatible with Typescript 4.4. After disabling tslint for that project it built fine.


iclanton:
We should maybe consider warning about the latest tested minor version here: https://github.com/microsoft/rushstack/blob/master/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts#L235-L240.

That can be in a separate PR.

iclanton (approved):

Issue #2874, opened 2021-08-26
[rush] Both incremental build and build cache are not working for this project structure


davidabap:
## Summary
Incremental build working fine in our project previously, but it doesn't work since we bump to `rush 5.50.0` I change a little in your demo project and the issue can be reproduced. https://github.com/davidabap/rush-example
In a word, I put your demo project into `inner` folder and introduce a new my-app2 which located in the `outer` folder

## Repro steps

1. Run `rush update` in inner
2. Run `rush build` in inner * 2
3. Incremental build doesn't work
4. Delete my-app2 in `rush.json`
5. Run `rush build` in inner * 2
6. Incremental build working fine

  **Expected result:** 
Incremental build and build cache should working for this rush project sturcture like before.
  **Actual result:** 
Both Incremental build and build cache are not working 
## Details
You can find the undefined error in the `rush build --verbose` mode, the issue come from https://github.com/microsoft/rushstack/blob/master/apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts#L243
![image](https://user-images.githubusercontent.com/25033055/130891490-fc298ddf-c248-4cb8-a435-5f6fd0bea363.png)

## Standard questions

Please answer these questions to help us investigate your issue more quickly:

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.52.0 |
| `rushVersion` from rush.json? | 5.52.0 |
| `useWorkspaces` from rush.json? | false |
| Operating system? | Mac |
| Would you consider contributing a PR? | No |
| Node.js version (`node -v`)? | 16.3.0 |

davidabap:
Hi @iclanton @dmichon-msft  is there any comments on that ? Is it some regression which will be fixed in new patch ? or Rush doesn't support "**project exists outside of rush.json parent folder**" ?

elliot-nelson:
Hi @davidabap!

I was looking into your issue with the folder structure above and I notice that as of https://github.com/microsoft/rushstack/pull/2838, the behavior of your situation has changed:

Instead of emitting an error/warning, it now succeeds, but the _outer_ app is actually not checked for changes.  This means the result is actually a false success -- no matter how many changes you've made to the outer app, `rush build` will never attempt to rebuild it.  (It incorrectly believes it has a "cache hit" and just restores from cache.)

[...]

PR #2892, opened 2021-09-06
[rush] Support project paths outside the rush.json parent folder

elliot-nelson:
## Summary

This PR is a potential approach for solving the issue described in #2874 by @davidabap.

The goal is to correctly support incremental builds (both build skipping and build caching) for projects that live _inside the git repo_ but _outside the rush.json parent folder_. Where possible, supporting this edge case shouldn't impact the performance for standard setups where all project folders are inside the rush.json parent folder.

## Details

 - For the most common use case, where all project folders are under the `rush.json` parent folder, there should be no changes at all to existing functionality.

 - If one or more project folders are not under the `rush.json` parent folder, find a new root folder that contains all of the project folders. All files inside this new root folder will be considered for incremental change analysis (but will be described relative to the `rush.json` parent folder, as normal).

 - With this change, it's possible to point at a relative folder path that ends up _outside_ the git repo, causing the new root folder not to be tracked by git. This is a user error that results in no change detection at all, and will likely print:

 ```console
Error calculating the state of the repo. (inner error: Error: git ls-tree exited with status 128: fatal: not a git repository (or any of the parent directories): .git
). Continuing without diffing files.

This workspace does not appear to be tracked by Git. Rush will proceed without incremental build, caching, and change detection.
```

(Overall, I feel like this would be the correct behavior in that scenario.)

 - ⚠️ This PR contains a breaking change for `package-deps-hash` (not strictly required I suppose, but 3 optional trailing parameters is beginning to stretch the bounds of a reasonable public API anyway).

## How it was tested

 - Added a couple unit tests describing the new functionality to avoid future regressions.
 - Tested before/after behavior on @davidabap's example repo.

### DIFF
--- a/libraries/package-deps-hash/src/getPackageDeps.ts
+++ b/libraries/package-deps-hash/src/getPackageDeps.ts
@@ -6,6 +6,30 @@ import * as path from 'path';
 import { Executable } from '@rushstack/node-core-library';
 
 /**
+ * Options for the `getPackageDeps` function.
+ *
+ * @public
+ */
+export interface IGetPackageDepsOptions {
+  /**
+   * An optional array of file path exclusions. If a file should be omitted from the list of
+   * dependencies, use this to exclude it. Paths must be relative to the `packagePath`.
+   */
+  excludedPaths?: string[];
+
+  /**
+   * An optional path to the `git` executable. Defaults to `git`.
+   */
+  gitPath?: string;
+
+  /**
+   * An optional path string. If provided, the returned file paths will be transformed to be
+   * relative to this path, rather than the `packagePath` passed in.
+   */
+  relativeToPath?: string;
+}
+
+/**
  * Parses a quoted filename sourced from the output of the "git status" command.
  *
  * Paths with non-standard characters will be enclosed with double-quotes, and non-standard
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by dmichon-msft** [comment]
I believe the most robust solution would be to have rush ask Git for the repo root folder during initialization, and then use paths relative to the repo root for all interactions with Git.
I started experimenting here:
https://github.com/dmichon-msft/rushstack/blob/git-repo-relative/libraries/package-deps-hash/src/getRepoDeps.ts

Part of the incentive is that the some of the stable output formats from Git commands will only return paths relative to the repo root anyway, so it is the natural folder to work with for Git operations.

Essentially the main part of the change would be to alter the internal mapping in `findProjectByPosixRelativePath` to instead be `findProjectByGitPath` to be explicit about its purpose. Then use the Git root folder for all the Git commands performed by Rush.

File: libraries/package-deps-hash/src/getPackageDeps.ts [hunk 1]
```diff
@@ -225,36 +250,43 @@ export function gitStatus(path: string, gitPath?: string): string {
...
 
   // Add all the checked in hashes
-  const result: Map<string, string> = parseGitLsTree(gitLsOutput);
+  let result: Map<string, string> = parseGitLsTree(gitLsOutput);
+
+  // If requested, transform all paths relative to a new root path
+  if (options.relativeToPath && options.relativeToPath !== packagePath) {
+    const newResult: Map<string, string> = new Map();
+    for (const [key, value] of result.entries()) {
+      newResult.set(path.relative(options.relativeToPath, path.join(packagePath, key)), value);
```
dmichon-msft, line 273: This call is both expensive in terms of performance and destroys the previously portable nature of this path (Git always returns '/' as the delimiter)

dmichon-msft: Hence my recommendation that we do the transform to the `projectRelativeFolder` to normalize it to the Git root, then look up everything in terms of Git path.

elliot-nelson:
> I believe the most robust solution would be to have rush ask Git for the repo root folder during initialization, and then use paths relative to the repo root for all interactions with Git.
> I started experimenting here:
> https://github.com/dmichon-msft/rushstack/blob/git-repo-relative/libraries/package-deps-hash/src/getRepoDeps.ts
> 
> Part of the incentive is that the some of the stable output formats from Git commands will only return paths relative to the repo root anyway, so it is the natural folder to work with for Git operations.
> 
> Essentially the main part of the change would be to alter the internal mapping in `findProjectByPosixRelativePath` to instead be `findProjectByGitPath` to be explicit about its purpose. Then use the Git root folder for all the Git commands performed by Rush.

This makes sense to me.  Since my PR is kind of a response to a corner case, I was worrying about affecting other corner cases (for example, imagine a really large corporate repo that is organized with Lerna, and a small subfolder is beginning an experiment to transition to Rush... going from "git-ls all files in the rush.json parent folder" to "git-ls the git root" would be a big increase in files in that case).

But if we're making that shift anyway for different reasons, I agree it would make things simpler.

libraries/package-deps-hash/src/getPackageDeps.ts
elliot-nelson: 🤔 You're right about the portability, although I could change to `path.posix` to fix that.

I wasn't expecting the performance comment, I was imagining both `path.` calls as essentially string operations under the covers.  (Not much slower than interacting with the Map.)

I guess I would expect that no matter what, we'd want `projectRelativeFolder` to always be relative to `rush.json`, because that's how it would behave if there was no git repo at all.  (I agree with the having package-deps-hash retrieve everything relative to git root, but then I still think you need some kind of speedy way to remap them all to be relative to rush.json).

PR #2897, opened 2021-09-08
[Rush] Allow commands to have fallbacks for build cache read

dmichon-msft:
## Summary
Allows custom bulk script commands to specify fallback alternatives for build cache reads.
For example a command "build-incremental" might fall back to "build".

Allows custom parameters to specify that they can be ignored when reading from the build cache.
For example a parameter "--no-test" could be ignored when restoring from cache.

## Details
Couples the command schemas somewhat more strongly with the concept of the build cache, in order to allow for fallback keys for cache reads.

## How it was tested
Modified the local command-line.json to have a command `build-incremental` that falls back to `build`, and make the `--no-color` flag have `ignoreForCacheRead: true`. Verified that `rush build-incremental --no-color` could be satisifed by the cache entry created by `rush build`.

### DIFF
--- a/apps/rush-lib/src/api/CommandLineJson.ts
+++ b/apps/rush-lib/src/api/CommandLineJson.ts
@@ -28,6 +28,8 @@ export interface IBulkCommandJson extends IBaseCommandJson {
   allowWarningsInSuccessfulBuild?: boolean;
   watchForChanges?: boolean;
   disableBuildCache?: boolean;
+
+  fallbackCommandsForCacheRead?: string[];
 }
 
 /**
@@ -50,6 +52,8 @@ export interface IBaseParameterJson {
   description: string;
   associatedCommands: string[];
   required?: boolean;
+
+  ignoreForCacheRead?: boolean;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


elliot-nelson:
I'd love to see this draft move forward.

Our scenario: there's a team in our monorepo that would like to use https://rushjs.io/pages/advanced/watch_mode/, but it's not tenable today because creating a `"build:watch"` bulk command means that each morning, every developer has to pull and all of the library changes they pull are not cached in the cloud build cache.

In this monorepo, we have dependencies `A -> B -> C -> D -> App`.  `A` and `B` are deep dependencies that the library developers usually, and `C` and `D` are the shallow dependencies that the app developers are touching while they touch the App.

We'd like to recommend the developers do this in their terminal windows:

```console
# Terminal 1
rush build:watch --to-except App

# Terminal 2
cd apps/App && heft start
```

But this approach adds an extra 3-5 minutes compared to their current approach:

```console
# Terminal 1
cd libraries/D && heft build --watch

# Terminal 2
cd apps/App && heft start
```

dmichon-msft:
@elliot-nelson , the main reason this is paused is that it gets almost completely obviated by phased commands, assuming that your `build:watch` command does a strict subset of the phases executed by `build`. I have a discussion issue regarding possible other optimizations to the interaction between the build cache and the scheduling engine at #3029.

iclanton:

PR #2901, opened 2021-09-10
chore: update `@types/node` version to v12

MichaelDeBoey:
Since the repo requires Node v12, we could change at least have that version of typings.


apps/api-documenter/package.json
```diff
@@ -31,7 +31,7 @@
     "@rushstack/heft-node-rig": "workspace:*",
     "@types/heft-jest": "1.0.1",
     "@types/js-yaml": "3.12.1",
-    "@types/node": "10.17.13",
+    "@types/node": "^12.20.24",
```
iclanton, line 34: Can you change these to just the version without the carat? Typings packages aren't as stable as other packages.

MichaelDeBoey: Done! 👍

iclanton (approved):

PR #2912, opened 2021-09-19
Update to TypeScript 4.4

crisbeto:
## Summary

Updates the required version of TypeScript to 4.4.2.

## Details

Projects that currently depend on packages like `api-extractor` can't use the new `useUnknownInCatchVariables` compiler option, because the version of TypeScript is set to 4.3.5. These changes update to 4.4.2.

### DIFF
--- a/apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
+++ b/apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
@@ -43,6 +43,17 @@ export interface IAzureStorageBuildCacheProviderOptions {
 
 const SAS_TTL_MILLISECONDS: number = 7 * 24 * 60 * 60 * 1000; // Seven days
 
+interface IBlobError extends Error {
+  statusCode: number;
+  code: string;
+  response?: {
+    status: string;
+    parsedHeaders?: {
+      errorCode: string;
+    }
+  }
+}
+
 export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase {
   private readonly _storageAccountName: string;
   private readonly _storageContainerName: string;
@@ -110,7 +121,8 @@ export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase 
       } else {
         return undefined;
       }
-    } catch (e) {
+    } catch (err) {
+      const e: IBlobError = err as IBlobError;
       const errorMessage: string =
         'Error getting cache entry from Azure Storage: ' +
         [e.name, e.message, e.response?.status, e.response?.parsedHeaders?.errorCode]
@@ -174,7 +186,9 @@ export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase 
 
     try {
       blobAlreadyExists = await blockBlobClient.exists();
-    } catch (e) {
+    } catch (err) {
+      const e: IBlobError = err as IBlobError;
+
       // If RUSH_BUILD_CACHE_CREDENTIAL is set but is corrupted or has been rotated
       // in Azure Portal, or the user's own cached credentials have been corrupted or
       // invalidated, we'll print the error and continue (this way we don't fail the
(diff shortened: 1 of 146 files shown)


crisbeto:
The PR resolves almost all of the errors that came up as a result of the upgrade, but I'm stuck on the these two:

```
==[ FAILURE: 2 projects ]======================================================

--[ FAILURE: heft-jest-reporters-test ]-------------[ 1 minute 44.4 seconds ]--

SyntaxError: An error occurred while adding the reporter at path "/home/vsts/work/1/s/build-tests/heft-jest-reporters-test/lib/test/customJestReporter.js".Unexpected token 'export'

--[ FAILURE: heft-typescript-composite-test ]---------------[ 58.69 seconds ]--

Error: File does not exist: /home/vsts/work/1/s/build-tests/heft-typescript-composite-test/node_modules/@types/jest
ENOENT: no such file or directory, lstat '/home/vsts/work/1/s/build-tests/heft-typescript-composite-test/node_modules/@types/jest'
```

Does anybody have ideas on what might be causing them?

crisbeto:
Question for reviewers: am I supposed to bump the package version or will it be done by whoever is publishing to npm?

apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
```diff
@@ -110,7 +121,10 @@ export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase
       } else {
         return undefined;
       }
-    } catch (e) {
+    } catch (err) {
+      // `err` here is `unknown` so we have to cast it.
+      // eslint-disable-next-line @typescript-eslint/typedef
+      const e = err as IBlobError;
```
iclanton, line 127:
```suggestion
      const e: IBlobError = err as IBlobError;
```

apps/rush-lib/src/logic/buildCache/AzureStorageBuildCacheProvider.ts
```diff
@@ -174,7 +188,11 @@ export class AzureStorageBuildCacheProvider extends CloudBuildCacheProviderBase
 
     try {
       blobAlreadyExists = await blockBlobClient.exists();
-    } catch (e) {
+    } catch (err) {
+      // `err` here is `unknown` so we have to cast it.
+      // eslint-disable-next-line @typescript-eslint/typedef
+      const e = err as IBlobError;
```
iclanton, line 194:
```suggestion
      const e: IBlobError = err as IBlobError;
```

common/changes/@microsoft/api-documenter/ts-4.4_2021-09-19-10-42.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-documenter",
+      "comment": "Update to TypeScript 4.4",
```
iclanton, line 5: Is this changelog message useful to consumers of these packages?

iclanton:
>Question for reviewers: am I supposed to bump the package version or will it be done by whoever is publishing to npm?

That will happen automatically on publish.

iclanton:
TS 3.9 -> 4.4 is a fairly big jump. Will this change make these packages incompatible with older versions of TS? Specifically, will the declarations that we ship end up containing syntax that older versions of TS won't understand. If they will, what is the oldest compatible version?

common/changes/@microsoft/api-documenter/ts-4.4_2021-09-19-10-42.json
crisbeto: Not really, but the CI was complaining that the changelog wasn't generated.

crisbeto:
The feedback has been addressed. Regarding the breaking changes between 3.9 and 4.4, I don't think that it'll be too breaking since consumers use generated code. Off the top of my head, these can be potential breakages:
* A higher version of `tslib` may be required since TS generates slightly different calls to helper functions. I _think_ that TS actually keeps backwards-compatible version of the helpers though.
* TS 4.3 allows for the type of a setter to differ from the one on the getter.
* TS 4.4 introduces `static` code blocks on classes.

common/changes/@microsoft/api-documenter/ts-4.4_2021-09-19-10-42.json
iclanton: You can generate blank changelog messages with `rush change --bulk --bump-type none`

crisbeto: Done.

iclanton:
I just pulled your branch and diffed the generated `.d.ts` files against what's built in the existing `master` branch and there aren't any differences, so this won't break compatibility.

iclanton (approved):

PR #2917, opened 2021-09-21
Lock node-fetch to avoid breaking change in 2.6.3 version

dmichon-msft:
## Summary
Version 2.6.3 of `node-fetch` made a breaking change to the encoding state of URLs that is incompatible with the expectations of `@azure/core-http` with regards to including SAS tokens in URLs. This locks the version used by Rush to 2.6.2 to avoid breaking customers who use the Azure cloud cache provider.

## How it was tested
Verified locally that installing with node-fetch 2.6.2 does not reproduce the authentication error encountered during cloud cache writes. Fresh installs started failing ~7 hours ago, which lines up with when version 2.6.3 of node-fetch was released.

### DIFF
--- a/apps/rush-lib/package.json
+++ b/apps/rush-lib/package.json
@@ -42,7 +42,7 @@
     "jszip": "~3.7.1",
     "lodash": "~4.17.15",
     "minimatch": "~3.0.2",
-    "node-fetch": "~2.6.1",
+    "node-fetch": "2.6.2",
     "npm-package-arg": "~6.1.0",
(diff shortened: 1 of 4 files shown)


sachinjoseph (approved):

D4N14L:
Should an associated issue be filed with the owners of node-fetch? Looks like this is similar: https://github.com/node-fetch/node-fetch/issues/1300

D4N14L (approved):

apostolisms (approved):

PR #2919, opened 2021-09-21
[heft] Fix directory separators in generated tsBuildInfoFile

dmichon-msft:
## Summary
Fixes incremental emit of TypeScript to only emit changed files.

## Details
TypeScript internally normalizes all directory separators to `/` so that it can use exact string comparisons to test for changes. However, it only applies this transformation to the `tsBuildInfoFile` value when serializing the tsBuildInfo JSON to disk. It expects the input in the tsconfig to have already been normalized.
As a result, constructing the path using `path.join()` on Windows produces an incorrect path that will always fail to match the serialized value.

## How it was tested
Run:
```
heft build --verbose
heft build --verbose
```
Verify that the second time does not emit any files.

### DIFF
--- a/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
+++ b/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
@@ -147,7 +147,9 @@ export class TypeScriptBuilder extends SubprocessRunnerBase<ITypeScriptBuilderCo
         .replace(/\+/g, '-')
         .replace(/\//g, '_');
 
-      this.__tsCacheFilePath = `${this._configuration.buildMetadataFolder}/ts_${serializedConfigHash}.json`;
+      // Paranoia. If backslashes sneak it in breaks incremental compilation
+      const normalizedCacheFolder: string = Path.convertToSlashes(this._configuration.buildMetadataFolder);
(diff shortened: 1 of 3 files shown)


apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
```diff
@@ -147,7 +147,9 @@ export class TypeScriptBuilder extends SubprocessRunnerBase<ITypeScriptBuilderCo
         .replace(/\+/g, '-')
         .replace(/\//g, '_');
 
-      this.__tsCacheFilePath = `${this._configuration.buildMetadataFolder}/ts_${serializedConfigHash}.json`;
+      // Paranoia. If backslashes sneak it in breaks incremental compilation
```
halfnibble, line 150: I think you meant "in it" not "it in".
Also, we can probably remove the "Paranoia" part. 😁

D4N14L (approved):

halfnibble (approved):

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
dmichon-msft: Fixed.

PR #2934, opened 2021-10-02
[node-core-library] Expose an ITerminal interface.

iclanton:
## Summary

Expose an `ITerminal` interface to allow APIs that expose or take a terminal object to use a different version of node-core-library than the version the project was built with.

## How it was tested

Built.


halfnibble (approved):
Looks good to me!

D4N14L (approved):

PR #2936, opened 2021-10-03
[rush] Add the phased command feature-related properties to command-line.json

iclanton:
## Summary

This PR adds properties and the related validation for the phased commands feature (https://github.com/microsoft/rushstack/issues/2300) to `command-line.json`.

## How it was tested

Ensured that these files validate correctly on this repo: https://github.com/elliot-nelson/rush-phased-builds-example/tree/phased


apps/rush-lib/src/schemas/command-line.schema.json
```diff
@@ -149,19 +149,117 @@
...
+          "type": "boolean"
+        },
+
+        "allowWarningsOnSuccess": {
+          "title": "Allow Warnings on Success",
+          "description": "By default, Rush returns a nonzero exit code if errors or warnings occur during a command. If this option is set to \"true\", Rush will return a zero exit code if warnings occur during the execution of this phase.",
+          "type": "boolean"
+        }
+      }
+    },
```
elliot-nelson, line 257: For phased commands, just looking at the list of properties moved over from the command to the phase:

### incremental

Is `incremental` property missing for phased commands? I don't see it on either the command or the phase. (We discussed assuming that the build cache would be enabled first, but I'm not sure if that means we don't want the ability to _force_ a command not to be incremental -- otherwise how can we support rebuild.)

### enableParallelism

On a phase, `enableParallelism` means that two instances of _that phase_ have to run sequentially (while other phases can run at the same time), or that the phase must be run "alone" (blocking ALL other phases)?

I guess I can see it working either way, and I almost wonder if we need to consider both options.

### ignoreMissingScript

Makes sense. In my initial discussion with @octogonz I was very worried about "onboarding" -- how to slowly bring up projects in a monorepo with lots of project owners and different build frameworks. But, as long as you're allowed to define a phase as "" (empty string), the easy way to onboard is just collapse everything to the first phase.

For example, a new phased system with 3 phases (`compile`, `lint`, `test`), you could copy/paste this boilerplate into every single project's `package.json`:

```js
    "_phase:compile": "[insert entire contents of 'build' script]",
    "_phase:lint": "",
    "_phase:test": ""
```

This way the entire monorepo is immediately valid, even if you haven't figured out how to break apart "compile" and "lint" for every project yet (or even if there are some projects that CANNOT break apart compile and lint).

### allowWarningsOnSuccess

:+1: A net win for phased builds, especially if you're one of those "don't care about lint warnings" people.  Now you can ignore warnings just for one phase of a build instead of the entire build.

apps/rush-lib/src/schemas/command-line.schema.json
```diff
@@ -214,6 +319,22 @@
...
+              "title": "Add phases to command",
+              "description": "Include the phases listed here in the command's execution if this parameter is specified.",
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "skipPhasesForCommand": {
+              "title": "Skip phases for command",
+              "description": "Exclude the phases listed here from the command's execution if this parameter is specified. Note that this may break dependency relationships between phases.",
```
elliot-nelson, line 333: > Note that this may break dependency relationships between phases.

This caught my eye.  I suppose nothing stops a maintainer from accidentally creating a command made up of phases `[lint, test]`, both of depend on `compile`, yet that phase is not part of the command.  This command would, I guess, exit without doing _anything_, or it would need to bail out early with some kind of error, telling the user the definition of the command doesn't make sense.

If we choose the failure path, then I think we could fall right into the same path -- `rush broken-build` would behave the same as a theoretical `rush build --no-compile`.

apps/rush-lib/src/schemas/command-line.schema.json
```diff
@@ -214,6 +319,22 @@
           "properties": {
             "parameterKind": {
               "enum": ["flag"]
+            },
+            "addPhasesToCommand": {
+              "title": "Add phases to command",
+              "description": "Include the phases listed here in the command's execution if this parameter is specified.",
```
elliot-nelson, line 325: Similar to below, adding phases to commands could suffer the same fate (a broken command).

For example, from my example repo, the command `rush push-notes --and-lint` would either not be able to lint anything, or it would need to exit with a failure because `lint` depends on `compile`, which isn't part of the command.

elliot-nelson: (FWIW, since the ability to create "broken commands" isn't unique to these new properties, I don't think it's a stumbling block for them. But informing the user outright that the set of phases they are about to execute is not valid is probably the right call in all of these situations.)

apps/rush-lib/src/schemas/command-line.schema.json
iclanton: > I suppose nothing stops a maintainer from accidentally creating a command made up of phases `[lint, test]`, both of which depend on `compile`, yet that phase is not part of the command. This command would, I guess, exit without doing anything, or it would need to bail out early with some kind of error, telling the user the definition of the command doesn't make sense.

So I said [in the description for `command.phases`](https://github.com/microsoft/rushstack/blob/f1577cd4cf311c0508b93be1f9d106c75febb37e/apps/rush-lib/src/schemas/command-line.schema.json#L175): `List the phases associated with this command. Note that phases with dependencies will be implicitly included even if they aren't explicitly enumerated in this property.` Perhaps that isn't what we want?

I guess we have a few questions to answer before this goes in, as part of the design:
- Do the phases listed in a command's `phases` property need to be a complete graph? i.e. - if `test` depends on a `compile`, can you just list `test` and it'll implicitly include `compile`? If all phases need to be listed, what happens if one isn't? Validation error?
- Isomorphic question for `addPhasesToCommand`: Let's say we have phases `D->C->B->A` (where `->` means "depends on) and we have a `build` command that runs `B` and `A`, and let's say we have a `addPhasesToCommand` that adds `D`. Should that be a validation error? What if the flag with the `addPhasesToCommand` property can be associated with a command that doesn't run any of `A`, `B`, `C`, or `D`? Should this flag's association with any command require a complete set of phases? Should it just add the phases, ignoring the dependencies? Should we have a richer object for the entries in `addPhasesToCommand` like `{ phaseName: string, ignoreDependencyGraph?: boolean }`?
- Isomorphic question for `skipPhasesForCommand`.

elliot-nelson: 🤯 

"Internally consistent" isn't the only criteria here, but if we want selecting phases to be as close as possible to selecting projects (https://rushjs.io/pages/developer/selecting_subsets/), then I guess I would expect:

### Treating "add" like "--to" and "skip" like "--to-except"

- Using `"addPhase": "lint"` _or_ `"phases": ["lint"]` is like saying `--to lint` -- Rush automatically includes any other phases you intended.  That's not what I was thinking yesterday but I can't think of a good reason not to have Rush help out the user

- Using `"skipPhase": "compile"` is _maybe_ like saying `--to-except compile`? One way to interpret this is: take the raw list of phases for command, add any `addPhase` phases as appropriate, _then_ compute the DAG of phases and add in any missing ones, _then_ remove the skipped phases.  Now you run the DAG, and any time you encounter a request to run a phase that isn't in your list, you just pretend it didn't happen (do not check build cache, do not build, do not pass go).

This feels consistent to me because _like_ `--to-except`, you could use this in situations where you understand more about the current context than Rush does.  As an example, this might be a valid way to use Rush if you were hunting for non-deterministic test cases:

```console
# "Ignore" compile after first time
rush build # => [_phase:compile, _phase:lint, _phase:test]
rush build --no-compile # => [_phase:lint, _phase:test]
rush build --no-compile # => [_phase:lint, _phase:test]

# Similar conceptually to
rush build --to @acme/dashboard
rush build --only @acme/dashboard
rush build --only @acme/dashboard
```

Something I don't like about this approach is that it's impossible to build the set of phases `[_phase:lint, _phase:test]` for a command -- you can only do it by also introducing some kind of parameter. That feels inconsistent to me in a different way.

elliot-nelson: OK, now that I've got that answer out of my system, I don't know if I love it.  Another option:

### Aggressive validation

The _project selection_ options are flags used by all developers in repo at any time, so having them do "what you expect" most of the time is important. But structuring these phases (and parameters that can modify them) happens in a configuration file and is probably relatively rarely touched. Detecting _mistakes in what the user specified_ might be much more useful then assisting them, because it helps them discover gaps in their mental model when configuring phases.

To allow aggressive validation _and_ allow the user to create "incomplete" builds, I think you'd need to change the definition of the parameters:

- `addPhaseToCommand`: add additional phase to the phase list, if it doesn't already exist
- `skipPhaseForCommand`: remove phase from the phase list, if it exists, AND add it a list of phases skipped on sight (that is, DAG Validation will not fail because this phase is missing).

Then we'd also need to add `skipPhasesForCommand` as a parameter in the `phased` command definition.

```javascript
// This is an INVALID phase definition for a command
//
// If I define it and run "rush lint", I should get:
// ERROR: Phased command includes "_phase:lint" which depends on "_phase:compile", which is not part of this command.
"phases": ["_phase:lint"]

// This is a VALID definition for the command
"phases": ["_phase:lint"],
"skipPhasesForCommand": ["_phase:compile"]
```

So far, this is my favorite approach, because:

1. The set of outputs you can construct in a phased command and using add/remove parameter flags is the same set.
2. Defining and organizing phases could be confusing, and having immediate feedback if you've constructed a command that doesn't make sense is more useful than magically helping me in potentially unintended ways.
3. I can override any errors in (2) if I really want to, either with or without parameter flags.

iclanton:
Hmm the build vs rebuild thing is a good point. I think the simplest option will be to move `incremental` to the command instead.

Made that change, and made a PR against the example repo: https://github.com/elliot-nelson/rush-phased-builds-example/pull/4

PR #2937, opened 2021-10-04
Create rush command to run project scaffolding

Claudiazhaoya:
## Summary

This PR is to create rush command to run yeoman to create project template for our inner-loop repo. 
## How it was tested

It is expected to run 'rush scaffold' so that 'yo odspweb --skip-install' will be invoked.


iclanton:
Seems like this went into the wrong repo.

PR #2953, opened 2021-10-08
Bump cyclic dependencies.

iclanton:


dmichon-msft:
@iclanton , could you also turn on the frozen lockfile flags in rush.json?

iclanton:
@dmichon-msft I turned on a few options in `experiments.json`. Are those what you were referring to?

PR #2959, opened 2021-10-11
Custom provider for Build Cache feature

manrueda:
## Summary

Add support to load any node.js module as a provider for Build Cache even if does not live inside Rush.

## Details

Different consumers/companies have different requirements around network reachable shared resources that make something simple like the Build Cache hard to implement when it does not fit into S3 or Azure Blob directly.

With this new custom provider feature any consumer of Rush will be able to use a provider that fits their needs without having to get it contributed to Rush (mostly because their case is so unique).

For example, my need is to write cache to S3 but read cache from CloudFront. This is an incredibly unique flow because of company guidelines around security. I will be able to implement this as a private package for my company without having to add code that is unique to Rush.

Also adds a new environment variable to be able to change providers, this allows to force a different behavior around caching between local and a build server/CI process.

## How it was tested

I run the buld command locally using the `rushtest` alias patter to make sure that the new provider is called.


iclanton:
Take a look at this change: https://github.com/microsoft/rushstack/pull/2900. It may handle the scenario you're implementing here.

iclanton:
@chengcyber is making some pretty quick progress, so I'd expect it to go in and be released fairly soon. There's a [Zulip channel](https://rushstack.zulipchat.com/#narrow/stream/296580-rush-hour/topic/.F0.9F.92.AC.20Breakout.20Topic.3A.20Plugins.20for.20Rush) about this feature if you're interested.

PR #2966, opened 2021-10-13
[rush] Streamline rushx startup banner

### LINKED ISSUE #2815: [rush] Add --verbose flag to rushx cut down on current line-noise
## Summary

`rushx` currently outputs a good amount of text when you run anything with it. When trying to run CLI projects this can be a bit distracting. `rushx` should have a `--verbose` flag behind which these current lines of output would be hidden.

Example,

```
rushx start hosts disable bar
Found configuration in /home/me/ext/dnsbot2/rush.json


Rush Multi-Project Build Tool 5.47.0 - https://rushjs.io
Node.js version is 14.17.3 (LTS)

Found configuration in /home/me/ext/dnsbot2/rush.json

Executing: "node --no-warnings dist/index.js hosts disable bar"

Updating all records...
```
Only the last line is output from my actual project.

## Repro steps

  **Expected result:** 

  **Actual result:** 

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.48.0 |
| `rushVersion` from rush.json? | 5.47.0 |
| Operating system? | linux |
| Node.js version (`node -v`)? | 14.17.3 |

elliot-nelson:
## Summary

 - Streamline the startup banner and command line information displayed at startup of `rushx`, to make it more inline with other package managers.
 - Add optional `-q/--quiet` argument to suppress the banner completely.
 - Fixes #2815

## Details

 - Simplified version of my initial PR https://github.com/microsoft/rushstack/pull/2864
 - This PR does not affect `rush` behavior, only `rushx`

## How it was tested

 - Using local monorepo, confirmed expected behavior:
   - `rushx foo` displays streamlined banner
   - `rushx --quiet foo` does not display banner info
   - `rush` behavior is not affected by this PR

### DIFF
--- a/apps/rush-lib/src/cli/test/RushXCommandLine.test.ts
+++ b/apps/rush-lib/src/cli/test/RushXCommandLine.test.ts
@@ -12,6 +12,7 @@ import { RushXCommandLine } from '../RushXCommandLine';
 
 describe('RushXCommandLine', () => {
   let $argv: string[];
+  let $versions: NodeJS.ProcessVersions;
   let executeLifecycleCommandMock: jest.SpyInstance | undefined;
   let logMock: jest.SpyInstance | undefined;
   let rushConfiguration: RushConfiguration | undefined;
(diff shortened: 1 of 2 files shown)


elliot-nelson:
@octogonz I was doing more digging on npm/pnpm and I think probably "-q/--quiet" (conceptually, _reducing output_) is probably way more accurate than "-s/--silent".  NPM itself actually has both along with a pile of other verbosity settings, and even "silent" is not ACTUALLY silent for most commands.

 - Introduced streamline banner for rushx, very similar to Yarn's.
 - Added `-q/--quiet` flag to suppress rushx banner.
 - Updated help and added some basic unit tests (mostly just to get baseline snapshots of the usage screen and the new banner).

PR #2972, opened 2021-10-18
[heft-webpack5-plugin] Upgrade webpack 5 to latest

scamden:
Fixes a bug between webpack-sources and webpack-license-plugin (along with keeping the library up to date with latest fixes etc.)

Tested against current webpack build with no issues.

### DIFF
--- a/common/changes/@rushstack/heft-webpack5-plugin/master_2021-10-18-21-07.json
+++ b/common/changes/@rushstack/heft-webpack5-plugin/master_2021-10-18-21-07.json
@@ -2,8 +2,8 @@
   "changes": [
     {
       "packageName": "@rushstack/heft-webpack5-plugin",
-      "comment": "upgrade webpack 5 to latest",
-      "type": "patch"
+      "comment": "Update to latest version of Webpack 5 and change the dependency specifier to a carat (^) to automatically take the latest minor version.",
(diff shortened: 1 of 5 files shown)


iclanton:
In the past, we've generally avoided carat specifiers for external dependencies, but in these packages it may actually be the expected behavior.

@octogonz, thoughts?

iclanton:
>any reasonably complex webpack setup will require plugins from webpack's common tools or 3rd party plugins that rely on having webpack installed

Yeah that's true. What would your thoughts be on adding an additional parameter with the webpack package instance to [this function](https://github.com/microsoft/rushstack/blob/master/heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts#L23-L24)? So your `webpack.config.js` could look something like this:

```JS
module.exports = function(env, webpack) {
  return {
    // webpack config stuff
   plugins: [
    new webpack.DefinePlugin({
      myVariable: 123
    })
  ]
}
```

I know that deviates from the standard webpack config, but it would allow us to provide webpack to a consumer without forcing the consumer to take a dependency on Webpack. WRT the peerdep, I can't think of a way to express "consider this a peer dependency if it's provided by a dependency, otherwise consider it a normal dependency."

iclanton:

PR #2974, opened 2021-10-21
[heft] Add environment variables for common heft test parameters

elliot-nelson:
## Summary

Add `ts-command-line` environment variable support for a few `heft test` parameters.

Command Line Option | Environment Variable Added
--- | ---
~`--update-snapshots`~ | ~`HEFT_UPDATE_SNAPSHOTS=1`~
`--test-timeout-ms=15000` | `HEFT_TEST_TIMEOUT_MS=15000`
`--detect-open-handles` | `HEFT_DETECT_OPEN_HANDLES=1`
`--max-workers=1` | `HEFT_MAX_WORKERS=1`

## Details

All of these happen to be just pass-through options for Jest.  In fact, if Jest were more amenable to the idea of extending support for this it probably makes more sense there.

My use case for these in particular are "I want to temporarily run several projects at once with some options", for example:

```
# Update all the affected projects so I can review the snapshots
HEFT_UPDATE_SNAPSHOTS=1 rush build --to @acme/acme-lib

# Run in band which is _much_ faster if each of your projects has <100 unit tests
HEFT_MAX_WORKERS=1 rush build

# Look for open handles on multiple projects
HEFT_DETECT_OPEN_HANDLES=1 rush build --to . --verbose

# etc. etc.
```

There's some I specifically left out, for example, test name / test file matching doesn't seem that useful in this context.

Another way to attack this, in a specific monorepo, might be to start adding configurable parameters in `command-line.json` for rush build.  This isn't ideal for two reasons: (1) projects that _don't_ use heft will get the param and possibly barf; (2) you don't have the flexibility of an env var (for example, I can't tack HEFT_MAX_WORKERS=1 onto one job of a CI pipeline that runs on a low-core VM).

Because it uses `ts-command-line`, you'll get aggressive validation of the _value_ of your variables... i.e. `HEFT_UPDATE_SNAPSHOTS=true` will inform you that you should pass `0` or `1`.  Mistyping / misspelling a `HEFT_` env var does not net you a Rush-style error; that's something that could go in a separate PR if it seems useful.

## How it was tested

 - Updated snapshots
 - Tested each env var on our local monorepo

### DIFF
--- a/apps/heft/src/cli/actions/TestAction.ts
+++ b/apps/heft/src/cli/actions/TestAction.ts
@@ -107,7 +107,7 @@ export class TestAction extends BuildAction {
     this._testTimeout = this.defineIntegerParameter({
       parameterLongName: '--test-timeout-ms',
       argumentName: 'INTEGER',
-      environmentVariable: 'HEFT_TEST_TIMEOUT_MS',
+      environmentVariable: 'HEFT_JEST_TEST_TIMEOUT_MS',
       description:
         "Change the default timeout for tests; if a test doesn't complete within this many" +
         ' milliseconds, it will fail. Individual tests can override the default. If unspecified, ' +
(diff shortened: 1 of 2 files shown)


D4N14L:
Something you may want to consider before merging is validating in Heft that all `HEFT_` prefixed environment vars are defined by one plugin or another, similar to how the `RUSH_` prefix is validated by Rush. This would avoid issues with "why isn't <my env var> being recognized by <some external plugin>" questions.

PR #3005, opened 2021-11-03
[heft-esbuild-plugin] Adds initial implementation of heft-esbuild-plugin

benjamind:
## Summary

Added an initial implementation of an [esbuild](https://esbuild.github.io/) plugin for heft.

## Details

This plugin currently uses the `compile` `run` hook to execute an esbuild build command. I used the compile hook because esbuild can be used to strip and compile TS code to JS. In this use-case its best to have this happen at the compile hook since its before bundling (though often in this scenario you also use the `bundle` option too).

Config files look something like this:

```json
{
  "$schema": "../../../node_modules/@diwi/heft-esbuild-plugin/schemas/esbuild.schema.json",
  "builds": [
    {
      "entryPoints": [
        "src/main/index.ts",
        "src/worker/worker.ts",
        "src/lib/index.ts",
        "src/service-worker/sw.ts"
      ],
      "outdir": "dist",
      "outbase": "src",
      "bundle": true,
      "format": "esm",
      "target": [
        "es2020",
        "chrome90"
      ]
    }
  ]
}
```

The configuration supports multiple 'builds', which can each define their own esbuild settings and entrypoints. This is useful for targetting different targets or formats in library projects.

The `bundle` option does open the question of whether or not this should cause the build to only happen on the bundle hook though, and am open to exploring the impact of that change.

## How it was tested

Currently this has been manually tested through publishing on an internal npm repository and then having a rig project use the above config. I will likely follow up with another PR for some other plugins and can then link to a public repo for the rig project that I'm using right now.

Would be interested to hear what the proposed automated test strategy is for heft plugins as I'm not seeing many examples of how to do jest testing for these plugins right now?


benjamind:
Obviously more work to be done on this. I'd like to expose more of the esbuild configuration options, and I think we should explore running in both `compile` and `bundle` stages.  Interested to hear the teams feedback on this, we're not currently using it in any production projects, but I have a working example project that is using a rigged config project and a couple of other plugins that I'll submit separately (browsersync, workbox) to produce an example app.

common/config/rush/browser-approved-packages.json
```diff
@@ -2,6 +2,10 @@
 {
   "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/approved-packages.schema.json",
   "packages": [
+    {
+      "name": "esbuild",
+      "allowedCategories": [ "libraries" ]
+    },
```
iclanton, line 8: This should go in `nonbrowser-approved-packages.json`

heft-plugins/heft-esbuild-plugin/LICENSE
```diff
@@ -0,0 +1,24 @@
+@rushstack/heft-webpack4-plugin
```
iclanton, line 1:
```suggestion
@rushstack/heft-esbuild-plugin
```

common/changes/@rushstack/heft-esbuild-plugin/delarre-heft-esbuild-plugin_2021-11-03-21-06.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-esbuild-plugin",
+      "comment": "Adds initial implementation of heft-esbuild-plugin",
```
iclanton, line 5:
```suggestion
      "comment": "Initial implementation of heft-esbuild-plugin",
```

heft-plugins/heft-esbuild-plugin/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@rushstack/heft-esbuild-plugin",
+  "version": "0.0.1",
```
iclanton, line 3:
```suggestion
  "version": "0.0.0",
```

heft-plugins/heft-esbuild-plugin/schemas/esbuild.schema.json
```diff
@@ -0,0 +1,63 @@
...
+                  "type": "string"
+                }
+              }
+            ]
+          },
+          "format": {
+            "description": "The output format for generated code: https://esbuild.github.io/api/#format",
+            "type": "string"
+          }
+        }
```
iclanton, line 59: Should these accept enums?

heft-plugins/heft-esbuild-plugin/schemas/esbuild.schema.json
```diff
@@ -0,0 +1,63 @@
...
+    "extends": {
+      "description": "Optionally specifies another JSON config file that this file extends from.  This provides a way for standard settings to be shared across multiple projects.",
+      "type": "string"
+    },
+    "builds": {
+      "type": "array",
+      "description": "The array of builds that Esbuild should run",
+      "items": {
+        "type": "object",
+        "properties": {
```
iclanton, line 21: It seems like some of these should be required.

heft-plugins/heft-esbuild-plugin/schemas/esbuild.schema.json
```diff
@@ -0,0 +1,63 @@
...
+          "entryPoints": {
```
iclanton, line 22:
```suggestion
          "entrypointPaths": {
```

heft-plugins/heft-esbuild-plugin/schemas/esbuild.schema.json
```diff
@@ -0,0 +1,63 @@
...
+            "type": "array",
+            "description": "The array of entrypoints that Esbuild should compile",
+            "items": {
+              "type": "string"
+            }
+          },
+          "outdir": {
```
iclanton, line 29:
```suggestion
          "outputFolder": {
```

heft-plugins/heft-esbuild-plugin/src/EsbuildPlugin.ts
```diff
@@ -0,0 +1,145 @@
...
+  HeftConfiguration,
+  HeftSession,
+  ICompileSubstage,
+  IBuildStageContext,
+  IHeftPlugin,
+  ScopedLogger
+} from '@rushstack/heft';
+
+import { configurationFileLoader, IEsbuildConfigurationJson } from './EsbuildConfigLoader.js';
+import { BuildFailure, BuildResult } from 'esbuild';
```
iclanton, line 15:
```suggestion
import type { BuildFailure, BuildResult } from 'esbuild';
```

heft-plugins/heft-esbuild-plugin/src/EsbuildPlugin.ts
```diff
@@ -0,0 +1,145 @@
...
+import type {
+  HeftConfiguration,
+  HeftSession,
+  ICompileSubstage,
+  IBuildStageContext,
+  IHeftPlugin,
+  ScopedLogger
+} from '@rushstack/heft';
+
+import { configurationFileLoader, IEsbuildConfigurationJson } from './EsbuildConfigLoader.js';
```
iclanton, line 14:
```suggestion
import { configurationFileLoader, IEsbuildConfigurationJson } from './EsbuildConfigLoader';
```

heft-plugins/heft-esbuild-plugin/src/EsbuildPlugin.ts
```diff
@@ -0,0 +1,145 @@
...
+                  heftConfiguration.globalTerminal.writeErrorLine(
+                    `An error occurred in an afterRecompile hook: ${error}`
+                  );
+                });
+              }
+            })
+              .then(resolve)
+              .catch(reject);
+          });
+        });
```
iclanton, line 60: This is a little funky.

heft-plugins/heft-esbuild-plugin/src/EsbuildPlugin.ts
```diff
@@ -0,0 +1,145 @@
...
+  private _esbuildConfigurationFileCache: Map<string, IEsbuildConfigurationFileCacheEntry> = new Map<
+    string,
+    IEsbuildConfigurationFileCacheEntry
+  >();
+
+  public apply(heftSession: HeftSession, heftConfiguration: HeftConfiguration): void {
+    const logger: ScopedLogger = heftSession.requestScopedLogger('esbuild');
+    heftSession.hooks.build.tap(PLUGIN_NAME, (build: IBuildStageContext) => {
+      build.hooks.compile.tap(PLUGIN_NAME, (compile: ICompileSubstage) => {
+        compile.hooks.run.tapPromise(PLUGIN_NAME, async () => {
```
iclanton, line 40: Do we want this in compile or bundle?

heft-plugins/heft-esbuild-plugin/src/EsbuildPlugin.ts
```diff
@@ -0,0 +1,145 @@
...
+          terminal,
+          buildFolder,
+          heftConfiguration.rigConfig
+        )
+      };
+
+      this._esbuildConfigurationFileCache.set(buildFolder, esbuildConfigurationFileCacheEntry);
+    }
+
+    return esbuildConfigurationFileCacheEntry.configurationFile;
```
iclanton, line 86:
```suggestion
    const buildFolder: string = heftConfiguration.buildFolder;
    return await configurationFileLoader().tryLoadConfigurationFileForProjectAsync(
      terminal,
      buildFolder,
      heftConfiguration.rigConfig
    );
```

heft-plugins/heft-esbuild-plugin/src/EsbuildPlugin.ts
```diff
@@ -0,0 +1,145 @@
...
+            callback();
+          })
+          .catch((err) => {
+            logger.terminal.writeError('esbuild error', err);
+          });
+      }
+    );
+    return Promise.all(buildPromises).then(() => {
+      return;
+    });
```
iclanton, line 143:
```suggestion

    await Promise.all(buildPromises);
```

heft-plugins/heft-esbuild-plugin/src/EsbuildPlugin.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (!esbuildConfigurationJson) {
+      throw new Error('Unable to find an esbuild.json config');
+    }
+    const { builds } = esbuildConfigurationJson;
+    logger.terminal.writeLine(`Using Esbuild version ${esbuild.version}`);
+    if (watchMode) {
+      logger.terminal.writeLine('Watching files...');
+    }
+    const buildPromises: Promise<void>[] = builds.map(
+      ({ entryPoints, outdir, outbase, bundle, target, format }): Promise<void> => {
```
iclanton, line 106: Can you convert this to `async`/`await`-style?

rush.json
```diff
@@ -767,7 +767,12 @@
...
       "shouldPublish": true
     },
-
+    {
+      "packageName": "@rushstack/heft-esbuild-plugin",
+      "projectFolder": "heft-plugins/heft-esbuild-plugin",
+      "reviewCategory": "libraries",
+      "shouldPublish": true
+    },
     // "libraries" folder (alphabetical order)
```
iclanton, line 776:
```suggestion
    },
    
    // "libraries" folder (alphabetical order)
```

iclanton (changes requested):

heft-plugins/heft-esbuild-plugin/src/EsbuildPlugin.ts
benjamind: This is a fundamental question. ESBuild can be used both as a compiler (for stripping Typescript to produce JS) and/or as a bundler.

Right now our primary use-case has been 'both', i.e. we compile the TS to JS with ESbuild and also bundle the output in one step. I attached Esbuild to the compile task in this case to allow us to do that compilation before running the Workbox plugin which is attached to the afterCompile hook. We basically then do not have a bundle task essentially since the output of our compile step is effectively bundled already.

I understand this is probably a little bit of a deviation from the traditional flow though so am open to other ideas of how to orchestrate this.

heft-plugins/heft-esbuild-plugin/schemas/esbuild.schema.json
benjamind: I'd prefer to keep these parameters matching the esbuild parameters so the docs are easily understood I think? e.g. https://esbuild.github.io/api/#outdir

heft-plugins/heft-esbuild-plugin/src/EsbuildPlugin.ts
benjamind: Yeah agreed, is there a cleaner solution here?

iclanton:

FullStackSteve:
@benjamind @iclanton we would love to use esbuild in heft - anything we can do to speed this up?

benjamind:
> @benjamind @iclanton we would love to use esbuild in heft - anything we can do to speed this up?

@FullStackSteve I'm unlikely to complete this plugin at this point, feel free to take over the PR though!

PR #3007, opened 2021-11-03
[heft-workbox-plugin] Adds initial implementation of heft-workbox-plugin

benjamind:
## Summary

Added an initial implementation of a [Workbox](https://developers.google.com/web/tools/workbox) plugin for heft.

## Details

Workbox is a suite of tools and libraries to help assist in producing serviceworkers for PWAs. It has a wide range of potential configuration, but this plugin aims to only achieve one thing: use the [workbox-build](https://developers.google.com/web/tools/workbox/modules/workbox-build) module to inject a manifest into a pre-existing serviceworker file. This is similar in many respects to webpacks InjectManifest plugin, but separated out from webpacks bundle environment.

Configuration looks something like this:

```json
{
  "$schema": "../../../node_modules/@diwi/heft-workbox-plugin/schemas/workbox.schema.json",
  "swSrc": "./dist/service-worker/sw.js",
  "swDest": "./dist/sw.js",
  "globDirectory": "./dist/",
  "globPatterns": [
    "**/*.js",
    "!service-worker/sw.js",
    "**/*.html"
  ],
  "globIgnores": [
    "network-only.html"
  ]
}
```

The config options here map to the workbox [inject manifest configurations](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.injectManifest).

The plugin is setup to run as a `compile` task in the `afterCompile` and `afterRecompile` hooks. In development mode it will only copy the existing serviceworker file into destination location, but in production mode it will use the `injectManifest` api from workbox to gather all the files according to config, hash them and inject them into the sw precache manifest. This has been done because the manifest creation can take some time and slows down development workflox. In development we often run without serviceworker precaching turned on anyway, and so it can be best to just skip this step for development builds.

An example serviceworker file that works with this can look something like this:

```ts
import { clientsClaim, setCacheNameDetails } from "workbox-core";
import { precacheAndRoute, matchPrecache } from "workbox-precaching";
import { setCatchHandler, registerRoute } from "workbox-routing";
import { NetworkOnly } from "workbox-strategies";

declare let self: ServiceWorkerGlobalScope;

setCacheNameDetails({
  prefix: "heft-example",
});

clientsClaim();
self.skipWaiting();

// Register routes and strategies to handle them
registerRoute("/network-only.html", new NetworkOnly());

// Use with precache injection
precacheAndRoute(self.__WB_MANIFEST || []);

// Catch routing errors, like if the user is offline
setCatchHandler(async ({ request }) => {
  // Return the precached offline page if a document is being requested
  if (request.destination === "document") {
    return matchPrecache("/offline.html") as Promise<Response>;
  }

  return Response.error();
});
```

## How it was tested

Currently this has been manually tested through publishing on an internal npm repository and then having a rig project use the above config. I will likely follow up with another PR for some other plugins and can then link to a public repo for the rig project that I'm using right now.


heft-plugins/heft-workbox-plugin/src/WorkboxConfigLoader.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Returns the loader for the `config/workbox.json` config file.
+ */
+export function configurationFileLoader(): ConfigurationFile<IWorkboxConfigurationJson> {
+  if (!workboxConfigurationFileLoader) {
+    const schemaPath: string = path.resolve(__dirname, '..', 'schemas', 'workbox.schema.json');
+    workboxConfigurationFileLoader = new ConfigurationFile<IWorkboxConfigurationJson>({
+      projectRelativeFilePath: 'config/workbox.json',
+      jsonSchemaPath: schemaPath
+    } as IConfigurationFileOptions<IWorkboxConfigurationJson>);
```
iclanton, line 24: Why is this cast necessary?

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
```diff
@@ -0,0 +1,121 @@
...
+  HeftSession,
+  ICompileSubstage,
+  IBuildStageContext,
+  IHeftPlugin,
+  ScopedLogger
+} from '@rushstack/heft';
+
+import { configurationFileLoader, IWorkboxConfigurationJson } from './WorkboxConfigLoader.js';
+
+import { copyFile } from 'fs';
```
iclanton, line 16: Use `FileSystem` from `node-core-library`.

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * @internal
+ */
+export class WorkboxPlugin implements IHeftPlugin {
+  public readonly pluginName: string = PLUGIN_NAME;
+
+  private _workboxConfigurationFileCache: Map<string, IWorkboxConfigurationFileCacheEntry> = new Map<
+    string,
+    IWorkboxConfigurationFileCacheEntry
+  >();
```
iclanton, line 35: `heft-config-file` has an internal cache, so you shouldn't need this.

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
```diff
@@ -0,0 +1,121 @@
...
+          buildFolder,
+          heftConfiguration.rigConfig
+        )
+      };
+
+      this._workboxConfigurationFileCache.set(buildFolder, workboxConfigurationFileCacheEntry);
+    }
+
+    return workboxConfigurationFileCacheEntry.configurationFile;
+  }
```
iclanton, line 75:
```suggestion
  private async _ensureConfigFileLoadedAsync(
    terminal: ITerminal,
    heftConfiguration: HeftConfiguration
  ): Promise<IWorkboxConfigurationJson | undefined> {
    const buildFolder: string = heftConfiguration.buildFolder;
    return await configurationFileLoader().tryLoadConfigurationFileForProjectAsync(
      terminal,
      buildFolder,
      heftConfiguration.rigConfig
    );
  }
```

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
```diff
@@ -0,0 +1,121 @@
...
+  private async _runWorkboxAsync(
+    heftSession: HeftSession,
+    heftConfiguration: HeftConfiguration,
+    isProduction: boolean
+  ): Promise<void> {
+    const workboxConfigurationJson: IWorkboxConfigurationJson | undefined =
+      await this._ensureConfigFileLoadedAsync(this._logger.terminal, heftConfiguration);
+    if (!workboxConfigurationJson) {
+      throw new Error('Unable to find an workbox.json config');
+    }
```
iclanton, line 86:
```suggestion
    if (!workboxConfigurationJson) {
      this._logger.emitError(new Error('Unable to find an workbox.json config'));
      return;
    }
```

Should this be a no-op instead?

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
```diff
@@ -0,0 +1,121 @@
...
+      const copyPromise: Promise<void> = new Promise((resolve, reject): void => {
+        copyFile(swSrc, swDest, (err) => {
+          if (err) {
+            reject(err);
+          } else {
+            resolve();
+          }
+        });
+      });
+      return copyPromise;
```
iclanton, line 102:
```suggestion
      await FileSystem.copyFileAsync({ source: swSrc, target: swDest });
```

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
```diff
@@ -0,0 +1,121 @@
...
+        globDirectory,
+        globPatterns,
+        globIgnores,
+        modifyURLPrefix
+      })
+      .then(({ count, size }) => {
+        this._logger.terminal.writeLine(
+          `Generated ${swDest}, which will precache ${count} files, totaling ${size} bytes.`
+        );
+      });
```
iclanton, line 119:
```suggestion
    const { count, size } = await workboxBuild
      .injectManifest({
        swSrc,
        swDest,
        globDirectory,
        globPatterns,
        globIgnores,
        modifyURLPrefix
      })

      this._logger.terminal.writeLine(
        `Generated ${swDest}, which will precache ${count} files, totaling ${size} bytes.`
      );
```

heft-plugins/heft-workbox-plugin/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./node_modules/@rushstack/heft-node-rig/profiles/default/tsconfig-base.json",
+
+  "compilerOptions": {
+    "types": ["node"],
+    "lib": ["WebWorker"]
```
iclanton, line 6: What uses WebWorker?

rush.json
```diff
@@ -767,6 +767,12 @@
       "reviewCategory": "libraries",
       "shouldPublish": true
     },
+    {
+      "packageName": "@rushstack/heft-workbox-plugin",
```
iclanton, line 771: It'd be useful to have a test project.

iclanton (changes requested):

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
benjamind: Ah good question, I wasn't sure if configuration as a plugin in the heft.json should then cause an error if the plugin can't find its config file, or if it should just no-op. I kinda lean more towards the format since it can be confusing when you think you've set something up but haven't? But I guess if we're logging the error too then its clear?

heft-plugins/heft-workbox-plugin/tsconfig.json
benjamind: ServiceWorker code requires the WebWorker lib due to the way the TS declarations are done.

rush.json
benjamind: Agreed. Should this be several different test projects for different classes of application? I have a small test project I was using to test out the workbox, esbuild and browsersync plugins together. I can bring that in as another PR if the team wants?

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
iclanton: Either way works. Just depends on how the plugin is expected to be consumed.

rush.json
iclanton: Sounds useful to me.

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
```diff
@@ -0,0 +1,112 @@
...
+import type {
+  HeftConfiguration,
+  HeftSession,
+  ICompileSubstage,
+  IBuildStageContext,
+  IHeftPlugin,
+  ScopedLogger
+} from '@rushstack/heft';
+
+import { configurationFileLoader, IWorkboxConfigurationJson } from './WorkboxConfigLoader.js';
```
iclanton, line 14:
```suggestion
import { configurationFileLoader, IWorkboxConfigurationJson } from './WorkboxConfigLoader';
```

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
benjamind: I was following the example in the Typescript plugin: https://github.com/microsoft/rushstack/blob/master/apps/heft/src/plugins/TypeScriptPlugin/TypeScriptPlugin.ts#L159

Is this definitely not needed?

heft-plugins/heft-workbox-plugin/schemas/workbox.schema.json
```diff
@@ -0,0 +1,50 @@
...
+        "type": "string"
+      }
+    },
+    "globIgnores": {
+      "description": "A set of patterns matching files to always exclude when generating the precache manifest",
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    }
```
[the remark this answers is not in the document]
benjamind: Yes good call, have switched to an array of strings with the same pattern match.

heft-plugins/heft-workbox-plugin/src/WorkboxConfigLoader.ts
benjamind: It wasn't good catch, removed.

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
benjamind: Have merged this in 669a988

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
benjamind: Merged in 669a988

heft-plugins/heft-workbox-plugin/src/WorkboxPlugin.ts
benjamind: Went for your way, and error in the console is sufficient.

iclanton:

PR #3013, opened 2021-11-06
[rush-lib] throw package install when `git config core.hooksPath` is set

### LINKED ISSUE #3004: [rush] Rush doesn't acknowledge `git config core.hooksPath`
I've been migrating a monorepo from lerna + yarn workspaces, and it had git hooks before, except they were implemented by `husky`.  And `husky` change the hooks installation path to it's own.

And `rush` doesn't acknowledge the setting in any way, so if you're unaware that `hooksPath` is a thing, debugging is really unfun. It would be great if `rush` could check the setting, and either respect it and install the hooks where it says, or warn to run `git config --unset core.hooksPath`

Faithfinder:
## Summary

Fixes #3004 

## Details

The method to retrieve the hooks path now uses `git rev-parse` to ask git to resolve the hooks path by itself. There's a fallback to the old default route in case we're on an older version of git where it errors or something.
I also print out the path where the hooks are installed during the installation process to reduce surprise.

This should solve the problem of hooks suddenly not working, and if the user is surprised to see a `.husky` folder in their `rush` directory at least they can see that rush installs it by design. There's nothing to indicate where the path comes from, but I don't think it's rush's job to educate users on the details of git configuration.

@D4N14L proposed an alternative to point `core.hooksPath` to `{rush-repo}/common/git-hooks`, however I don't like when tools hijack configuration of other tools. I do skip installation if source and destination are the same.

I have a couple of questions, I'll add comments to the source.

## How it was tested

Locally with various kinds of potential paths. However, I did test on only one git version, so this doesn't cover something like "Your git is from the 90s". The config variable is from 2016, so hopefully it shouldn't be a real issue? Couldn't find what year the command I'm using to resolve it is from (`git rev-parse --git-path hook`).

Also, didn't find any real examples for testing git, so didn't write any new tests.


apps/rush-lib/src/logic/Git.ts
```diff
@@ -29,6 +29,7 @@ export class Git {
   private _gitInfo: gitInfo.GitRepoInfo | undefined;
 
   private _gitEmailResult: IResultOrError<string> | undefined = undefined;
+  private _gitHooksPath: IResultOrError<string> | undefined = undefined;
```
Faithfinder, line 32: I've monkeyed the data structure here from `_gitEmailResult` above, but I'm not actually doing anything with any errors. I wonder if I should use something simpler, or this is fine?

apps/rush-lib/src/logic/Git.ts
```diff
@@ -133,11 +134,15 @@ export class Git {
...
   public getHooksFolder(): string | undefined {
     const repoInfo: gitInfo.GitRepoInfo | undefined = this.getGitInfo();
-    if (repoInfo && repoInfo.worktreeGitDir) {
-      return path.join(repoInfo.worktreeGitDir, 'hooks');
+    const hooksResult: IResultOrError<string> = this._tryGetGitHooksPath();
+    const hooksPath: string = hooksResult.result ?? '.git/hooks';
+
+    if (repoInfo?.root) {
+      return path.resolve(repoInfo.root, hooksPath);
     }
```
Faithfinder, line 146: AFAIK git can be complicated and I don't know anything but the most basic usage, so I'm not sure this covers 100% of possibilities? Should be no worse than how it was before, I guess

apps/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -315,13 +315,21 @@ export abstract class BaseInstallManager {
...
     const git: Git = new Git(this.rushConfiguration);
     const hookDestination: string | undefined = git.getHooksFolder();
+    const hooksSourceSameAsDestination: boolean = hookSource === hookDestination;
 
-    if (FileSystem.exists(hookSource) && hookDestination) {
+    if (hooksSourceSameAsDestination) {
+      console.log(
+        `git config core.hooksPath points to "common/git-hooks" folder, no need for separate git hooks installation`
+      );
+    }
```
Faithfinder, line 324: There's some cleanup going on in the copy script below, so I'm not sure whether skipping it is actually a good idea. On the other hand, making copies write over their source is probably bad as well.

apps/rush-lib/src/logic/Git.ts
```diff
@@ -133,11 +134,15 @@ export class Git {
   /**
    * Get the folder where Git hooks should go for the current working tree.
    * Returns undefined if the current path is not under a Git working tree.
+   * Respects `git config core.hooksPath`.
    */
   public getHooksFolder(): string | undefined {
     const repoInfo: gitInfo.GitRepoInfo | undefined = this.getGitInfo();
-    if (repoInfo && repoInfo.worktreeGitDir) {
-      return path.join(repoInfo.worktreeGitDir, 'hooks');
+    const hooksResult: IResultOrError<string> = this._tryGetGitHooksPath();
```
D4N14L, line 141: You could probably handle the error in a method similar to how we handle the error in `getGitEmail()` here: https://github.com/microsoft/rushstack/blob/0f2d6a586a16a6970f7d6614d652b3bdd5b086bf/apps/rush-lib/src/logic/Git.ts#L100

apps/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -315,13 +315,21 @@ export abstract class BaseInstallManager {
     const hookSource: string = path.join(this._rushConfiguration.commonFolder, 'git-hooks');
     const git: Git = new Git(this.rushConfiguration);
     const hookDestination: string | undefined = git.getHooksFolder();
+    const hooksSourceSameAsDestination: boolean = hookSource === hookDestination;
 
-    if (FileSystem.exists(hookSource) && hookDestination) {
+    if (hooksSourceSameAsDestination) {
+      console.log(
+        `git config core.hooksPath points to "common/git-hooks" folder, no need for separate git hooks installation`
```
D4N14L, line 322:
```suggestion
        `git config core.hooksPath points to "common/git-hooks" folder, skipping git hooks installation`
```

apps/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -315,13 +315,21 @@ export abstract class BaseInstallManager {
...
+      );
+    }
+
+    if (FileSystem.exists(hookSource) && hookDestination && !hooksSourceSameAsDestination) {
       const allHookFilenames: string[] = FileSystem.readFolder(hookSource);
       // Ignore the ".sample" file(s) in this folder.
       const hookFilenames: string[] = allHookFilenames.filter((x) => !/\.sample$/.test(x));
       if (hookFilenames.length > 0) {
         console.log(os.EOL + colors.bold('Found files in the "common/git-hooks" folder.'));
+        console.log(colors.bold('Hooks will be installed into: ') + hookDestination);
```
D4N14L, line 332: May want to make sure `hookDestination` is bolded as well.

apps/rush-lib/src/logic/Git.ts
D4N14L: You could probably handle the error in a method similar to how we handle the error in `getGitEmail()` here: https://github.com/microsoft/rushstack/blob/0f2d6a586a16a6970f7d6614d652b3bdd5b086bf/apps/rush-lib/src/logic/Git.ts#L100

D4N14L (changes requested):
Just a couple changes, otherwise looks good!

apps/rush-lib/src/logic/base/BaseInstallManager.ts
Faithfinder: Oh this was intentional, to make a clearer separation between the static message and the dynamic info. Matter of taste I suppose. I can change it back?

D4N14L: @octogonz thoughts? I don't mind either way (a differentiation would be helpful, I think?)

apps/rush-lib/src/logic/Git.ts
D4N14L: Apparently there's quite a bit more involved in finding the `worktreeGitDir` vs. just finding the `root`. See: https://github.com/rwjblue/git-repo-info/blob/d3ab418ef8b392eabbe911a37871708b15201b70/index.js#L13

So it seems it would be safer to default to `worktreeGitDir` instead of doing a `path.join(...)` with the `.git` folder being expected at the root.

D4N14L (changes requested):
One more change... you brought up a good point about use of the `worktreeGitDir` value.

apps/rush-lib/src/logic/Git.ts
Faithfinder: So, I need some help here. 

In your average repo (like the `rushstack`) `git rev-parse` returns me the hook directory relative to the root, not the `.git`. Since `rev-parse` is doing it relative to the directory where it's run (i.e. if I run it manually 5 folders deep in the repo it returns `../../../../../.git/hooks` for unset param).

I assume this just means that `rush` always runs `git` commands at the repo root. If there is a variable for "Directory where we run git" I'd rather use that, but I haven't found anything like it.

This would mean that It's safe to leave this as is even in the worktree case. However, it would be great if someone with the actual setup tested it. If there's no one and it has to be me, it'll take me some time to understand what those things are and how to set them up.

Another thing that can be done is use [-path-format=absolute](https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---path-formatabsoluterelative) on `rev-parse`, and not guess anything, BUT it's a feature from Q1 2021, and wasn't available on my machine, so I decided that it's too early to depend on it.

Relevant reading:
The git docs on the [param itself](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corehooksPath) are quite laconic.
The docs on [rev-parse](https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse.txt---git-pathltpathgt) are barely coherent if you ask me (or maybe just require deeper git knowledge).

___
P.S. - It's `path.resolve`, not `.join` - `hooksPath` can be absolute instead, and `resolve` will respect that unlike `join`

D4N14L: @Faithfinder Agree that depending on `-path-format=absolute` isn't the right solution here.

As for the first part, I agree with your logic here. I'm not too personally familiar with the reasoning behind why we used `worktreeGitDir`, but looking into the docs, it may not even be the correct value to use here. The git docs you provided state that the `core.hooksPath` should be postfixed to `GIT_DIR` to get the absolute hooks path, which implies it should be located in the main worktree, so really we shouldn't need to worry about the worktree at all.

We run the git commands in this class with a working dir of `this._rushConfiguration.rushJsonFolder`. So, maybe we should have 2 cases here where:
- `path.resolve(this._rushConfiguration.rushJsonFolder, hooksPath)` is used in the case where we have the `hooksPath`, since it's returned relative to the working directory
- `path.resolve(repoInfo.root, '.git', 'hooks')` is used in the case where we don't

It's extremely rare to see a repo with a Rush configuration somewhere other than the root of the repo (and it would likely cause other issues... but that's beside the point), but it is possible. This feels like the _most correct_ way of doing it as far as I can tell... Thoughts?

Faithfinder: Yeah, sounds reasonable. I've made the changes and pushed them

D4N14L (approved):

octogonz:
@D4N14L and I chatted about this PR today.  In effect it will cause Rush's Git hooks to get installed under the `.husky/` folder, by honoring a stray `<repo>/.git/config` setting left behind by the Husky tool. That seems odd, and has a risk of [nondeterministic build errors](https://github.com/microsoft/rushstack/issues/3004#issuecomment-964524613).

We came around to the position that:

- If you have a branch that uses Husky, that's okay.
- If you have a branch that uses Rush's **common/git-hooks** feature, great, that's recommended. 
- But no branch should be using BOTH Husky and **common/git-hooks** together.
- When you run `rush install` on a branch with **common/git-hooks**, Rush owns the hooks installation, not Husky.  To avoid confusing breaks like what @Faithfinder  reported in https://github.com/microsoft/rushstack/issues/3004, Rush can automatically perform `git config --unset core.hooksPath` to clean up the old state.  (For smooth sailing on CI machines, I don't even think we need to ask the user to do it.)
- However if someone has a compelling case for using `core.hooksPath` with Rush, we could expose that as an explicit feature (e.g. a setting in **rush.json**, a `RUSH_GIT_HOOKS_PATH` environment variable, etc)

I apologize for proposing to change course after this PR was already underway.

octogonz:
BTW the `core.hooksPath` setting seems to be a potential security concern: 

An important aspect of locating hooks under the `.git/` folder was to ensure that unknown hook scripts won't be executed blindly.  Instead, the user needs to manually consent by invoking a command to copy them into that folder.  Whereas if you reconfigure `core.hooksPath` to be a Git-tracked folder such as `.husky/post-checkout`, then you are inadvertently consenting to run arbitrary code from unknown branches.  That seems like a significant weakening of Git's security design.

Faithfinder:
It's reasonable enough. I guess the main problem is that `hooks` are inherently a local machine state that doesn't get synced by git, even though both `rush` and `husky` and I imagine the developer community in general mostly want to use them to enforce some standards in repos, meaning that syncing is needed.

I'm not entirely sure that `rush` should unset the variable automatically though. I'm afraid to surprise a user the other way - "Why are my git hooks path unset, I didn't do it". Even if I add a BOLD RED WARNING when it happens, it would probably be drowned by the wall of text that `rush update` produces. I would instead report an error and stop the update process

I don't mind at all not supporting the case when `core.hooksPath` is set until someone shows up who needs it. That's certainly not us, my goal is just stopping people from being surprised by hooks not running. I don't think it's reasonable to expect your average developer to know that there's such a setting in git. I would say `husky` is pretty naughty for messing with my local config.

>BTW the core.hooksPath setting seems to be a potential security concern

I mean, you can even point it outside of the repo at a network share where your admin team drops the scripts they want you to run, how's that for a setup?

octogonz:
> > BTW the core.hooksPath setting seems to be a potential security concern
> 
> I mean, you can even point it outside of the repo at a network share where your admin team drops the scripts they want you to run, how's that for a setup?

From the Git docs, that sounds like the intended usage, and it's probably not a security concern because we implicitly trust the admin who set up the machine.  Whereas the exploit involves scripts from an untrusted source, something like:

1. I run `husky install` which assigns `core.hooksPath` in my `repo/.git/config` file
2. Our `main` branch has a `.husky/pre-commit` file, and everything seems fine
3. A bad person creates a Git branch in their own private fork, where `.husky/post-checkout` is a malicious script.  They push their branch and share it, e.g as a PR.
4. I use `git checkout` to inspect their branch.  Since it originates from an untrusted party, normally I might think twice before executing the code there.  But `git checkout` normally is safe, thanks to Git's security protection.  However ***before I even had a chance to look at the code or decide whether to build the code,*** the `.husky/post-checkout` script now gets executed by Git.

elliot-nelson:
I guess my take is Rush should do the "right thing" even if other tools aren't, and resetting your `core.hooksPath` on a `rush install` doesn't seem like the "right thing" to me.

FWIW, my preference would be that if I run `rush install`, and Rush is about to install some hooks from "common/git-hooks", and it detects a non-standard `core.hooksPath`, it immediately stops with a new error message.  I can then go and change a new boolean (maybe `allowNonStandardHooksPath` in `rush.json`) if I really want to keep that setting, or I can delete the `core.hooksPath` config.

This approach assumes that most developers probably don't want Rush installing hooks into some unknown location, and that they probably don't even know about that unknown location; but if they do want it, they can choose to use it.

This behavior doesn't "automatically heal" a busted CI environment, but I don't know think that's a requirement -- if you are a team that is converting from Husky to Rush and you know that old & new branches are going to run on self-hosted machines, you can always add `git config --unset core.hooksPath` to your own CI pipeline in the new branch.  (I don't think Rush wants to get into the business of fixing that itself; this type of situation only occurs when you're running semi-trusted branches from your own team on self-hosted machines, and there's _so many_ things an unthinking developer can do to alter your CI state in ways worse than Husky -- errant regedit calls, installing software, creating config files in shared paths, etc.).

octogonz:
> I'm not entirely sure that `rush` should unset the variable automatically though. I'm afraid to surprise a user the other way - "Why are my git hooks path unset, I didn't do it". Even if I add a BOLD RED WARNING when it happens, it would probably be drowned by the wall of text that `rush update` produces. I would instead report an error and stop the update process

@Faithfinder this makes sense.  With Elliot's input, here's a revised proposal:

1. **No validation should be performed unless Rush's Git hooks are enabled.**  This ensures that it's possible to use alternative Git hook installers in a Rush monorepo.  Looking [at the code](https://github.com/microsoft/rushstack/blob/aef95129bad762edabb81f8ef4a33532c398b6a5/apps/rush-lib/src/logic/base/BaseInstallManager.ts#L322), Rush's feature is considered to be "enabled" if-and-only-if the `common/git-hooks` folder exists and contains at least one file that is not a `rush init` sample file.  (BTW note that when the feature is enabled, Rush will always delete any non-Rush scripts from `.git/config`. If you think about it, that's essential for correct operation.)
2. **Rush should fail hard if core.hooksPath is incorrect.** If someone explicitly sets `core.hooksPath` to the default value of `$GIT_DIR/hooks` (or `.git/hooks`), we should probably allow that.  For any other value, Rush should print an error in red instructing the user to run `git config --unset core.hooksPath` themselves.  The `rush install` operation is aborted.
3. **But a person can temporarily ignore the error.**  Rush has a `--bypass-policy` option whose intent is *"Hey, I'm just fiddling around, don't make me fix that problem, let me proceed anyway."*  (This option is not meant to be used in a production situation such as a CI pipeline.)  So as an escape hatch, if the person does `rush install --bypass-policy`, then the error should be downgraded to a yellow STDERR warning, so that `rush install` can proceed.  Hooks will NOT be installed into the `.husky` folder, however.
4. **No hooksPath support for now.** Chatting with Elliot, we decided NOT to implement the `allowNonStandardHooksPath` or `RUSH_ALLOW_NON_STANDARD_HOOKS_PATH` ideas.  Nobody has actually asked for that yet, and moreover it's not a practice that we want to recommend.  If you really need `core.hooksPath`, most likely you should disable Rush's Git hooks feature.


For #2, I'm suggesting for the error message to look like this:

```
Rush cannot install the "common/git-hooks" scripts because your Git configuration 
specifies "hooksPath=.husky".  To repair your Git configuration, please run:

  git config --unset core.hooksPath

(Or, to temporarily ignore this problem, invoke Rush with the "--bypass-policy" option.)
```

Faithfinder:
I agree with most points except maybe the error message. The git configuration isn't broken, so it shouldn't be repaired, it's just that rush doesn't support the custom path, so:

```
Cannot install the "common/git-hooks". Rush doesn't support custom git hooks path and your Git configuration 
specifies "core.hooksPath=.husky".  You can unset it by running:

  git config --unset core.hooksPath

(Or, to temporarily ignore this problem, invoke Rush with the "--bypass-policy" option.)
```

I also have an urge to write `Contact you repository maintainer to check for other options`, but that gives me a Windows XP vibe

image

octogonz:
> ```
> Cannot install the "common/git-hooks". Rush doesn't support custom git hooks path and your Git configuration 
> specifies "core.hooksPath=.husky".  You can unset it by running:
> 
>   git config --unset core.hooksPath
> 
> (Or, to temporarily ignore this problem, invoke Rush with the "--bypass-policy" option.)
> ```

The phrasing *"Rush doesn't support custom git hooks"* for me has a connotation that Rush is lacking a feature.  In fact we had a ready-to-go PR implementing support, 😉 but after careful thought we found several reasons to recommend against usage of `core.hooksPath` in a large production monorepo.  You have a good point however that the word "repair" perhaps too strongly implies that Husky has damaged the state for other tools.  Maybe the wording could be:

```
Rush cannot install the "common/git-hooks" scripts because your Git configuration 
specifies "hooksPath=.husky". You can remove that setting by running:
```

apps/rush-lib/src/logic/Git.ts
dmichon-msft: I recently made changes to a family of git commands to, in fact, always run them from the git root, specifically to support having rush.json in a subfolder. I use this function here to obtain said root for comparison purposes: https://github.com/microsoft/rushstack/blob/fa7f808d23e1f77047fffed56ef9390168c8111f/libraries/package-deps-hash/src/getRepoState.ts#L138

Faithfinder:
Alright, take 2!

apps/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -473,6 +438,74 @@ export abstract class BaseInstallManager {
...
+                ' ',
+                `Rush cannot install the "common/git-hooks" scripts because your Git configuration specifies "core.hooksPath=${git.getConfigHooksPath()}".`,
+                'You can remove the setting by running:',
+                ' ',
+                '    git config --unset core.hooksPath',
+                ' '
+              ].join(os.EOL)
+            )
+          );
+          if (!this.options.bypassPolicy) {
```
octogonz, line 470: If the person chooses "--bypass-policy", we had specified that the Rush's hooks should NOT be installed into the .husky folder.  Whereas this code falls through.

I can fix that.

octogonz: Fixed ✔

octogonz (approved):
Test cases

- in a clean repo, `rush install` correctly installs hooks and deletes any irrelevant hook files
- when `core.hooksPath` is corrupted by Husky, `rush install` fails and reports an error that is nicely formatted
- `rush install --bypass-policy` instead prints a yellow warning to STDERR and proceeds without installing hooks
- after copy+pasting+invoking the recommended command (`--unset`), `rush install` now succeeds and correctly installs hooks

PR #3019, opened 2021-11-11
[set-webpack-public-path-plugin] Some typings cleanup.

iclanton:
This is just some cleanup, especially removing the "webpack4" and "webpack5" from the typings.

### DIFF
--- a/webpack/set-webpack-public-path-plugin/src/SetPublicPathPlugin.ts
+++ b/webpack/set-webpack-public-path-plugin/src/SetPublicPathPlugin.ts
@@ -2,8 +2,7 @@
 // See LICENSE in the project root for license information.
 
 import { EOL } from 'os';
-import { cloneDeep, escapeRegExp } from 'lodash';
-import type * as Webpack4 from 'webpack4';
+import type * as Webpack from 'webpack';
 import type * as Webpack5 from 'webpack5';
 import type * as Tapable from 'tapable';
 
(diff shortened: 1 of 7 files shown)


webpack/set-webpack-public-path-plugin/src/SetPublicPathPlugin.ts
```diff
@@ -130,13 +125,17 @@ const ASSET_NAME_TOKEN: string = '-ASSET-NAME-c0ef4f86-b570-44d3-b210-4428c5b782
 
 const ASSET_NAME_TOKEN_REGEX: RegExp = new RegExp(ASSET_NAME_TOKEN);
 
+function escapeRegExp(str: string): string {
+  return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
D4N14L, line 129: Why not the version from the actual package? `/[\\^$.*+?()[\]{}|]/g`

D4N14L (approved):

webpack/set-webpack-public-path-plugin/src/SetPublicPathPlugin.ts
D4N14L: Honestly I think it's the same, just in a different order. Feel free to ignore.

iclanton: This is what GitHub Copilot suggested. They're equivalent, so I'll just leave this as-is.

PR #3023, opened 2021-11-16
[localization-plugin] Accept ".resx.json" as a valid strings file extension

dmichon-msft:
## Summary
Adds ".resx.json" as a supported strings file extension to simplify transitioning from existing ".resx" files to a JSON-based format without having to alter TypeScript files that refer to the strings.

## Details
Uses the existing designed ".loc.json" file structure for the ".resx.json" files.
Updates the loader injection to mark all strings files as side-effect free and to be treated as JSON.

## How it was tested
Updated the localization-plugin-test-03 to include a file with the .resx.json extension


webpack/localization-plugin/src/utilities/Constants.ts
```diff
@@ -23,8 +23,9 @@ export class Constants {
   public static STRING_PLACEHOLDER_PREFIX: string = '_LOCALIZED_STRING_f12dy0i7_n4bo_dqwj_39gf_sasqehjmihz9';
 
   public static RESX_REGEX: RegExp = /\.resx$/i;
+  public static RESX_JSON_REGEX: RegExp = /\.resx$/i;
   public static LOC_JSON_REGEX: RegExp = /\.loc\.json$/i;
```
iclanton, line 27:
```suggestion
```

These don't seem to be in use anymore.

PR #3024, opened 2021-11-16
[module-minifier-plugin] Upgrade terser to 5.10.0

dmichon-msft:
## Summary
Upgrades ModuleMinifierPlugin to use terser 5.10.0, removing the need to monkey patch identifier allocation.

## Details
Uses the async minifier API on terser and takes advantage of the new `mangle.nth_identifier` hook to customize identifier allocation and ensure uniformity across minified modules.
Updates NoopMinifier to emit an empty source map if source maps are requested.

## How it was tested
Using the localization-plugin-02 build test.


common/reviews/api/module-minifier-plugin.api.md
```diff
@@ -205,14 +215,6 @@ export const STAGE_AFTER: 100;
 // @public
 export const STAGE_BEFORE: -100;
 
-// @public
-export class SynchronousMinifier implements IModuleMinifier {
-    constructor(options: ISynchronousMinifierOptions);
-    minify(request: IModuleMinificationRequest, callback: IModuleMinificationCallback): void;
-    // (undocumented)
-    readonly terserOptions: MinifyOptions;
```
D4N14L, line 213: Should probably be documented

webpack/module-minifier-plugin/src/NoopMinifier.ts
```diff
@@ -18,13 +18,23 @@ export class NoopMinifier implements IModuleMinifier {
...
-      map: undefined
+      // If source maps are requested, provide an empty source map
+      map: nameForMap
+        ? {
+            version: 3,
+            names: [],
+            file: nameForMap,
+            sources: [nameForMap],
+            sourcesContent: [code],
+            mappings: 'AAAA'
```
D4N14L, line 35: This seems like an odd default for `mappings

webpack/module-minifier-plugin/src/LocalMinifier.ts
```diff
@@ -56,9 +56,24 @@ export class SynchronousMinifier implements IModuleMinifier {
       return callback(cached);
     }
 
-    const result: IModuleMinificationResult = minifySingleFile(request, this.terserOptions);
-    this._resultCache.set(hash, result);
-
-    callback(result);
+    minifySingleFile(request, this.terserOptions)
+      .then((result: IModuleMinificationResult) => {
+        this._resultCache.set(hash, result);
```
D4N14L, line 61: Looking at `minifySingleFile` it seems like it's designed already to catch errors and re-output like below. Would we want to cache if there's an error present in the `result`?

webpack/module-minifier-plugin/src/LocalMinifier.ts
```diff
@@ -56,9 +56,24 @@ export class SynchronousMinifier implements IModuleMinifier {
...
+        });
+      });
   }
 }
+
+/**
+ * Use the name "LocalMinifier" instead for technical correctness.
+ * @deprecated
+ */
+export { LocalMinifier as SynchronousMinifier };
```
D4N14L, line 79: This seems odd... especially since `minify` isn't synchronous anymore.

D4N14L (approved):

common/reviews/api/module-minifier-plugin.api.md
```diff
@@ -205,14 +211,6 @@ export const STAGE_AFTER: 100;
 // @public
 export const STAGE_BEFORE: -100;
 
-// @public
-export class SynchronousMinifier implements IModuleMinifier {
```
iclanton, line 209: This is a breaking change.

D4N14L: David and I discussed this earlier, but I think a breaking change is fair here. For one, it no longer runs synchronously so labeling it as such is a bit misleading. Also, it's still a `0.x.x` package so the minor bump behaves like a major bump in terms of `~` and `^` versions in `package.json`.

iclanton: Sounds reasonable, then.

iclanton (approved):

PR #3025, opened 2021-11-18
[rush] Stop deleting pnpm store after failed workspace installs

dmichon-msft:
## Summary
Stops deleting the pnpm store after max retries have been reached for a workspace install.

## Details
Breaking the pnpm store is a situation that in practice does not happen. The two most common causes of failed install are that the requested package does not exist in the registry (happens a lot when mirroring packages, or due to user error in package.json), or a network error on the machine. Neither situation is helped by deleting the local cache.

### DIFF
--- a/apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
+++ b/apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
@@ -329,57 +329,42 @@ export class WorkspaceInstallManager extends BaseInstallManager {
       );
     }
 
-    try {
-      Utilities.executeCommandWithRetry(
-        {
-          command: packageManagerFilename,
-          args: installArgs,
-          workingDirectory: this.rushConfiguration.commonTempFolder,
-          environment: packageManagerEnv,
-          suppressOutput: false
-        },
-        this.options.maxInstallAttempts,
-        () => {
-          if (this.rushConfiguration.packageManager === 'pnpm') {
-            console.log(colors.yellow(`Deleting the "node_modules" folder`));
-            this.installRecycler.moveFolder(commonNodeModulesFolder);
+    Utilities.executeCommandWithRetry(
+      {
+        command: packageManagerFilename,
+        args: installArgs,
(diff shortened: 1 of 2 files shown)


apps/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
```diff
@@ -329,57 +329,42 @@ export class WorkspaceInstallManager extends BaseInstallManager {
...
       }
+    );
+
+    // If all attempts fail we just terminate. No special handling needed.
+
+    // Ensure that node_modules folders exist after install, since the timestamps on these folders are used
+    // to determine if the install can be skipped
+    const projectNodeModulesFolders: string[] = [
+      path.join(this.rushConfiguration.commonTempFolder, RushConstants.nodeModulesFolderName),
+      ...this.rushConfiguration.projects.map((x) => {
```
halfnibble, line 361: Can we name this variable `rushProject` or something more descriptive than "x"?

halfnibble:
Do we then have a command to manually delete the pnpm-store if we suspect a corrupt system? And if so, should we print out a warning on multiple failures like, `` `pnpm was unable to successfully install after ${maxRetries} attempts. If you suspect the pnpm-store is corrupt, try the following command: "rush <whatever>" (Note: this erases all cached package downloads!)` ``

D4N14L:
> Do we then have a command to manually delete the pnpm-store if we suspect a corrupt system? And if so, should we print out a warning on multiple failures like, `` `pnpm was unable to successfully install after ${maxRetries} attempts. If you suspect the pnpm-store is corrupt, try the following command: "rush <whatever>" (Note: this erases all cached package downloads!)` ``

We've talked in the past about having a `rush purge --full` or something along those lines, but it's never been fully spec'd out.

dmichon-msft:
If the pnpm store becoming corrupted is really a thing, take it up with Zoltan? I'm pretty sure this is not a thing that happens, especially since part of the boot process of pnpm is integrity checking any parts of the store it is going to use.

halfnibble (approved):

PR #3026, opened 2021-11-22
[module-minifier-plugin] Export the list of javascript reserved words.

iclanton:
## Summary

Exports the list of javascript reserved words from the `@rushstack/module-minifier-plugin` package.

### DIFF
--- a/webpack/module-minifier-plugin/src/MinifiedIdentifier.ts
+++ b/webpack/module-minifier-plugin/src/MinifiedIdentifier.ts
@@ -140,7 +140,6 @@ const RESERVED_ORDINALS: number[] = ((): number[] => {
 
 /**
  * Gets a base54 string suitable for use as a JavaScript identifier, omitting those that are valid ECMAScript keywords
- * Not guaranteed not to collide if `ordinal` \>= 100000
  *
  * @param ordinal - The number to convert to a base54 identifier
  *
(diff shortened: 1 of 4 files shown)


webpack/module-minifier-plugin/src/MinifiedIdentifier.ts
```diff
@@ -24,6 +29,7 @@ const RESERVED_KEYWORDS: string[] = [
   'double',
   'else',
   'enum',
+  'eval',
```
dmichon-msft, line 32: "eval" isn't a reserved word in the spec. It's perfectly valid to do `var eval = 4;`

webpack/module-minifier-plugin/src/MinifiedIdentifier.ts
```diff
@@ -140,9 +140,11 @@ const RESERVED_ORDINALS: number[] = ((): number[] => {
 
 /**
  * Gets a base54 string suitable for use as a JavaScript identifier, omitting those that are valid ECMAScript keywords
- * Not guaranteed not to collide if `ordinal` >= 100000
+ * Not guaranteed not to collide if `ordinal` \>= 100000
```
dmichon-msft, line 143: I think this statement no longer holds.

dmichon-msft (approved):
Just one minor note on the old comments.

PR #3041, opened 2021-11-29
[heft-jest-plugin] update to jest 27

rzyns:
## Summary

This updates the Jest dependency to the latest (`~27.3.1`) jest libraries.

## Details

There are some breaking changes, which is to be expected when jumping two whole major versions. I'm not sure I've uncovered them all, the only one that seems to have affected the projects I've tested this on is disallowing `done()` callbacks in tests defined by asynchronous functions. I haven't measured performance quantitatively, but it _feels_ like my builds are running faster with this change :)

## How it was tested

- ran `rush update && rush rebuild` in a fresh clone
- proxied the package in my own Rush project of 27 packages (mix of browser and node), and ran `rush update && rush rebuild` many times to iteratively address the breaking changes in jest 27
- ran `rush rebuild` several times after addressing the breaking changes, and it consistently builds cleanly now


ErikMikkelson:
Would it be worth updating jest-build-transform.ts to iterate over jestOptions.config.roots instead of hardcoding to package src?  I have a modified version of the jest plugin for 27.  which does that so we can keep our deployment code separate from our shipping code.

rzyns:
> Would it be worth updating jest-build-transform.ts to iterate over jestOptions.config.roots instead of hardcoding to package src? I have a modified version of the jest plugin for 27. which does that so we can keep our deployment code separate from our shipping code.

Yeah, that's a good idea, @ErikMikkelson. Maybe that should be done in a separate PR, though? At this point, I just really, _really_ want modern Jest support in main-line Rush, so I figured I'd try to change as little as possible in this PR. But if you have a specific example of what you want to see and you're inclined to share it, please do send it my way. In either case, I'll poke around a bit more in `jest-build-transform.ts` and see what I can do

ErikMikkelson:
> Yeah, that's a good idea, @ErikMikkelson. Maybe that should be done in a separate PR, though? 

You're right, @rzyns , I don't want to derail your PR.  I can set up a PR later.  

If you're curious, the code was in jest-build-transform, you just wrap the process with something like

`
    // CHANGECHANGE: Looping through all jestOptions.roots instead of assuming 'src'
    for (const root of (jestOptions.config.roots || [])) {
      const [subdir] = root.split('/').slice(-1);
      // Is the input file under the "src" folder?
      const srcFolder: string = path.join(jestOptions.config.rootDir, subdir);
`

The only other place to change is the jest config, of course.

D4N14L:
Hey! So I took some inspiration from your PR and created a separate one with some additional changes. Feel free to take a look and try out this branch? #3076

iclanton:
Jest 27 has been released with @rushstack/heft-jest-plugin@0.2.0

PR #3043, opened 2021-11-30
[rush] Rework execution engine

dmichon-msft:
## Summary
Reworks the task execution engine in Rush to interact with the task queue using the ECMAScript async iteration protocol.

## Details
Creates a new `AsyncTaskQueue` component that implements the async iteration protocol (`Symbol.asyncIterator`)

Updates the TaskSelector to be reusable within a watch mode process for multiple build passes, moving the pass-specific parameters to the `createTasks()` method. Ensures that all dependency relationships between tasks are preserved, even if the across tasks that are excluded from a build.

Adds a local vscode launch configuration to debug a unit test file in a Heft project.

Updates interaction with the build cache such that, if a dependency task is skipped for any reason, the task will not write to the build cache.

## How it was tested
Various local build commands.
Verified that unsafe selection parameters (e.g. `-o rush-lib`) work as expected.


apps/rush-lib/src/logic/taskRunner/AsyncTaskQueue.ts
```diff
@@ -0,0 +1,160 @@
...
+import { TaskStatus } from './TaskStatus';
+
+/**
+ * @internal
+ *
+ * Implmentation of the async iteration protocol for a collection of Task objects.
+ * The async iterator will wait for a task to be ready for execution, or terminate if there are no more tasks.
+ *
+ * @remarks
+ * It is imperative that the caller update the status of tasks prior to invoking `next()` on the iterator again,
```
D4N14L, line 14: Maybe there should be a pseudo-state you set it to before returning the task, to avoid requiring this be done by the implementer? Something like `TaskStatus.waitingForAction` or something (not sure what a good name would be, this is just the first thing that I didn't hate)? Just seems odd to place the ownership of setting the task state outside of the thing that cares about it.

D4N14L (approved):

apps/rush-lib/src/logic/taskRunner/AsyncTaskQueue.ts
D4N14L: Maybe you can pass in the state to set it to in the constructor, and set the state as they're iterated and returned from the `next()` call?

dmichon-msft: To phrase this a bit differently, calling `next()` on the iterator is the operation that scans the queue for ready tasks and assigns them to pending iterators. The task status is not the issue, the dependency count is.

apps/rush-lib/src/logic/taskRunner/AsyncTaskQueue.ts
```diff
@@ -0,0 +1,160 @@
...
+      }
+    );
+
+    // By iterating in reverse order we do less array shuffling when removing tasks
+    for (let i: number = taskQueue.length - 1; waitingIterators.length > 0 && i >= 0; i--) {
+      const task: Task = taskQueue[i];
+
+      if (task.status !== TaskStatus.Ready) {
+        // It shouldn't be on the queue, remove it
+        // This should be a blocked task
```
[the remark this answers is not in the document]
dmichon-msft: That would make sense. Technically tasks in the queue should only exist in the states "Ready" and "Blocked", because the former is the initial state and the latter is the only one that gets set by execution of other tasks. However, this particular check was migrated directly from the old code. I can update it, but it's worth checking that tasks only are in one of those two states.

apps/rush-lib/src/logic/taskRunner/AsyncTaskQueue.ts
D4N14L: Then I would say the comment needs a bit of updating to better reflect that. I was a bit confused and assumed it meant the literal `Task.status`

apps/rush-lib/src/logic/taskRunner/AsyncTaskQueue.ts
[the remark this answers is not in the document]
D4N14L: I'd say it's fine as-is then, but yes checking for only those two states would be a good call. Otherwise it should blow up with an explicit error message. Would make it clearer when someone tries to re-call `next()` what exactly happened.

.vscode/launch.json
```diff
@@ -43,6 +43,24 @@
...
+        "--debug",
+        "test",
+        "--test-path-pattern",
+        "${fileBasenameNoExtension}"
+      ],
+      "skipFiles": ["<node_internals>/**"],
+      "outFiles": [],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen"
+    },
```
iclanton, line 63:
```suggestion
		{
			"type": "node",
			"request": "launch",
			"name": "Debug Selected Test File (Heft)",
			"cwd": "${fileDirname}",
			"runtimeArgs": [
				"--inspect-brk",
				"${workspaceFolder}/apps/heft/lib/start.js",
				"--debug",
				"test",
				"--test-path-pattern",
				"${fileBasenameNoExtension}"
			],
			"skipFiles": ["<node_internals>/**"],
			"outFiles": [],
			"console": "integratedTerminal",
			"internalConsoleOptions": "neverOpen"
		},
```

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -153,43 +167,72 @@ export class CommandLineConfiguration {
...
+      const relatedPhaseSet: Set<IPhase> = new Set([phase]);
+      for (const relatedPhase of relatedPhaseSet) {
+        for (const dependency of relatedPhase.phaseDependencies.self) {
+          relatedPhaseSet.add(dependency);
+        }
+
+        for (const dependency of relatedPhase.phaseDependencies.upstream) {
+          relatedPhaseSet.add(dependency);
         }
       }
```
iclanton, line 230: `safePhases` stays empty through these loops. Is this code missing something?

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -433,38 +483,42 @@ export class CommandLineConfiguration {
...
+      return;
+    }
+
+    for (const dependency of phase.phaseDependencies.self) {
+      if (searchedPhases.has(dependency)) {
+        const dependencyNameForError: string =
+          typeof dependency.name === 'string' ? dependency.name : '<synthetic phase>';
+        throw new Error(
+          `In ${RushConstants.commandLineFilename}, there exists a cycle within the ` +
+            `set of ${dependencyNameForError} dependencies: ${Array.from(
```
iclanton, line 509:
```suggestion
        throw new Error(
          `In ${RushConstants.commandLineFilename}, there exists a cycle within the ` +
            `set of ${dependency.name} dependencies: ${Array.from(
```

iclanton: `dependencyNameForError` was a remnant of an old design that I didn't fully clean up.

apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -22,8 +22,13 @@ import { RushConfigurationProject } from '../../api/RushConfigurationProject';
 import { BuildCacheConfiguration } from '../../api/BuildCacheConfiguration';
 import { SelectionParameterSet } from '../SelectionParameterSet';
 import { CommandLineConfiguration, IPhase, IPhasedCommand } from '../../api/CommandLineConfiguration';
-import { IProjectTaskSelectorOptions, ProjectTaskSelector } from '../../logic/ProjectTaskSelector';
+import {
```
iclanton, line 25:
```suggestion
import type {
```

apps/rush-lib/src/logic/ProjectTaskSelector.ts
```diff
@@ -33,18 +37,15 @@ export interface IProjectTaskSelectorOptions {
...
 
   public constructor(options: IProjectTaskSelectorOptions) {
     this._options = options;
-    this._projectChangeAnalyzer =
-      options.projectChangeAnalyzer || new ProjectChangeAnalyzer(options.rushConfiguration);
-    this._phasesToRun = options.phasesToRun;
-    this._phases = options.phases;
-    this._customParametersByPhaseName = new Map();
+    this._phasesToRun = new Set(options.phasesToRun);
+    this._phases = Array.from(options.phases.values());
```
iclanton, line 47: How do we know the values in a map are going to be in a particular order?

apps/rush-lib/src/logic/ProjectTaskSelector.ts
```diff
@@ -66,120 +67,142 @@ export class ProjectTaskSelector {
     }
   }
 
-  public registerTasks(): TaskCollection {
-    const projects: ReadonlySet<RushConfigurationProject> = this._options.projectSelection;
-    const taskCollection: TaskCollection = new TaskCollection();
+  public createTasks(createTasksOptions: ICreateTasksOptions): Set<Task> {
+    const { _options: options } = this;
```
iclanton, line 71:
```suggestion
    const options: IProjectTaskSelectorOptions = this._options;
```

apps/rush-lib/src/logic/ProjectTaskSelector.ts
```diff
@@ -66,120 +67,142 @@ export class ProjectTaskSelector {
...
   }
 
-  private _getPhaseDisplayNameForProject(phase: IPhase, project: RushConfigurationProject): string {
+  private _getTaskDisplayName(phase: IPhase, project: RushConfigurationProject): string {
     if (phase.isSynthetic) {
       // Because this is a synthetic phase, just use the project name because there aren't any other phases
       return project.packageName;
     } else {
-      const phaseNameWithoutPrefix: string = phase.name.substring(RushConstants.phaseNamePrefix.length);
+      const phaseNameWithoutPrefix: string = phase.name.slice(RushConstants.phaseNamePrefix.length);
```
iclanton, line 190: Is slice more performant?

apps/rush-lib/src/logic/ProjectTaskSelector.ts
```diff
@@ -66,120 +67,142 @@ export class ProjectTaskSelector {
...
+     * task_ordinal = (project_count * phase_index) + project.index
+     */
+    function* getRawDependencies(taskOrdinal: number): Iterable<number> {
+      // eslint-disable-next-line no-bitwise
+      const projectIndex: number = taskOrdinal % projectCount | 0;
+
+      const {
+        phaseDependencies: { self, upstream }
+      }: // eslint-disable-next-line no-bitwise
+      IPhase = ordinalPhases[(taskOrdinal / projectCount) | 0];
```
iclanton, line 131: Looking phases up by index seems far more complicated than looking them up by name. What's the benefit to this change?

apps/rush-lib/src/logic/taskExecution/ProjectTaskRunner.ts
```diff
@@ -54,7 +54,7 @@ export interface IProjectTaskRunnerOptions {
   projectChangeAnalyzer: ProjectChangeAnalyzer;
   allowWarningsInSuccessfulBuild?: boolean;
   taskName: string;
-  phase: IPhase;
+  phase: Pick<IPhase, 'name' | 'logFilenameIdentifier'>;
```
iclanton, line 57: Why the computed type?

common/reviews/api/rush-lib.api.md
```diff
@@ -590,14 +590,16 @@ export class RushConfigurationProject {
...
+    constructor(projectJson: IRushConfigurationProjectJson, rushConfiguration: RushConfiguration, tempProjectName: string, index: number);
     // @internal
     readonly _consumingProjectNames: Set<string>;
     get consumingProjects(): ReadonlySet<RushConfigurationProject>;
     get cyclicDependencyProjects(): Set<string>;
     get dependencyProjects(): ReadonlySet<RushConfigurationProject>;
     // @deprecated
     get downstreamDependencyProjects(): string[];
+    // @internal
+    get index(): number;
```
iclanton, line 602: Prefix this property name with an underscore.

libraries/heft-config-file/src/test/__snapshots__/ConfigurationFile.test.ts.snap
```diff
@@ -185,7 +185,7 @@ Object {
 `;
 
 exports[`ConfigurationFile error cases Throws an error when the file isn't valid JSON 1`] = `
-[Error: In config file "<project root>/src/test/errorCases/invalidJson/config.json": SyntaxError: Unexpected token '}' at 2:19
+[Error: In config file "<project root>/src/test/errorCases/invalidJson/config.json": SyntaxError: Unexpected token '\\n' at 2:19
```
iclanton, line 188: Revert.

apps/rush-lib/src/api/CommandLineConfiguration.ts
dmichon-msft: safePhases is mutated by _checkForPhaseSelfCycles and is used to skip redundant work while performing a depth-first search of the dependency graph.

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -202,24 +245,31 @@ export class CommandLineConfiguration {
...
                 );
               }
+
+              phases.add(phase);
             }
 
             if (normalizedCommand.skipPhasesForCommand) {
               for (const phaseName of normalizedCommand.skipPhasesForCommand) {
-                if (!this.phases.has(phaseName)) {
+                const phase: IPhase | undefined = this.phases.get(phaseName);
```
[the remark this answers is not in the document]
dmichon-msft: The value is supposed to be used, just needed to determine correct evaluation order.

apps/rush-lib/src/logic/ProjectTaskSelector.ts
dmichon-msft: The ECMAScript Map specification says that the iterator enumerates entries in insertion order.

apps/rush-lib/src/logic/taskExecution/ProjectTaskRunner.ts
dmichon-msft: This had been done because I was translating the phase type in ProjectTaskSelector, but is now no longer necessary.

libraries/heft-config-file/src/test/__snapshots__/ConfigurationFile.test.ts.snap
dmichon-msft: This change is here because without it I get a build error from the repository.

apps/rush-lib/src/logic/ProjectTaskSelector.ts
dmichon-msft: Nah, just habitual consistency

apps/rush-lib/src/api/CommandLineJson.ts
```diff
@@ -31,12 +31,19 @@ export interface IBulkCommandJson extends IBaseCommandJson {
 }
 
 /**
- * "phasedCommand" from command-line.schema.json
+ * Base interface shared by the "phasedCommand" JSON entries and the post-processed
+ * "IPhase" interface in the CommandLineConfiguration
  */
-export interface IPhasedCommandJson extends IBaseCommandJson {
+export interface IPhasedCommandWithoutPhasesJson extends IBaseCommandJson {
```
iclanton, line 37: I still think this would be cleaner with a generic type on `phases` and `skipPhasesForCommand`

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -153,43 +170,72 @@ export class CommandLineConfiguration {
...
+            self: new Set(),
+            upstream: new Set()
+          }
+        };
+
+        this.phases.set(phase.name, processedPhase);
       }
     }
 
+    // Resolve the string links
```
iclanton, line 189:
```suggestion
    // Resolve phase names
```

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -153,43 +170,72 @@ export class CommandLineConfiguration {
...
           }
+          phase.phaseDependencies.upstream.add(dependency);
+        }
+      }
+    }
+
+    // Do the recursive stuff after the dependencies have been converted
+    const safePhases: Set<IPhase> = new Set();
+    for (const phase of this.phases.values()) {
+      const relatedPhaseSet: Set<IPhase> = new Set([phase]);
```
iclanton, line 224: `relatedPhaseSet` doesn't appear to be used anywhere.

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -202,29 +248,51 @@ export class CommandLineConfiguration {
...
                     `"${normalizedCommand.name}" command, the phase "${phaseName}" does not exist.`
                 );
               }
+
+              commandPhases.add(phase);
+            }
+
+            // Apply implicit phase dependency expansion
+            // The equivalent of the "--to" operator used for projects
+            for (const phase of commandPhases) {
```
iclanton, line 274: Recursive expansion?

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -433,38 +501,47 @@ export class CommandLineConfiguration {
...
+  /**
+   * Performs a depth-first search to detect cycles in the directed graph of phase "self" dependencies.
+   */
+  private _checkForPhaseSelfCycles(
+    phase: IPhase,
+    // Phases in the current depth-first-search path
+    phasesInPath: Set<IPhase>,
+    // Phases that have already been fully walked and confirmed to not be in any cycles
+    cycleFreePhases: Set<IPhase>
+  ): void {
```
iclanton, line 521:
```suggestion
   * Performs a depth-first search to detect cycles in the directed graph of phase "self" dependencies.
   * 
   * @param phasesInPath - Phases in the current depth-first-search path
   * @param cycleFreePhases - Phases that have already been fully walked and confirmed to not be in any cycles
   */
  private _checkForPhaseSelfCycles(phase: IPhase, phasesInPath: Set<IPhase>, cycleFreePhases: Set<IPhase>): void {
```

apps/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -72,11 +73,13 @@ export class RushConfigurationProject {
   public constructor(
```
iclanton, line 73: These args have gotten complex enough that I think this warrants an `options` object.

apps/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -216,6 +219,14 @@ export class RushConfigurationProject {
     return this._rushConfiguration;
   }
 
+  /**
+   * This index of this project in the Rush configuration's projects array
+   * @internal
+   */
+  public get _index(): number {
```
iclanton, line 226: I know we already have the getter convention, but it's probably fine to just create this as a `public readonly` property.

We can realistically convert the rest to that, but I'd want to do that in either a separate chagne or during the next major release.

apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -7,7 +7,7 @@ import colors from 'colors/safe';
 import { AlreadyReportedError, Terminal } from '@rushstack/node-core-library';
 import { CommandLineFlagParameter, CommandLineStringParameter } from '@rushstack/ts-command-line';
 
-import { Event } from '../../index';
+import { Event, ProjectChangeAnalyzer } from '../../index';
```
iclanton, line 10: Can you update these to import from their actual files?

apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -21,10 +21,12 @@ import { LastLinkFlag, LastLinkFlagFactory } from '../../api/LastLinkFlag';
 import { RushConfigurationProject } from '../../api/RushConfigurationProject';
 import { BuildCacheConfiguration } from '../../api/BuildCacheConfiguration';
 import { SelectionParameterSet } from '../SelectionParameterSet';
-import { CommandLineConfiguration, IPhase, IPhasedCommand } from '../../api/CommandLineConfiguration';
-import { IProjectTaskSelectorOptions, ProjectTaskSelector } from '../../logic/ProjectTaskSelector';
-import { IFlagParameterJson } from '../../api/CommandLineJson';
+import type { CommandLineConfiguration, IPhase, IPhasedCommand } from '../../api/CommandLineConfiguration';
+import type { IProjectTaskSelectorOptions } from '../../logic/ProjectTaskSelector';
+import { ProjectTaskSelector } from '../../logic/ProjectTaskSelector';
```
iclanton, line 26: Does our TS version support 

```suggestion
import { ProjectTaskSelector , type IProjectTaskSelectorOptions } from '../../logic/ProjectTaskSelector';
```

yet?

apps/rush-lib/src/logic/taskExecution/TaskExecutionManager.ts
```diff
@@ -156,16 +154,14 @@ export class TaskExecutionManager {
    * tasks are completed successfully, or rejects when any task fails.
    */
   public async executeAsync(): Promise<void> {
-    this._currentActiveTasks = 0;
     this._completedTasks = 0;
-    this._totalTasks = this._tasks.length;
+    const totalTasks: number = (this._totalTasks = this._tasks.size);
```
iclanton, line 158: Why not just set `this._totalTasks` in the constructor?

apps/rush-lib/src/logic/taskExecution/TaskExecutionManager.ts
```diff
@@ -174,7 +170,36 @@ export class TaskExecutionManager {
...
+        // laneId can be used in logging to examine concurrency
+        const laneId: number = index + 1;
+        // The taskQueue is a singular async iterable that stalls until a task is available, and marks itself
+        // done when the queue is empty.
+        for await (const task of taskQueue) {
+          // Take a task, execute it, wait for it to finish, wait for a new task
+          await this._executeTaskAsync(task, laneId);
+        }
+      })
+    );
```
iclanton, line 202: This feels fairly similar to `Async.forEachAsync(..., () => { ... }, { concurrency: maxParallelism });`

apps/rush-lib/src/logic/taskExecution/TaskExecutionManager.ts
```diff
@@ -363,6 +344,17 @@ export class TaskExecutionManager {
...
       dependent.dependencies.delete(task);
     });
+
+    const invalidationQueue: Set<Task> = new Set(task.dependents);
+    for (const consumer of invalidationQueue) {
+      // If a task is skipped, state is not guaranteed in downstream tasks, so block cache write
+      consumer.runner.isCacheWriteAllowed = false;
+
+      // Propagate through the entire build queue applying cache write prevention.
+      for (const indirectConsumer of consumer.dependents) {
```
iclanton, line 354: Seems like this needs to recurse.

apps/rush-lib/src/logic/taskExecution/ProjectTaskFactory.ts
```diff
@@ -0,0 +1,133 @@
...
+        projectChangeAnalyzer: factoryOptions.projectChangeAnalyzer,
+        phase
+      }),
+      TaskStatus.Ready
+    );
+
+    return task;
+  }
+
+  public static getScriptToRun(
```
iclanton, line 70: This doesn't appear to be used outside of this class.

apps/rush-lib/src/logic/taskExecution/ProjectTaskFactory.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    if (!script) {
+      return '';
+    } else {
+      const taskCommand: string = `${script} ${customParameterValues.join(' ')}`;
+      return process.platform === 'win32' ? convertSlashesForWindows(taskCommand) : taskCommand;
+    }
+  }
+
+  public static getTaskDisplayName(phase: IPhase, project: RushConfigurationProject): string {
```
iclanton, line 89: Same with this.

apps/rush-lib/src/logic/taskExecution/AsyncTaskQueue.ts
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import { Task } from './Task';
+import { TaskStatus } from './TaskStatus';
+
+/**
+ * @internal
```
iclanton, line 8: This isn't exposed from `index.ts`, so `@internal` is implied.

apps/rush-lib/src/logic/taskExecution/AsyncTaskQueue.ts
```diff
@@ -0,0 +1,160 @@
...
+      }
+    );
+
+    // By iterating in reverse order we do less array shuffling when removing tasks
+    for (let i: number = taskQueue.length - 1; waitingIterators.length > 0 && i >= 0; i--) {
+      const task: Task = taskQueue[i];
+
+      if (task.status !== TaskStatus.Ready) {
+        // It shouldn't be on the queue, remove it
+        // This should be a blocked task
```
iclanton, line 52: Should we perform a sanity check for this?

apps/rush-lib/src/logic/ProjectTaskSelector.ts
iclanton: It'd be safer with future refactors to just record the phases into an ordered array in the first place and pass that in.

apps/rush-lib/src/api/CommandLineConfiguration.ts
dmichon-msft: Relic from my merge with your previous PR

apps/rush-lib/src/api/CommandLineConfiguration.ts
dmichon-msft: This does a full breadth-first search. Updated the comment. Same algorithm is used to apply the `--to` operator in `Selection.expandAllDependencies`

apps/rush-lib/src/api/CommandLineJson.ts
dmichon-msft: It's a bit tricky to use a generic to discard a property from an object; `skipPhasesForCommand` gets trimmed during resolution.

dmichon-msft: That would make this:
```ts
export interface IPhasedCommandJson<TPhases = string[]> extends IBaseCommandJson {
  ...
  phases: TPhases;
  skipPhasesForCommand?: TPhases extends string[] ? TPhases : never;
}
```
Which I think is generally less readable.

apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
dmichon-msft: Needs 4.5+

apps/rush-lib/src/logic/taskExecution/TaskExecutionManager.ts
dmichon-msft: It is similar, but this version was created to support the DevTools tracing style logging to keep track of which CPU core or "parallelism lane" a given task executes on; information which does not exist in the confines of `Async.forEachAsync`. By tracking this information it makes the parallelism at any point during the execution considerably easier to view. It also currently has slightly different error handling behavior, though that's more due to laziness than being a design objective.

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -36,19 +36,33 @@ export interface IPhase extends IPhaseJson {
...
+    self: Set<IPhase>;
+    upstream: Set<IPhase>;
+  };
 }
 
 export interface ICommandWithParameters {
   associatedParameters: Set<Parameter>;
 }
-export interface IPhasedCommand extends IPhasedCommandJson, ICommandWithParameters {
+export interface IPhasedCommand extends Omit<IPhasedCommandJson, 'phases'>, ICommandWithParameters {
```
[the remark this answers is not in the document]
dmichon-msft: Used a common base to avoid trying to discard `skipPhasesForCommand` in a child interface.

apps/rush-lib/src/logic/ProjectTaskSelector.ts
dmichon-msft: Switching to strings, since the performance effects are negligible

apps/rush-lib/src/logic/ProjectTaskSelector.ts
```diff
@@ -1,220 +1,170 @@
...
-  private _getPhaseDisplayNameForProject(phase: IPhase, project: RushConfigurationProject): string {
-    if (phase.isSynthetic) {
-      // Because this is a synthetic phase, just use the project name because there aren't any other phases
-      return project.packageName;
-    } else {
-      const phaseNameWithoutPrefix: string = phase.name.substring(RushConstants.phaseNamePrefix.length);
-      return `${project.packageName} (${phaseNameWithoutPrefix})`;
-    }
-  }
+      for (const dep of getRawDependencies(node)) {
```
dmichon-msft, line 127: If I change the signature to `const [depKey, depProject, depPhase] of getRawDependencies(project, phase)` and likewise add the `project, phase` parameters to `getFilteredDependencies(key, project, phase)` I can remove the need to ever deserialize the keys, which is probably a performance win.

dmichon-msft: Though since this whole execution takes on the order of 25 ms for 2 phases with 600 projects, might not be terribly relevant.

apps/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -33,50 +26,67 @@ export interface IRushConfigurationProjectJson {
...
   /**
-   * A set of projects within the Rush configuration which directly consume this package.
-   *
-   * @remarks
-   * Writable because it is mutated by RushConfiguration during initialization.
+   * This index of this project in the Rush configuration's projects array
    * @internal
    */
-  public readonly _consumingProjectNames: Set<string>;
+  public readonly _index: number;
```
iclanton, line 61: This is no longer used anywhere.

dmichon-msft: Removed.

iclanton (approved):

PR #3050, opened 2021-12-01
[rush] Use lockfile diffing for more granular Git changed project detection

### LINKED ISSUE #3053: [rush] In 5.57.0-dev.2, "rush change -v" sometimes incorrectly considers ALL projects to be changed
## Summary

While testing **5.57.0-dev.2** in our monorepo, we encountered a regression where `rush change -v` incorrectly detects that every project in the monorepo has been changed.

👉This appears to be a regression introduced by PR #2968.

## Repro steps

The repro steps are in a private monorepo, and I don't have time to prepare them.

## Details

The regression is in this `ProjectChangeAnalyzer.getChangedProjectsAsync()`:

https://github.com/microsoft/rushstack/blob/7d8ff23ab4b500123434e7b5a0ca0400278aece1/apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts#L200-L207

The [getRepoChanges()](https://github.com/microsoft/rushstack/blob/7d8ff23ab4b500123434e7b5a0ca0400278aece1/apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts#L188) call invokes `git diff-index` and correctly identifies the list of files with changes.

However PR #2968 added some extra logic (`TODO: Implement shrinkwrap diffing here`) that basically says _"If the shrinkwrap file has been modified, then assume every project in the monorepo has changes."_

That might make sense for an incremental operation, where processing a few extra projects is okay.  But not for `rush change`.

More fundamentally, I think the code is conflating two subtly different concepts of "impacted by":

1. `rush build`:  Which build tasks **depend upon** files touched by this diff?  ...versus...
2. `rush change`:  Which projects **are parent folders of** source files that are touched by this diff?

The second concept is much more narrow.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `rushVersion` from rush.json? | 5.57.0-dev.2 |

dmichon-msft:
## Summary
When running a command like `rush list --only git:HEAD`, if the set of touched files includes the lockfile and the package manager is `pnpm`, compares the old and new lockfiles to determine affected projects instead of assuming that the change affects all projects.

## Details
Adds a `getFileContent` method to the `Git` wrapper to read the content of a file at a previous commit.
Adds a `loadFromString` API to the `ShrinkwrapFileFactory` and underlying package-manager-specific files.
Updates `ProjectChangeAnalyzer` to, upon seeing a lockfile modification, if the package manager is `pnpm` (since project shrinkwraps are not implemented for other package managers), load both the old and current lockfiles and compare the project shrinkwrap entries.
Adds a `hasChanged` method to the `PnpmProjectShrinkwrapFile` class to abstract the comparison operation.
Optimizes bulk computation of `PnpmProjectShrinkwrapFile` content by caching the dependencies of external packages for the entire lockfile, so that the per-package computations can be reused across projects. The cache lives as long as the lockfile object, which is always thrown out as soon as the per-project shrinkwraps have been computed.

Fixes #3053

## How it was tested
Staged various temporary changes to `common/config/rush/pnpm-lock.yaml` and verified the set of projects selected by `rush list --only git:HEAD`

### DIFF
--- a/apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
+++ b/apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
@@ -282,9 +282,9 @@ export class BulkScriptAction extends BaseScriptAction {
     this._selectionParameters = new SelectionParameterSet(this.rushConfiguration, this, {
       // Include lockfile processing since this expands the selection, and we need to select
       // at least the same projects selected with the same query to "rush build"
-      includeLockfile: true,
+      includeExternalDependencies: true,
       // Enable filtering to reduce evaluation cost
-      enableFiltering: false
+      enableFiltering: true
     });
 
     this._verboseParameter = this.defineFlagParameter({
--- a/apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
+++ b/apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
@@ -131,8 +131,7 @@ export class BulkScriptAction extends BaseScriptAction {
     }
 
     const selection: Set<RushConfigurationProject> = await this._selectionParameters.getSelectedProjectsAsync(
-      terminal,
-      true
+      terminal
     );
 
     if (!selection.size) {
@@ -280,7 +279,13 @@ export class BulkScriptAction extends BaseScriptAction {
       });
     }
 
-    this._selectionParameters = new SelectionParameterSet(this.rushConfiguration, this);
+    this._selectionParameters = new SelectionParameterSet(this.rushConfiguration, this, {
+      // Include lockfile processing since this expands the selection, and we need to select
+      // at least the same projects selected with the same query to "rush build"
+      includeLockfile: true,
+      // Enable filtering to reduce evaluation cost
+      enableFiltering: false
+    });
 
     this._verboseParameter = this.defineFlagParameter({
       parameterLongName: '--verbose',
--- a/apps/rush-lib/src/logic/Git.ts
+++ b/apps/rush-lib/src/logic/Git.ts
@@ -21,6 +21,11 @@ interface IResultOrError<TResult> {
   result?: TResult;
 }
 
+export interface IGetBlobOptions {
+  blobSpec: string;
+  repositoryRoot: string;
+}
+
 export class Git {
   private readonly _rushConfiguration: RushConfiguration;
   private _checkedGitPath: boolean = false;
@@ -207,11 +212,11 @@ export class Git {
     return result;
   }
 
-  public getFileContent(targetBranch: string, gitRelativePath: string, repositoryRoot: string): string {
+  public getBlobContent({ blobSpec, repositoryRoot }: IGetBlobOptions): string {
     const gitPath: string = this.getGitPathOrThrow();
     const output: string = Utilities.executeCommandAndCaptureOutput(
       gitPath,
-      ['cat-file', 'blob', `${targetBranch}:${gitRelativePath}`, '--'],
+      ['cat-file', 'blob', blobSpec, '--'],
       repositoryRoot
     );
 
--- a/apps/rush-lib/src/logic/Git.ts
+++ b/apps/rush-lib/src/logic/Git.ts
@@ -207,6 +207,17 @@ export class Git {
     return result;
   }
 
+  public getFileContent(targetBranch: string, gitRelativePath: string, repositoryRoot: string): string {
+    const gitPath: string = this.getGitPathOrThrow();
+    const output: string = Utilities.executeCommandAndCaptureOutput(
+      gitPath,
+      ['cat-file', 'blob', `${targetBranch}:${gitRelativePath}`, '--'],
+      repositoryRoot
+    );
+
+    return output;
+  }
+
   /**
    * @param pathPrefix - An optional path prefix "git diff"s should be filtered by.
    * @returns
--- a/apps/rush-lib/src/logic/Git.ts
+++ b/apps/rush-lib/src/logic/Git.ts
@@ -191,7 +191,7 @@ export class Git {
     const gitPath: string = this.getGitPathOrThrow();
     const output: string = Utilities.executeCommandAndCaptureOutput(
       gitPath,
-      ['--no-optional-locks', 'merge-base', 'HEAD', `${targetBranch}`, '--'],
+      ['--no-optional-locks', 'merge-base', 'HEAD', targetBranch, '--'],
       this._rushConfiguration.rushJsonFolder
     );
     const result: string = output.trim();
--- a/apps/rush-lib/src/logic/Git.ts
+++ b/apps/rush-lib/src/logic/Git.ts
@@ -183,6 +183,22 @@ export class Git {
     }
   }
 
+  public getMergeBase(targetBranch: string, terminal: ITerminal, shouldFetch: boolean = false): string {
+    if (shouldFetch) {
+      this._fetchRemoteBranch(targetBranch, terminal);
+    }
+
+    const gitPath: string = this.getGitPathOrThrow();
+    const output: string = Utilities.executeCommandAndCaptureOutput(
+      gitPath,
+      ['--no-optional-locks', 'merge-base', 'HEAD', `${targetBranch}`, '--'],
+      this._rushConfiguration.rushJsonFolder
+    );
+    const result: string = output.trim();
+
+    return result;
+  }
+
   public getChangedFolders(
     targetBranch: string,
     terminal: ITerminal,
--- a/apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
+++ b/apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
@@ -224,7 +224,7 @@ export class ProjectChangeAnalyzer {
 
       if (shrinkwrapStatus) {
         if (shrinkwrapStatus.status !== 'M') {
-          // The lockfile was created or deleted. Affects all projects.
+          terminal.writeLine(`Lockfile was created or deleted. Assuming all projects are affected.`);
           return new Set(rushConfiguration.projects);
         }
 
@@ -255,7 +255,9 @@ export class ProjectChangeAnalyzer {
             }
           }
         } else {
-          // Diffing only supported for pnpm. Must assume it affects all projects since we can't be sure.
+          terminal.writeLine(
+            `Lockfile has changed and lockfile content comparison is only supported for pnpm. Assuming all projects are affected.`
+          );
           return new Set(rushConfiguration.projects);
         }
       }
@@ -287,7 +289,7 @@ export class ProjectChangeAnalyzer {
     }
 
     if (enableFiltering) {
-      // Reading rush-project.json may be problematic if, e.g.
+      // Reading rush-project.json may be problematic if, e.g. rush install has not yet occurred and rigs are in use
       await Async.forEachAsync(
         changesByProject,
         async ([project, projectChanges]) => {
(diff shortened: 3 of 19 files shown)


apps/rush-lib/src/cli/scriptActions/BulkScriptAction.ts
```diff
@@ -280,7 +279,13 @@ export class BulkScriptAction extends BaseScriptAction {
       });
     }
 
-    this._selectionParameters = new SelectionParameterSet(this.rushConfiguration, this);
+    this._selectionParameters = new SelectionParameterSet(this.rushConfiguration, this, {
+      // Include lockfile processing since this expands the selection, and we need to select
+      // at least the same projects selected with the same query to "rush build"
+      includeLockfile: true,
+      // Enable filtering to reduce evaluation cost
+      enableFiltering: false
```
iclanton, line 287:
```suggestion
      enableFiltering: true
```

apps/rush-lib/src/logic/Git.ts
```diff
@@ -207,6 +207,17 @@ export class Git {
     return result;
   }
 
+  public getFileContent(targetBranch: string, gitRelativePath: string, repositoryRoot: string): string {
```
iclanton, line 210:
```suggestion
  public getFileContentAtRef(targetRef: string, gitRelativePath: string, repositoryRoot: string): string {
```

iclanton: Might also make sense to pass an options object to this function to prevent someone from messing up the parameter order.

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -29,6 +30,9 @@ export interface IGetChangedProjectsOptions {
   targetBranchName: string;
   terminal: ITerminal;
   shouldFetch?: boolean;
+
+  includeLockfile: boolean;
+  enableFiltering: boolean;
```
iclanton, line 35:
```suggestion
  /**
   * If set to `true`, consider the project's external dependencies when determining if it has changed.
   */
  includeLockfile: boolean;
  
  /**
   * If set to `true`, ignore files specified in the project's `rush-project.json`'s `incrementalBuildIgnoredGlobs` property.
   */
  enableFiltering: boolean;
```

`includeLockfile` might also make more sense as `includeExternalDependencies` or `includeProjectDependenciesFromLockfile`.

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -178,69 +182,99 @@ export class ProjectChangeAnalyzer {
...
+              currentShrinkwrap
+                .getProjectShrinkwrap(project)
+                .hasChanges(oldShrinkWrap.getProjectShrinkwrap(project))
+            ) {
+              changedProjects.add(project);
+            }
+          }
+        } else {
+          // Diffing only supported for pnpm. Must assume it affects all projects since we can't be sure.
+          return new Set(rushConfiguration.projects);
```
iclanton, line 247: Might be good to print a note about this.

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -178,69 +182,99 @@ export class ProjectChangeAnalyzer {
...
-          changesByProject.set(project, (projectChanges = new Map()));
+        if (changedProjects.has(project)) {
+          // Lockfile changes cannot be ignored via rush-project.json
+          continue;
+        }
+
+        if (enableFiltering) {
+          let projectChanges: Map<string, IFileDiffStatus> | undefined = changesByProject.get(project);
+          if (!projectChanges) {
+            changesByProject.set(project, (projectChanges = new Map()));
```
iclanton, line 267:
```suggestion
            projectChanges = new Map();
            changesByProject.set(project, projectChanges);
```

apps/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -178,69 +182,99 @@ export class ProjectChangeAnalyzer {
...
+          changedProjects.add(project);
         }
-        projectChanges.set(file, diffStatus);
       }
     }
 
-    if (forIncrementalBuild) {
-      const changedProjects: Set<RushConfigurationProject> = new Set();
+    if (enableFiltering) {
+      // Reading rush-project.json may be problematic if, e.g.
```
iclanton, line 277: Looks like this comment is incomplete.

apps/rush-lib/src/logic/npm/NpmShrinkwrapFile.ts
```diff
@@ -47,28 +47,29 @@ export class NpmShrinkwrapFile extends BaseShrinkwrapFile {
...
   }
 
+  public static loadFromString(shrinkwrapContent: string): NpmShrinkwrapFile {
+    // We don't use JsonFile/jju here because shrinkwrap.json is a special NPM file format
+    // and typically very large, so we want to load it the same way that NPM does.
+    // strip BOM
+    const data: string =
+      shrinkwrapContent.charCodeAt(0) === 0xfeff ? shrinkwrapContent.slice(1) : shrinkwrapContent;
+
+    return new NpmShrinkwrapFile(JSON.parse(data));
```
iclanton, line 70:
```suggestion
    // strip BOM
    const data: string =
      shrinkwrapContent.charCodeAt(0) === 0xfeff ? shrinkwrapContent.slice(1) : shrinkwrapContent;

    // We don't use JsonFile/jju here because shrinkwrap.json is a special NPM file format
    // and typically very large, so we want to load it the same way that NPM does.
    return new NpmShrinkwrapFile(JSON.parse(data));
```

apps/rush-lib/src/logic/selectors/GitChangedProjectSelectorParser.ts
```diff
@@ -1,41 +1,35 @@
 import type { RushConfiguration } from '../../api/RushConfiguration';
 import type { RushConfigurationProject } from '../../api/RushConfigurationProject';
-import { EvaluateSelectorMode, IEvaluateSelectorOptions, ISelectorParser } from './ISelectorParser';
-import { ProjectChangeAnalyzer } from '../ProjectChangeAnalyzer';
+import { IEvaluateSelectorOptions, ISelectorParser } from './ISelectorParser';
+import { IGetChangedProjectsOptions, ProjectChangeAnalyzer } from '../ProjectChangeAnalyzer';
+
+export interface IGitSelectorParserOptions {
+  includeLockfile: boolean;
+  enableFiltering: boolean;
```
iclanton, line 8: Same comment here as in `ProjectChangeAnalyzer`'s new options.

apps/rush-lib/src/logic/yarn/YarnShrinkwrapFile.ts
```diff
@@ -166,19 +166,18 @@ export class YarnShrinkwrapFile extends BaseShrinkwrapFile {
   }
 
   public static loadFromFile(shrinkwrapFilename: string): YarnShrinkwrapFile | undefined {
-    let shrinkwrapString: string;
-    let shrinkwrapJson: YarnPkgLockfileTypes.ParseResult;
     try {
-      if (!FileSystem.exists(shrinkwrapFilename)) {
+      return YarnShrinkwrapFile.loadFromString(shrinkwrapFilename);
```
iclanton, line 170:
```suggestion
      const shrinkwrapString: string = FileSystem.readFile(shrinkwrapFilename);
      return YarnShrinkwrapFile.loadFromString(shrinkwrapString);
```

apps/rush-lib/src/utilities/Utilities.ts
```diff
@@ -759,63 +766,25 @@ export class Utilities {
...
+    const options: IExecutableSpawnSyncOptions = {
+      currentWorkingDirectory: workingDirectory,
       stdio: stdio,
-      env: keepEnvironment
+      environment: keepEnvironment
         ? environment
         : Utilities._createEnvironmentForRushCommand({ initialEnvironment: environment })
     };
 
-    // This is needed since we specify shell=true below.
```
iclanton, line 780: Ensure removing this doesn't break lifecycle scripts and global commands.

dmichon-msft: Reverted this part of the change to be scoped to a different PR. It's still something we should do, since running commands in shell mode is bad practice, but will be dealt with separately.

apps/rush-lib/src/logic/Git.ts
```diff
@@ -178,6 +183,22 @@ export class Git {
...
 
+  public getMergeBase(targetBranch: string, terminal: ITerminal, shouldFetch: boolean = false): string {
+    if (shouldFetch) {
+      this._fetchRemoteBranch(targetBranch, terminal);
+    }
+
+    const gitPath: string = this.getGitPathOrThrow();
+    const output: string = Utilities.executeCommandAndCaptureOutput(
+      gitPath,
+      ['--no-optional-locks', 'merge-base', 'HEAD', `${targetBranch}`, '--'],
```
iclanton, line 194:
```suggestion
      ['--no-optional-locks', 'merge-base', 'HEAD', targetBranch, '--'],
```

apps/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -497,6 +501,49 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
+    let integrityMap: Map<string, string> | undefined = this._integrities.get(importerKey);
+    if (!integrityMap) {
+      const importer: IPnpmShrinkwrapImporterYaml | undefined = this.getImporter(importerKey);
+      if (importer) {
+        integrityMap = new Map();
+        this._integrities.set(importerKey, integrityMap);
+
+        const sha256Digest: string = crypto
+          .createHash('sha256')
+          .update(JSON.stringify(importerKey))
```
iclanton, line 516: Why do we need to `JSON.stringify` this.

apps/rush-lib/src/logic/pnpm/PnpmProjectShrinkwrapFile.ts
```diff
@@ -71,16 +81,18 @@ export class PnpmProjectShrinkwrapFile extends BaseProjectShrinkwrapFile {
...
       ...Object.entries(parentShrinkwrapEntry.dependencies || {}),
       ...Object.entries(parentShrinkwrapEntry.optionalDependencies || {})
-    ].filter((d) => d[0].indexOf('@rush-temp/') === -1);
+    ];
 
     const projectShrinkwrapMap: Map<string, string> = new Map();
-    for (const [name, version] of externalDependencies) {
-      this._addDependencyRecursive(projectShrinkwrapMap, name, version, parentShrinkwrapEntry);
+    for (const [name, version] of allDependencies) {
+      if (name.indexOf('@rush-temp/') < 0) {
```
iclanton, line 91:
```suggestion
      if (name.indexOf(`${RushConstants.rushTempNpmScope}/`) < 0) {
```

common/changes/@microsoft/rush/lockfile-diff_2021-11-30-23-57.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "When computing the set of projects affected by a Git change, if the change contains the lockfile and the package manager is pnpm, diff the content of the lockfile to determine which projects were affeted by the lockfile change, instead of assuming all projects are affected.",
```
iclanton, line 5:
```suggestion
      "comment": "When computing the set of projects affected by a Git change, if the change includes changes to the lockfile and the package manager is pnpm, diff the content of the lockfile to determine which projects were affected by the lockfile change, instead of assuming all projects are affected.",
```

iclanton (approved):

PR #3051, opened 2021-12-01
[rush] Fix an issue where parameter values containing spaces aren't correctly passed to global commands.

iclanton:
## Summary

Fixes https://github.com/microsoft/rushstack/issues/3027

## Details

Escapes quotemarks in args and wraps args containing spaces in quotemarks before passing them to the shell.

## How it was tested

Tested with https://github.com/iclanton/rushstack/commit/44c1948519788c7b13b21939c5402473623ef2da and running `rush my-global-command --my-string "x \"y z"`

### DIFF
--- a/apps/rush-lib/src/cli/scriptActions/GlobalScriptAction.ts
+++ b/apps/rush-lib/src/cli/scriptActions/GlobalScriptAction.ts
@@ -104,6 +104,17 @@ export class GlobalScriptAction extends BaseScriptAction {
       customParameter.appendToArgList(customParameterValues);
     }
 
+    for (let i: number = 0; i < customParameterValues.length; i++) {
+      let customParameterValue: string = customParameterValues[i];
+      customParameterValue = customParameterValue.replace(/"/g, '\\"');
+
+      if (customParameterValue.indexOf(' ') >= 0) {
+        customParameterValue = `"${customParameterValue}"`;
+      }
+
+      customParameterValues[i] = customParameterValue;
+    }
+
     let shellCommand: string = this._shellCommand;
     if (customParameterValues.length > 0) {
       shellCommand += ' ' + customParameterValues.join(' ');
--- a/common/changes/@microsoft/rush/fix-global-command-escaping_2021-12-01-00-30.json
+++ b/common/changes/@microsoft/rush/fix-global-command-escaping_2021-12-01-00-30.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix an issue where parameter values containing spaces are incorrectly passed to global scripts.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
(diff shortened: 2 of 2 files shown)


libraries/ts-command-line/src/test/__snapshots__/CommandLineParameter.test.ts.snap
```diff
@@ -556,15 +644,15 @@ a longer description
 
 Optional arguments:
   -h, --help            Show this help message and exit.
-  -c {one,two,three,default}, --choice {one,two,three,default}
+  -c {one,two,three,choice with spaces,default}, --choice {one,two,three,choice with spaces,default}
```
dmichon-msft, line 647: The help docs should really be quoting anything that needs to be quoted.

common/reviews/api/ts-command-line.api.md
```diff
@@ -30,7 +30,7 @@ export class CommandLineChoiceListParameter extends CommandLineParameter {
     constructor(definition: ICommandLineChoiceListDefinition);
     readonly alternatives: ReadonlyArray<string>;
     // @override
-    appendToArgList(argList: string[]): void;
+    appendToArgList(argList: string[], wrapValueInQuotesIfNecessary?: boolean): void;
```
dmichon-msft, line 33: I think quoting values should ultimately be the responsibility of the entity using the argument list. It's a little weird to have it be the responsibility of the parser when the thing consuming the argument list can easily quote the arguments with a single call to `.map()`.

iclanton: Fair enough.

libraries/ts-command-line/src/test/__snapshots__/CommandLineParameter.test.ts.snap
iclanton: We should fix that separately.

dmichon-msft (approved):

apps/rush-lib/src/cli/scriptActions/GlobalScriptAction.ts
```diff
@@ -104,6 +104,17 @@ export class GlobalScriptAction extends BaseScriptAction {
       customParameter.appendToArgList(customParameterValues);
     }
 
+    for (let i: number = 0; i < customParameterValues.length; i++) {
+      let customParameterValue: string = customParameterValues[i];
+      customParameterValue = customParameterValue.replace(/"/g, '\\"');
+
+      if (customParameterValue.indexOf(' ') >= 0) {
```
D4N14L, line 111: Are there any other reasons we would want to wrap the provided string in quotes (special chars?)

D4N14L (approved):

apps/rush-lib/src/cli/scriptActions/GlobalScriptAction.ts
iclanton: I'm pretty sure spaces are the only characters that matter.

PR #3057, opened 2021-12-01
[eslint-plugin] Support ESLint 8

### LINKED ISSUE #2890: Support ESLint 8.x
ESLint v8.0.0 is [released](https://eslint.org/blog/2021/10/eslint-v8.0.0-released) 🎉

It would be awesome to have official ESLint 8 support. 👊
I'm happy to help where I can of course 🙂

D4N14L:
## Summary

Updates the Rushstack repo to support and use ESLint v8 for linting.

Fixes #2890

cc @lukashass 

## Details

There are some aspects to this upgrade that deserve to be called out:
- `@rushstack/tree-pattern` is temporarily remaining on `eslint@7.30.0`. This is due to it's dependency on an older version of `@rushstack/eslint-config`/`@rushstack/eslint-plugin` since it consumes these packages as cyclic dependencies. Once the new packages are published and the cyclic dependencies are bumped, this project should be able to be updated to `8.3.0` and have the common-versions.json changes removed.
- `@rushstack/eslint-config` temporarily disables `eslint-plugin-promise` rules (of which there was only one). This is required since the package does not yet support ESLint v8 (see this issue: https://github.com/xjamundx/eslint-plugin-promise/issues/218).

## How it was tested

Build while consuming the new version of ESLint and the updated config/plugin in dependent projects.


apps/rush-lib/src/logic/taskRunner/BaseBuilder.ts
```diff
@@ -24,21 +24,21 @@ export abstract class BaseBuilder {
   /**
    * Name of the task definition.
    */
-  abstract readonly name: string;
+  public abstract readonly name: string;
```
iclanton, line 27: Why did this have to change in this PR?

iclanton:
Can you add a project that continues to test ESLint 7?

common/changes/@rushstack/eslint-config/user-danade-eslint8_2021-12-01-20-53.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/eslint-config",
+      "comment": "Temporarily disable eslint-plugin-promise until ESLint v8 support is added (https://github.com/xjamundx/eslint-plugin-promise/issues/218)",
+      "type": "patch"
```
iclanton, line 6:
```suggestion
      "type": "minor"
```

common/changes/@rushstack/heft-web-rig/user-danade-eslint8_2021-12-01-20-54.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-web-rig",
+      "comment": "Bump ESLint to v8",
+      "type": "patch"
```
iclanton, line 6:
```suggestion
      "type": "minor"
```

common/config/rush/common-versions.json
```diff
@@ -59,6 +59,11 @@
    * This design avoids unnecessary churn in this file.
    */
   "allowedAlternativeVersions": {
+    /**
+     * Temporary addition to allow for bumping ESLint in other projects. Only used by @rushstack/tree-pattern
+     */
```
iclanton, line 64: Update this comment when a test project is added.

stack/eslint-plugin/src/no-new-null.ts
```diff
@@ -54,7 +55,7 @@ const noNewNullRule: TSESLint.RuleModule<MessageIds, Options> = {
       switch ((node as TSESTree.Node).type) {
         case AST_NODE_TYPES.MethodDefinition:
           return true;
-        case AST_NODE_TYPES.ClassProperty:
+        case AST_NODE_TYPES.PropertyDefinition:
```
iclanton, line 58: This doesn't break compatibility with 6 or 7, right?

iclanton:
IIRC you mentioned that `eslint-plugin-promise` is just missing some manifest file. Can we add something in `eslint-patch` to fix that as a stopgap?

apps/rush-lib/src/logic/taskRunner/BaseBuilder.ts
D4N14L: This was caught by the rule `@typescript-eslint/explicit-member-accessibility`. Not sure why it wasn't caught before.... but it is now.

stack/eslint-plugin/src/no-new-null.ts
D4N14L: No, it shouldn't be. This type is sourced from the @typescript-eslint/ast-spec package, which in-turn is used by the parser. The type that we get back would come from `@typescript-eslint/typescript-estree`, which also uses that package as long as we keep them updated to the same version (which we do in this PR).

Here's the breaking change that necessitated this update: https://github.com/typescript-eslint/typescript-eslint/pull/3806

D4N14L:
> IIRC you mentioned that `eslint-plugin-promise` is just missing some manifest file. Can we add something in `eslint-patch` to fix that as a stopgap?

I would prefer not to go that route. The change on the eslint-plugin-promise is relatively minor, just taking a while to go through... on our side, we currently only consume one rule, and it's a pretty simple one (confusing resolve and reject callbacks). Patching for eslint to add a new feature is one thing... patching a rules package where we only consume 1 rule seems excessive. Especially since `eslint-patch` is now used by so many people: https://www.npmjs.com/package/@rushstack/eslint-patch

iclanton (approved):

D4N14L:
@iclanton I added a small test project, could you take a look to make sure that it's sufficient?

iclanton (approved):

PR #3074, opened 2021-12-06
[api-extractor-model] Remove "const" keyword before enum to use with typescript isolatedModules

samisayegh:
## Summary

When a typescript project is using [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules), referencing exported `const` enums  is not allowed.

## Details

Setting `isolatedModules` to `true` is recommended for projects transpiled with babel or [esbuild](https://esbuild.github.io/content-types/#isolated-modules). Unlike `tsc`, these tools evaluate files independently. Setting the `isolatedModules` flag e.g. forces developers to indicate whether they are exporting a type or value, since transpilers evaluating modules independently cannot infer this information.

However, the flag also alerts when a `const` enum is used. It seems that `tsc` evaluates `const enums` across modules and inlines the constants at compile. No `enum` object is created, which is problematic of independent transpilation.

I found [this article helpful](https://ncjamieson.com/dont-export-const-enums/) for understanding the effect of specifying `const`.

By removing the `const` keyword, the enums can safely be used in projects using `isolatedModules`. This change is similar to [this recent PR](https://github.com/microsoft/rushstack/pull/3060).

## How it was tested

I used `npm link` and no longer saw the error. Note also that I did not face the error with the `ReleaseTag` enum which is not using `const`.

### DIFF
--- a/common/changes/@microsoft/api-extractor-model/master_2021-12-06-21-04.json
+++ b/common/changes/@microsoft/api-extractor-model/master_2021-12-06-21-04.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor-model",
+      "comment": "remove \"const\" keyword before enums to allow use with typescript isolatedModules",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@microsoft/api-extractor-model"
+}
--- a/common/changes/@microsoft/api-extractor-model/master_2021-12-06-21-04.json
+++ b/common/changes/@microsoft/api-extractor-model/master_2021-12-06-21-04.json
@@ -2,8 +2,8 @@
   "changes": [
     {
       "packageName": "@microsoft/api-extractor-model",
-      "comment": "remove \"const\" keyword before enums to allow use with typescript isolatedModules",
-      "type": "patch"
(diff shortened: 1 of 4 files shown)


samisayegh:
When running `rush change` on the forked repo, I see the following error:

rush_change_error

Not sure what to do hmm

dmichon-msft (approved):

common/changes/@microsoft/api-extractor-model/master_2021-12-06-21-04.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor-model",
+      "comment": "remove \"const\" keyword before enums to allow use with typescript isolatedModules",
```
iclanton, line 5:
```suggestion
      "comment": "Replace const enums with conventional enums to allow for compatibility with JavaScript consumers.",
```

common/changes/@microsoft/api-extractor-model/master_2021-12-06-21-04.json
```diff
@@ -0,0 +1,10 @@
...
+      "type": "patch"
```
iclanton, line 6:
```suggestion
      "type": "minor"
```

iclanton (approved):

Issue #3073, opened 2021-12-06
[heft] Unable to separate invocations for unit tests, coverage collection, and integration tests


bryandrenner:
## Summary

When I invoke standard Jest, I'm able to distinguish between three test-running scenarios:

- **Running all unit tests, collecting code coverage metrics.** It is useful to distinguish this scenario, because it gives me fast and holistic feedback without depending on integrated external services.
- **Selectively running unit tests in the file that I'm currently authoring.** It is useful to distinguish this scenario, because it gives me very fast feedback about the portion of implementation that I'm currently focusing on.
- **Running slower integration tests.** It is useful to distinguish this scenario, because integration tests have a dedicated global setup and teardown, they don't inform unit test coverage metrics, they are generally slow, and they require distinct setup and connectivity.

Heft-Jest doesn't appear to support these scenarios, because it only supports a single Jest config file, and it does not support Jest's standard variety of CLI arguments (such as `--coverage`). This necessitates cumbersome workflows for authoring and running tests. For example, if coverage is enabled in my configuration, then I get code coverage metrics and warnings when I only wanted to run tests from a single file. If I provide a global setup and teardown, then slow, integration-oriented code executes even for unit tests.

## Details

Heft-Jest should support more standard Jest CLI options, especially `--coverage`.

I know Heft's design philosophy is to support only a single config file, so perhaps Heft-Jest could provide its own global-setup-and-teardown strategy.

## Standard questions

Please answer these questions to help us investigate your issue more quickly:

| Question | Answer |
| -------- | -------- |
| `@rushstack/heft` version? | ^0.42.2 |
| Operating system? | Mac |
| Would you consider contributing a PR? | Yes |
| Node.js version (`node -v`)? | 14.18.1 |

D4N14L:
Could you provide a list of args that you believe should be funneled through? Other than the `--coverage` arg, I'm guessing the `--config` arg would probably be something that you'd want.

Originally the `--config` arg was able to be provided via the plugin options in `heft.json`, but this was removed after discussion that this path really shouldn't need to be changed for the standard scenario. However, I can see a valid use-case for multiple configs supporting different scenarios that can be pointed to using `heft test --config ...`. @pgonzal thoughts?

[...]

PR #3076, opened 2021-12-07
[heft-jest-plugin] Update to Jest 27

### LINKED ISSUE #2825: [heft] Upgrade Jest dependency
Would be nice if heft could see and use a locally installed version of jest on a per-package basis, or if I could set the jest version from a rig spec.

D4N14L:
## Summary

Upgrades `@rushstack/heft-jest-plugin` to use Jest 27.

Fixes #2825

Takes inspiration from #3041

## Details

- Updated the transformers to implement the Jest-provided interface and fixed export format for Jest 27 compatibility
- Improved transformer cache key creation
- Updated the resolver to provide correct typings for Jest 27
- Created utils class to contain copies of non-exported methods used by Jest for resolution
- Cleaned up config file loading

## How it was tested

Building entire project since it is consumed by other Rushstack projects


heft-plugins/heft-jest-plugin/src/JestPlugin.ts
```diff
@@ -330,34 +314,12 @@ export class JestPlugin implements IHeftPlugin<IJestPluginOptions> {
         '$.moduleLoader': jestResolveMetadata,
         '$.prettierPath': jestResolveMetadata,
         '$.resolver': jestResolveMetadata,
-        '$.runner': JestPlugin._getJsonPathMetadata({
```
iclanton, line 333: These are safe to remove? Did Jest change its convention here?

iclanton:
Is this easy to address too? https://github.com/microsoft/rushstack/issues/3073

common/config/rush/browser-approved-packages.json
```diff
@@ -2,6 +2,18 @@
...
+      "allowedCategories": [ "libraries" ]
+    },
+    {
+      "name": "jest-watch-master",
+      "allowedCategories": [ "libraries" ]
+    },
+    {
+      "name": "jest-watch-select-projects",
+      "allowedCategories": [ "libraries" ]
+    },
```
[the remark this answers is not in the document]
D4N14L: Oh, thought I did put them there... oops.

heft-plugins/heft-jest-plugin/src/exports/jest-build-transform.ts
```diff
@@ -1,4 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-export * from '../jest-build-transform';
+import { BuildTransformer } from '../transformers/BuildTransformer';
```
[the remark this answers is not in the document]
D4N14L: They changed how they import the transformers and broke our current way of doing it :(

heft-plugins/heft-jest-plugin/src/JestPlugin.ts
D4N14L: It's a mix of cleanup and a change on the Jest side.

The args that were being provided that were different than what the normal `metadata` provided were unused. The other part of what's changed is that we're now looking at the property name to determine which type of resolution we use (either the generic Jest resolution, or the `jest-resolve`-provided resolution functions.

heft-plugins/heft-jest-plugin/src/HeftJestDataFile.ts
```diff
@@ -82,6 +82,11 @@ export class HeftJestDataFile {
     return JsonFile.load(jsonFilePath);
   }
 
+  public static async loadForProjectAsync(projectFolder: string): Promise<IHeftJestDataFileJson> {
+    const jsonFilePath: string = HeftJestDataFile.getConfigFilePath(projectFolder);
+    return JsonFile.loadAsync(jsonFilePath);
```
iclanton, line 87:
```suggestion
    return await JsonFile.loadAsync(jsonFilePath);
```

D4N14L:
> Is this easy to address too? #3073

The problem there is that the issue is a bit under-defined, in the sense of that I'm not quite sure what args would be required to support what they're asking for. I've left a comment on that issue, and for now I've added a `--config` arg to the Jest plugin CLI, but I'm leaning towards calling that issue "out of scope" and addressing it in a future PR after the requirements are clarified. As such, feel free to tell me to remove the `--config` arg if you think it would be better suited for that PR.

common/changes/@rushstack/heft/user-danade-Jest27_2021-12-08-03-32.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft",
+      "comment": "Remove Jest-specific CLI arguments from Heft",
```
iclanton, line 5:
```suggestion
      "comment": "Remove Jest-specific CLI arguments from Heft. These parameters have been moved to @rushstack/heft-jest-plugin.",
```

heft-plugins/heft-jest-plugin/src/transformers/BuildTransformer.ts
```diff
@@ -0,0 +1,449 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
```
iclanton, line 1: This is mostly a rename, right? What parts of this file changed?

heft-plugins/heft-jest-plugin/src/transformers/StringMockTransformer.ts
```diff
@@ -0,0 +1,22 @@
...
+ * environment.
+ */
+export class StringMockTransformer implements SyncTransformer {
+  public process(sourceText: string, sourcePath: Config.Path, options: TransformOptions): TransformedSource {
+    // Double-escape "'" and "\" characters in the filename because this is going to be serialized
+    // in a string in generated JS code, bounded by single quotes
+    const escapedSourcePath: string = sourcePath.replace(/\'/g, "\\'").replace(/\\/g, '\\\\');
+
+    // For a file called "myImage.png", this will generate a JS module that exports the literal string "myImage.png"
+    return `module.exports = '${escapedSourcePath}';`;
```
iclanton, line 20:
```suggestion
    // For a file called "myImage.png", this will generate a JS module that exports the literal string "myImage.png"
    return `module.exports = ${JSON.stringify(sourcePath)};`;
```

Any reason not to just do that?

heft-plugins/heft-jest-plugin/src/transformers/BuildTransformer.ts
D4N14L: Rename + refactor. I added the cache key methods and the async methods as well, with the only functional difference being the use of async file I/O

heft-plugins/heft-jest-plugin/src/transformers/StringMockTransformer.ts
D4N14L: I don't see why not. It should be just as valid.

heft-plugins/heft-jest-plugin/src/transformers/BuildTransformer.ts
```diff
@@ -0,0 +1,449 @@
...
+ * Implements SyncTransformer interface instead of AsyncTransformer since async is only supported
+ * in ESM, which is still considered experimental:
+ * https://github.com/facebook/jest/issues/11226#issuecomment-804449688
+ */
+export class BuildTransformer implements AsyncTransformer, SyncTransformer {
+  /**
+   * Read heft-jest-data.json, which is created by the JestPlugin.  It tells us
+   * which emitted output folder to use for Jest.
+   */
+  private static _getHeftJestDataFileJson(rootDir: string): IHeftJestDataFileJson {
```
dmichon-msft, line 63: Why a blend of free functions and private static functions? The former are better for if we ever run this through a minifier; some folks prefer the latter for casting and mocking.

dmichon-msft (commented):
Found a few things to update

heft-plugins/heft-jest-plugin/src/JestUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+  optional?: boolean;
+}
+
+// See: https://github.com/facebook/jest/blob/3093c18c428d962eb959437b322c6a5b0ae0e7a2/packages/jest-config/src/utils.ts#L58
+export const replaceRootDirInPath = (rootDir: Config.Path, filePath: Config.Path): string => {
+  if (!/^<rootDir>/.test(filePath)) {
+    return filePath;
+  }
+
+  return path.resolve(rootDir, path.normalize('./' + filePath.substring('<rootDir>'.length)));
```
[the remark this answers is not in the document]
D4N14L: It should. I was just keeping it as close to stock as possible (that call is in the code from Jest). I'll strip it down.

heft-plugins/heft-jest-plugin/src/transformers/BuildTransformer.ts
D4N14L: Sticking with the statics. I like having the class be able to implement the correct types.

heft-plugins/heft-jest-plugin/src/transformers/BuildTransformer.ts
```diff
@@ -0,0 +1,449 @@
...
+  }
+
+  /**
+   * @override
+   */
+  public getCacheKey(sourceText: string, sourcePath: Config.Path, options: TransformOptions): string {
+    const heftJestDataFile: IHeftJestDataFileJson = BuildTransformer._getHeftJestDataFileJson(
+      options.config.rootDir
+    );
+    const cacheKeyFunction: NewGetCacheKeyFunction = createCacheKeyFunction(
```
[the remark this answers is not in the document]
D4N14L: Unfortunately no, and looking at it, it also makes synchronous file I/O calls. I'll just re-implement in JestUtils

heft-plugins/heft-jest-plugin/src/transformers/BuildTransformer.ts
```diff
@@ -0,0 +1,449 @@
...
+      }
+      delayMs(2000);
+    }
+  }
+
+  private static async _waitForTranspiledFileAsync(
+    sourcePath: string,
+    transpiledPath: string
+  ): Promise<void> {
+    let stalled: boolean = false;
```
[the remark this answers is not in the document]
D4N14L: Talked offline, this is intended as-is until we can change how we call into Jest directly to run tests.

common/config/rush/pnpm-lock.yaml
```diff
@@ -442,7 +437,7 @@ importers:
       '@storybook/addon-actions': 6.3.12_d1f22ff477f871af5a90a42227cbf464
       '@storybook/addon-essentials': 6.3.12_88ce35e0c338e7352df42621a356b850
       '@storybook/addon-links': 6.3.12_react-dom@16.13.1+react@16.13.1
-      '@storybook/cli': 6.3.12_jest@25.4.0
+      '@storybook/cli': 6.3.12
```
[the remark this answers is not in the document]
D4N14L: I'll add it back for now. It doesn't do anything anyways afaik

heft-plugins/heft-jest-plugin/src/JestUtils.ts
```diff
@@ -0,0 +1,149 @@
...
+};
+
+// Adapted from @jest/create-cache-key-function to add memoization and support for async I/O.
+// See: https://github.com/facebook/jest/blob/86e64611c98dd3a6656be27dc5c342d53f8e7c30/packages/jest-create-cache-key-function/src/index.ts#L43
+const getGlobalCacheKeyAsync = async (files: string[], values: string[]): Promise<string> => {
+  const components: (string | undefined)[] = [
+    process.env.NODE_ENV,
+    process.env.BABEL_ENV,
+    ...values,
+    ...files
```
dmichon-msft, line 98: We can't safely memoize the file -> hash lookup unless something wipes the cache across executions

dmichon-msft (approved):

iclanton (approved):

PR #3078, opened 2021-12-08
Update to TypeScript 4.5

chadhietala:
## Summary

This PR updates all packages to depend on or test against TS 4.5.

One thing that I was wondering is that in all the packages that hardcode `@rushstack/heft-rig-node` as a devDep I had to explicitly override the typescript version to get the tests to pass. I'm not sure if this is the correct approach or not.

I'm also not sure what the version policy is for TypeScript version bumps so I haven't ran `rush change` yet.

## Details

Updated the deps and ran tests

## How it was tested

Locally ran the tests

### DIFF
--- a/build-tests/install-test-workspace/workspace/typescript-v3-test/package.json
+++ b/build-tests/install-test-workspace/workspace/typescript-v3-test/package.json
@@ -11,7 +11,7 @@
   "devDependencies": {
     "@rushstack/eslint-config": "*",
     "@rushstack/heft": "*",
-    "typescript": "~4.4.2",
+    "typescript": "~4.5.2",
     "tslint": "~5.20.1",
     "eslint": "~8.3.0"
   }
--- a/apps/api-extractor-model/package.json
+++ b/apps/api-extractor-model/package.json
@@ -21,9 +21,10 @@
   },
   "devDependencies": {
     "@rushstack/eslint-config": "workspace:*",
-    "@rushstack/heft": "0.42.3",
-    "@rushstack/heft-node-rig": "1.2.32",
+    "@rushstack/heft": "0.42.5",
+    "@rushstack/heft-node-rig": "1.3.0",
     "@types/heft-jest": "1.0.1",
-    "@types/node": "12.20.24"
+    "@types/node": "12.20.24",
+    "typescript": "~4.5.2"
   }
 }
--- a/apps/api-extractor/package.json
+++ b/apps/api-extractor/package.json
@@ -46,16 +46,17 @@
     "resolve": "~1.17.0",
     "semver": "~7.3.0",
     "source-map": "~0.6.1",
-    "typescript": "~4.4.2"
+    "typescript": "~4.5.2"
   },
   "devDependencies": {
     "@rushstack/eslint-config": "workspace:*",
-    "@rushstack/heft": "0.42.3",
-    "@rushstack/heft-node-rig": "1.2.32",
+    "@rushstack/heft": "0.42.5",
+    "@rushstack/heft-node-rig": "1.3.0",
     "@types/heft-jest": "1.0.1",
     "@types/lodash": "4.14.116",
     "@types/node": "12.20.24",
     "@types/resolve": "1.17.1",
-    "@types/semver": "7.3.5"
+    "@types/semver": "7.3.5",
+    "typescript": "~4.5.2"
   }
 }
--- a/apps/heft/package.json
+++ b/apps/heft/package.json
@@ -50,8 +50,8 @@
   "devDependencies": {
     "@microsoft/api-extractor": "workspace:*",
     "@rushstack/eslint-config": "workspace:*",
-    "@rushstack/heft": "0.42.3",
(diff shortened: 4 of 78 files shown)


iclanton:
The packages that have a specific version of `heft-node-rig` are projects that `heft-node-rig` directly or indirectly depends on. Including a devDep on typescript for those packages is fine. After this PR gets published, we can just update the cyclic dependency again and remove that dependency.

iclanton:
Changefiles: projects that just build with TypeScript but don't pass it on to their consumers should get a `"none"` bump type and no change description. Packages that do pass it on (like api-extractor and the rigs) should get a minor bump and a change description.

One concern about bumping the TypeScript dependency is the introduction of new syntax or keywords in the `.d.ts` files that will break compatability with older versions of the compiler. I usually build the repo both with and without the TS upgrade, collect all of the declaration files, and diff the two and see what changes. If there are breaking changes that will require consumers to upgrade to the latest TypeScript, we should hold off.

build-tests/install-test-workspace/workspace/typescript-v3-test/package.json
```diff
@@ -11,7 +11,7 @@
   "devDependencies": {
```
iclanton, line 11: This package name is different from the folder it lives in. Is this supposed to be a project that uses TS v3? @octogonz?

chadhietala:
@iclanton Let me know if the [change files](https://github.com/microsoft/rushstack/pull/3078/commits/d49ecdf983a115086669a4ef3e72de7126a0bde4) look correct or not. I diffed the output of api-extractor on `master` and my branch and they seem to be identical output.

Screen Shot 2021-12-08 at 9 43 56 AM

iclanton:
Here's a tool I just put together to diff the `.d.ts` files in two repos' projects' `lib` folders. Looks like there aren't any differences, so we should be good to move forward:

```TypeScript
import { RushConfiguration } from "@microsoft/rush-lib";
import { FileSystem, FileWriter, Text } from "@rushstack/node-core-library";
import * as fs from "fs";
import * as Diff from "diff";

function getArgValue(argName: string): string {
  const argIndex: number = process.argv.indexOf(argName);
  if (argIndex < 0) {
    throw new Error(`The ${argName} parameter is required.`);
  }

  return process.argv[argIndex + 1];
}

const repoAFolderPath: string = getArgValue("--repo-a");
const repoBFolderPath: string = getArgValue("--repo-b");
const outFolder: string = getArgValue("--out-folder");

const rushConfiguration: RushConfiguration =
  RushConfiguration.loadFromDefaultLocation({
    startingFolder: repoAFolderPath,
  });

FileSystem.ensureEmptyFolder(outFolder);

function compareDtsFiles(
  repoASubfolderPath: string,
  repoBSubfolderPath: string,
  destinationFolderPath: string
): void {
  for (const dirent of fs.readdirSync(repoASubfolderPath, {
    withFileTypes: true,
  })) {
    const sourceAPath: string = `${repoASubfolderPath}/${dirent.name}`;
    const sourceBPath: string = `${repoBSubfolderPath}/${dirent.name}`;

    const destinationPath: string = `${destinationFolderPath}/${dirent.name}`;
    if (dirent.isDirectory()) {
      compareDtsFiles(sourceAPath, sourceBPath, destinationPath);
    } else if (dirent.name.endsWith(".d.ts")) {
      const sourceA: string = Text.convertToLf(
        FileSystem.readFile(sourceAPath)
      );
      const sourceB: string = Text.convertToLf(
        FileSystem.readFile(sourceBPath)
      );

      const diff: Diff.Change[] = Diff.diffLines(sourceA, sourceB);
      const fileHasNoDifferences: boolean =
        diff.length === 1 && !diff[0].added && !diff[0].removed;

      if (!fileHasNoDifferences) {
        FileSystem.ensureFolder(destinationFolderPath);
        const diffFileWriter: FileWriter = FileWriter.open(destinationPath, {
          append: false,
        });
        for (const change of diff) {
          const lines: string[] = change.value.trimEnd().split("\n");
          let linePrefix: string;
          if (change.added) {
            linePrefix = "+ ";
          } else if (change.removed) {
            linePrefix = "- ";
          } else {
            linePrefix = "  ";
          }

          for (const line of lines) {
            diffFileWriter.write(`${linePrefix}${line}\n`);
          }

          console.log(`Wrote diff to: ${destinationPath}`);
        }

        diffFileWriter.close();
      }
    }
  }
}

for (const project of rushConfiguration.projects) {
  const repoALibFolderPath: string = `${repoAFolderPath}/${project.projectRelativeFolder}/lib`;

  if (FileSystem.exists(repoALibFolderPath)) {
    compareDtsFiles(
      repoALibFolderPath,
      `${repoBFolderPath}/${project.projectRelativeFolder}/lib`,
      `${outFolder}/${project.projectRelativeFolder}/lib`
    );
  }
}

```

iclanton (approved):

iclanton:
Here's the cyclics update: https://github.com/microsoft/rushstack/pull/3079

PR #3124, opened 2022-01-02
[rush] Add an option to not pass phase-control flag parameters on to project commands.

iclanton:
## Summary

Currently the design for flag parameters that manipulate which phases are executed conflicts with the design for caching. Using a flag to omit a phase from execution suggests that the phase may not be necessary to get a developer who has just pulled changes into their local clone ready to start work (a `_phase:test` phase excluded by a `--lite` flag for example). However, if the flag is also passed on to the `_phase:build` phase, it ends up being part of the `_phase:build`'s cache entry's hash, meaning that `_phase:build` won't be able to be restored from cache unless the cache was populated by a `rush build --lite` run.

I expect that a repo with phased commands and a remote cache will commonly run `rush build` in CI and then expect developers to run `rush build --lite` to get going in the morning on their local machines. `rush build --lite` should pull the output of the `_phase:build` phase from the cache that was populated by the CI run. This doesn't work if `_phase:build`'s cache entry IDs include the `--lite` flag.

## Details

To address this issue, I propose adding a `"doNotIncludeInProjectCommandLine"` property to `"flag"` parameters in `common/config/rush/command-line.json` that prevents the parameter from being passed on to projects' individual script executions. This property is only allowed for flags that control which phases are executed.

## How it was tested

TBD.

### DIFF
--- a/apps/rush-lib/src/api/CommandLineConfiguration.ts
+++ b/apps/rush-lib/src/api/CommandLineConfiguration.ts
@@ -329,6 +329,20 @@ export class CommandLineConfiguration {
           case 'flag': {
             const addPhasesToCommandSet: Set<string> = new Set<string>();
 
+            if (
+              normalizedParameter.doNotIncludeInProjectCommandLine &&
+              !normalizedParameter.addPhasesToCommand?.length &&
+              !normalizedParameter.skipPhasesForCommand?.length
+            ) {
+              throw new Error(
+                `In ${RushConstants.commandLineFilename}, the parameter "${normalizedParameter.longName}" ` +
+                  'has the "doNotIncludeInProjectCommandLine" property set to true, but does not specify ' +
+                  'any phases in the "addPhasesToCommand" or "skipPhasesForCommand" properties. The ' +
+                  '"doNotIncludeInProjectCommandLine" parameter only applies to flag parameters that ' +
+                  'control phase execution.'
+              );
+            }
+
             if (normalizedParameter.addPhasesToCommand) {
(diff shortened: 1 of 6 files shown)


elliot-nelson:
@iclanton Thanks for this PR, I'll respond here (to the PR and your question in the ticket).

I'm still working through the use case here, but in general, _no_ flags are passed to a phase unless that phase is explicitly listed in the `associatedPhases` property (or at least, that's how I was envisioning it).

So imagine in `command-line.json`:

```
{
    "parameterKind": "flag",
    "longName": "--prod",

    "associatedCommands": ["build"],

    "associatedPhases": ["_phase:compile", "_phase:deploy"],
    "addPhasesToCommand": ["_phase:deploy", "_phase:tweet"],
    "skipPhasesForCommand": ["_phase:test"]
}
```

The flag `--prod` is going to:

(1) Add phases `deploy` and `tweet` to whatever command you run (`build` in this case).
(2) Skip phase `test` in whatever command you run (`build` in this case).
(3) Add the flag `--prod` to the command line for the `compile` and `deploy` phases.

The flag `--prod` is _not_ added to the `tweet` phase, for example, because it is not in the `associatedPhases` list.  (Despite the fact we are running `--build`, the `associatedCommands` property is really only controlling the help CLI that decides whether `rush build --prod` is a valid command, not whether to add `--prod` to each command we run.)

Does that make sense?

iclanton:
Ah I misinterpreted the design. Real fix is here: https://github.com/microsoft/rushstack/pull/3126

PR #3126, opened 2022-01-04
[rush] Fix an issue with parameter associations with phases.

iclanton:
## Summary

There is a mistake in the current implementation of phase/parameter associations.

## Details

The design calls for parameters to be associated with commands. If a parameter is specified when a phased command is invoked, the parameter is passed only to the phases it is associated with. The current implementation implicitly associates parameters with all commands with phases they're directly or indirectly associated with, and their values are passed on to all phases.

## How it was tested

Tested in https://github.com/elliot-nelson/rush-phased-builds-example and this repo.

### DIFF
--- a/apps/rush-lib/src/api/CommandLineConfiguration.ts
+++ b/apps/rush-lib/src/api/CommandLineConfiguration.ts
@@ -308,6 +308,8 @@ export class CommandLineConfiguration {
 
         this.parameters.push(normalizedParameter);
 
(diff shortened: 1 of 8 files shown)


elliot-nelson (approved):

D4N14L (approved):

dmichon-msft (approved):
I'm trying to switch everything to key the maps off of the IPhase object instead of the name, but I can update my PR again to compensate for the logic change here.

PR #3134, opened 2022-01-06
Fix some instances of "confusable" characters.

iclanton:
VSCode called out a character ([here](https://github.com/microsoft/rushstack/blob/master/libraries/debug-certificate-manager/src/CertificateManager.ts#L426)) that is "ambiguous," which is an actual bug. I wrote a little script to replace those in a repo: https://gist.github.com/iclanton/6956d1a0e98737393d70c5d7f70484ec

This PR replaces them across the repo.


D4N14L (approved):
Seems like something that would be a good pre-commit hook.

octogonz (approved):

PR #3140, opened 2022-01-07
[rush-lib] Fix regression in bulk commands with unsafe selectors

dmichon-msft:
## Summary
Fixes behavior of the unsafe `--only` `--impacted-by` and `--impacted-by-except` selection parameters with respect to the task selector. A regression was introduced during the phased command implementation that project dependencies were forcibly included.

## Details
Restores the filtering of the project graph to elide projects that are not included by the selection.

## How it was tested
Verified `node ./apps/rush-lib/lib/start.js build -o module-minifier-plugin` selects exactly one project

### DIFF
--- a/apps/rush-lib/src/logic/ProjectTaskSelector.ts
+++ b/apps/rush-lib/src/logic/ProjectTaskSelector.ts
@@ -112,7 +112,7 @@ export class ProjectTaskSelector {
         );
         if (commandToRun === undefined && !phase.ignoreMissingScript) {
           throw new Error(
-            `The project [${project.packageName}] does not define a '${phase.name}' command in the 'scripts' section of its package.json`
+            `The project '${project.packageName}' does not define a '${phase.name}' command in the 'scripts' section of its package.json`
           );
         }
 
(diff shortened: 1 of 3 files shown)


apps/rush-lib/src/logic/ProjectTaskSelector.ts
```diff
@@ -70,84 +70,94 @@ export class ProjectTaskSelector {
...
+        const commandToRun: string | undefined = ProjectTaskSelector.getScriptToRun(
           project,
-          dependencyPhase,
-          taskCollection
+          phase.name,
+          customParameterValues
         );
+        if (commandToRun === undefined && !phase.ignoreMissingScript) {
+          throw new Error(
+            `The project [${project.packageName}] does not define a '${phase.name}' command in the 'scripts' section of its package.json`
```
D4N14L, line 115: The `[ ... ]` square brackets are a bit odd here. I think we normally just put the package name in other locations.

D4N14L (approved):

apps/rush-lib/src/logic/ProjectTaskSelector.ts
iclanton:
```suggestion
            `The project '${project.packageName}' does not define a '${phase.name}' command in the 'scripts' section of its package.json`
```

iclanton (approved):

PR #3145, opened 2022-01-07
Convert the rushstack repo to a phased build.

iclanton:
This change converts the repo to build with two phases: `_phase:build` and `_phase:test`.

On my machine, `rush rebuild` goes from 5:45 without this change to 4:30 with this change.

### DIFF
--- a/apps/rush/package.json
+++ b/apps/rush/package.json
@@ -24,8 +24,10 @@
   "engineStrict": true,
   "homepage": "https://rushjs.io",
   "scripts": {
-    "build": "heft test --clean",
-    "start": "node lib/start-dev-docs.js"
+    "build": "heft build --clean",
+    "start": "node lib/start-dev-docs.js",
+    "_phase:build": "heft build --clean",
+    "_phase:test": "heft test --no-build"
   },
   "bin": {
     "rush": "./bin/rush",
--- a/common/config/rush/command-line.json
+++ b/common/config/rush/command-line.json
@@ -11,6 +11,17 @@
    * "rush my-global-command --help".
    */
   "commands": [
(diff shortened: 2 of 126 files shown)


common/config/rush/command-line.json
```diff
@@ -11,6 +11,17 @@
    * "rush my-global-command --help".
    */
   "commands": [
+    {
+      "commandKind": "phased",
+      "name": "build",
+      "summary": "Phased build",
+      "description": "Ditto",
```
iclanton, line 18: Update.

common/config/rush/rush-plugins.json
```diff
@@ -0,0 +1,25 @@
+/**
```
D4N14L, line 1: Is there a template for this file in the `rush init` files?

D4N14L (approved):

apps/rush/package.json
```diff
@@ -24,8 +24,10 @@
   "engineStrict": true,
   "homepage": "https://rushjs.io",
   "scripts": {
-    "build": "heft test --clean",
-    "start": "node lib/start-dev-docs.js"
+    "build": "heft build --clean",
+    "start": "node lib/start-dev-docs.js",
```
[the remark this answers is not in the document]
iclanton: It just prints a note describing how to debug Rush.

common/config/rush/rush-plugins.json
iclanton: Yeah https://github.com/microsoft/rushstack/blob/master/apps/rush-lib/assets/rush-init/common/config/rush/rush-plugins.json

PR #3170, opened 2022-01-19
[module-minifier-plugin] Async import compression

dmichon-msft:
## Summary
Adds an option `compressAsyncImports: true` to `ModuleMinifierPlugin` that, if specified, reduces the API call size of webpack async import expressions by combining all chunk imports and the final module import into a single call to the webpack runtime. Also creates a lookup table of frequently used chunk sets to further reduce code size.

## Details
Converts expressions like `Promise.resolve([__webpack_require__.e(1), __webpack_require__.e(2)]).then(__webpack_require__.bind(23))` to the more concise `__webpack_require__.ee([1,2 ], 23)`

## How it was tested
Build test project `localization-plugin-test-02`
Linking inside a private repo with a large quantity of async imports.

### DIFF
--- a/webpack/module-minifier-plugin/src/AsyncImportCompressionPlugin.ts
+++ b/webpack/module-minifier-plugin/src/AsyncImportCompressionPlugin.ts
@@ -1,0 +1,277 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import webpack, { Compiler, Plugin } from 'webpack';
+import { ReplaceSource } from 'webpack-sources';
+import { Tapable, TapOptions } from 'tapable';
+
+import Template from 'webpack/lib/Template';
+
+import { STAGE_AFTER } from './Constants';
+import {
+  IExtendedModule,
+  IModuleMinifierPluginHooks,
+  IPostProcessFragmentContext
+} from './ModuleMinifierPlugin.types';
+
+const PLUGIN_NAME: 'AsyncImportCompressionPlugin' = 'AsyncImportCompressionPlugin';
+
+const TAP_AFTER: TapOptions<'sync'> = {
+  name: PLUGIN_NAME,
+  stage: STAGE_AFTER
+};
+
+const ASYNC_IMPORT_PREFIX: '__IMPORT_ASYNC' = '__IMPORT_ASYNC';
+const ASYNC_IMPORT_REGEX: RegExp = /__IMPORT_ASYNC[^\)]+/g;
+
+declare class WebpackImportDependency extends webpack.compilation.Dependency {
+  public module: webpack.compilation.Module;
+  public block: {
+    chunkGroup: webpack.compilation.ChunkGroup;
+    range: [number, number];
+  };
+}
+
+interface IAsyncImportMetadata {
+  chunkCount: number;
+  chunkIds: number[];
+  count: number;
+  index: number;
+}
+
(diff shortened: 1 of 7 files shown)


webpack/module-minifier-plugin/src/PortableMinifierIdsPlugin.ts
```diff
@@ -28,7 +29,7 @@ const TAP_AFTER: TapOptions<'sync'> = {
 };
 
 const STABLE_MODULE_ID_PREFIX: '__MODULEID_SHA_' = '__MODULEID_SHA_';
-const STABLE_MODULE_ID_REGEX: RegExp = /['"]?(__MODULEID_SHA_[0-9a-f]+)['"]?/g;
+const STABLE_MODULE_ID_REGEX: RegExp = /(?<!C)['"]?(__MODULEID_SHA_[0-9a-f]+)['"]?/g;
```
iclanton, line 32: Can you leave a comment describing what this matches?

webpack/module-minifier-plugin/src/PortableMinifierIdsPlugin.ts
```diff
@@ -65,29 +66,36 @@ export class PortableMinifierModuleIdsPlugin implements Plugin {
 
     const stableIdToFinalId: Map<string | number, string | number> = new Map();
 
-    this._minifierHooks.finalModuleId.tap(PLUGIN_NAME, (id: string | number | undefined) => {
+    const { finalModuleId } = this._minifierHooks;
+
+    finalModuleId.tap(PLUGIN_NAME, (id: string | number | undefined) => {
```
iclanton, line 71: This is less readable.

webpack/module-minifier-plugin/src/AsyncImportCompressionPlugin.ts
```diff
@@ -0,0 +1,277 @@
...
+  module: webpack.compilation.Module,
+  originModule: webpack.compilation.Module
+): string {
+  const exportsType: string | undefined = module.buildMeta?.exportsType;
+  const strict: boolean | undefined = originModule.buildMeta?.strictHarmonyModule;
+
+  if (exportsType === 'namespace') {
+    return '';
+  } else if (exportsType === 'named') {
+    return ',3';
```
iclanton, line 67: Where do these numbers come from? Can you stick them in consts with meaningful names?

webpack/module-minifier-plugin/src/AsyncImportCompressionPlugin.ts
```diff
@@ -0,0 +1,277 @@
...
+  for (const chunkGroup of chunk.groupsIterable) {
+    if (chunkGroup.getNumberOfChildren() > 0) {
+      return true;
+    }
+  }
+  return false;
+}
+
+/**
+ * Plugin that replaces `Promise.all([__webpack_require__.e(1), __webpack_require__,e(12)]).then(__webpack_require__.t.bind(123,7))`
```
iclanton, line 85:
```suggestion
 * Plugin that replaces `Promise.all([__webpack_require__.e(1), __webpack_require__.e(12)]).then(__webpack_require__.t.bind(123,7))`
```

iclanton (approved):

webpack/module-minifier-plugin/src/AsyncImportCompressionPlugin.ts
dmichon-msft: They come from webpack's source code. I'll add some comments or at least a reference.

PR #3177, opened 2022-01-20
[package-deps-hash] Reduce minimum git version to 2.20.0, fix working tree state computation

### LINKED ISSUE #3109: [rush] Git version requirements are a bit high?
So, I'm working on my company's build pipeline right now, and I've noticed one unfortunate thing - official docker `node` images, which I think are most likely to be used with `rush`, all have git version `2.11` installed rather than anything later. After some digging you can also use `node:x-buster` images which contain git `2.20`.
Since `rush` wants `2.30`, I'd need to install it into the image manually, and with images based on Debian rather than Ubuntu it's not a straightforward process (I think I need to build from source? Maybe I'll find a simpler way, but it doesn't google easily).

While I understand that it's not entirely rush's problem, I do wonder if minimal version _really_ needs to be that high?

### LINKED ISSUE #3163: [rush] Error: git hash-object exited with status 128
## Summary

When running `rush build` I'm encountering the following error:
```
Error calculating the state of the repo. (inner error: Error: git hash-object exited with status 128: fatal: could not open 'libraries/*****/src/****.ts' for reading: No such file or directory
). Continuing without diffing files.
```
This results in a full rebuild each time `rush build` is called.

## Repro steps

Unfortunately specific repro steps are a bit thin given I'm not sure which exact recent change caused the issue. 
I do know it only started occurring after a recent Rush version update.

1. Set **Rush** to version `5.59.2` and **pnpm** to `6.25.1`
2. Run `rush build`

  **Expected result:**  
Cached build to occur.

  **Actual result:**
Upon first package (whichever it is) receive WARN error:
```
Error calculating the state of the repo. (inner error: Error: git hash-object exited with status 128: fatal: could not open 'libraries/*****/src/****.ts' for reading: No such file or directory
). Continuing without diffing files.
```

## Details

Root cause:
https://github.com/microsoft/rushstack/blob/3a872ed093aaa4a56d19b4b0a96cc44ab88f3e40/libraries/package-deps-hash/src/getRepoState.ts#L222-L225

Similar historic issues:
https://github.com/microsoft/rushstack/issues/2007
https://github.com/microsoft/rushstack/pull/2033 

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.59.2 |
| `rushVersion` from rush.json? | 5.59.2 |
| Operating system? |  Windows  |
| Would you consider contributing a PR? | Yes   |
| Node.js version (`node -v`)? | 16.13.0 |

dmichon-msft:
## Summary
Fixes #3163 
Fixes #3109

## Details
Reduces the minimum version of Git required by `getRepoState` to `2.20.0` by removing `--merge-base` from `getRepoChanges` and documenting the behavior change.

Fixes a bug in `getRepoState` with regards to parsing the output of `git status`. Added unit tests for `parseGitStatus`.

## How it was tested
New unit tests for `parseGitStatus` covering a range of statuses.
Verified that `node ./apps/rush-lib/lib/start.js build -o git:HEAD` correctly handles a file being renamed in the index.

### DIFF
--- a/libraries/package-deps-hash/src/test/getRepoState.test.ts
+++ b/libraries/package-deps-hash/src/test/getRepoState.test.ts
@@ -3,7 +3,7 @@
 
 import { parseGitStatus, parseGitVersion } from '../getRepoState';
 
-describe('parseGitVersion', () => {
+describe(parseGitVersion.name, () => {
   it('Can parse valid git version responses', () => {
     expect(parseGitVersion('git version 2.30.2.windows.1')).toEqual({
       major: 2,
@@ -38,7 +38,7 @@ describe('parseGitVersion', () => {
   });
 });
 
-describe('parseGitStatus', () => {
+describe(parseGitStatus.name, () => {
   it('Finds index entries', () => {
     const files: string[] = [`A.ts`, `B.ts`, `C.ts`];
     const input: string = [`A  ${files[0]}`, `D  ${files[1]}`, `M  ${files[2]}`, ''].join('\0');
--- a/libraries/package-deps-hash/src/test/getRepoState.test.ts
+++ b/libraries/package-deps-hash/src/test/getRepoState.test.ts
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-import { parseGitVersion } from '../getRepoState';
+import { parseGitStatus, parseGitVersion } from '../getRepoState';
 
-describe('getRepoState', () => {
+describe('parseGitVersion', () => {
   it('Can parse valid git version responses', () => {
     expect(parseGitVersion('git version 2.30.2.windows.1')).toEqual({
       major: 2,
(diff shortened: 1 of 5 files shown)


D4N14L (approved):

libraries/package-deps-hash/src/test/getRepoState.test.ts
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-import { parseGitVersion } from '../getRepoState';
+import { parseGitStatus, parseGitVersion } from '../getRepoState';
 
-describe('getRepoState', () => {
+describe('parseGitVersion', () => {
```
iclanton, line 6:
```suggestion
describe(parseGitVersion.name, () => {
```

libraries/package-deps-hash/src/test/getRepoState.test.ts
```diff
@@ -37,3 +37,53 @@ describe('getRepoState', () => {
     );
   });
 });
+
+describe('parseGitStatus', () => {
```
iclanton, line 41:
```suggestion
describe(parseGitStatus.name, () => {
```

PR #3186, opened 2022-01-24
[rush-lib] Add `"watchOptions"` with `watchPhases` to "phased" command kind

dmichon-msft:
## Summary
Adds a new property `"watchOptions"` with a subproperty `"watchPhases"` to `"commandKind": "phased"` in `command-line.json`. If `watchOptions` is specified and `watchPhases` is non-empty, then after the initial command execution, the command will boot up a file watcher and, upon detecting changes, run the phases listed in `"watchPhases"` for all affected projects.

## Details
The prior `"watchForChanges": true` option for `"commandKind": "bulk"` is now implemented by setting `watchPhases` to `phases` during command translation and `alwaysWatch` to `true`, or an empty set if `false`.

This logic change now treats the watch mode portion of command execution as a separate step after the initial execution. Build cache writes are allowed during the initial execution (since it should run all projects and is therefore "safe") but disabled during watch execution.

The expected usage is that the `"phases"` field is used to specify the set of build phases that need to be executed before any arbitrary changed project can be rebuilt, whereas the set in `"watchPhases"` are the phases used to validate a given local change. The example below runs `_phase:build` during the initial execution, then `_phase:build` and `_phase:test` in response to changes, so that a subsequent local change will get fully validated but the initial execution does the minimal necessary setup (and ideally pulls entirely from cache).

## How it was tested
Added the following to `command-line.json` temporarily:
```jsonc
    {
      "commandKind": "phased",
      "name": "start",
      "summary": "Build all projects that haven't been built or have changed since they were last built.",
      "description": "Build all projects that haven't been built or have changed since they were last built.",
      "safeForSimultaneousRushProcesses": false,

      "enableParallelism": true,
      "incremental": true,
      // Initial execution only uses the build phase so that all dependencies of watch phases have been built
      "phases": ["_phase:build"],
      "watchOptions": {
        // Act as though `--watch` is always passed. If false, adds support for passing `--watch`.
        "alwaysWatch": true
        // During watch recompilation run both build and test for affected projects
        "watchPhases": ["_phase:build", "_phase:test"]
      }
    },
```
Then ran `node ./apps/rush-lib/lib/start.js start -v` and verified the initial build of all projects (only build phase), followed by watch mode rebuilds of modified projects (build and test phases) in response to local changes.

### DIFF
--- a/apps/rush-lib/src/api/CommandLineConfiguration.ts
+++ b/apps/rush-lib/src/api/CommandLineConfiguration.ts
@@ -62,10 +62,10 @@ export interface IPhasedCommand extends IPhasedCommandWithoutPhasesJson, IComman
    * was not explicitly defined in the command-line.json file.
    */
   isSynthetic: boolean;
-  watchForChanges?: boolean;
   disableBuildCache?: boolean;
 
   phases: Set<IPhase>;
+  watchPhases: Set<IPhase>;
 }
 
 export interface IGlobalCommand extends IGlobalCommandJson, ICommandWithParameters {}
@@ -242,12 +242,14 @@ export class CommandLineConfiguration {
         switch (command.commandKind) {
           case RushConstants.phasedCommandKind: {
             const commandPhases: Set<IPhase> = new Set();
+            const watchPhases: Set<IPhase> = new Set();
 
             normalizedCommand = {
               ...command,
               isSynthetic: false,
               associatedParameters: new Set<Parameter>(),
-              phases: commandPhases
+              phases: commandPhases,
+              watchPhases
             };
 
             for (const phaseName of command.phases) {
@@ -272,6 +274,23 @@ export class CommandLineConfiguration {
 
               for (const dependency of phase.phaseDependencies.upstream) {
                 commandPhases.add(dependency);
+              }
+            }
+
+            const { watchOptions } = command;
+
+            if (watchOptions) {
+              // No implicit phase dependency expansion for watch mode.
+              for (const phaseName of watchOptions.watchPhases) {
+                const phase: IPhase | undefined = this.phases.get(phaseName);
+                if (!phase) {
+                  throw new Error(
+                    `In ${RushConstants.commandLineFilename}, in the "watchPhases" property of the ` +
+                      `"${normalizedCommand.name}" command, the phase "${phaseName}" does not exist.`
+                  );
+                }
+
+                watchPhases.add(phase);
               }
             }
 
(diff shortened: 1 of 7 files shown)


apps/rush-lib/src/schemas/command-line.schema.json
```diff
@@ -187,6 +187,28 @@
...
               }
+            },
+            "watchOptions": {
+              "title": "Watch Options",
+              "description": "Controls the file watching behavior of this command. If not specified, this command does not watch files.",
+              "type": "object",
+              "additionalProperties": false,
+              "required": ["watchForChanges", "watchPhases"],
+              "properties": {
+                "watchForChanges": {
```
apostolisms, line 198: do you really need a watchForChanges when you already define a watchOptions: {} ?

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -62,10 +62,11 @@ export interface IPhasedCommand extends IPhasedCommandWithoutPhasesJson, IComman
    * was not explicitly defined in the command-line.json file.
    */
   isSynthetic: boolean;
-  watchForChanges?: boolean;
   disableBuildCache?: boolean;
 
   phases: Set<IPhase>;
+
+  watchPhases: Set<IPhase>;
```
iclanton, line 69:
```suggestion
  watchPhases: Set<IPhase>;
```

apps/rush-lib/src/schemas/command-line.schema.json
iclanton: I like having an explicit "this is a watch command" setting.

apps/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -275,6 +277,23 @@ export class CommandLineConfiguration {
               }
             }
 
+            const { watchOptions } = command;
+
+            if (watchOptions) {
```
iclanton, line 282:
```suggestion
            if (watchOptions?.watchForChanges) {
```

apps/rush-lib/src/schemas/command-line.schema.json
```diff
@@ -187,6 +187,28 @@
...
+                  "title": "Watch For Changes",
+                  "description": "Indicates that this command will watch for changes after the initial execution.",
+                  "enum": [true]
```
iclanton, line 201: Does this implicitly make the type `boolean`? Should we allow someone to say `false` for whatever reason?

Maybe this should be a boolean in the schema and if they say `false`, we throw a more meaningful error message than "this doesn't match the schema?"

dmichon-msft: I'd rather drop the property altogether, since is it redundant. The presence of `watchOptions` makes it a watch action (what's the point of defining `watchOptions` for an action that doesn't watch?). Making it only accept `true` was a compromise with the request to include the property.

apps/rush-lib/src/api/CommandLineConfiguration.ts
dmichon-msft: No longer relevant.

apps/rush-lib/src/schemas/command-line.schema.json
dmichon-msft: Switched to the `alwaysWatch` option, which, when `false` allows a `--watch` CLI flag to enable the watcher (provided that `watchPhases` is non-empty).

iclanton (approved):

PR #3188, opened 2022-01-25
Update cyclic dependencies.

iclanton:


common/config/rush/command-line.json
```diff
@@ -14,12 +14,6 @@
     {
       "commandKind": "phased",
       "name": "build",
-      "summary": "Build all projects that haven't been built, or have changed since they were last built.",
```
D4N14L, line 17: No longer necessary?

D4N14L (approved):

common/config/rush/command-line.json
iclanton: Yeah, it gets missing properties for "build" and "rebuild" from the default. Phased commands broke this, and then it was fixed in 5.61.1

PR #3193, opened 2022-01-27
[rush-lib] Optimizations for interacting with `tar` utility

dmichon-msft:
## Summary
Passes input file paths via `stdin` instead of creating a temporary file.
Use multiple parallel async file system operations when scanning for symbolic links.
Create a separate log file for each phase in a project that writes to the build cache.
Modify local TAR generation to write to a temp file, then move it to the build cache, to ensure that the cache doesn't get polluted with partial entries.

## Details
Leverages`Async.forEachAsync` with a queue to scan output files.
Uses the tarball name as part of the log file path, so that each phase gets its own log file.
Logs the list of files to cache to the tar log.

## How it was tested
Local Rush builds, writing to and reading from the build cache.

### DIFF
--- a/apps/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
+++ b/apps/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
@@ -268,7 +268,9 @@ export class ProjectBuildCache {
 
     const tarUtility: TarExecutable | undefined = await ProjectBuildCache._tryGetTarUtility(terminal);
     if (tarUtility) {
-      const tempLocalCacheEntryPath: string = this._localBuildCacheProvider.getCacheEntryPath(cacheId);
+      const finalLocalCacheEntryPath: string = this._localBuildCacheProvider.getCacheEntryPath(cacheId);
+      // Derive the temp file from the destination path to ensure they are on the same volume
+      const tempLocalCacheEntryPath: string = `${finalLocalCacheEntryPath}.temp`;
       const logFilePath: string = this._getTarLogFilePath();
       const tarExitCode: number = await tarUtility.tryCreateArchiveFromProjectPathsAsync({
         archivePath: tempLocalCacheEntryPath,
@@ -276,8 +278,15 @@ export class ProjectBuildCache {
         project: this._project,
         logFilePath
       });
+
       if (tarExitCode === 0) {
-        localCacheEntryPath = tempLocalCacheEntryPath;
+        // Move after the archive is finished so that if the process is interrupted we aren't left with an invalid file
+        await FileSystem.moveAsync({
+          sourcePath: tempLocalCacheEntryPath,
+          destinationPath: finalLocalCacheEntryPath,
+          overwrite: true
+        });
+        localCacheEntryPath = finalLocalCacheEntryPath;
       } else {
         terminal.writeWarningLine(
           `"tar" exited with code ${tarExitCode} while attempting to create the cache entry. ` +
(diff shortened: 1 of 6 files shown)


iclanton:
This was tested on Windows, common Linux distros, and macOS?

dmichon-msft:
Verified on Windows and WSL Ubuntu; needs more testing on Mac OSX. Note that the `--dereference` parameter by design means that there will be a subtle difference in the output between replaying from cache and a fresh build if the output contains symbolic links, particularly in the event that a tool reading the output reads links or uses `realPath`. The `--dereference` parameter is akin to `--preserveSymlinks` in NodeJS and treats the link as a real folder.

apps/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
```diff
@@ -272,10 +265,11 @@ export class ProjectBuildCache {
       const logFilePath: string = this._getTarLogFilePath();
       const tarExitCode: number = await tarUtility.tryCreateArchiveFromProjectPathsAsync({
         archivePath: tempLocalCacheEntryPath,
-        paths: filesToCache.outputFilePaths,
+        paths: foldersToCache,
```
iclanton, line 268: The `paths` property should probably get renamed to make it clear that it's a list of folder names.

apps/rush-lib/src/utilities/TarExecutable.ts
```diff
@@ -143,6 +142,8 @@ export class TarExecutable {
         `Start time: ${new Date().toString()}`,
         `Invoking "${this._tarExecutablePath} ${args.join(' ')}"`,
         '',
+        `======= BEGIN PROCESS INPUT ======`,
+        input || '',
```
iclanton, line 146:
```suggestion
        input || '',
        '======== END PROCESS INPUT ========',
```

apps/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
dmichon-msft: It can be folders or files, the Tar exe doesn't care. It will walk folders, or write files.

apps/rush-lib/src/utilities/TarExecutable.ts
```diff
@@ -89,27 +89,26 @@ export class TarExecutable {
         archivePath,
         // [Windows bsdtar 3.3.2] -z, -j, -J, --lzma  Compress archive with gzip/bzip2/xz/lzma
         '-z',
-        // [Windows bsdtar 3.3.2] -C <dir>  Change to <dir> before processing remaining files
-        '-C',
-        projectFolderPath,
+        // [Windows bsdtar 3.3.2] Dereference symbolic links and store the underlying content
+        '--dereference',
```
iclanton, line 93: We want to error if someone tries to cache a symlink, not just dereference it.

apps/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
```diff
@@ -358,80 +366,86 @@ export class ProjectBuildCache {
...
-      }
-    }
+  private _getTarLogFilePath(): string {
+    return path.join(this._project.projectRushTempFolder, `${this._cacheId}.log`);
   }
 
-  private _getTarLogFilePath(): string {
-    return path.join(this._project.projectRushTempFolder, 'build-cache-tar.log');
+  private _getTempTarFilePath(): string {
+    return path.join(this._project.projectRushTempFolder, `${this._cacheId}.tgz`);
```
iclanton, line 448: What if a user's global cache is on a different volume than the repo? Moving this will potentially take some time.

Can we instead append `-temp_${Date.now()}` or something to the final cache file path?

iclanton (approved):

PR #3194, opened 2022-01-27
[rush-lib]: getChangedFiles should not only show added files

rccoder:
## Summary

`getChangedFiles` should not only show added files

## Details

delete `--diff-filter=A` params

## How it was tested

Verified that rush build still works.


iclanton:
This API isn't exposed on `rush-lib`'s public surface area. I'm wondering what prompted this fix.

This change would likely break `rush change`, which is the only code inside `rush-lib` that uses this function.

PR #3203, opened 2022-02-04
Keep original version when publishing prereleases

### LINKED ISSUE #3201: [rush] --prerelease-name bumps the packages making semver incompatible with prerelease versions
`--prerelease-name` is increasing the semver with a patch before adding the prerelease and it breaks selectors like `^1.0.0` where a prerelease version `1.0.1-something` isn't assignable to it

isc30:
When `--prerelease-name` is specified for `rush publish`, it will bump the packages before applying the prerelease suffix.
This leads to edge cases where a peerDependency selector doesn't satisfy the prereleases anymore:

 `1.0.1-something` is **not** assignable to `^1.0.0`
 `1.0.0-something` **is** assignable to `^1.0.0`

In order to solve this and make prerelease packages more usable, prereleases must not bump version numbers.

closes #3201


isc30:
just realized that `--sufix` doesn't bump the versions, no need for this

PR #3213, opened 2022-02-07
[rush] lockstep version policy with no nextBump doesn't update downstream dependencies

isc30:
Hotfixes https://github.com/microsoft/rushstack/pull/3059

> this issue is a blocker for us

When using a `lockstep` version policy with no `nextBump`, the downstream dependencies weren't affected.

## Added extra unit test covering the specific edge case
- project `h` and `i` use a `lockstep` version policy with no `nextBump`
- project `j` depends on `i`
- project `h` gets a changefile
- run `rush publish --apply`

### Expected
- the changefile to `h` causes the `lockstep` version policy to bump
- project `i` is also bumped due to the version policy
- project `j` is a downstream dependency of `i`, so it gets a patch update too

### Actual
- the changefile to `h` causes the `lockstep` version policy to bump
- project `i` is also bumped due to the version policy
- project `j` is not affected in any way


isc30:
test are failing in the pipelines but work locally, i need some assistance with that

iclanton:
Shouldn't `i` and `j` also get bumped in [this](https://github.com/microsoft/rushstack/blob/a5c7c3c2da57a43655db75f5c6d8bdc1656d3fe8/apps/rush-lib/src/logic/test/PublishUtilities.test.ts#L64-L80) test?

iclanton:
I pulled your branch and poked around a bit with the tests, and it seems like there are some inconsistencies in what gets bumped.

isc30:
hmm okay

> Shouldn't `i` and `j` also get bumped in [this](https://github.com/microsoft/rushstack/blob/a5c7c3c2da57a43655db75f5c6d8bdc1656d3fe8/apps/rush-lib/src/logic/test/PublishUtilities.test.ts#L64-L80) test?

true, it should list all as dependency bumps.
I will investigate, thanks!

isc30:
> Shouldn't `i` and `j` also get bumped in [this](https://github.com/microsoft/rushstack/blob/a5c7c3c2da57a43655db75f5c6d8bdc1656d3fe8/apps/rush-lib/src/logic/test/PublishUtilities.test.ts#L64-L80) test?

After some investigation, it seems that `j` shouldn't be bumped. In that specific test, `h` gets a `dependency` change, which also triggers a dependency change for `i`.

Dependency changes don't propagate up since they don't change the version, so `j` isn't touched (and that's ok since `i` will be `1.0.0` and that matches the version selector for `j`)

see https://github.com/microsoft/rushstack/blob/master/apps/rush-lib/src/logic/PublishUtilities.ts#L503

# The PR is ready to be reviewed again, thanks!

apps/rush-lib/src/logic/PublishUtilities.ts
```diff
@@ -72,17 +71,57 @@ export class PublishUtilities {
...
+      // For each requested package change, ensure downstream dependencies are also updated.
+      allChanges.packageChanges.forEach((change, packageName) => {
+        hasChanges ||= PublishUtilities._updateDownstreamDependencies(
+          change,
+          allChanges,
+          allPackages,
+          rushConfiguration,
+          prereleaseToken,
+          projectsToExclude
+        );
```
iclanton, line 89:
```suggestion
        const changeHasDownstreamChanges: boolean = PublishUtilities._updateDownstreamDependencies(
          change,
          allChanges,
          allPackages,
          rushConfiguration,
          prereleaseToken,
          projectsToExclude
        );
        hasChanges = hasChanges || changeHasDownstreamChanges;
```

```TypeScript
function y(): boolean {
  return true;
}
let x: boolean = false;
x ||= y();
```

compiles to 

```JavaScript
function y() {
    return true;
}
var x = false;
x || (x = y());
```

which isn't what you want here.

apps/rush-lib/src/logic/PublishUtilities.ts
```diff
@@ -72,17 +71,57 @@ export class PublishUtilities {
...
+
+        if (hasChanged) {
+          console.log(
+            `${EOL}* APPLYING: update ${pkg.packageName} to version ${versionPolicyChange.newVersion}`
+          );
+        }
+
+        hasChanges ||= hasChanged;
```
iclanton, line 122:
```suggestion
        hasChanges = hasChanges || hasChanged;
```

apps/rush-lib/src/logic/PublishUtilities.ts
```diff
@@ -696,7 +697,7 @@ export class PublishUtilities {
         for (const dependency of downstream) {
           const pkg: IPackageJson = dependency.packageJson;
 
-          PublishUtilities._updateDownstreamDependency(
+          hasChanges ||= PublishUtilities._updateDownstreamDependency(
```
iclanton, line 700:
```suggestion
          const dependenciesHaveDownstreamChanges: boolean = PublishUtilities._updateDownstreamDependency(
```

isc30:
I really like how the tests look now, way cleaner!

common/changes/@microsoft/rush/fix-suffix-publish_2022-02-07-12-51.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "hotfix #3059",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue where the dependencies of lockstep-versioned projects aren't bumped when the lockstep-versioned projects' versions are bumped. This issue only arose in the scenario when the lockstep-versioned projects' version bumps are driven by the change type in the changefiles and not by the \"nextBump\" property in \"common/config/rush/version-policies.json\"",
```

iclanton (approved):

PR #3216, opened 2022-02-08
[rush-lib] Add `buildCacheVersion` cache bust constant

dmichon-msft:
## Summary
Adds a new `buildCacheVersion` constant used by the build cache during the cache id computation. This field will be incremented whenever a change is made to the caching algorithm to ensure that cache tarballs are not reused across algorithm versions.

## Details
The version is injected into the cache id hash as another field.

## How it was tested
Ran a local build and verified was not a cache hit. PR build also not a cache hit. Running twice with the new binary is a cache hit.


apps/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
```diff
@@ -484,6 +484,9 @@ export class ProjectBuildCache {
 
     const sortedProjectStates: string[] = projectStates.sort();
     const hash: crypto.Hash = crypto.createHash('sha1');
+    // This value is used to force cache bust when the build cache algorithm changes
+    hash.update(`${RushConstants.buildCacheVersion}`);
+    hash.update(RushConstants.hashDelimiter);
```
iclanton, line 489:
```suggestion
    if (RushConstants.buildCacheVersion !== 1) {
      // This value is used to force cache bust when the build cache algorithm changes
      hash.update(`${RushConstants.buildCacheVersion}`);
      hash.update(RushConstants.hashDelimiter);
    }
```

As to not break existing caches - unless that's the idea.

dmichon-msft: @iclanton This change is intended to break existing caches to ensure that #3211 fully replaces any cache entries from the preceding release.

common/changes/@microsoft/rush/cache-bust_2022-02-08-00-59.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Add \"buildCacheVersion\" field to build cache hash computation to forcibly replace bad cache entries. Will be incremented when changes are made to the cache algorithm in the future.",
```
iclanton, line 5:
```suggestion
      "comment": "Add the ability to forcibly replace bad cache entries generated by issues in a previous Rush release.",
```

iclanton (approved):

D4N14L (approved):

PR #3223, opened 2022-02-09
[rush-lib] fix bug in `rush change --verify` for nested Rush repos

robertf224:
This command would fail before for nested Rush repos (Rush repos that aren't at the root of the git repo) even after adding changelogs -- the code gathering the changed projects worked properly but the code gathering the changelogs did not.  It was using a hardcoded `common/changes` path from the git repo root.  Instead we pull the resolved changes folder off the Rush configuration.

Not sure how to test this exactly but hoping the bug and fix are obvious to maintainers.

### DIFF
--- a/common/changes/@microsoft/rush/rf-nested-rush-change-verify_2022-02-09-22-07.json
+++ b/common/changes/@microsoft/rush/rf-nested-rush-change-verify_2022-02-09-22-07.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix bug in `rush change --verify` for Rush repos not at the root of a git repo",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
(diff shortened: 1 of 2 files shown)


robertf224:
ok I think this should work properly -- deriving a relative path to the changes folder from the absolute repo root and the absolute changes folder

common/changes/@microsoft/rush/rf-nested-rush-change-verify_2022-02-09-22-07.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix bug in `rush change --verify` for Rush repos not at the root of a git repo",
```
iclanton, line 5:
```suggestion
      "comment": "Fix a bug where `rush change --verify` would not find the correct `common/changes` folder if the `rush.json` is not in the Git repo's root folder.",
```

iclanton:
Thanks for fixing this!

PR #3225, opened 2022-02-10
[rush-lib] Add lifecycle hooks for global and phased commands

dmichon-msft:
## Summary
Adds new Rush lifecycle hooks for the execution of specific and general global and phased commands.
Adds new lifecycle hooks to phased command execution.

## Details
Adds top-level Rush lifecycle hooks that will be invoked after `initialize` when invoking any global script command, or any phased script command.
Adds hook maps in Rush that can be used to tap specific global script commands or phased script commands by name,

Adds hooks to phased command execution for modifying the Operation Graph prior to execution for both the initial pass and watch-mode runs.
Adds hooks to phased command execution that will be invoked after execution of the initial pass, and after execution of watch-mode runs.

Refactors internals to avoid confusion in the public API surface, e.g. by removing execution-time metadata from the Operation Graph and rerouting global configuration away from the `executeAsync` method of `IOperationRunner`.

## How it was tested
Local builds and unit tests to ensure consistency of behavior in `rush build`. Using the hooks to make changes currently left to a subsequent PR.

### DIFF
--- a/apps/rush-lib/src/cli/scriptActions/GlobalScriptAction.ts
+++ b/apps/rush-lib/src/cli/scriptActions/GlobalScriptAction.ts
@@ -10,6 +10,8 @@ import { BaseScriptAction, IBaseScriptActionOptions } from './BaseScriptAction';
 import { Utilities } from '../../utilities/Utilities';
 import { Autoinstaller } from '../../logic/Autoinstaller';
 import { IGlobalCommand, IShellCommandTokenContext } from '../../api/CommandLineConfiguration';
+import { AsyncSeriesHook } from 'tapable';
+import { IGlobalScriptAction } from '../../pluginFramework/RushLifeCycle';
 
 /**
  * Constructor parameters for GlobalScriptAction.
@@ -87,6 +89,18 @@ export class GlobalScriptAction extends BaseScriptAction<IGlobalCommand> {
   }
 
   public async runAsync(): Promise<void> {
+    const { hooks } = this.rushSession;
+    if (hooks.anyGlobalScriptCommand.isUsed()) {
+      await hooks.anyGlobalScriptCommand.promise(this);
+    }
+
+    const specificHook: AsyncSeriesHook<IGlobalScriptAction> | undefined = hooks.globalScriptCommand.get(
+      this.actionName
+    );
+    if (specificHook) {
+      await specificHook.promise(this);
+    }
+
     const additionalPathFolders: string[] =
       this.commandLineConfiguration?.additionalPathFolders.slice() || [];
 
--- a/apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
+++ b/apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
@@ -3,6 +3,7 @@
 
 import * as os from 'os';
 import colors from 'colors/safe';
+import { AsyncSeriesHook } from 'tapable';
 
 import { AlreadyReportedError, Terminal } from '@rushstack/node-core-library';
 import { CommandLineFlagParameter, CommandLineStringParameter } from '@rushstack/ts-command-line';
@@ -22,10 +23,13 @@ import { BuildCacheConfiguration } from '../../api/BuildCacheConfiguration';
 import { SelectionParameterSet } from '../SelectionParameterSet';
 import type { CommandLineConfiguration, IPhase, IPhasedCommand } from '../../api/CommandLineConfiguration';
 import { OperationSelector } from '../../logic/operations/OperationSelector';
-import { Selection } from '../../logic/Selection';
+import { Operation } from '../../logic/operations/Operation';
 import { IOperationFactoryOptions, OperationFactory } from '../../logic/operations/ShellOperationFactory';
+import { Selection } from '../../logic/Selection';
 import { Event } from '../../api/EventHooks';
 import { ProjectChangeAnalyzer } from '../../logic/ProjectChangeAnalyzer';
+import { IPhasedScriptAction } from '../../pluginFramework/RushLifeCycle';
+import { PhasedScriptActionHooks } from '../../pluginFramework/PhasedScriptActionHooks';
 
 /**
  * Constructor parameters for BulkScriptAction.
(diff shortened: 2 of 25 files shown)


common/reviews/api/rush-lib.api.md
```diff
@@ -300,11 +308,34 @@ export class IndividualVersionPolicy extends VersionPolicy {
 export interface _INpmOptionsJson extends IPackageManagerOptionsJsonBase {
 }
 
+// @beta
+export interface IOperationRunner {
+    executeAsync(context: IOperationRunnerContext): Promise<OperationStatus>;
```
iclanton, line 313: Would a plugin be allowed/expected to call this? That seems dangerous.

apps/rush-lib/src/pluginFramework/PhasedScriptActionHooks.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @beta
+ */
+export class PhasedScriptActionHooks {
+  /**
+   * The hook to run after the initial run completes.
+   */
+  public afterRun: AsyncSeriesHook<void> = new AsyncSeriesHook<void>(undefined, 'afterRun');
+  /**
+   * The hook to run after a watch-mode run completes.
+   */
```
iclanton, line 18: Watch-mode runs don't "complete." Maybe expand this comment to clarify that this is fired after each set of operations completes after being invoked when a change is detected.

common/reviews/api/rush-lib.api.md
```diff
@@ -678,9 +744,13 @@ export class _RushGlobalFolder {
...
 
-// @beta (undocumented)
+// @beta
 export class RushLifecycleHooks {
-    initialize: AsyncSeriesHook<void>;
+    anyGlobalScriptCommand: AsyncSeriesHook<IGlobalScriptAction>;
+    anyPhasedScriptComamnd: AsyncSeriesHook<IPhasedScriptAction>;
+    globalScriptCommand: HookMap<AsyncSeriesHook<IGlobalScriptAction>>;
+    initialize: AsyncSeriesHook<IRushAction>;
+    phasedScriptCommand: HookMap<AsyncSeriesHook<IPhasedScriptAction>>;
```
iclanton, line 753: These hook names don't really describe when they're called? Are these called when these things are constructed? Invoked? What's the difference between `anyGlobalScriptCommand` and `globalScriptCommand`?

Hook names should probably include a verb that describes when they're invoked.

apps/rush-lib/src/cli/scriptActions/GlobalScriptAction.ts
```diff
@@ -87,6 +89,18 @@ export class GlobalScriptAction extends BaseScriptAction<IGlobalCommand> {
   }
 
   public async runAsync(): Promise<void> {
+    const { hooks } = this.rushSession;
+    if (hooks.anyGlobalScriptCommand.isUsed()) {
+      await hooks.anyGlobalScriptCommand.promise(this);
+    }
```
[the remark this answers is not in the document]
dmichon-msft: It's an async no-op, as opposed to a sync no-op if it isn't called.

apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -143,33 +171,72 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommand> {
...
+      suppressErrors: isWatch,
+      executionManagerOptions,
+      terminal
+    };
+
+    await this._executeOperations(initialOptions);
+
+    if (hooks.afterRun.isUsed()) {
+      await hooks.afterRun.promise();
     }
```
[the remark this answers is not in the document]
dmichon-msft: Invoking the hook minimally injects an async tick. If we don't care about that tiny bit of overhead, then sure.

common/reviews/api/rush-lib.api.md
dmichon-msft: Call it? No. However, a plugin may provide new `IOperationRunner` instances and need to _implement_ it, so the API needs to be public.

common/reviews/api/rush-lib.api.md
dmichon-msft: I would love for them to be invoked the moment the parser cares about them, such that you could, for example, modify the CLI parameters. Unfortunately, `ts-command-line` doesn't support deferring the parser specification for an action until it is the action being parsed (or `--help` was requested, in which case effectively all actions are the action being parsed).

I can prefix these all with `run` if you want to make them exact.

If you have a global script command `rush foo`, the following execute in order when you invoke it:
```
// Run the general hook for any Rush action
await hooks.initialize.promise(fooAction);
// Run the general hook for a global script action
await hooks.anyGlobalScriptCommand.promise(fooAction);
// Run the specific hook for a global script action named "foo"
await hooks.globalScriptCommand.get('foo')?.promise(fooAction);
```

The purpose of the increasing levels of specificity is so that the lazy-install/lazy-load of Rush plugins can be scoped to:
- Any Rush action
- Any global script action
- A global script action named "foo"
- etc.

apps/rush-lib/src/pluginFramework/PhasedScriptActionHooks.ts
dmichon-msft: Event flow:
1) Initial run starts
2) Initial run completes
3) Rush waits for a file system change
4) Watch run starts
5) Watch run completes
6) Goto (3)

common/reviews/api/rush-lib.api.md
dmichon-msft: By this convention, `initialize` is really `anyCommand`

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -394,13 +349,14 @@ export class OperationExecutionManager {
           break;
         default:
           // This should never happen
-          throw new InternalError('Unexpected task status: ' + operation.status);
+          throw new InternalError(`Unexpected task status: ${status}`);
```
D4N14L, line 352:
```suggestion
          throw new InternalError(`Unexpected operation status: ${status}`);
```

apps/rush-lib/src/logic/operations/OperationExecutionRecord.ts
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * The set of operations that depend on this operation.
+   */
+  public readonly consumers: Set<OperationExecutionRecord> = new Set();
+
+  public readonly stopwatch: Stopwatch = new Stopwatch();
+  public readonly stdioSummarizer: StdioSummarizer = new StdioSummarizer();
+
+  public readonly runner: IOperationRunner;
+  public readonly weight: number;
```
D4N14L, line 82: May need some description on how weight plays into order of execution

dmichon-msft: The comment is in `Operation`, which is where it is a mutable field, but I can copy it here for clarity.

apps/rush-lib/src/cli/scriptActions/GlobalScriptAction.ts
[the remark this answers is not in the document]
dmichon-msft: Adding the if also skips the dynamic function compilation; tapable will always compile the function the first time you invoke it, and that does a bunch of unnecessary work if the hook isn't used. That's why `isUsed()` exists.

dmichon-msft:
I'm going to split this into a few different PRs with smaller scopes.

dmichon-msft:
The internal refactoring is now #3245

PR #3231, opened 2022-02-11
Feat rush plugin feat kind

chengcyber:
## Summary

Fixes https://github.com/microsoft/rushstack/issues/3190

## Details

Adding a `initializeFeatureKind` lifecycle to PluginManager. Plugin defines `featureKinds: ['build-cache-provider']` will be initialized right before trying to load `BuildCacheConfiguration`.


apps/rush-lib/src/cli/test/rush-fake-build-cache-plugin/index.js
```diff
@@ -0,0 +1,22 @@
...
+  trySetCacheEntryBufferAsync: jest.fn(),
+  updateCachedCredentialAsync: jest.fn(),
+  updateCachedCredentialInteractiveAsync: jest.fn(),
+  deleteCachedCredentialsAsync: jest.fn()
+};
+
+class RushFakeBuildCachePlugin {
+  static provider = rushBuildCacheProvider;
+  apply(rushSession) {
+    rushSession.hooks.initialize.tap('rush-fake-build-cache-plugin', function () {
```
dmichon-msft, line 14: If plugin registration is based on feature kind, the plugin shouldn't be able to access any APIs of the executing Rush session until it makes a call to Rush to hand it `FeatureKinds.buildCacheProvider` and the name of the plugin, upon which Rush should hand back the `BuildCacheProviderHooks` object that contains exactly the API surface associated with being a "build cache provider".

apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -116,6 +117,9 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommand> {
     const terminal: Terminal = new Terminal(this.rushSession.terminalProvider);
     let buildCacheConfiguration: BuildCacheConfiguration | undefined;
     if (!this._disableBuildCache) {
+      await this.parser.pluginManager.hooks.initializeFeatureKind
+        .get(FeatureKind.buildCacheProvider)
+        ?.promise();
       buildCacheConfiguration = await BuildCacheConfiguration.tryLoadAsync(
         terminal,
         this.rushConfiguration,
```
dmichon-msft, line 125: I'd expect this to look more like:
```ts
const buildCacheProviderHooks: BuildCacheProviderHooks = await this.parser.pluginManager.initializeAsync(FeatureKind.buildCacheProvider);
buildCacheConfiguration  = await BuildCacheConfiguration.tryLoadAsync(terminal, this.rushConfiguration, buildCacheProviderHooks);
```

Accessing the APIs that are relevant to any given `FeatureKind` should require handing the plugin infrastructure the `FeatureKind`. The plugin requests the relevant API surface by handing over the `FeatureKind`, and the internals of Rush that support plugins hand the plugin manager the `FeatureKind` to get the fully-populated hooks.

dmichon-msft (commented):
My main concern around `FeatureKind` is that I want whatever is in the plugin manifest to be an exact match to the API surface, so that we don't give plugin authors (or ourselves in the process of maintaining Rush) a footgun wherein we can touch a "build cache provider" API without either registering with the Rush plugin manager that "I am a build cache provider" or "I want to consume registered build cache providers".
If we're adding the abstraction of feature kinds, then I expect each `FeatureKind` to be an access key to the relevant set of hooks:
- A plugin must hand Rush `FeatureKind.buildCacheProvider` to get its hands on `BuildCacheProviderHooks` to tap into
- The Rush build cache itself, when it wants providers, must hand the plugin manager `FeatureKind.buildCacheProvider` to get the fully-initialized `BuildCacheProviderHooks` object to execute the hooks

Note that the API surface is no longer required to be implemented using hooks in this model, the only important part is that the calling entity uses the `FeatureKind` to get access to the relevant API from either direction.

chengcyber:
@dmichon-msft I see. You are requesting the `featureKinds:["build-cache-provider"]` indicates both "activiation" and "register functionality" in Rush.js. For now, current implementation covers "activiation" only. Let me see how to take care of "register functionality" part.

Thought:

```
  const buildCacheProviderHooks: BuildCacheProviderHooks = await this.parser.pluginManager.initializeAsync(FeatureKind.buildCacheProvider);
  buildCacheConfiguration  = await BuildCacheConfiguration.tryLoadAsync(terminal, this.rushConfiguration, buildCacheProviderHooks);
```

A edge case would be like two custom build cache plugins are used. one declares `build-cache-provider` feature kind, which triggers Rush initialize `buildCacheProviderHooks`, but the other doesn't declare `build-cache-provider` feature kind.

Maybe this logic should be handled in `rushSession`

```
// plugin code
apply(rushSession) {
  const buildCacheProviderAPI = rushSession.getFeatureKindAPI(FeatureKind.BuildCacheProvider); // <-- validate rush-plugin-manifest.json here
  buildCacheProviderAPI.registerCloudBuildCacheProviderFactory('<build_cache_provider_name', () => ...)
}
```

dmichon-msft:
@chengcyber 
> If a plugin should be activiated sometime, it takes by "activationEvents" property, For instance, only activiated when `rush build`: `"activiationEvents": ["onCommand:build"]`. Currently, this property names `asscoiatedCommands: ["build"]`

I'm strongly of the opinion that we shouldn't separate the "when does this plugin activate?" question from the "what extensibility points does this plugin use?" question. Since the only mechanism we provide for plugins to do anything is to interact with the `RushSession` object, I see the plugin manifest as simply a forward declaration of the interactions the plugin has with `RushSession`. In the particular example of a build cache provider, the expected interaction is the plugin registering itself as a build cache provider, so that's what we expect to see in the plugin manifest.

For simplicity of API usage, I see it as Rush tracking all calls into `RushSession` (and potentially nested API objects) during the synchronous invocation of the plugin's `apply` function. The `RushSession` instance is a proxy around the real `RushSession` that is scoped to the particular plugin instance.
As soon as `apply` finishes executing, Rush compares the set of APIs that the plugin registered callbacks with to the plugin manifest and errors if it is misaligned (or, if the manifest is owned by the consuming repo instead of the plugin repo, it updates the file on disk and requires a git commit).
After the comparison is finished, all of methods on the `RushSession` proxy switch state to immediately throwing errors if invoked, instructing the developer that the plugin is malformed and that all hooks/API extension points must be accessed during the call to `apply` (these extension points are expected to be handed async lambda functions that are responsible for importing the actual runtime implementation of the plugin).

In the theoretical case where, rather than _providing_ a `build-cache-provider`, a plugin wants to _consume_ a `build-cache-provider`, this shouldn't be mechanically any different than Rush itself consuming the set of providers. The plugin will still need to tap into the Rush execution lifecycle, and it is only the action of tapping into the Rush execution lifecycle that we need to be aware of in the plugin manifest, so that we know when to install and load the plugin.

chengcyber:
> I'm strongly of the opinion that we shouldn't separate the "when does this plugin activate?" question from the "what extensibility points does this plugin use?" question. Since the only mechanism we provide for plugins to do anything is to interact with the RushSession object, I see the plugin manifest as simply a forward declaration of the interactions the plugin has with RushSession. In the particular example of a build cache provider, the expected interaction is the plugin registering itself as a build cache provider, so that's what we expect to see in the plugin manifest.

I am cool with this. The `"activiationEvents"` property won't be added at this point. So no worry about the separation!

> In the theoretical case where, rather than providing a build-cache-provider, a plugin wants to consume a build-cache-provider.

Oh, this is really a good point. The constraints provided by `contributes` is bidirectional. Not only checking a plugin can use a API in rush.js, but also ensure the plugin MUST implement the contributions.

I will start to do some changes soon. Thank you for reviewing! @dmichon-msft

apps/rush-lib/src/pluginFramework/RushSessionForPlugin.ts
```diff
@@ -0,0 +1,118 @@
...
+    this._pluginManifest = pluginManifest;
+
+    if (!Array.isArray(pluginManifest.contributes)) {
+      return;
+    }
+
+    const hookForPlugin: RushLifecycleHooks = new RushLifecycleHooks();
+
+    // Plugins specified contribution point should be initialize lazily.
+    this.hooks.initialize = hookForPlugin.initialize;
```
dmichon-msft, line 44: As of #3252, `initialize` is no longer the only hook for plugins, there are now specific hooks for phased/global command execution, with the purpose of allowing the plugin access to the executing command (e.g. the PhasedScriptAction or GlobalScriptAction instance).

apps/rush-lib/src/pluginFramework/RushSessionForPlugin.ts
```diff
@@ -0,0 +1,118 @@
...
+      switch (contribute) {
+        case ContributionPoint.buildCacheProvider: {
+          const contributeCallInfo: IContributeCallInfo = {
+            expectedFunctionNameSet: new Set<string>(),
+            calledFunctionNameSet: new Set<string>()
+          };
+          this._contributeInfo.set(contribute, contributeCallInfo);
+          this.getContributeAPIForBuildCacheProvider = () => {
+            const api: IContributeAPIForBuildCacheProvider = super.getContributeAPIForBuildCacheProvider();
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dmichon-msft, line 56: This seems like overkill. It should be sufficient to validate that the plugin calls `getContributeAPIForBuildCacheProvider`.

chengcyber:
Updates:

1. rebase master, supports more rush life cycles on latest master
2. use a proxy-based rush session for plugins which contributes sth.
3. validate `getContributesAPIForXXX' be called when invoking apply function instead of an overkill validation of detail API.

iclanton:

PR #3236, opened 2022-02-16
[module-minifier-plugin] Update webpack hash calculations

dmichon-msft:
## Summary
Ensures that `@rushstack/module-minifier-plugin` is accounted for in the webpack chunk hash and compilation hash.

## Details
Includes the plugin version and all top-level options as part of the webpack chunk hash calculation. Added documentation about leveraging the webpack `output.hashSalt` field if necessary to change the hash for changes to, e.g. the minifier configuration, since that is out of the direct scope of information readily processable by this plugin.

## How it was tested
Verified that the compilation hash changed with the updated plugin. Unfortunately the compilation and chunk hashes without the plugin appear to already be unstable (several runs yielded distinct hashes).

### DIFF
--- a/webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
+++ b/webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
@@ -2,6 +2,7 @@
 // See LICENSE in the project root for license information.
 
 import { createHash, Hash } from 'crypto';
+
 import {
   CachedSource,
   ConcatSource,
@@ -12,6 +13,7 @@ import {
 } from 'webpack-sources';
 import * as webpack from 'webpack';
 import { AsyncSeriesWaterfallHook, SyncHook, SyncWaterfallHook, TapOptions } from 'tapable';
+
 import {
   CHUNK_MODULES_TOKEN,
   MODULE_WRAPPER_PREFIX,
@@ -21,6 +23,7 @@ import {
 } from './Constants';
 import { getIdentifier } from './MinifiedIdentifier';
 import {
+  IMinifierConnection,
   IModuleMinifier,
   IModuleMinifierPluginOptions,
   IModuleMinificationResult,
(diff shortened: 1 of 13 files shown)


webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
```diff
@@ -535,6 +545,29 @@ export class ModuleMinifierPlugin implements webpack.Plugin {
           }
         );
 
+        const { usePortableModules, compressAsyncImports } = this._optionsForHash;
+
+        function updateChunkHash(hash: Hash, chunk: webpack.compilation.Chunk): void {
+          // For now, automatically assume that any change to this plugin will alter the hash.
+          hash.update(pluginVersion as string);
+          // Using vs. not using source maps might affect output
+          hash.update(`${useSourceMaps}`);
```
iclanton, line 554: How difficult would it be to programmatically put all options into a stable structure and serializing and hashing that, so no one can forget to update this in the future?

webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
```diff
@@ -535,6 +545,29 @@ export class ModuleMinifierPlugin implements webpack.Plugin {
           }
         );
 
+        const { usePortableModules, compressAsyncImports } = this._optionsForHash;
+
+        function updateChunkHash(hash: Hash, chunk: webpack.compilation.Chunk): void {
+          // For now, automatically assume that any change to this plugin will alter the hash.
+          hash.update(pluginVersion as string);
```
iclanton, line 552: Every time this plugin gets updated, all hashes will change. That seems suboptimal.

iclanton (approved):

webpack/module-minifier-plugin/src/ModuleMinifierPlugin.ts
D4N14L: Agree with this. It's probably better to be over-cautious with what may affect the hash vs having to update this list every time a new setting that impacts the hash is added.

dmichon-msft: I would need to update the minifier implementation to include a method `updateHash(hash)` that allows it to add its contribution. In the case of the `MessagePortMinifier` it will need to do a postmessage to the other end to get said hash contribution, which will need some testing. So most options can be handled via `JSON.stringify`, but there is also the issue that this plugin exposes its own hooks that affect code generation. That means a further `updateChunkHash` hook will need to be added to the API surface.

Doable, but mostly it kicks the can down to any extensions.

dmichon-msft: `terser-webpack-plugin` uses a package called `serialize-javascript` to be able to serialize the TerserOptions structure, since it can contain arbitrary javascript. Likely, I would need to do the same.

dmichon-msft: Updated to do the above.

iclanton (approved):

PR #3237, opened 2022-02-16
[localization-plugin] Add an option to ignore missing RESX comments.

iclanton:


webpack/localization-plugin/README.md
```diff
@@ -186,6 +186,10 @@ clones on Windows can end up with CRLF newlines and clones on 'nix operating sys
 newlines. This option can be used to help make compilations run on different platforms produce the same
 result.
 
+#### `localizedData.ignoreMissingResxComments = 'crlf' | 'lf'`
```
D4N14L, line 189: Value doesn't match description

iclanton:
```suggestion
#### `localizedData.ignoreMissingResxComments = true | false | undefined
```

D4N14L (approved):

PR #3240, opened 2022-02-16
[rush-lib] Disable implicit expansion of included phases

dmichon-msft:
## Summary
Modifies evaluation of the `"phases"` field for `"commandKind": "phased"` in `command-line.json` to be an exact list of the phases to execute for the command, instead of quietly injecting any phases that were declared as dependencies of listed phases.

## Details
The main motivation for this is authoring of a watch mode command that performs an asset deployment phase at the end of both the initial execution and subsequent delta executions; however, the build phase used for the initial execution and for the delta executions are different phases (the latter skips cleaning, for example) and so in order to build the dependency graph correctly, the only options are:
1) Disable implicit phase dependency expansion, such that a `_phase:deploy` can be written that depends on both `_phase:build` and `_phase:build-incremental`. In the initial build, `_phase:build` runs, followed by `_phase:deploy`. In the watch rebuild, `_phase:build-incremental` runs, followed by `_phase:deploy`.
2) Write a plugin that taps the (not yet implemented) `prepareOperations` hooks as in #3225
3) Create a separate phase (and thereby a separate npm script) that does exactly the same work but has different declared dependencies.

This PR is option (1).

## How it was tested
Added a sample command `rush test-only` to the Rush Stack repo and validated that running it only invokes the `test` phase.

### DIFF
--- a/apps/rush-lib/src/api/CommandLineConfiguration.ts
+++ b/apps/rush-lib/src/api/CommandLineConfiguration.ts
@@ -273,25 +273,11 @@ export class CommandLineConfiguration {
               commandPhases.add(phase);
             }
 
-            // Apply implicit phase dependency expansion
-            // The equivalent of the "--to" operator used for projects
-            // Appending to the set while iterating it accomplishes a full breadth-first search
-            for (const phase of commandPhases) {
-              for (const dependency of phase.phaseDependencies.self) {
-                commandPhases.add(dependency);
-              }
-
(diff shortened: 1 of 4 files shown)


dmichon-msft:
@octogonz , @iclanton , @elliot-nelson , thoughts?

dmichon-msft:
Abandoning in favor of using option (3) for the scenario. Will revisit how to declare commands that skip phases later.

PR #3245, opened 2022-02-17
[rush-lib] Split out `OperationExecutionRecord` from `Operation`

dmichon-msft:
## Summary
Refactors internals of phased commands to prepare a cleaner public API surface for plugins that impact scheduling or provide custom operation runners.

## Details
Separates the concerns of `Operation`, which is a graph node while declaring what will be run, from `OperationExecutionRecord`, which is the runtime entity at execution time that includes the result state.

## How it was tested
Local phased build executions.

### DIFF
--- a/apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
+++ b/apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
@@ -22,10 +22,17 @@ import { LastLinkFlag, LastLinkFlagFactory } from '../../api/LastLinkFlag';
 import { RushConfigurationProject } from '../../api/RushConfigurationProject';
 import { BuildCacheConfiguration } from '../../api/BuildCacheConfiguration';
 import { SelectionParameterSet } from '../SelectionParameterSet';
-import type { CommandLineConfiguration, IPhase, IPhasedCommandConfig } from '../../api/CommandLineConfiguration';
+import type {
+  CommandLineConfiguration,
+  IPhase,
+  IPhasedCommandConfig
+} from '../../api/CommandLineConfiguration';
 import { createOperations } from '../../logic/operations/OperationSelector';
 import { Operation } from '../../logic/operations/Operation';
-import { IOperationFactoryOptions, OperationFactory } from '../../logic/operations/ShellOperationFactory';
+import {
+  IShellOperationRunnerFactoryOptions,
+  ShellOperationRunnerFactory
+} from '../../logic/operations/ShellOperationRunnerFactory';
 import { Selection } from '../../logic/Selection';
 import { Event } from '../../api/EventHooks';
 import { ProjectChangeAnalyzer } from '../../logic/ProjectChangeAnalyzer';
@@ -48,7 +55,7 @@ export interface IPhasedScriptActionOptions extends IBaseScriptActionOptions<IPh
 interface IExecuteInternalOptions {
   executionManagerOptions: IOperationExecutionManagerOptions;
   isWatch: boolean;
-  operationFactoryOptions: IOperationFactoryOptions;
+  operationFactoryOptions: IShellOperationRunnerFactoryOptions;
   projectSelection: ReadonlySet<RushConfigurationProject>;
   stopwatch: Stopwatch;
   terminal: Terminal;
@@ -59,7 +66,7 @@ interface IExecutionOperationsOptions {
   ignoreHooks: boolean;
   operations: Set<Operation>;
   stopwatch: Stopwatch;
-  suppressErrors: boolean;
+  isWatch: boolean;
   terminal: Terminal;
 }
 
@@ -156,7 +163,7 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
       return;
     }
 
-    const operationFactoryOptions: IOperationFactoryOptions = {
+    const operationFactoryOptions: IShellOperationRunnerFactoryOptions = {
       rushConfiguration: this.rushConfiguration,
       buildCacheConfiguration,
       commandLineConfiguration: this._repoCommandLineConfiguration,
@@ -205,7 +212,9 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
       terminal
     } = options;
 
-    const operationFactory: OperationFactory = new OperationFactory(operationFactoryOptions);
+    const operationFactory: ShellOperationRunnerFactory = new ShellOperationRunnerFactory(
+      operationFactoryOptions
+    );
 
     const operations: Set<Operation> = createOperations(new Set(), {
       operationFactory,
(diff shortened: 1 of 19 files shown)


apps/rush-lib/src/logic/operations/OperationSelector.ts
```diff
@@ -3,176 +3,95 @@
...
+  createOperationRunner(options: IOperationOptions): IOperationRunner;
 }
 
 /**
- * This class is responsible for transforming a set of selected phases and selected projects into an operation dependency graph.
+ * This function creates operations for the set of selected projects and phases, using the provided factory
  */
-export class OperationSelector {
-  private readonly _phasesToRun: ReadonlySet<IPhase>;
+export function createOperations(
```
dmichon-msft, line 34: The reason for this function signature is for use with `phasedScriptAction.hooks.createOperations.tap('default', createOperations)` and `phasedScriptAction.hooks.createDeltaOperations.tap('default', createOperations)` when those lifecycle hooks are added.

apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -382,23 +409,24 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
...
-        console.log(colors.red(`rush ${this.actionName} - Errors! (${stopwatch.toString()})`));
+        terminal.writeErrorLine(colors.red(`rush ${this.actionName} - Errors! (${stopwatch.toString()})`));
       }
 
       if (!ignoreHooks) {
         this._doAfterTask(stopwatch, false);
       }
-      if (!isWatch) {
+
+      if (!suppressErrors) {
```
D4N14L, line 429: I like the original naming here.

apps/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -409,7 +437,7 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
       this.actionName !== RushConstants.buildCommandName &&
       this.actionName !== RushConstants.rebuildCommandName
     ) {
-      // Only collects information for built-in tasks like build or rebuild.
+      // Only collects information for built-in actions like build or rebuild.
```
D4N14L, line 440: Terminology

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -208,182 +222,141 @@ export class OperationExecutionManager {
...
+        if (message) {
+          terminal.writeStderrLine(message);
+        }
+        terminal.writeStderrLine(colors.red(`"${name}" failed to build.`));
+        const blockedQueue: Set<OperationExecutionRecord> = new Set(record.consumers);
+        for (const blockedRecord of blockedQueue) {
+          if (blockedRecord.status === OperationStatus.Ready) {
+            this._completedOperations++;
+
+            //
```
D4N14L, line 287: Tell me more

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -208,182 +222,141 @@ export class OperationExecutionManager {
...
+        break;
     }
-    operation.status = OperationStatus.Success;
 
-    operation.dependents.forEach((dependent: Operation) => {
-      if (!this._changedProjectsOnly) {
-        dependent.runner.isSkipAllowed = false;
+    // Apply status changes to direct dependents
+    for (const item of record.consumers) {
+      if (blockCacheWrite) {
```
D4N14L, line 339: Would be good if we could do something like pass in an OperationCacheManager that could separately control cache management. Not a blocker since this is still in keeping with existing functionality, but a thought for the future.

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -394,13 +367,19 @@ export class OperationExecutionManager {
           break;
         default:
           // This should never happen
-          throw new InternalError('Unexpected task status: ' + operation.status);
+          throw new InternalError(`Unexpected task status: ${status}`);
```
D4N14L, line 370: Phrasing

apps/rush-lib/src/logic/operations/OperationSelector.ts
```diff
@@ -3,176 +3,95 @@
...
-          phase,
-          project,
-          operation
-        };
+      const runner: IOperationRunner = isIncluded
+        ? operationFactory.createOperationRunner({
+            phase,
+            project
+          })
+        : new NullOperationRunner(key, OperationStatus.Skipped, true);
```
D4N14L, line 58: May want to pass in a params object to make it clear what we're passing around

apps/rush-lib/src/logic/operations/ShellOperationFactory.ts
```diff
@@ -2,28 +2,28 @@
...
   rushConfiguration: RushConfiguration;
   buildCacheConfiguration?: BuildCacheConfiguration | undefined;
+  commandLineConfiguration: CommandLineConfiguration;
   isIncrementalBuildAllowed: boolean;
   customParameters: Iterable<IRegisteredCustomParameter>;
   projectChangeAnalyzer: ProjectChangeAnalyzer;
 }
 
-export class OperationFactory implements IOperationFactory {
+export class OperationFactory implements IOperationRunnerFactory {
```
D4N14L, line 26: Class name doesn't match file name

D4N14L (approved):

apps/rush-lib/src/logic/operations/OperationStatus.ts
```diff
@@ -2,15 +2,40 @@
...
+   */
   SuccessWithWarning = 'SUCCESS WITH WARNINGS',
+  /**
+   * The Operation was skipped via the legacy incremental build logic
+   */
   Skipped = 'SKIPPED',
+  /**
+   * The Operation had its outputs restored from the build cache
+   */
   FromCache = 'FROM CACHE',
```
D4N14L, line 32: This one still seems to stick out to me as being not in-keeping with what the status' are for. Kinda in the same vein as splitting the caching stuff out of the OperationExecutionManager. Not blocking, since it's maintaining existing functionality, but maybe something to consider before making this API public

D4N14L (approved):

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
dmichon-msft: Will consider in the future as part of trying to separate out the cache into a plugin.

apps/rush-lib/src/logic/operations/OperationStatus.ts
dmichon-msft: Will need some thought, definitely.

PR #3248, opened 2022-02-22
[rush-amazon-s3-build-cache-plugin] retry failed S3 requests with the same behavior as the the azure blob storage

### LINKED ISSUE #3198: [rush] transient S3 errors updating build cache causes CI to fail
## Summary

Somewhat frequently we get issues in our CI jobs during the phase that compiles dependencies. We make use of the S3 cloud caching functionality and sometimes (I haven't been able to work out specifically why, it seems to happen sometimes when there is little activity on the bucket) that compile step will fail because one of the packages fails to upload to S3 with the error

```
Error uploading cache entry to S3: Error: Amazon S3 responded with status code 503 (Slow Down)

Unable to set cloud cache entry.
```

Rerunning the job always fixes it. Ideally Rush would try and re-upload the cache entry itself automatically (or with a flag if required), a delay before retrying seems sensible.

Alternatively, a flag to not make this error fail the compile step would be fine too.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.59.2 |
| `rushVersion` from rush.json? | 5.59.2 |
| Operating system? | Linux |
| Node.js version (`node -v`)? | 14.17.5 |

ricsam:
## Summary

Fixes #3198, transient S3 errors updating build cache causes CI to fail

## How it was tested

Added one unit test with mocks to test that the retries are called.

Added support for starting a proxy server which will return 500 on every second request to test that the build-cache works when running rush build. This is documented in `build-tests/rush-amazon-s3-build-cache-plugin-integration-test/README.md`

## Todo

- [x] Rush change
- [ ] Rush version --bump

### DIFF
--- a/common/changes/@microsoft/rush/retry-cloud-cache-requests_2022-03-10-22-16.json
+++ b/common/changes/@microsoft/rush/retry-cloud-cache-requests_2022-03-10-22-16.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "fix transient errors in the aws s3 cloud cache plugin",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/common/changes/@microsoft/rush/retry-cloud-cache-requests_2022-03-10-22-16.json
+++ b/common/changes/@microsoft/rush/retry-cloud-cache-requests_2022-03-10-22-16.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "fix transient errors in the aws s3 cloud cache plugin",
+      "comment": "Improve retry logic in the Amazon S3 cloud cache plugin and improve reporting when the user is not authenticated.",
       "type": "none"
     }
   ],
--- a/rush-plugins/rush-amazon-s3-build-cache-plugin/src/AmazonS3Client.ts
+++ b/rush-plugins/rush-amazon-s3-build-cache-plugin/src/AmazonS3Client.ts
@@ -402,9 +402,8 @@ export class AmazonS3Client {
   }
 
   private async _sendCacheRequestWithRetries<T>(sendRequest: () => Promise<T>): Promise<T> {
-    const terminal: ITerminal = this._terminal;
-
     type TryResponse = { hasNetworkError: false; response: T } | { hasNetworkError: true; error: unknown };
+
     const trySendRequest: () => Promise<TryResponse> = async (): Promise<TryResponse> => {
       try {
         const response: T = await sendRequest();
@@ -422,20 +421,20 @@ export class AmazonS3Client {
 
     const response: TryResponse = await trySendRequest();
 
+    const log: (...messageParts: (string | IColorableSequence)[]) => void = this._writeDebugLine.bind(this);
+
     if (response.hasNetworkError) {
       if (storageRetryOptions && storageRetryOptions.maxTries > 1) {
-        terminal.writeVerboseLine(
-          'Network request failed. Will retry request as specified in storageRetryOptions'
-        );
-        const { retryDelayInMs, retryPolicyType, maxTries, maxRetryDelayInMs } = storageRetryOptions;
+        log('Network request failed. Will retry request as specified in storageRetryOptions');
         async function retry(retryAttempt: number): Promise<T> {
+          const { retryDelayInMs, retryPolicyType, maxTries, maxRetryDelayInMs } = storageRetryOptions;
           let delay: number = retryDelayInMs;
           if (retryPolicyType === StorageRetryPolicyType.EXPONENTIAL) {
             delay = retryDelayInMs * Math.pow(2, retryAttempt - 1);
           }
           delay = Math.min(maxRetryDelayInMs, delay);
 
-          terminal.writeVerboseLine(`Will retry request in ${delay}s...`);
+          log(`Will retry request in ${delay}s...`);
           await new Promise<void>((resolve) => {
             setTimeout(() => {
               resolve();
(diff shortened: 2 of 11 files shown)


dmichon-msft:
Retry logic should be handled internally by the build cache provider plugin. For example, the Azure SDK has built-in retry logic and defaults to 4 retries:
https://docs.microsoft.com/en-us/javascript/api/@azure/storage-blob/storageretryoptions?view=azure-node-latest#@azure-storage-blob-storageretryoptions-maxtries
It is only the Amazon S3 plugin that has an issue with single-attempt, which is due to it having been refactored to use `node-fetch` and directly contact the APIs to avoid the enormously heavy Amazon S3 SDK (in terms of both raw dependency count and code size). As such, retry handling needs to be added specifically for that plugin.

ricsam:
Thank you! The smallest PR to solve the issue would be to give the S3 plug-in the same default retry logic as the azure blob package.

I can go for that to begin with.

What do you think about having the retry options configurable? What do you prefer: each plug-in having their own retry options, or one global retry options which both plug-ins consume?

rush-plugins/rush-amazon-s3-build-cache-plugin/src/AmazonS3Client.ts
```diff
@@ -103,27 +124,35 @@ export class AmazonS3Client {
...
-    } else {
-      this._throwS3Error(response, await this._safeReadResponseText(response));
-    }
+    const sendRequest: () => Promise<Buffer | undefined> = async (): Promise<Buffer | undefined> => {
+      const response: fetch.Response = await this._makeRequestAsync('GET', objectName);
+      if (response.ok) {
+        return await response.buffer();
+      } else if (response.status === 404) {
+        return undefined;
+      } else if (response.status === 403 && !this._credentials) {
```
dmichon-msft, line 133: `400` and `401` are not retriable

ricsam: I added 2 tests to verify that 400 and 401 are retriable:

https://github.com/microsoft/rushstack/pull/3248/commits/773aaccc80740fb4cb4fd6509494557b42315a97

dmichon-msft: To clarify, I meant that they should be considered to be not retriable, since missing/incorrect auth is not something that can be corrected by sending the request again, nor is a malformed request payload.

ricsam: Ah that makes sense 😅. I'll update so that the code can either throw error or retry

ricsam: Do you know the rationale behind only throwing 403 errors if credentials are provided? This code: `if (response.status === 403 && !this._credentials) return undefined` 

Should 401 and 400 be handled in the same way?

ricsam: Also, in `uploadObjectAsync` response codes are not handled (it is just `if (!response.ok) throw err`). Do you think response codes in `uploadObjectAsync` should be handled in the same way as in `getObjectAsync`?

ricsam: To answer my question myself:

I'd say 400, 401, 403 should be handled equally.

When no credentials are provided unauth errors should be silenced as the ones potentially running into this issue usually are human users (not CI) that don't have credentials and they should still be able to build without access to the cloud storage.

Usually human users users don't have writeAccess either to cloud storage so the `uploadObjectAsync` can simply throw if invalid response.

ricsam:
I pushed a commit where 400, 401 and 403 are handled the same

rush-plugins/rush-amazon-s3-build-cache-plugin/src/test/AmazonS3Client.test.ts
```diff
@@ -218,21 +222,31 @@ describe(AmazonS3Client.name, () => {
...
+    let realSetTimeout: typeof setTimeout;
     beforeEach(() => {
+      // mock date
       realDate = global.Date;
       global.Date = MockedDate as typeof Date;
+
+      // mock setTimeout
+      realSetTimeout = global.setTimeout;
+      global.setTimeout = ((callback: () => void, time: number) => {
+        return realSetTimeout(callback, 1);
```
dmichon-msft, line 234: You could use `process.nextTick(callback)` as an alternative.

dmichon-msft (approved):

rush-plugins/rush-amazon-s3-build-cache-plugin/src/AmazonS3Client.ts
```diff
@@ -103,27 +134,65 @@ export class AmazonS3Client {
...
+      } else if (
+        (response.status === 400 || response.status === 401 || response.status === 403) &&
+        !this._credentials
+      ) {
+        // unauthorized due to not providing credentials,
+        // silence error for better DX when e.g. running locally without credentials
+        return {
+          hasNetworkError: false,
+          response: undefined
+        };
```
iclanton, line 158: If there are no credentials and we get a 401, or if we get a 403, we should tell the user that they need to authenticate. This looks the same as a cache miss, so a user with missing, expired, or insufficient credentials will never get entries from cache and will never know that they need to do something about that.

iclanton (changes requested):

iclanton:
You need to run `rush change`.

ricsam:
This is ready for review 😊

common/changes/@microsoft/rush/retry-cloud-cache-requests_2022-03-10-22-16.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "fix transient errors in the aws s3 cloud cache plugin",
```
iclanton, line 5:
```suggestion
      "comment": "Improve retry logic in the Amazon S3 cloud cache plugin and improve reporting when the user is not authenticated.",
```

iclanton (approved):

PR #3252, opened 2022-02-25
[rush-lib] Add hooks for phased/global commands

dmichon-msft:
## Summary
Adds Rush lifecycle hooks for global command and phased command execution. Allows to tap for all commands or by name.
Updates the Rush lifecycle hooks to pass the currently executing command object.

Split from #3225 .

## Details
Adds a `runAnyPhasedScriptCommand` hook that runs after `initialize` but before the more specific `runPhasedScriptCommand.for(commandName)` hook.
Adds a `runAnyGlobalScriptCommand` hook that runs after `initialize` but before the more specific `runGlobalScriptCommand.for(commandName)` hook.
Passes the currently executing command to the relevant hook, so that properties (and minimally the name) can be read by the plugin.

Waiting for #3245 before adding hooks to phased commands directly.

## How it was tested
Verified that existing phased and global commands still work. Using the new hooks is for a future PR.

### DIFF
--- a/apps/rush-lib/src/pluginFramework/RushLifeCycle.ts
+++ b/apps/rush-lib/src/pluginFramework/RushLifeCycle.ts
@@ -1,14 +1,78 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-import { AsyncSeriesHook } from 'tapable';
+import { AsyncSeriesHook, HookMap } from 'tapable';
 
 /**
+ * Information about the currently executing action provided to plugins.
+ * @beta
+ */
+export interface IRushAction {
+  /**
+   * The name of this action, as seen on the command line
+   */
+  readonly actionName: string;
+}
+
+/**
+ * Information about the currently executing global script action (as defined in command-line.json) provided to plugins.
+ * @beta
+ */
+export interface IGlobalScriptAction extends IRushAction {
+  // Nothing added.
+}
+
+/**
+ * Information about the currently executing phased script action (as defined in command-line.json, or default "build" or "rebuild") provided to plugins.
+ * @beta
+ */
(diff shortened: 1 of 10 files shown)


common/reviews/api/rush-lib.api.md
```diff
@@ -7,6 +7,7 @@
 /// <reference types="node" />
 
 import { AsyncSeriesHook } from 'tapable';
+import { HookMap } from 'tapable';
```
D4N14L, line 10:
```suggestion
import { AsyncSeriesHook, HookMap } from 'tapable';
```

apps/rush-lib/src/pluginFramework/RushLifeCycle.ts
```diff
@@ -1,14 +1,78 @@
...
 // See LICENSE in the project root for license information.
 
-import { AsyncSeriesHook } from 'tapable';
+import { AsyncSeriesHook, HookMap } from 'tapable';
 
 /**
+ * Information about the currently executing action provided to plugins.
+ * @beta
+ */
+export interface IRushAction {
```
D4N14L, line 10: Is `Action` the phrasing we want to go with here? That may confuse with Heft actions.

dmichon-msft: Considering that I called them commands in the hook names, will update

dmichon-msft: Though they mean the same thing here as in Heft

common/reviews/api/rush-lib.api.md
dmichon-msft: Generated file, it'll get automatically revered next time it builds.

D4N14L: Ah, didn't realize what file it was on.

iclanton (approved):

PR #3262, opened 2022-03-06
Include a new project "@rushstack/hashed-folder-copy-plugin"

iclanton:
## Summary

This project is a new Webpack plugin. For more information, see the README in this PR.

## How it was tested

Included a test project.


webpack/hashed-folder-copy-plugin/src/HashedFolderCopyPlugin.ts
```diff
@@ -0,0 +1,377 @@
...
+          }
+
+          const assetContents: Buffer = FileSystem.readFileToBuffer(globResultFullPath);
+          assetsToAdd.set(globResult, { size: () => assetContents.byteLength, source: () => assetContents });
+          module.buildInfo.fileDependencies.add(globResultFullPath);
+        }
+      }
+    }
+
+    const hash: crypto.Hash = crypto.createHash('md5');
```
TheLarkInn, line 303: I thought webpack has an api for hashing. Maybe on this.buildInfo https://cs.github.com/webpack/webpack/blob/0456807b72d914a507361d183c87380f740b40e1/lib/NormalModule.js#L646

TheLarkInn: AKA if we have a hash available from the build as a whole, can we use that instead of generating a new one.

webpack/hashed-folder-copy-plugin/src/HashedFolderCopyPlugin.ts
```diff
@@ -0,0 +1,377 @@
...
+          packageName = globsBase;
+          pathInsidePackage = '';
+        } else {
+          packageName = globsBase.substr(0, slashAfterPackageNameIndex);
+          pathInsidePackage = globsBase.substr(slashAfterPackageNameIndex + 1);
+        }
+
+        let packagePath: string;
+        try {
+          const resolveResult: string = resolver.resolveSync(
```
TheLarkInn, line 270: Can we resolveAsync instead and then catch in the callback?

build-tests/hashed-folder-copy-plugin-test/src/index.ts
```diff
@@ -0,0 +1,23 @@
+import type { IRequireFolderOptions } from '@rushstack/hashed-folder-copy-plugin';
+
+declare function requireFolder(options: IRequireFolderOptions): string;
```
TheLarkInn, line 3: I've never used this plugin before so I'm still understanding its functionality. Do you think we could add another source file imported by index.ts, that uses `requireFolder()` but also specifies different outputPath for the test case.

TheLarkInn:
Does this work on webpack 5? Should we throw if it doesn't?

webpack/hashed-folder-copy-plugin/src/HashedFolderCopyPlugin.ts
iclanton: Unfortunately, no. This hook is a `SyncBailHook`.

webpack/hashed-folder-copy-plugin/src/HashedFolderCopyPlugin.ts
iclanton: Webpack does have a hash API, but I don't think it'll offer much beyond just using `crypto`.

> if we have a hash available from the build as a whole, can we use that instead of generating a new one.

The general use case for this plugin is for including a folder of static assets, which may not change between builds. If we use the whole compilation's hash, any change to the compilation will be a cache break on these assets, even if they didn't change.

build-tests/hashed-folder-copy-plugin-test/src/index.ts
iclanton: Sure, added that.

TheLarkInn (approved):
Looks good. This second test case is pretty helpful.

PR #3263, opened 2022-03-06
[rush-lib] Added Parallelism percentage value support

wbern:
## Summary

Adds support for percentage values, see https://github.com/microsoft/rushstack/issues/3257

## Details

This is the first pull request I'm introducing in an effort to avoid draining resources of local developer machines.

Since developers still have to specify the --parallelism flag, my next idea would be to add some functionality that let's us use all cores minus 1. There's no specific issue for this yet.

Jest has support for percentage-values, so I figured this should be a welcome change for flexibility regardless. This code was also heavily inspired by jest's implementation, found here: https://github.com/facebook/jest/blob/e0b33b74b5afd738edc183858b5c34053cfc26dd/packages/jest-config/src/getMaxWorkers.ts#L29

## How it was tested

I added a test that simply ensures that the generation won't fail. I found that creating a snapshot should be unnecessary in this case.

I actually did not run this against any repo.. Maybe I need to do that. 🙃

### DIFF
--- a/apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
+++ b/apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
@@ -159,22 +159,21 @@ export class OperationExecutionManager {
       } else {
         const parsed: number = parseInt(parallelism, 10);
 
-        if (isNaN(parsed)) {
+        let parallelismInt: number = parsed > 0 ? parsed : 1;
+
+        if (typeof parallelism === 'string' && parallelism.trim().endsWith('%')) {
+          if (parsed <= 0 || parsed > 100) {
+            throw new Error(
+              `Invalid percentage value of '${parallelism}', value cannot be less than '0%' or more than '100%'`
+            );
+          }
+
+          const workers: number = Math.floor((parsed / 100) * numberOfCores);
+          parallelismInt = Math.max(workers, 1);
+        } else if (isNaN(parsed)) {
           throw new Error(
             `Invalid parallelism value of '${parallelism}', expected a number, a percentage, or 'max'`
           );
-        }
-
-        let parallelismInt: number = parsed > 0 ? parsed : 1;
-
-        if (
-          typeof parallelism === 'string' &&
-          parallelism.trim().endsWith('%') &&
-          parsed > 0 &&
-          parsed <= 100
-        ) {
-          const workers: number = Math.floor((parsed / 100) * numberOfCores);
-          parallelismInt = Math.max(workers, 1);
         }
 
         this._parallelism = parallelismInt;
--- a/apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
+++ b/apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
@@ -159,10 +159,8 @@ export class OperationExecutionManager {
       } else {
         const parallelismAsNumber: number = Number(parallelism);
 
-        const getParallelismValueHighThanZero = (value: number) => (value > 0 ? value : 1);
-
         if (typeof parallelism === 'string' && parallelism.trim().endsWith('%')) {
-          const parsedPercentage = Number(parallelism.trim().replace(/\%$/, ''));
+          const parsedPercentage: number = Number(parallelism.trim().replace(/\%$/, ''));
 
           if (parsedPercentage <= 0 || parsedPercentage > 100) {
             throw new Error(
@@ -171,9 +169,9 @@ export class OperationExecutionManager {
           }
 
           const workers: number = Math.floor((parallelismAsNumber / 100) * numberOfCores);
-          this._parallelism = getParallelismValueHighThanZero(workers);
+          this._parallelism = Math.max(workers, 1);
         } else if (!isNaN(parallelismAsNumber)) {
-          this._parallelism = getParallelismValueHighThanZero(parallelismAsNumber);
+          this._parallelism = Math.max(parallelismAsNumber, 1);
         }
 
         throw new Error(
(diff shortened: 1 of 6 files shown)


apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -95,10 +95,24 @@ export class OperationExecutionManager {
...
+          );
+        }
+
+        let parallelismInt: number = parsed > 0 ? parsed : 1;
+
+        if (
+          typeof parallelism === 'string' &&
+          parallelism.trim().endsWith('%') &&
+          parsed > 0 &&
+          parsed <= 100
```
iclanton, line 112: We should error if someone specifies something <0% or >100%.

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -95,10 +95,24 @@ export class OperationExecutionManager {
...
-        const parallelismInt: number = parseInt(parallelism, 10);
+        const parsed: number = parseInt(parallelism, 10);
 
-        if (isNaN(parallelismInt)) {
-          throw new Error(`Invalid parallelism value of '${parallelism}', expected a number or 'max'`);
+        if (isNaN(parsed)) {
+          throw new Error(
+            `Invalid parallelism value of '${parallelism}', expected a number, a percentage, or 'max'`
+          );
+        }
```
iclanton, line 104: This will always throw if someone specifies a percent.

wbern: OK, should be fixed now.

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
wbern: Added a check and error message for this.

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -95,10 +95,23 @@ export class OperationExecutionManager {
       if (parallelism === 'max') {
         this._parallelism = numberOfCores;
       } else {
-        const parallelismInt: number = parseInt(parallelism, 10);
-
-        if (isNaN(parallelismInt)) {
-          throw new Error(`Invalid parallelism value of '${parallelism}', expected a number or 'max'`);
+        const parsed: number = parseInt(parallelism, 10);
```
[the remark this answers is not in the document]
iclanton: Yeah, probably...

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -95,13 +95,28 @@ export class OperationExecutionManager {
...
+          const parsedPercentage = Number(parallelism.trim().replace(/\%$/, ''));
+
+          if (parsedPercentage <= 0 || parsedPercentage > 100) {
+            throw new Error(
+              `Invalid percentage value of '${parallelism}', value cannot be less than '0%' or more than '100%'`
+            );
+          }
+
+          const workers: number = Math.floor((parallelismAsNumber / 100) * numberOfCores);
+          this._parallelism = getParallelismValueHighThanZero(workers);
```
dmichon-msft, line 112:
```suggestion
          this._parallelism = Math.max(workers, 1);
```

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -95,13 +95,28 @@ export class OperationExecutionManager {
       if (parallelism === 'max') {
         this._parallelism = numberOfCores;
       } else {
-        const parallelismInt: number = parseInt(parallelism, 10);
+        const parallelismAsNumber: number = Number(parallelism);
 
-        if (isNaN(parallelismInt)) {
-          throw new Error(`Invalid parallelism value of '${parallelism}', expected a number or 'max'`);
+        const getParallelismValueHighThanZero = (value: number) => (value > 0 ? value : 1);
+
```
dmichon-msft, line 101:
```suggestion
```

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -95,13 +95,28 @@ export class OperationExecutionManager {
...
+          if (parsedPercentage <= 0 || parsedPercentage > 100) {
+            throw new Error(
+              `Invalid percentage value of '${parallelism}', value cannot be less than '0%' or more than '100%'`
+            );
+          }
+
+          const workers: number = Math.floor((parallelismAsNumber / 100) * numberOfCores);
+          this._parallelism = getParallelismValueHighThanZero(workers);
+        } else if (!isNaN(parallelismAsNumber)) {
+          this._parallelism = getParallelismValueHighThanZero(parallelismAsNumber);
```
dmichon-msft, line 114:
```suggestion
          this._parallelism = Math.max(parallelismAsNumber, 1);
```

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -95,13 +95,28 @@ export class OperationExecutionManager {
...
+        const parallelismAsNumber: number = Number(parallelism);
 
-        if (isNaN(parallelismInt)) {
-          throw new Error(`Invalid parallelism value of '${parallelism}', expected a number or 'max'`);
+        const getParallelismValueHighThanZero = (value: number) => (value > 0 ? value : 1);
+
+        if (typeof parallelism === 'string' && parallelism.trim().endsWith('%')) {
+          const parsedPercentage = Number(parallelism.trim().replace(/\%$/, ''));
+
+          if (parsedPercentage <= 0 || parsedPercentage > 100) {
```
dmichon-msft, line 105: Do we actually need to cap it at number of cores? If you specify an integer we allow you to set as many parallel processes as you want, which may be relevant if you have a lot of packages that spend a lot of time waiting rather than using the CPU.

dmichon-msft (approved):
Left a few minor optimization suggestions. Not sure if we actually need to clamp the percentage at 100%, but we could always revisit that later.

apps/rush-lib/src/logic/operations/OperationExecutionManager.ts
iclanton: Let's tackle that separately if someone requests it.

iclanton (approved):

Issue #3181, opened 2022-01-21
[heft] Design Proposal: Alignment with Rush phased commands


dmichon-msft:
This is a proposal for aligning the architecture of Heft to be more compatible with Rush "phased" commands in the interests of improving parallelism, customizability for other tools (esbuild, swc, etc.), reducing Heft aggregate boot time, and optimizing multi-project watching.

# Goal 1: Increased Parallelism and Configurability

## Current state
Today `heft test` runs a sequence of hardcoded pipeline stages:
```
[Clean] (if --clean) -> [Build] (unless --no-build) -> [Test]
```
Where the `Build` stage is further subdivided into hardcoded sub-stages:
```
[Pre-Compile] -> [Compile] -> [Bundle] -> [Post-Build]
```

This limits the ability of Rush to exploit task parallelism to running `heft build --clean` and `heft test --no-build` for each project, i.e. if:
```
A <-(depends on)- B
```
Then the `test` phase for `A` can run concurrently with the `build` phase for `B`.

The `heft.json` file provides event actions and plugins to inject build steps at various points within this pipeline, but the pipeline itself is not particularly customizable.

When run from the command line, Heft loads a single `HeftConfiguration` object and creates a `HeftSession` that corresponds to the command line session.

## Desired state
In future build rigs that exploit the `isolatedModules` contract to allow transpilation of each and every module from TypeScript -> JavaScript to be an independent operation, we instead have stages more like the following, each of which handles cleaning internally:
- **Compile**: Converts TypeScript -> ECMAScript.
  - Dependencies: none
- **SASS**: Convert SASS -> CSS, emit .d.ts files.
  - Dependencies: none
- **Analyze**: Type Check, Lint, emit .d.ts files.
  - Dependencies: Analyze in dependency projects, SASS in current project
- **Test**: Run unit tests.
  - Dependencies: Compile in self and dependency projects
- **Bundle**: Combine ECMAScript/CSS/etc. into more compact bundled form.
  - Dependencies: Compile in self and dependency projects. Potentially bundle in dependency projects.

Custom rigs may require more or fewer stages to accommodate other build steps, and importantly, may alter the dependency relationship between the stages. For example a rig may opt to run its tests on bundled output, and therefore have the "test" stage depend on the "bundle" stage.

# Goal 2: Reduce time booting Heft repeatedly in a large Rush monorepo

## Current state
[...]

PR #3276, opened 2022-03-14
[heft-config-file] Add inline `inheritanceType` specification in configuration files

D4N14L:
## Summary

Implements inline `inheritanceType` properties, largely following the spec from #3181.

## Details

When extending a base configuration file, this feature allows a developer to specify `$<targetProperty>.inheritanceType` properties to customize how the specified property will be inherited into the configuration file. Previously, this was only possible at the configuration file class level, and only for top-level properties. Additionally, keyed `object` types are now able to be merged.

Currently, support is provided for the following `inheritanceType` values:
- `replace`
- `append` (Only `Array` properties)
- `merge` (Only keyed `object` properties)

For example, if you have a file configB.json that extends configA.json, the following can now be accomplished:

configA.json:
```javascript
{
  "propA": [ 1, 2, 3 ],
  "propB": {
    "propC": {
      "propD": "From configA"
      "propE": [ "From configA" ]
      "propF": "Also from configA"
    }
  }
}
```

configB.json:
```javascript
{
  "propA": [ 4, 5, 6 ],
  "$propB.inheritanceType": "merge",
  "propB": {
    "$propC.inheritanceType": "merge",
    "propC": {
      "propD": "From configB",
      "$propE.inheritanceType": "append",
      "propE": [ "From configB" ]
    }
  }
}
```

Result:
```javascript
{
  "propA": [ 1, 2, 3, 4, 5, 6 ],
  "propB": {
    "propC": {
      "propD": "From configB"
      "propE": [ "From configA", "From configB" ],
      "propF": "Also from configA"
  }
}
```
By default, the same behavior as the existing `@rushstack/heft-config-file` will be used for inheritance at all levels, making this change backwards compatible. The current behavior is:
- Arrays: `append`
- All other types: `replace`

Strictly speaking, this makes explicitly specifying the `append` value redundant in cases where top-level arrays aren't set to `replace` by the configuration file class, since the only other inheritance type available on arrays would be to `replace`. However, it is included for clarity during configuration file specification. In a somewhat similar manner, specifying `replace` on an object would be equally redundant, however the `replace` value would still be required for Arrays.

One notable change to existing behavior is that schema validation is now done only after the configuration object is fully constructed through the extends-stack. Previously, each configuration file would be individually validated against the schema. This would have made it impossible to inherit values from a previous object while modifying other existing values.

## How it was tested

Added multiple test cases to validate inheritance behavior in a variety of scenarios, including handling of error cases.


common/changes/@rushstack/heft-config-file/user-danade-AllowCustomMergeBehavior_2022-03-14-22-51.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-config-file",
+      "comment": "Add inline \"inheritanceType\" property support. This feature allows for configuration files to specify how object and array properties are inherited, overriding the default inheritance behavior provided by the configuration file class.",
```
iclanton, line 5:
```suggestion
      "comment": "Add support for an inline \"$<propertyName>.inheritanceType\" property. This feature allows for configuration files to specify how object and array properties are inherited, overriding the default inheritance behavior provided by the configuration file class.",
```

iclanton:
What if we said the magic properties are objects, like:

```JSON
{
  "$propName": {
    "inheritaneType": "merge",
    // other options we add someday
  },
  "propName": ...
}
```

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -395,7 +401,7 @@ export class ConfigurationFile<TConfigurationFile> {
       });
     }
 
-    let parentConfiguration: Partial<TConfigurationFile> = {};
```
iclanton, line 398: Hmm this should still be `Partial`, shouldn't it?

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -668,6 +557,248 @@ export class ConfigurationFile<TConfigurationFile> {
...
+    const filteredObjectPropertyNames: string[] = [];
+    const inheritanceTypeMap: Map<string, IPropertyInheritance<InheritanceType>> = new Map();
+    for (const propertyName of currentObjectPropertyNames) {
+      if (ignoreProperties && ignoreProperties.has(propertyName)) {
+        continue;
+      }
+      const inheritanceTypeMatches: RegExpMatchArray | null = propertyName.match(
+        CONFIGURATION_FILE_MERGE_BEHAVIOR_FIELD_REGEX
+      );
+      if (inheritanceTypeMatches && inheritanceTypeMatches.length === 2) {
```
iclanton, line 601: Length should always be `2`, right?

Do you think we should fail if something looks like a magic property, but isn't? We should also probably have a way to escape these. Maybe `$$` is an escaped `$`.

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -668,6 +557,248 @@ export class ConfigurationFile<TConfigurationFile> {
...
+        const mergeTargetPropertyName: string = inheritanceTypeMatches[1];
+        const inheritanceTypeRaw: unknown | undefined = (currentObject || {})[propertyName];
+        if (!currentObjectPropertyNames.has(mergeTargetPropertyName)) {
+          throw new Error(
+            `Issue in processing configuration file property "${propertyName}". ` +
+              `An inheritance type was provided but no matching property was found`
```
iclanton, line 607:
```suggestion
              `An inheritance type was provided but no matching property was found in the parent.`
```

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -668,6 +557,248 @@ export class ConfigurationFile<TConfigurationFile> {
...
+          );
+        } else if (typeof inheritanceTypeRaw !== 'string') {
+          throw new Error(
+            `Issue in processing configuration file property "${propertyName}". ` +
+              `An unsupported inheritance type was provided: ${JSON.stringify(inheritanceTypeRaw)}`
+          );
+        } else if (typeof (currentObject || {})[mergeTargetPropertyName] !== 'object') {
+          throw new Error(
+            `Issue in processing configuration file property "${propertyName}". ` +
+              `An inheritance type was provided for a property that is not an object`
```
iclanton, line 617:
```suggestion
              `An inheritance type was provided for a property that is not an object or array.`
```

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -668,6 +557,248 @@ export class ConfigurationFile<TConfigurationFile> {
...
+      configurationFilePath: resolvedConfigurationFilePath,
+      originalValues: {} as TField
+    };
+    const result: TField = {
+      [CONFIGURATION_FILE_FIELD_ANNOTATION]: resultAnnotation
+    } as unknown as TField;
+
+    // Do a first pass to gather and strip the inheritance type annotations from the merging object
+    const currentObjectPropertyNames: Set<string> = new Set(Object.keys(currentObject || {}));
+    const filteredObjectPropertyNames: string[] = [];
```
iclanton, line 592:
```suggestion
    // An array of property names that should be included in the resulting object
    const filteredObjectPropertyNames: string[] = [];
```

D4N14L:
> What if we said the magic properties are objects, like:
> 
> ```json
> {
>   "$propName": {
>     "inheritaneType": "merge",
>     // other options we add someday
>   },
>   "propName": ...
> }
> ```

Given the limited additional properties we would be adding here, I feel this would be better handled by stacking the properties, so something like:
```json
{
  "$propName.inheritanceType": "merge",
  "$propName.otherOption": "...",
  "propName": "..."
}
```
While yours is absolutely more flexible, I'm not sure that the added flexibility trumps the conciseness of the proposed format.

Or, I guess, we could support both styles, such that the simplified `$propName.<option>` maps to an object with an `<option>` property internally? So that the two formats are functionally equivalent?

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -668,6 +557,248 @@ export class ConfigurationFile<TConfigurationFile> {
...
+                `An unsupported inheritance type was provided: "${inheritanceTypeRaw}"`
+            );
+        }
+      } else {
+        filteredObjectPropertyNames.push(propertyName);
+      }
+    }
+
+    // We only filter the currentObject because the parent object should already be filtered
+    const propertyNames: Set<string> = new Set<string>([
```
iclanton, line 642: Change these `string` types to `keyof TField` and you should be able to remove the `as any`s below. You'll need to cast some values to `keyof TField`.

If that doesn't work, cast to `Record<string, IPropertyInheritance<InheritanceType>>` instead below.

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -668,6 +557,248 @@ export class ConfigurationFile<TConfigurationFile> {
...
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          configuredPropertyInheritance && (configuredPropertyInheritance as any)[propertyName] !== undefined
+            ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              (configuredPropertyInheritance as any)[propertyName]
+            : defaultInheritanceType;
+      }
+
+      let newValue: unknown;
+      const usePropertyValue: () => void = () => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
iclanton, line 671: You should be able to remove these too if `propertyName` is `keyof TField`.

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -668,6 +557,248 @@ export class ConfigurationFile<TConfigurationFile> {
...
+            break;
+          }
+
+          default: {
+            throw new Error(`Unknown inheritance type "${propertyInheritance}"`);
+          }
+        }
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
iclanton, line 795: Same here.

libraries/heft-config-file/src/ConfigurationFile.ts
D4N14L: The way I see it, if we don't have an `extends` field, there is no parent configuration file, so it should be `undefined`. Below we pass in an empty object in the case that it's still `undefined` in order to satisfy the `Partial<TConfigurationFile>` arg, but that's not strictly the "parent configuration file", more an artifact of how we call the method.

Likewise, `_mergeObjects` will only take what it assumes are partial values for the field, and return a partial result, since we can't be sure that the type is entirely complete. The _only_ time that we strictly know that everything is complete is after we've validated against a provided schema, since that's the source of truth for the object structure.

libraries/heft-config-file/src/ConfigurationFile.ts
D4N14L: It should always be `2` since the regex is intentionally tight, correct. I can remove that check since it's not really necessary.

Not so sure we should fail on things that look like magic properties though, at least not yet. If we're sticking to the `$<propertyName>.inheritanceType` style for now, then the only way to collide with this would be to be of that exact format. If we were to generalize the magic properties to be like what you described above (eg. `"$<propertyName": { ... }`) then it becomes more of a problem, and would likely necessitate some form of error case (and escaping from that error case... I like the `$$` to escape the `$` in that scenario). For now though, that doesn't really seem like a problem given the current constraints.

These are really good points though and I'm going to drop a comment in to point them out in the future, should we go that route.

iclanton (approved):

PR #3277, opened 2022-03-15
[rush-lib]: update readme, fix broken links

kompasbank-lasse:
Fixes broken links for the rush-lib documentation and api reference.


iclanton:
Can you run `rush change`?

PR #3288, opened 2022-03-21
[rush] Make --quiet flags more consistent

elliot-nelson:
## Summary

Make support for `-q/--quiet` consistent for both direction execution and for `install-run-rush` scripts.

The goal is that whether you're running `rush`, `rushx`, or one of their `install-run` variants, you can specify `-q / --quiet` to suppress any startup output.

## Details

For `rushx`, the `install-run-rushx` script now honors the `-q/--quiet` flag in the same way as direct execution:

(In this example, script `"hello"` is defined as `"echo hello"`.)

```console
> rushx --quiet hello
hello

> node ../../common/scripts/install-run-rushx.js --quiet hello
hello

```

For `rush`, the `-q/--quiet` flag has been _added_ as a pre-command flag, similar to `-d/--debug`. It operates in the same way it does with rushx, which is to suppress all informational and diagnostic information printed at startup.

```
> rush --quiet list
a
b
c

> node common/scripts/install-run-rush.js --quiet list
a
b
c

```

As part of this change, the `shouldRestrictConsoleOutput` utility was moved into `RushCommandLineParser`, so that its implementation is closer to the other code processing args for Rush.

For suppressing output in `install-run`, which has quite a few places it can log output, we now pass around a very small logger object (rather than inserting new if statements in every function).  It's possible you could use some of `ITerminal` for this instead, although it seems pretty heavy-duty here.

## How it was tested

Tested all 4 commands (rush, rushx, install-run-rush, install-run-rushx), with and without -q/--quiet flags.

### DIFF
--- a/apps/rush-lib/src/cli/RushCommandLineParser.ts
+++ b/apps/rush-lib/src/cli/RushCommandLineParser.ts
@@ -155,8 +155,11 @@ export class RushCommandLineParser extends CommandLineParser {
    * Used during Rush startup.
    */
   public static peekIsQuiet(): boolean {
-    const flags: string[] = process.argv.slice(2);
-    return flags.includes('-q') || flags.includes('--quiet');
+    for (let i = 2; i < process.argv.length; i++) {
+      const arg = process.argv[i];
+      if (arg === '-q' || arg === '--quiet') return true;
+    }
+    return false;
   }
 
   /**
--- a/apps/rush-lib/src/cli/RushCommandLineParser.ts
+++ b/apps/rush-lib/src/cli/RushCommandLineParser.ts
@@ -5,7 +5,7 @@ import colors from 'colors/safe';
 import * as os from 'os';
 import * as path from 'path';
 
-import { CommandLineParser, CommandLineFlagParameter } from '@rushstack/ts-command-line';
+import { CommandLineParser, CommandLineFlagParameter, CommandLineHelper } from '@rushstack/ts-command-line';
 import {
   InternalError,
   AlreadyReportedError,
@@ -22,7 +22,6 @@ import {
   IGlobalCommandConfig,
   IPhasedCommandConfig
 } from '../api/CommandLineConfiguration';
-import { Utilities } from '../utilities/Utilities';
 
 import { AddAction } from './actions/AddAction';
 import { ChangeAction } from './actions/ChangeAction';
@@ -72,6 +71,7 @@ export class RushCommandLineParser extends CommandLineParser {
   public readonly pluginManager: PluginManager;
 
   private _debugParameter!: CommandLineFlagParameter;
+  private _quietParameter!: CommandLineFlagParameter;
   private readonly _rushOptions: IRushCommandLineParserOptions;
   private readonly _terminalProvider: ConsoleTerminalProvider;
   private readonly _terminal: Terminal;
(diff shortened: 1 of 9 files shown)


apps/rush-lib/src/cli/RushCommandLineParser.ts
```diff
@@ -145,6 +145,31 @@ export class RushCommandLineParser extends CommandLineParser {
...
+  }
+
+  /**
+   * Peek at the -q/--quiet mode flag, before the command line parser is executed.
+   *
+   * Used during Rush startup.
+   */
+  public static peekIsQuiet(): boolean {
+    const flags: string[] = process.argv.slice(2);
+    return flags.includes('-q') || flags.includes('--quiet');
```
dmichon-msft, line 159: Since this runs on literally every boot of the process it may be worth not allocating a new array and iterating only once.

dmichon-msft (approved):

apps/rush-lib/src/cli/RushCommandLineParser.ts
elliot-nelson: Converted to a single for loop and re-tested the 4 basic entry points.

apps/rush-lib/src/cli/RushCommandLineParser.ts
```diff
@@ -145,6 +145,34 @@ export class RushCommandLineParser extends CommandLineParser {
...
+  public get isQuiet(): boolean {
+    return this._quietParameter.value;
+  }
+
+  /**
+   * Peek at the -q/--quiet mode flag, before the command line parser is executed.
+   *
+   * Used during Rush startup.
+   */
+  public static peekIsQuiet(): boolean {
```
iclanton, line 157: It kinda feels like this should be private.

apps/rush-lib/src/cli/RushCommandLineParser.ts
```diff
@@ -145,6 +145,34 @@ export class RushCommandLineParser extends CommandLineParser {
...
+  }
+
+  /**
+   * Utility to determine if the app should restrict writing to the console.
+   */
+  public static shouldRestrictConsoleOutput(): boolean {
+    return (
+      CommandLineHelper.isTabCompletionActionRequest(process.argv) ||
+      RushCommandLineParser.peekIsQuiet() ||
+      process.argv.indexOf('--json') !== -1
```
iclanton, line 172: Can this be included in the `peekIsQuiet` loop? Maybe rename that function.

common/changes/@microsoft/rush/enelson-quiet_2022-03-21-16-19.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Add support for -q/--quiet to install-run-rush scripts",
```
iclanton, line 5:
```suggestion
      "comment": "Add support for suppressing startup information to invocations of `rush`, `rushx`, and the `install-run-rush` scripts.",
```

iclanton (approved):

apps/rush-lib/src/cli/RushCommandLineParser.ts
elliot-nelson: 🤔 I was trying to keep it as a separate, useful method, but I suppose we'd really prefer that if an actual Action wants to know if quiet mode is on, they can ask the (instantiated and evaluated) RushCommandLineParser instance.

I think I'll collapse it all into the single function and make it as tight as possible.

iclanton (approved):

PR #3289, opened 2022-03-22
[api-extractor] Type members

chadhietala:
## Summary

This supersedes #3003 but only does so for simple type aliases (no type algebra)

## Details

See #3002 

## How it was tested

I extended the example projects to confirm that only simple type aliases get expanded out


chadhietala:
@octogonz any chance you can review this?

iclanton:

chadhietala:
@iclanton or @octogonz any chance you can take a look at this?

apps/api-documenter/src/documenters/MarkdownDocumenter.ts
```diff
@@ -753,9 +753,9 @@ export class MarkdownDocumenter {
   }
 
   /**
-   * GENERATE PAGE: INTERFACE
+   * GENERATE PAGE: ITEM-CONTAINER
    */
-  private _writeInterfaceTables(output: DocSection, apiClass: ApiInterface): void {
+  private _writeItemContainerTables(output: DocSection, apiClass: ApiItemContainerMixin): void {
```
zelliott, line 758: Update the second parameter to be `apiInterfaceOrTypeAlias: ApiInterface|ApiTypeAlias`? (I know that the parameter name being incorrect is preexisting).

apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -934,6 +934,13 @@ export class ApiModelGenerator {
 
       parentApiItem.addMember(apiTypeAlias);
     }
+
+    if (
```
zelliott, line 938: I'm not 100% confident about this, but don't we only want to process child declarations if `apiTypeAlias === undefined`? (i.e. move this chunk into the `if` block above)

apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -934,6 +934,13 @@ export class ApiModelGenerator {
...
+      ts.isTypeAliasDeclaration(astDeclaration.declaration) &&
+      ts.isTypeLiteralNode(astDeclaration.declaration.type)
```
zelliott, line 940: It's not clear to me from line 940 which type aliases will have their children processed and which will not. I think a comment providing some supported vs not supported examples could go a long way here.

build-tests/api-documenter-test/src/index.ts
```diff
@@ -52,6 +52,28 @@ export function exampleFunction(x: ExampleTypeAlias, y: number): IDocInterface1
...
+  /**
+   * A name example
+   */
+  name: string;
+};
+
+/**
+ * @public
+ */
+export type MyUnionType =
```
zelliott, line 68: This type alias will not have its members documented, correct? If that's the case, can we say so in its doc comment to make it clear to readers what this type alias is testing?

build-tests/api-extractor-test-01/src/TypeLiteral.ts
```diff
@@ -0,0 +1,9 @@
+/**
```
zelliott, line 1: I think the `api-extractor-scenarios/apiItemKinds` test project could be a better fit for this test case. In the past, I've used the `api-extractor-scenarios` project to hold different test cases that don't require a standalone project set up. We could create a new file called `typeAliases.ts` there, and add some supported vs unsupported type alias examples. I think the `api-extractor-test-01` test project tests something very specific (despite its ambiguous name). WDYT?

common/changes/@microsoft/api-extractor-model/type-members_2021-11-03-08-38.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor-model",
+      "comment": "Support documenting members of type aliases",
+      "type": "patch"
```
zelliott, line 6: Doesn't this add new functionality and thus this should be `minor` (same for the other two changelog files)? Can we also clarify in the change comment that this only supports documenting members of a very simple shape of type alias?

zelliott (commented):
A few more additional comments:

1. I think we'll need to bump the `.api.json` version in `DeserializerContext.ts`.
2. Can we update the comment at https://github.com/microsoft/rushstack/blob/67963a13fd227ebc6e73323b391bc4e9a8b767f9/libraries/api-extractor-model/src/model/ApiPropertySignature.ts#L20 to reflect that `ApiPropertySignature` items are no longer unique to `ApiInterface` containers? I think there are a few additional API items that need to have comments updated as well (e.g. `ApiMethodSignature`, `ApiConstructSignature`, etc).

PR #3297, opened 2022-03-25
[rush-lib] Fix cache write behavior

### LINKED ISSUE #3287: [rush] Build Cache feature is broken in rush@5.63.0
## Summary

It seems `rush@5.63.0` build cache is totally broken...

## Repro steps

Use rushstack repo

update rush.json to use `"rushVersion": "5.63.0",`

remove the build-cache folder at `common/temp/build-cache`

run `rush build --to rush-lib -v`

and run again.

**Expected result:** 

the second time of `rush build` hit caches.

**Actual result:** 

every projects are logging `This project was not found in the build cache.`, which means local build cache is broken now.

Ideally, local build cache should present under `common/temp/build-cache` but nothing now.

___

When you switch back to `  "rushVersion": "5.62.4",`, everything is fine.
  

## Details

It feels like `Successfully set cache entry.` is not logging. So, it might be related with incorrect value of the `isCacheWriteAllowed` property of `ShellOperationRunner`.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.62.2 |
| `rushVersion` from rush.json? | 5.63.0 |
| Operating system? | MacOS |

dmichon-msft:
## Summary
Fixes cache writes during build.
Fixes #3287 

## Details
The `ShellOperationRunner` (which is currently the only runner that supports the build cache) now defaults cache write eligibility to `true` instead of `false`. Disabling is handled via propagation from upstream operations.

## How it was tested
Ran `rush build`.
Deleted the build cache.
Ran `node ./apps/rush-lib/lib/start.js build`
Ran `node ./apps/rush-lib/lib/start.js build` again and verified that the result came from the build cache

### DIFF
--- a/common/changes/@microsoft/rush/fix-cache-write_2022-03-25-22-29.json
+++ b/common/changes/@microsoft/rush/fix-cache-write_2022-03-25-22-29.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix build cache write eligibility.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/common/changes/@microsoft/rush/fix-cache-write_2022-03-25-22-29.json
+++ b/common/changes/@microsoft/rush/fix-cache-write_2022-03-25-22-29.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Fix build cache write eligibility.",
+      "comment": "Fix an issue where the build cache is never written to.",
       "type": "patch"
     }
   ],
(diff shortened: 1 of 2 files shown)


D4N14L (approved):

common/changes/@microsoft/rush/fix-cache-write_2022-03-25-22-29.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix build cache write eligibility.",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue where the build cache is never written to.",
```

iclanton (approved):

PR #3300, opened 2022-03-28
[rush-lib] Support tags in rush.json and selectors

### LINKED ISSUE #1241: [rush] Generalize "review categories" into tags for project selection
The concept of "categories" or groups of packages is fairly powerful, and has much use outside of just indicating which dependencies are approved for a given group of packages. For example, they could be used to name a group of packages to build on the command line, rather than passing `-t` multiple times. This is the specific situation I'm personally interested in (selecting a subset of packages to build) but I believe there are likely other uses as well.

Additionally, the way the term "category" is used in Rush today requires that a given package be assigned to just a single category. I propose that instead, we use "tags" - the terminology differs only in that tags are implied to be assignable to multiple packages, and that a package may likewise assigned to multiple tags. Set-based operations could then be performed on tagged packages, for example to select a union (packages with tags "production" or "internal"), an intersection (packages with tags "production" and "library"), or a difference (packages with the tag "production" but not "utility").

I propose modifying `rush.json` so that a package can have an array of `tags`.  This would be mutually exclusive with `reviewCategory` which can be kept for backwards compatibility and would be equivalent to `tags: ["myCategory"]`:

```
 "projects": [
    {
      "packageName": "application",
      "projectFolder": "application",
      "tags": ["production", "apps"]
    },
    {
      "packageName": "lib1",
      "projectFolder": "lib1",
      "tags": ["internal", "libs"]
    },
    {
      "packageName": "lib2",
      "projectFolder": "lib2",
      "tags": ["production", "libs"]
    }
  ]
```

I propose introducing the concept of a tag expression, which can be used to select packages by their tags:

A single tag: "production"
A union: "production,internal"
An intersection: "production,&library"
A difference: "production,-utility"

The syntax is totally up for debate, this is just to give an example.

I then propose modifying the behavior of commands which take a package specifier, such as `rush build -t <package>`. These commands would be modified to also accept a tag expression. I feel like simply allowing a tag expression to be used in place of a package name is a better alternative to introducing a new flag, but I'm open to opinions either way. If we do reuse the flag, it might be a good idea to have a prefix to distinguish a tag from just a package that happens to have the same name as a tag.

It might also be worth exploring allowing the user to configure a default tag specifier which would be applied to commands like `rush build`, so that for example, certain packages are excluded by default unless one of their tags is explicitly included.

dmichon-msft:
## Summary
Fixes #1241

## Details
Adds a new array field `tags` to the JSON for a `project` in `rush.json`. This field is a set of identifiers that can be used with the standard set of selection parameters to select projects.

### Example:
```json
"projects": [
    {
      "packageName": "a",
      "projectFolder": "a",
      "tags": ["foo", "bar"]
    },
    {
      "packageName": "b",
      "projectFolder": "b",
      "tags": ["bar", "baz"]
    },
]
```

`rush list --only tag:foo` selects `a`
`rush list --only tag:bar` selects `a` and `b`
`rush list --only tag:baz` selects `b`

This plugs into the existing extensible parser for scoped selector arguments, adding to the existing supported protocols:
- `name:PACKAGE_NAME` - select a project whose name matches `PACKAGE_NAME`. Implied for backwards compatibility if no protocol is specified
- `git:REVISION` - selects all projects that contain code changes since `REVISION`, according to Git. Will include dependency changes in the lockfile.
- `version-policy:POLICY_NAME` - selects all projects that have version policy with name `POLICY_NAME`.
Added missing license headers to parser files.
Fixed a gap in the auto-complete logic for the default `name:` parser.

## How it was tested
Added custom tags "webpack", "rig" and "rush-plugin" into `rush.json` in the @rushstack repository and used them with `rush list --only tag:webpack` (and similar) to select the affected projects.

### DIFF
--- a/apps/rush-lib/src/logic/selectors/TagProjectSelectorParser.ts
+++ b/apps/rush-lib/src/logic/selectors/TagProjectSelectorParser.ts
@@ -7,7 +7,7 @@ import type { RushConfiguration } from '../../api/RushConfiguration';
 import type { RushConfigurationProject } from '../../api/RushConfigurationProject';
 import type { IEvaluateSelectorOptions, ISelectorParser } from './ISelectorParser';
 
-export class VersionPolicyProjectSelectorParser implements ISelectorParser<RushConfigurationProject> {
+export class TagProjectSelectorParser implements ISelectorParser<RushConfigurationProject> {
   private readonly _rushConfiguration: RushConfiguration;
 
   public constructor(rushConfiguration: RushConfiguration) {
@@ -19,25 +19,18 @@ export class VersionPolicyProjectSelectorParser implements ISelectorParser<RushC
     terminal,
     parameterName
   }: IEvaluateSelectorOptions): Promise<Iterable<RushConfigurationProject>> {
-    const selection: Set<RushConfigurationProject> = new Set();
-
-    if (!this._rushConfiguration.versionPolicyConfiguration.versionPolicies.has(unscopedSelector)) {
+    const selection: ReadonlySet<RushConfigurationProject> | undefined =
+      this._rushConfiguration.projectsByTag.get(unscopedSelector);
+    if (!selection) {
       terminal.writeErrorLine(
-        `The version policy "${unscopedSelector}" passed to "${parameterName}" does not exist in version-policies.json.`
+        `The tag "${unscopedSelector}" passed to "${parameterName}" does not exist in rush.json.`
       );
       throw new AlreadyReportedError();
     }
-
-    for (const project of this._rushConfiguration.projects) {
-      if (project.versionPolicyName === unscopedSelector) {
-        selection.add(project);
-      }
-    }
-
     return selection;
   }
 
   public getCompletions(): Iterable<string> {
-    return this._rushConfiguration.versionPolicyConfiguration.versionPolicies.keys();
+    return this._rushConfiguration.projectsByTag.keys();
   }
 }
--- a/apps/rush-lib/src/logic/selectors/TagProjectSelectorParser.ts
+++ b/apps/rush-lib/src/logic/selectors/TagProjectSelectorParser.ts
@@ -23,7 +23,7 @@ export class TagProjectSelectorParser implements ISelectorParser<RushConfigurati
       this._rushConfiguration.projectsByTag.get(unscopedSelector);
     if (!selection) {
       terminal.writeErrorLine(
-        `The tag "${unscopedSelector}" passed to "${parameterName}" does not exist in rush.json.`
+        `The tag "${unscopedSelector}" passed to "${parameterName}" is not specified for any projects in rush.json.`
       );
       throw new AlreadyReportedError();
     }
--- a/common/changes/@microsoft/rush/rush-tags_2022-03-28-20-28.json
+++ b/common/changes/@microsoft/rush/rush-tags_2022-03-28-20-28.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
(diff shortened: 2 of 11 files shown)


NickDarvey:
@dmichon-msft, I'm guessing this is the first step (of many) toward [this](https://github.com/pnpm/pnpm/discussions/4483#discussioncomment-2455852)?

dmichon-msft:
@NickDarvey This PR isn't relevant to version groups for dependencies. My current focus is on multi-project watch scenarios, and tags provide an easy mechanism for developers to group together related projects that they want to watch all at once without having to build complex selector queries.

elliot-nelson:
👍 Love the tags, we've been talking about that on our team for a while.

My only comment about the tags is that until (someday?) it's possible to combine selector parameters in away that produces an intersection, rather than a union, tags will only be useful if all the context is baked in.

(Since you can't ask for the set of projects "A && B", you end up needing tags to represent "A with B" and "A without B", causing an explosion of tags if your team buys into them.)

No need to solve this problem today but as soon as we let end users create tags, we should at least have a plan for it.

common/changes/@microsoft/rush/rush-tags_2022-03-28-20-28.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Support \"tags\" field in project definitions in rush.json. These may be used to select projects with, for example, \"rush list --only tag:my-custom-tag\".",
```
iclanton, line 5:
```suggestion
      "comment": "Add a \"tags\" field to project definitions in rush.json. These may be used to select projects, for example, \"rush list --only tag:my-custom-tag\".",
```

apps/rush-lib/src/logic/selectors/TagProjectSelectorParser.ts
```diff
@@ -0,0 +1,36 @@
...
+  public async evaluateSelectorAsync({
+    unscopedSelector,
+    terminal,
+    parameterName
+  }: IEvaluateSelectorOptions): Promise<Iterable<RushConfigurationProject>> {
+    const selection: ReadonlySet<RushConfigurationProject> | undefined =
+      this._rushConfiguration.projectsByTag.get(unscopedSelector);
+    if (!selection) {
+      terminal.writeErrorLine(
+        `The tag "${unscopedSelector}" passed to "${parameterName}" does not exist in rush.json.`
```
iclanton, line 26:
```suggestion
        `The tag "${unscopedSelector}" passed to "${parameterName}" is not specified for any projects in rush.json.`
```

iclanton (approved):
Looks good overall.

One potential tweak, though: Would it make sense to have all of the allowed tags specified in one field in `rush.json`, similar to the way review categories are enumerated in the `approvedPackagesPolicy.reviewCategories` field? This prevents someone from mistyping a tag when specifying it for a project.

dmichon-msft:
> One potential tweak, though: Would it make sense to have all of the allowed tags specified in one field in `rush.json`, similar to the way review categories are enumerated in the `approvedPackagesPolicy.reviewCategories` field? This prevents someone from mistyping a tag when specifying it for a project.
I can go either way on this. On the one hand having the field makes it a bit more work to add a tag; on the other hand catching typos is useful. We could easily add this as a follow up and even have it be backwards compatible, with the model that if you don't define an enumeration of allowed tags, you can use anything, but if you do specify such a field, the tags on projects need to match an entry in the list.

iclanton:
@dmichon-msft - I just went ahead and added that field in this PR.

RIP21:
Lovely. I've been waiting for it for soooo long.

Thanks a lot for this.

PR #3302, opened 2022-03-30
[heft-jest-plugin] Add --run-in-band flag

josephjunker:
## Summary

The `--runInBand` flag in Jest can provide a significant speedup when running tests in some circumstances. This PR adds a flag to `heft test` to enable it. It also adds an environment variable, `HEFT_JEST_RUN_IN_BAND` to enable the flag in CI.

## Details

I added this flag to the `JestPlugin` class inside `heft-jest-plugin` because that's where other Jest flags are. I saw that some flags are specified in `apps/heft/src/cli/actions/TestAction.ts`, but that most of them have recently moved into `JestPlugin`.

This change does not break backwards compatibility.

## How it was tested

I ran tests on a single project in a private Rush monorepo, using the `--no-build` flag. Without `--run-in-band` tests take 13-14 seconds. With `--run-in-band` tests take just over 3 seconds. I confirmed that this works when using the `HEFT_JEST_RUN_IN_BAND=1` environment variable as well. I also verified that using the `--debug` flag still causes the tests to run in-band.

### DIFF
--- a/common/changes/@rushstack/heft-jest-plugin/junker-add-run-in-band_2022-03-30-16-10.json
+++ b/common/changes/@rushstack/heft-jest-plugin/junker-add-run-in-band_2022-03-30-16-10.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-jest-plugin",
+      "comment": "Add \"--run-in-band\" flag to trigger Jest's \"--runInBand\" option",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@rushstack/heft-jest-plugin"
+}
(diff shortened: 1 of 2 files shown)


elliot-nelson:
> We already support the `maxWorkers` parameter, and setting `maxWorkers: 1` will cause Jest to run in proc. https://github.com/facebook/jest/blob/main/packages/jest-core/src/testSchedulerHelper.ts#L35-L44

@dmichon-msft This code snippet you linked is kind of baffling because it doesn't reflect what we see locally.

Consistently across all projects, try running `heft test --no-build --maxWorkers=1` and `heft test --no-build --debug`; the debug (run in band) version is 2x-3x faster on my MacOS laptop than the maxWorkers=1 version.

Using Joseph's branch, the `heft test --run-in-band` perf is the same as the `--debug` perf.

🤔 I don't have an immediate explanation though... perhaps this code has changed and the version of Jest in current use by Heft doesn't reflect this code snippet?

josephjunker:
We're working on upgrading to the latest Heft. Once we do, I'll retest and make sure there's no difference in test time with and without the `runInBand` flag.

PR #3305, opened 2022-03-31
[rush-serve-plugin] Implement Rush serve plugin

dmichon-msft:
## Summary
Defines a new rush plugin at `@rushstack/rush-serve-plugin`. This plugin interfaces with watch mode phased commands to set up and express server and serve project outputs subject to project-level configuration.

## Details
Supports:
- [x] Riggable project-level configuration of where to serve project outputs
- [x] Workspace-level configuration to specify which named commands the plugin applies to, and to provide the name of a command line parameter that should define the port number
- [x] Logs the served asset path to the console whenever Rush is waiting for changes, to make it maximally visible.

Not yet supported:
- [ ] Socket-based refresh in response to changes (this might not prove to be feasible, since it involves injecting script into code that is out of the control of the plugin)

## How it was tested
Temporarily configured the rushstack repository to enable the plugin and utilized it with the existing `rush start` command and the `localization-plugin-test-01` project to serve dev and prod outputs. Verified changes picked up on refresh.


rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
```diff
@@ -0,0 +1,134 @@
...
+      if (portParameter) {
+        const rawValue: string | undefined = portParameter.value;
+        const parsedValue: number = rawValue ? parseInt(rawValue, 10) : 0;
+        if (!isNaN(parsedValue)) {
+          requestedPort = parsedValue;
+        }
+      }
+
+      // TODO: Update this to read from a custom parameter with parameterLongName matching portParameterLongName,
+      // if it exists.
```
dmichon-msft, line 72:
```suggestion
```
This has been implemented above.

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
```diff
@@ -0,0 +1,134 @@
...
+      server.listen(requestedPort);
+      await once(server, 'listening');
+
+      const address: AddressInfo | undefined = server.address() as AddressInfo;
+      activePort = address?.port;
+
+      logger.terminal.writeLine(`Content is being served from:\n  https://localhost:${activePort}/`);
+      if (portParameter) {
+        // Hack the value of the parsed command line parameter to the actual runtime value.
+        (portParameter as unknown as { _value?: string })._value = `${activePort}`;
```
dmichon-msft, line 124: I don't like having to do this, but the CLI parameter contracts only define a getter for the `value` property, I guess under the assumption that there isn't a use case for mutating the values after parse time. However, there is utility to it here on account of the values being read back out and forwarded.

rush-plugins/rush-serve-plugin/src/RushProjectServeConfigFile.ts
```diff
@@ -0,0 +1,73 @@
...
+import type { RushConfigurationProject } from '@rushstack/rush-sdk';
+
+export interface IRushProjectServeJson {
+  routing: IRoutingRuleJson[];
+}
+
+export interface IRoutingRuleJson {
+  projectRelativeFolder: string;
+  servePath: string;
+  immutable?: boolean;
```
iclanton, line 18: This isn't included in the schema.

rush-plugins/rush-serve-plugin/src/schemas/rush-serve-plugin-options.schema.json
```diff
@@ -0,0 +1,29 @@
...
+      "type": "array",
+      "description": "Rush commands for which files should be served. Serving will be skipped if the command is not running in watch mode.",
+      "items": {
+        "type": "string"
+      }
+    },
+
+    "portParameterLongName": {
+      "type": "string",
+      "description": "The name of a custom parameter in command-line.json that provides a port number for the server.",
```
iclanton, line 25: Are we concerned that someone may already have a `--port` parameter?

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+  RushConfigurationProject,
+  RushSession,
+  IPhasedCommand,
+  Operation,
+  ICreateOperationsContext
+} from '@rushstack/rush-sdk';
+
+import { PLUGIN_NAME } from './constants';
+import { IRoutingRule, RushServeConfiguration } from './RushProjectServeConfigFile';
+import type { CommandLineStringParameter } from '@rushstack/ts-command-line';
```
iclanton, line 23:
```suggestion
import { once } from 'events';
import https from 'https';
import type { AddressInfo } from 'net';
import express from 'express';
import type { CommandLineStringParameter } from '@rushstack/ts-command-line';
import { CertificateManager, ICertificate } from '@rushstack/debug-certificate-manager';
import type {
  ILogger,
  RushConfiguration,
  RushConfigurationProject,
  RushSession,
  IPhasedCommand,
  Operation,
  ICreateOperationsContext
} from '@rushstack/rush-sdk';

import { PLUGIN_NAME } from './constants';
import { IRoutingRule, RushServeConfiguration } from './RushProjectServeConfigFile';
```

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+
+      let requestedPort: number = 0;
+
+      const portParameter: CommandLineStringParameter | undefined = portParameterLongName
+        ? (customParameters.get(portParameterLongName) as CommandLineStringParameter)
+        : undefined;
+      if (portParameter) {
+        const rawValue: string | undefined = portParameter.value;
+        const parsedValue: number = rawValue ? parseInt(rawValue, 10) : 0;
+        if (!isNaN(parsedValue)) {
```
iclanton, line 66: Seems like we should error in this case.

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+    },
+    async (operations: Set<Operation>, context: ICreateOperationsContext) => {
+      if (!context.isInitial || !context.isWatch) {
+        return operations;
+      }
+
+      const certManager: CertificateManager = new CertificateManager();
+
+      const certificate: ICertificate = await certManager.ensureCertificateAsync(
+        true, // Allow generating a new certificate
```
iclanton, line 52: So we're always running in HTTPS mode? Might make sense to make that configurable.

rush-plugins/rush-serve-plugin/src/schemas/rush-serve-plugin-options.schema.json
iclanton: Or is the issue here that Rush plugins don't have a way to add new parameters?

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
iclanton: `ts-command-line` lets you specify a default value. Would that work?

Or maybe we should just expose a function for updating/overriding the parameter value.

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+        // This will cause the resolved value to be forwarded to operations that may use it.
+        (portParameter as unknown as { _value?: string })._value = `${activePort}`;
+      }
+
+      return operations;
+    }
+  );
+
+  command.hooks.waitingForChanges.tap(PLUGIN_NAME, () => {
+    logger.terminal.writeLine(`Content is being served from:\n  https://localhost:${activePort}/`);
```
iclanton, line 129: So will we reprint this whenever Rush has finished rebuilding something? Is that actually useful?

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
```diff
@@ -0,0 +1,131 @@
...
+      activePort = address?.port;
+
+      logger.terminal.writeLine(`Content is being served from:\n  https://localhost:${activePort}/`);
+      if (portParameter) {
+        // Hack the value of the parsed command line parameter to the actual runtime value.
+        // This will cause the resolved value to be forwarded to operations that may use it.
+        (portParameter as unknown as { _value?: string })._value = `${activePort}`;
+      }
+
+      return operations;
```
iclanton, line 124: Doesn't tapable have a thing where if you return `undefined`, it'll treat you as having not done anything with the waterfall value?

iclanton (approved):

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
dmichon-msft: This logic executes after the initial parse (since I read the value from the parameter at first), but before the parameter values get forwarded to the underlying operation runners. Would probably need to expose the function for changing the parameter value.

rush.json
```diff
@@ -993,7 +993,7 @@
       "packageName": "@rushstack/rush-serve-plugin",
       "projectFolder": "rush-plugins/rush-serve-plugin",
       "reviewCategory": "libraries",
-      "shouldPublish": false
+      "shouldPublish": true
```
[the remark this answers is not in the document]
dmichon-msft: The other first party plugins are lockstepped with Rush since Rush embeds them for backwards compatibility. This plugin should only need to have its version bumped when the API surface changes, though that may be equivalent to being lockstepped with Rush.

rush-plugins/rush-serve-plugin/package.json
```diff
@@ -9,20 +9,24 @@
     "directory": "rush-plugins/rush-serve-plugin"
   },
   "main": "lib/index.js",
-  "types": "dist/index.d.ts",
+  "types": "lib/index.d.ts",
```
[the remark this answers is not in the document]
dmichon-msft: It did not, due to it not being the default. Enabling.

common/config/rush-plugins/rush-serve-plugin.json
```diff
@@ -0,0 +1,3 @@
+{
```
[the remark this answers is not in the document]
dmichon-msft: Was left over from feature testing.

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
dmichon-msft: The purpose of this hook is to make this information available whenever Rush is paused waiting for file changes, so that the user doesn't have to hunt for it.

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
dmichon-msft: It does, but the typings don't seem to know about it.

rush-plugins/rush-serve-plugin/src/phasedCommandHandler.ts
dmichon-msft: Most browsers make using raw HTTP very difficult nowadays and we have support for HTTPS. I'd be open to adding it as a feature if anyone asks for it, but it adds complexity.

rush-plugins/rush-serve-plugin/src/schemas/rush-serve-plugin-options.schema.json
dmichon-msft: While being able to specify the port from the command line has some utility, the primary purpose is to allow the auto-assigned port to be forwarded to any operations that need that information, for example if they emit files that contain the local server URL (as our internal repo does). Since this is for forwarding I don't want to make assumptions about the naming convention in this plugin, similar to why the set of command names it is applied to is a plugin config option.

rush-plugins/rush-serve-plugin/src/RushProjectServeConfigFile.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public async loadProjectConfigsAsync(
+    projects: Iterable<RushConfigurationProject>,
+    terminal: ITerminal
+  ): Promise<Iterable<IRoutingRule>> {
+    const rules: IRoutingRule[] = [];
+
+    await Async.forEachAsync(projects, async (project: RushConfigurationProject) => {
+      const rigConfig: RigConfig = await RigConfig.loadForProjectFolderAsync({
```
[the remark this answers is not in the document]
dmichon-msft: RigConfig does caching internally

PR #3309, opened 2022-04-03
[rush] Fix a typo in output of `rush change -v`

yume-chan:
## Summary

Fixed a typo.


iclanton:
Technically, it's currently grammatically correct (see https://www.grammar.com/periods-with-quotation-marks/), but I can see how this may make someone believe that they're expected to run `rush change.` (including the trailing period). I think this tweak makes sense in this case.

iclanton (approved):

PR #3313, opened 2022-04-04
[rush-azure-storage-build-cache-plugin] Add plugin for Azure authentication prompt

dmichon-msft:
## Summary
Adds a separately callable plugin to the `@rushstack/rush-azure-storage-build-cache-plugin` that can be used to refresh Azure credentials at the start of a rush command. Expected usage is to prompt a user immediately if authentication is required to a dev-deployment storage account or similar.

## Details
Leverages the existing infrastructure in the plugin to perform an interactive auth prompt, but as a separately named plugin. Configuration for the storage account is read from the plugin configuration.

Until such time as the Azure Storage build cache plugin is removed from the default install (Rush 6 release), this plugin is enabled by default but doesn't take any action unless a config file is provided.

## How it was tested
Configured the plugin in the repo and bound it to the `rush build` action temporarily. Verified the presence of an auth prompt when running `rush build` but not when running `rush start` (since the plugin was not configured to run when `rush start` was invoked).

### DIFF
--- a/apps/rush-lib/src/pluginFramework/PluginManager.ts
+++ b/apps/rush-lib/src/pluginFramework/PluginManager.ts
@@ -53,8 +53,10 @@ export class PluginManager {
     const builtInPluginConfigurations: IBuiltInPluginConfiguration[] = options.builtInPluginConfigurations;
 
     const ownPackageJsonDependencies: Record<string, string> = require('../../package.json').dependencies;
-    function tryAddBuiltInPlugin(builtInPluginName: string): void {
-      const pluginPackageName: string = `@rushstack/${builtInPluginName}`;
+    function tryAddBuiltInPlugin(builtInPluginName: string, pluginPackageName?: string): void {
+      if (!pluginPackageName) {
+        pluginPackageName = `@rushstack/${builtInPluginName}`;
+      }
       if (ownPackageJsonDependencies[pluginPackageName]) {
         builtInPluginConfigurations.push({
           packageName: pluginPackageName,
@@ -69,6 +71,10 @@ export class PluginManager {
 
     tryAddBuiltInPlugin('rush-amazon-s3-build-cache-plugin');
     tryAddBuiltInPlugin('rush-azure-storage-build-cache-plugin');
+    tryAddBuiltInPlugin(
+      'rush-azure-interactive-auth-plugin',
+      '@rushstack/rush-azure-storage-build-cache-plugin'
+    );
 
     this._builtInPluginLoaders = builtInPluginConfigurations.map((pluginConfiguration) => {
       return new BuiltInPluginLoader({
--- a/apps/rush-lib/src/pluginFramework/PluginManager.ts
+++ b/apps/rush-lib/src/pluginFramework/PluginManager.ts
@@ -71,6 +71,9 @@ export class PluginManager {
 
     tryAddBuiltInPlugin('rush-amazon-s3-build-cache-plugin');
     tryAddBuiltInPlugin('rush-azure-storage-build-cache-plugin');
+    // This is a secondary plugin inside the `@rushstack/rush-azure-storage-build-cache-plugin`
+    // package. Because that package comes with Rush (for now), it needs to get registered here.
+    // If the necessary config file doesn't exist, this plugin doesn't do anything.
     tryAddBuiltInPlugin(
       'rush-azure-interactive-auth-plugin',
       '@rushstack/rush-azure-storage-build-cache-plugin'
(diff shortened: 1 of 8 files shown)


rush-plugins/rush-azure-storage-build-cache-plugin/src/schemas/azure-interactive-auth.schema.json
```diff
@@ -0,0 +1,44 @@
...
+    "storageContainerName": {
+      "type": "string",
+      "description": "(Required) The name of the container in the Azure storage account to use for build cache."
+    },
+
+    "azureEnvironment": {
+      "type": "string",
+      "description": "The Azure environment the storage account exists in. Defaults to AzurePublicCloud.",
+      "enum": ["AzurePublicCloud", "AzureChina", "AzureGermany", "AzureGovernment"]
+    },
```
iclanton, line 26: Shouldn't these just come from the `build-cache.json` config file?

rush-plugins/rush-azure-storage-build-cache-plugin/src/RushAzureInteractiveAuthPlugin.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      const logger: ILogger = rushSession.getLogger(PLUGIN_NAME);
+      logger.terminal.writeLine(`Refreshing credentials`);
+
+      await new AzureStorageBuildCacheProvider({
+        storageAccountName: options.storageAccountName,
+        storageContainerName: options.storageContainerName,
+        azureEnvironment: options.azureEnvironment,
+        isCacheWriteAllowed: true
+      }).updateCachedCredentialInteractiveAsync(logger.terminal);
```
iclanton, line 69: This doesn't check if the credentials are already present and non-expired. This should probably only happen if the credentials actually need to be updated.

rush-plugins/rush-azure-storage-build-cache-plugin/src/schemas/azure-interactive-auth.schema.json
dmichon-msft: These are a separate config because the purpose of this plugin is to be able to reuse the existing Azure auth stack for non-build-cache tasks that need to authenticate to Azure, such as deploying assets to a storage account for testing.

rush-plugins/rush-azure-storage-build-cache-plugin/src/RushAzureInteractiveAuthPlugin.ts
dmichon-msft: Added a configurable minimum expiration time. If not specified will always renew.

apps/rush-lib/src/pluginFramework/PluginManager.ts
```diff
@@ -69,6 +71,10 @@ export class PluginManager {
 
     tryAddBuiltInPlugin('rush-amazon-s3-build-cache-plugin');
     tryAddBuiltInPlugin('rush-azure-storage-build-cache-plugin');
+    tryAddBuiltInPlugin(
```
iclanton, line 74:
```suggestion
    // This is a secondary plugin inside the `@rushstack/rush-azure-storage-build-cache-plugin`
    // package. Because that package comes with Rush (for now), it needs to get registered here.
    // If the necessary config file doesn't exist, this plugin doesn't do anything.
    tryAddBuiltInPlugin(
```

rush-plugins/rush-azure-storage-build-cache-plugin/src/AzureStorageAuthentication.ts
```diff
@@ -96,15 +96,30 @@ export class AzureStorageAuthentication {
...
     await CredentialCache.usingAsync(
       {
         supportEditing: true
       },
       async (credentialsCache: CredentialCache) => {
+        if (expiresOn) {
+          const expiry: Date | undefined = credentialsCache.tryGetCacheEntry(
+            this._credentialCacheId
+          )?.expires;
+          if (expiry && expiry > expiresOn) {
```
iclanton, line 115:
```suggestion
   * @param onlyIfExistingCredentialExpiresAfter - If specified, and a cached credential exists that is still valid
   * after the date specified, no action will be taken.
   */
  public async updateCachedCredentialInteractiveAsync(terminal: ITerminal, onlyIfExistingCredentialExpiresAfter?: Date): Promise<void> {
    await CredentialCache.usingAsync(
      {
        supportEditing: true
      },
      async (credentialsCache: CredentialCache) => {
        if (onlyIfExistingCredentialExpiresAfter) {
          const existingCredentialExpiration: Date | undefined = credentialsCache.tryGetCacheEntry(
            this._credentialCacheId
          )?.expires;
          if (existingCredentialExpiration && existingCredentialExpiration > onlyIfExistingCredentialExpiresAfter) {
```

iclanton (approved):

PR #3318, opened 2022-04-06
[rush] Split dependency analysis out from RushConfiguration.

iclanton:
This a refactor suggested in https://github.com/microsoft/rushstack/pull/3314#discussion_r843197891

### DIFF
--- a/apps/rush-lib/src/api/RushConfiguration.ts
+++ b/apps/rush-lib/src/api/RushConfiguration.ts
@@ -12,7 +12,8 @@ import {
   Path,
   FileSystem,
   PackageNameParser,
-  FileSystemStats
+  FileSystemStats,
+  Import
 } from '@rushstack/node-core-library';
 import { trueCasePathSync } from 'true-case-path';
 
@@ -33,8 +34,13 @@ import { ExperimentsConfiguration } from './ExperimentsConfiguration';
 import { PackageNameParsers } from './PackageNameParsers';
 import { RepoStateFile } from '../logic/RepoStateFile';
 import { LookupByPath } from '../logic/LookupByPath';
-import { DependencyType, PackageJsonDependency } from './PackageJsonEditor';
 import { RushPluginsConfiguration } from './RushPluginsConfiguration';
+import type * as DependencyAnalyzerModuleType from '../logic/DependencyAnalyzer';
+
+const DependencyAnalyzerModule: typeof DependencyAnalyzerModuleType = Import.lazy(
+  '../logic/DependencyAnalyzer',
+  require
+);
 
 const MINIMUM_SUPPORTED_RUSH_JSON_VERSION: string = '0.0.0';
 const DEFAULT_BRANCH: string = 'master';
@@ -495,11 +501,7 @@ export class RushConfiguration {
   private _projectsByName: Map<string, RushConfigurationProject> | undefined;
 
   // variant -> common-versions configuration
-  private _commonVersionsConfigurations: Map<string, CommonVersionsConfiguration> | undefined;
-  // variant -> map of package name -> implicitly preferred version
-  private _implicitlyPreferredVersions: Map<string, Map<string, string>> | undefined;
-  // variant -> map of package name to a set of specified dependency versions
-  private _allDependencyVersions: Map<string, Map<string, Set<string>>> | undefined;
+  private _commonVersionsConfigurationsByVariant: Map<string, CommonVersionsConfiguration> | undefined;
 
   private _versionPolicyConfiguration: VersionPolicyConfiguration;
   private _versionPolicyConfigurationFilePath: string;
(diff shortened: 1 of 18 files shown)


apps/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -33,8 +34,13 @@ import { ExperimentsConfiguration } from './ExperimentsConfiguration';
 import { PackageNameParsers } from './PackageNameParsers';
 import { RepoStateFile } from '../logic/RepoStateFile';
 import { LookupByPath } from '../logic/LookupByPath';
-import { DependencyType, PackageJsonDependency } from './PackageJsonEditor';
 import { RushPluginsConfiguration } from './RushPluginsConfiguration';
+import type * as DependencyAnalyzerModuleType from '../logic/DependencyAnalyzer';
+
+const DependencyAnalyzerModule: typeof DependencyAnalyzerModuleType = Import.lazy(
+  '../logic/DependencyAnalyzer',
+  require
```
dmichon-msft, line 42: If we ever want to bundle Rush, we need a better pattern

apps/rush-lib/src/logic/DependencyAnalyzer.ts
```diff
@@ -0,0 +1,192 @@
...
+  private _rushConfiguration: RushConfiguration;
+  private _analysisByVariant: Map<string, IDependencyAnalysis> = new Map();
+
+  private constructor(rushConfiguration: RushConfiguration) {
+    this._rushConfiguration = rushConfiguration;
+  }
+
+  public static forRushConfiguration(rushConfiguration: RushConfiguration): DependencyAnalyzer {
+    if (!DependencyAnalyzer._dependencyAnalyzerByRushConfiguration) {
+      DependencyAnalyzer._dependencyAnalyzerByRushConfiguration = new Map();
```
dmichon-msft, line 43: If we don't need to be able to enumerate the collection, may as well use a WeakMap for better cleanup.

apps/rush-lib/src/logic/DependencyAnalyzer.ts
```diff
@@ -0,0 +1,192 @@
...
+        //    (Even if it's the only version specifier anywhere in the repo, we still ignore it, because
+        //    otherwise the rule would be difficult to explain.)
+        if (alternativesForThisDependency.indexOf(dependencyVersion) > 0) {
+          ignoreVersion = true;
+        } else {
+          // Is it a local project?
+          const localProject: RushConfigurationProject | undefined =
+            this._rushConfiguration.getProjectByName(dependencyName);
+          if (localProject) {
+            // 2. If it's a symlinked local project, then it's not a candidate, because the package manager will
```
dmichon-msft, line 130: This comment is no longer true when the package manager supports workspaces.

apps/rush-lib/src/logic/DependencyAnalyzer.ts
```diff
@@ -0,0 +1,192 @@
...
+        ...project.packageJsonEditor.devDependencyList
+      ];
+      for (const { name: dependencyName, version: dependencyVersion, dependencyType } of dependencies) {
+        if (dependencyType === DependencyType.Peer) {
+          // If this is a peer dependency, it isn't a real dependency in this context, so it shouldn't
+          // be included in the list of dependency versions.
+          continue;
+        }
+
+        const alternativesForThisDependency: ReadonlyArray<string> =
```
dmichon-msft, line 112: It might be more efficient to do the checks against this collection in a post-processing step rather than while iterating projects. That way we only have to check each used version against the alternatives collection once.

apps/rush-lib/src/logic/DependencyAnalyzer.ts
```diff
@@ -0,0 +1,192 @@
...
+   */
+  private _getImplicitlyPreferredVersionByPackageName(
+    commonVersionsConfiguration: CommonVersionsConfiguration,
+    allVersionsByPackageName: Map<string, Set<string>>
+  ): Map<string, string> {
+    // Only generate implicitly preferred versions for variants that request it
+    const useImplicitlyPreferredVersions: boolean =
+      commonVersionsConfiguration.implicitlyPreferredVersions !== undefined
+        ? commonVersionsConfiguration.implicitlyPreferredVersions
+        : true;
```
dmichon-msft, line 173:
```suggestion
      commonVersionsConfiguration.implicitlyPreferredVersions ?? true
```

apps/rush-lib/src/logic/DependencyAnalyzer.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    if (useImplicitlyPreferredVersions) {
+      // If any dependency has more than one version, then filter it out (since we don't know which version
+      // should be preferred).  What remains will be the list of preferred dependencies.
+      // dependency --> version specifier
+      const implicitlyPreferred: Map<string, string> = new Map<string, string>();
+      for (const [dep, versions] of allVersionsByPackageName) {
+        if (versions.size === 1) {
+          const version: string = Array.from(versions)[0];
```
dmichon-msft, line 182: The old way of doing this using `versions[Symbol.iterator]().next().value` was better for performance and GC pressure. Creating an array just to read the first entry creates a throwaway array and requires calling next() at least twice.

dmichon-msft (commented):
Few things I'd change.

apps/rush-lib/src/logic/DependencyAnalyzer.ts
iclanton: That's how I originally had it written, but Pete argued it was less readable.

apps/rush-lib/src/api/RushConfiguration.ts
iclanton: Yeah we will. This may be less of a problem in a Rush bundle, though.

dmichon-msft (approved):

PR #3322, opened 2022-04-07
[WIP] Refactor FileSystem for better tree-shaking

dmichon-msft:
## Summary
Structures the `FileSystem` collection of functions as a namespace internally and moves each implementation to its own file so that ESM-aware bundlers (like webpack or esbuild) can tree-shake them and split them into separate async chunks if needed, while preserving the API contract.

## Details
Each implementation receives its own file, which both reduces the amount of code that needs to be loaded for unit tests and allows ESM bundlers like webpack or esbuild to separate implementations into different chunks if they are used at different times.

## How it was tested
api-extractor is currently not happy.
Rest of the package (which includes use of FileSystem) builds fine without changes.


iclanton:

ThomasMichon (approved):

Issue #3333, opened 2022-04-08
[rush] using `watchForChanges` when `rush.json` is not in repoRoot, results in `ENOENT: no such file or directory`


alber70g:
## Summary

I've upgraded my Rush version from `5.56.0` to `5.64.0`
After that, the project assumes `rush.json` being in repo root, which it is not.
In my case, I have `<repoRoot>/app/rush.json` with `<repoRoot>/app/rush/lib/lib-x`

The error I get is: `Error: ENOENT: no such file or directory, watch '/Users/albert/projects/websend/frontend/app/app/rush/tools/eslint-plugin'`
As you can see it includes **`app/app`**, twice `app` instead of once

The full stack trace below

## Repro steps

Create a repo with `rush.json` in a sub-directory

  **Expected result:** run watcher for the files properly

  **Actual result:**  Stack trace
```sh
Error: ENOENT: no such file or directory, watch '/Users/<user>/projects/websend/frontend/app/app/rush/tools/eslint-plugin'
    at FSWatcher.<computed> (node:internal/fs/watchers:244:19)
    at Object.watch (node:fs:2247:34)
    at addWatcher (/Users/<user>/.rush/node-v16.13.1/rush-5.64.0/node_modules/@microsoft/rush-lib/lib/logic/ProjectWatcher.js:116:36)
    at watchedResult (/Users/<user>/.rush/node-v16.13.1/rush-5.64.0/node_modules/@microsoft/rush-lib/lib/logic/ProjectWatcher.js:162:17)
    at new Promise (<anonymous>)
    at ProjectWatcher.waitForChange (/Users/<user>/.rush/node-v16.13.1/rush-5.64.0/node_modules/@microsoft/rush-lib/lib/logic/ProjectWatcher.js:79:37)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async PhasedScriptAction._runWatchPhases (/Users/<user>/.rush/node-v16.13.1/rush-5.64.0/node_modules/@microsoft/rush-lib/lib/cli/scriptActions/PhasedScriptAction.js:194:48)
    at async PhasedScriptAction.runAsync (/Users/<user>/.rush/node-v16.13.1/rush-5.64.0/node_modules/@microsoft/rush-lib/lib/cli/scriptActions/PhasedScriptAction.js:147:13)
[...]

PR #3334, opened 2022-04-08
fix(watcher): use rootDir instead of rushJsonFolder (#3333)

alber70g:
## Summary

Fix for #3333 

I know this isn't complete 😇  It is probably not the right change even, because it is using the private property `_data` in `previousState._data.rootDir`.

But, can someone take over from here? 🙏


iclanton:

dmichon-msft:
Fixed in #3345

PR #3340, opened 2022-04-09
[api-extractor] Explicitly expose when a type parameter is optional

zelliott:
## Summary

Fixes https://github.com/microsoft/rushstack/issues/3266.

## Details

See https://github.com/microsoft/rushstack/issues/3266#issuecomment-1090770831 for implementation details. I didn't add any changes to api-documenter yet, however, as currently type parameters do not have a Markdown table within which to place "optional" information (instead they are solely rendered within the method/function signature, [see here](build-tests/api-documenter-test/etc/markdown/api-documenter-test.docclass1.genericwithconstraintanddefault.md)). I'm happy to add a change to api-documenter as well if we want it, but it might make sense for it to be in a separate PR as it could be a larger change.

## How it was tested

I amended the api-documenter-test test scenario with a method with a type parameter with type constraints and a default type. I then ran `rush rebuild` on the entire project to validate that the generated `.api.json` files all looked correct. I also debugged execution to validate that the `isOptional` property was set correctly.


libraries/api-extractor-model/src/model/TypeParameter.ts
```diff
@@ -78,12 +79,18 @@ export class TypeParameter {
    */
   public name: string;
 
+  /**
+   * Whether the type parameter is optional. True IFF there exists a `defaultTypeExcerpt`.
+   */
+  public isOptional: boolean;
```
iclanton, line 85: It seems like knowing what the default type is would also be relevant to documentation. If the constraint is something simpleish like `extends HTMLElement` and the default type is something pretty specific like `HTMLCanvasElement`, that's worth calling out in documentation.

iclanton: Actually nevermind, looks like that information is already there.

iclanton (approved):

PR #3346, opened 2022-04-12
[rush] Fix install-run-rush --help.

iclanton:
## Summary

`install-run-rush --help` currently throws an exception and does not install Rush. This broke one of rush-sdk's scenarios.

## Details

This fix adds a special case for the `-h` and `--help` flags.

## How it was tested

Ran `install-run-rush --help` and confirmed that it installs rush-lib.

### DIFF
--- a/libraries/rush-lib/src/scripts/install-run-rush.ts
+++ b/libraries/rush-lib/src/scripts/install-run-rush.ts
@@ -84,7 +84,7 @@ function _run(): void {
       };
     } else if (!arg.startsWith('-') || arg === '-h' || arg === '--help') {
       // We either found something that looks like a command (i.e. - doesn't start with a "-"),
-      // or we found the -h/--help flag flag, which can be run without a command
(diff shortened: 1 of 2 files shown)


dmichon-msft (approved):

libraries/rush-lib/src/scripts/install-run-rush.ts
```diff
@@ -82,10 +82,10 @@ function _run(): void {
         info: () => {},
         error: console.error
       };
-    }
-    if (!arg.startsWith('-')) {
+    } else if (!arg.startsWith('-') || arg === '-h' || arg === '--help') {
+      // We either found something that looks like a command (i.e. - doesn't start with a "-"),
+      // or we found the -h/--help flag flag, which can be run without a command
```
iclanton, line 87:
```suggestion
      // or we found the -h/--help flag, which can be run without a command
```

PR #3358, opened 2022-04-18
[webpack-preserve-dynamic-require-plugin] Add a new plugin useful for node builds on webpack 5

dmichon-msft:
## Summary
A number of node projects, such as Heft include dynamic calls to "require" for which the target is based on the end-user's local filesystem layout. Adds a new plugin that preserves such calls instead of allowing webpack to try to resolve them at build time.

## Details
The plugin, as implemented, is a blanket removal of "context require" calls in favor of leaving the raw "require" call in place.

## How it was tested
Added a snapshot test that runs webpack with the plugin on a dynamic require call and verifies that the output contains a dynamic require call.


TheLarkInn:
@dmichon-msft There is a community package that might be worth looking into for spoofing dynamic requires: https://github.com/sindresorhus/require-fool-webpack that can save us some work here.

dmichon-msft:
The problem with that tool (and `__non_webpack_require__`) are that they require us to change the underlying source code. This plugin works for scenarios where that is not an option (e.g. is in an external dependency) and ensures that the code still works when invoked directly (i.e. not through webpack compilation).

webpack/preserve-dynamic-require-plugin/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./node_modules/@rushstack/heft-node-rig/profiles/default/tsconfig-base.json",
+  "compilerOptions": {
+    "target": "ES2019",
+    "types": ["heft-jest", "node"],
+    "noImplicitAny": false // Some typings are missing
```
iclanton, line 6: Which ones?

webpack/preserve-dynamic-require-plugin/src/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        cb(err);
+      },
+      writeFile(path: string, content: Buffer, cb: (err?: NodeJS.ErrnoException) => void): void {
+        results.set(path, content.toString('utf8'));
+        cb();
+      }
+    } as unknown as typeof compiler.outputFileSystem;
+
+    compiler.inputFileSystem = inputFs;
+    compiler.outputFileSystem = outputFs;
```
iclanton, line 82: It'd be useful to extract this so it can be used in other plugin unit tests.

webpack/preserve-dynamic-require-plugin/src/index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    await new Promise<void>((resolve: () => void, reject: (err: Error) => void) =>
+      compiler.run((err?: Error | null) => {
+        if (err) {
+          return reject(err);
+        }
+        resolve();
+      })
+    );
```
iclanton, line 91:
```suggestion
    await LegacyAdapters.convertCallbackToPromise(compiler.run);
```

iclanton (approved):

webpack/preserve-dynamic-require-plugin/tsconfig.json
dmichon-msft: copypasta, might not be true.

raghavthind2005 (approved):

webpack/preserve-dynamic-require-plugin/src/index.test.ts
dmichon-msft: Will do in another PR when we start adding more webpack plugin unit tests.

webpack/preserve-dynamic-require-plugin/src/index.test.ts
dmichon-msft: This significantly expands the dependency footprint of the unit test (node-core-library contains a lot of stuff), which adds a lot of overhead to the test execution. Also requires adding another dependency to the package. Leaving as-is.

D4N14L (approved):

PR #3364, opened 2022-04-21
[ts-command-line] Add ScopedCommandLineAction class

D4N14L:
## Summary

Adds the `ScopedCommandLineAction` class, which allows for scoping on commands to influence the parsing of parameters used.

## Details

The `ScopedCommandLineAction` class was implemented using a 2-stage parsing technique. All scoped arguments must be provided as positional arguments specified after the pseudo-argument `--`. This pseudo-argument is _required_ when specifying scoped arguments, and forces all following arguments to be interpreted as positional. These positional arguments are dropped into the `ScopedCommandLineAction.remainder` property, where they are then fed to a secondary parser to validate against the defined scoped commands. This allows for argparse to validate and parse the underlying scoped arguments as if they were provided to a top-level action, and makes help documentation clear and understandable in this scenario.

Underlying scoped arguments can be varied depending on the arguments provided to the owning action by the caller. This allows for two different scopes to have different arguments, eg:

`<tool> <action> --scope A -- --scopedFlag1` works because `--scopedFlag1` is defined only when `--scope` is set to `A`
`<tool> <action> --scope B -- --scopedFlag2` works because `--scopedFlag2` is defined only when `--scope` is set to `B`
`<tool> <action> --scope A -- --scopedFlag2` does _not_ work because `--scopedFlag2` is not defined when `--scope` is set to `A`

Additionally, scopes are not necessarily required on scoped commands (this is left up to the developer to allow or prevent), eg:

`<tool> <action> --scope A` runs the action limited to scope `A`
`<tool> <action>` runs the action against all scopes

As mentioned, `--help` documentation is well-defined and easy to understand, eg.

```console
> example scoped-action --help
usage: example scoped-action [-h] [--scope SCOPE] ...

a longer description

Positional arguments:
  "..."          Scoped parameters. Must be prefixed with "--", ex. "-- 
                 --scopedParameter foo --scopedFlag". For more information 
                 on available scoped parameters, use "-- --help".

Optional arguments:
  -h, --help     Show this help message and exit.

Optional scoping arguments:
  --scope SCOPE  The scope.

> example scoped-action --scope foo -- --help
usage: example scoped-action --scope foo -- [-h] [--scoped-arg SCOPED]

a longer description

Optional arguments:
  -h, --help           Show this help message and exit.
  --scoped-arg SCOPED  The scoped argument.

For more information on available unscoped parameters, use "example 
scoped-action --help"
```

## How it was tested

Added tests and snapshot tests to ensure that validation of parameters works as expected, and that the provided action `onExecute` method gets called with all required parameters populated.


sachinjoseph:
Would a syntax like `--scope A:"--param1 value1 --param2 value2" --scope B:"--param3 value3 --param4 value4"` be a good idea to pass params to multiple `scopes` (`A` and `B` in this case), and while still remaining very explicit about which `scope` to pass the params/values to? To be able to escape quotes (for example, say `value1` contains spaces) may be a challenge though

D4N14L:
> Would a syntax like `--scope A:"--param1 value1 --param2 value2" --scope B:"--param3 value3 --param4 value4"` be a good idea to pass params to multiple `scopes` (`A` and `B` in this case), and while still remaining very explicit about which `scope` to pass the params/values to? To be able to escape quotes (for example, say `value1` contains spaces) may be a challenge though

Hmmm.... The nice part about the approach taken in the PR is that duplicate parameters across different scopes (which provide their own parameters) are handled naturally, since the args would simply be available in the `onExecute` method. It removes the verbosity that would be associated with scoped arguments that are common across most scopes (ex. if scopes A, B, C, D are selected, and A, B, C accept `--arg1` and D does not, it would not be required to pass the arg to each individual scope).

Also, the design is such that you could use the action to implement a single scope, with multiple different scoping parameters progressively narrowing the scope (ex. `--to <A> --except <B> --only-changed` selecting everything up to `A` that has changed, and excluding `B`), or you could use the action to implement N number of scopes (ex. `--to <A> --from <B> --only <C>` selecting everything up to `A`, everything after `B`, and `C`). The way it is currently designed, the action allows the developer the flexibility to choose how these scopes should be used and how scoped parameters should be applied within the selected scopes.

common/reviews/api/ts-command-line.api.md
```diff
@@ -323,4 +329,23 @@ export interface ICommandLineStringDefinition extends IBaseCommandLineDefinition
...
+    // @internal
+    protected _getScopedCommandLineParser(): CommandLineParser;
+    protected onDefineParameters(): void;
+    protected abstract onDefineScopedParameters(scopedParameterProvider: CommandLineParameterProvider): void;
+    protected abstract onDefineUnscopedParameters(): void;
+    protected abstract onExecute(): Promise<void>;
+    get parameters(): ReadonlyArray<CommandLineParameter>;
+    // @internal
+    _processParsedData(parserOptions: ICommandLineParserOptions, data: _ICommandLineParserData): void;
+    static ScopingParameterGroupName: 'scoping';
```
iclanton, line 348: Can we make this constant a `unique symbol`.

libraries/ts-command-line/src/test/CommandLineParameter.test.ts
```diff
@@ -361,31 +361,34 @@ describe(CommandLineParameter.name, () => {
...
 
-    it('raises an error if env var value is not valid json', () => {
+    it('raises an error if env var value is not valid json', async () => {
       const commandLineParser: CommandLineParser = createHelloWorldParser();
       const args: string[] = ['hello-world'];
       process.env.ENV_COLOR = '[u';
 
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
-      expect(commandLineParser.executeWithoutErrorHandling(args)).rejects.toThrowErrorMatchingSnapshot();
+      return expect(
```
iclanton, line 369:
```suggestion
      await expect(
```

libraries/ts-command-line/src/test/ScopedCommandLineAction.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    const commandLineParser: TestCommandLine = new TestCommandLine();
+    const helpText: string = colors.stripColors(
+      commandLineParser.getAction('scoped-action').renderHelpText()
+    );
+    expect(helpText).toMatchSnapshot();
+  });
+
+  it('prints the scoped action help', async () => {
+    const commandLineParser: TestCommandLine = new TestCommandLine();
+    // Execute the parser in order to populate the scoped action
```
iclanton, line 113:
```suggestion
    // Execute the parser in order to populate the scoped action to populate the help text.
```

libraries/ts-command-line/src/providers/ScopedCommandLineAction.ts
```diff
@@ -0,0 +1,231 @@
...
+import type { CommandLineParameterProvider, ICommandLineParserData } from './CommandLineParameterProvider';
+
+interface IInternalScopedCommandLineParserOptions extends ICommandLineParserOptions {
+  readonly actionOptions: ICommandLineActionOptions;
+  readonly unscopedActionParameters: ReadonlyArray<CommandLineParameter>;
+  readonly onDefineScopedParameters: (commandLineParameterProvider: CommandLineParameterProvider) => void;
+  readonly onExecute: () => Promise<void>;
+}
+
+class InternalScopedCommandLineParser extends CommandLineParser {
```
iclanton, line 17: Can you leave a comment here mentioning that this is generated after the first pass of the parser and when it's executed?

libraries/ts-command-line/src/providers/ScopedCommandLineAction.ts
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  protected onDefineParameters(): void {
+    // No-op. Parameters are manually defined in the constructor.
+  }
+
+  protected async onExecute(): Promise<void> {
+    await super.onExecute();
+    // Redirect action execution to the provided callback
+    await this._internalOptions.onExecute();
```
iclanton, line 47: If a consumer's `onExecute` throws, their stacktrace will include `InternalScopedCommandLineParser`. Can we see if there's a way to avoid that?

libraries/ts-command-line/src/providers/ScopedCommandLineAction.ts
```diff
@@ -0,0 +1,231 @@
...
+    super(options);
+
+    this._options = options;
+    this._scopingParameters = [];
+  }
+
+  /**
+   * {@inheritDoc CommandLineParameterProvider.parameters}
+   */
+  public get parameters(): ReadonlyArray<CommandLineParameter> {
```
iclanton, line 94: Note to self: debug through this.

libraries/ts-command-line/src/providers/ScopedCommandLineAction.ts
```diff
@@ -0,0 +1,231 @@
...
+          // model the message off of the built-in "unrecognized arguments" message
+          `${this._unscopedParserOptions.toolFilename} ${this.actionName}: error: Unrecognized ` +
+            `arguments: ${this.remainder.values[0]}.`
+        );
+      }
+      scopedArgs.push(...this.remainder.values.slice(1));
+    }
+
+    // Call the scoped parser using only the scoped args.
+    await this._scopedCommandLineParser.executeWithoutErrorHandling(scopedArgs);
```
iclanton, line 155: If this isn't `-h`/`--help`, seems like we should be able to just run `await this.onExecute()` right after this line.

common/changes/@rushstack/ts-command-line/user-danade-ScopedCommandLine_2022-04-21-22-00.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/ts-command-line",
+      "comment": "Add ScopedCommandLineAction class, which allows for the creation of actions that have variable arguments based on the provided scope.",
```
iclanton, line 5:
```suggestion
      "comment": "Add ScopedCommandLineAction class, which allows for the definition of actions that have dynamic arguments whose definition depends on a provided scope. See https://github.com/microsoft/rushstack/pull/3364",
```

iclanton (approved):

iclanton (approved):

PR #3367, opened 2022-04-23
[heft-jest-plugin] Add --disable-code-coverage option

### LINKED ISSUE #3307: [heft] heft test cannot skip code coverage
## Summary

For a large project which hasn't established or conformed to code coverage requirements yet, the code coverage results output can be significant.  When trying to iterate on a particular test with `heft test --watch` for example, you have to constantly scroll up past what could be pages of output to see what made the test fail.

Code coverage probably also adds time to the test iteration cycle.  (Of course one should assess coverage before finalizing the code, but it may be slowing down test iteration process.)

To work around this, one can add `"coverageReporters": ["none"]` to a jest.config.json.  However, local code changes for this sort of thing are not great (e.g. risk to accidentally commit it, and extra time spent to avoid that).  It would be great if we could accomplish the same thing temporarily with a command-line option like `--skip-code-coverage`.

## Repro steps
* With a project that has unit tests and produces a lot of code coverage report output...
* Break a couple tests.
* `heft test --watch`
* Try to fix a test (make a code change and save it).
* Look at the --watch output change.  Look for what failed your test(s)...

**Expected result:**
Easy to continue iterating on remaining broken tests.

**Actual result:**
Have to constantly scroll up past all the code coverage output to tackle the immediate task.

elliot-nelson:
## Summary

Add a command-line option for Heft that temporarily disables code coverage.

Fixes #3307

## Details

In some cases (for example, when using `--watch` on files that end up running tests in multiple files), the output of the code coverage reporter can be long and get in the way of the test failures the developer is focusing on.

A command-line option allows a developer to temporarily opt-out of code coverage while focusing on fixing, and then they can turn it back on (without the hassle of hand-editing a `jest.config.json` and possibly accidentally checking it in).

## How it was tested

 - Tested locally in projects with and without code coverage turned on
 - Confirmed coverage is ON if configured unless `--disable-code-coverage` is specified

### DIFF
--- a/common/changes/@rushstack/heft-jest-plugin/enelson-skip-code-coverage_2022-04-23-20-23.json
+++ b/common/changes/@rushstack/heft-jest-plugin/enelson-skip-code-coverage_2022-04-23-20-23.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-jest-plugin",
+      "comment": "Add command-line option for temporarily disabling code coverage",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/heft-jest-plugin"
+}
--- a/heft-plugins/heft-jest-plugin/src/JestPlugin.ts
+++ b/heft-plugins/heft-jest-plugin/src/JestPlugin.ts
@@ -66,6 +66,7 @@ export interface IJestPluginOptions {
   configurationPath?: string;
   debugHeftReporter?: boolean;
   detectOpenHandles?: boolean;
+  disableCodeCoverage?: boolean;
   disableConfigurationModuleResolution?: boolean;
   findRelatedTests?: ReadonlyArray<string>;
   maxWorkers?: string;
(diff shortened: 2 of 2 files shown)


heft-plugins/heft-jest-plugin/src/JestPlugin.ts
```diff
@@ -615,6 +620,14 @@ export class JestPlugin implements IHeftPlugin<IJestPluginOptions> {
         ' This corresponds to the "--detectOpenHandles" parameter in Jest\'s documentation.'
     });
 
+    const disableCodeCoverage: IHeftFlagParameter = heftSession.commandLine.registerFlagParameter({
+      associatedActionNames: ['test'],
+      parameterLongName: '--disable-code-coverage',
+      environmentVariable: 'HEFT_JEST_DISABLE_CODE_COVERAGE',
+      description:
+        'Disable any configured code coverage.' + '  If no code coverage is configured, has no effect.'
```
iclanton, line 628:
```suggestion
        'Disable any configured code coverage. If code coverage is not configured, this parameter has no effect.'
```

common/changes/@rushstack/heft-jest-plugin/enelson-skip-code-coverage_2022-04-23-20-23.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-jest-plugin",
+      "comment": "Add command-line option for temporarily disabling code coverage",
```
iclanton, line 5:
```suggestion
      "comment": "Add command-line flag for disabling code coverage.",
```

iclanton (approved):

PR #3368, opened 2022-04-24
Update 'rush build' to only run build, and introduce 'rush test' to build and run tests.

iclanton:


dmichon-msft (approved):

elliot-nelson:
I'm interested in this change! Is this a "our repo will work this way" suggestion, or a "all rush monorepos should work this way" suggestion?

We've had questions from several teams in our monorepo confused by the difference between `rush build` (build and test everything) and `rushx build` (which doesn't test). Lining up build/rebuild and test/retest makes more sense to me than the `build --lite` approach, even if it will take some getting used to for everyone.

(We still haven't made the leap to phased builds however, and it seems like implementing phased builds first is the easiest way to support this split.)

dmichon-msft:
@elliot-nelson This change is specific to the execution of the command in the `@rushstack` repository itself. It may be taken as a suggestion but this PR isn't modifying the default template.

PR #3370, opened 2022-04-25
[rush] Un-deprecate RushConfigurationProject.packageJson.

iclanton:
This removes the `@deprecated` label from `RushConfigurationProject`'s `packageJson` property. This is a useful property for tools integrating with `rush-lib`.


D4N14L (approved):
Agree with this. `PackageJsonEditor` is a nice utility, but it has some limitations. And sometimes you only need the Json.

PR #3372, opened 2022-04-26
Add a -- parameter to git commands to prevent inclusion of arbitrary arguments.

iclanton:
Addresses https://github.com/microsoft/rushstack/issues/3362.


libraries/package-deps-hash/src/getRepoState.ts
```diff
@@ -302,8 +302,8 @@ export function getRepoChanges(
       '--no-commit-id',
       '--cached',
       '-z',
-      revision,
-      '--'
+      '--',
```
dmichon-msft, line 305: For this one the `--` has a special meaning. Revision goes before the `--`.

dmichon-msft (approved):

PR #3376, opened 2022-04-27
[rush-lib] Disable legacy skip if build cache enabled

### LINKED ISSUE #3311: [rush] Fix negative interaction between build cache and legacy incremental build
## Summary

If a project is _skipped_ (legacy incremental build), it disables any downstream project's cache write step, because we can't rely on the state of the build. This makes sense.

However, it negatively interacts with the options `disableBuildCacheForProject` and `disableBuildCacheForOperation` available in `rush-project.json`.  You might use these flags for several reasons:

 - You intend a project to _always_ build (because you know its state cannot be successfully captured in the build cache).
 - The project actually is some kind of integration test, or tests something outside the state of the monorepo.
 - The project validates something inside the monorepo but outside the dependency tree (perhaps something in `common/config`).

Today, enabling either flag _falls back_ to checking for "skips" (legacy incremental build). This potentially interacts very negatively with other projects; instead of the desired behavior of building every time, it will see that the legacy incremental build doesn't need a build and skip it, forcing downstream projects not to write to the local build cache.  If the downstream projects are much bigger than the project in question, it's a net negative for the developer.

## Details

In my opinion, once you've jumped to the build cache, you really don't want incremental builds _at all_.

I would propose that either:

 - (Option 1) In `ShellOperationRunner`, as soon as any attempt (successful or not) is made to read from the build cache, `isSkipAllowed` is set to false for the rest of this build.
 - (Option 2) *Or*, `isSkipAllowed` could be set to false immediately if the build cache config is set up at all.

However, I don't want to discount that legacy incremental build might be useful for certain cases.  For example, your monorepo might contain a project that is _both_ (1) expensive to rebuild _and_ (2) can't be properly cached.  In this case, you'd be forced to disable caching for this project, and my suggestion would inflict the expensive rebuild on local developers instead of an incremental build skip.

 - (Option 3) Add a _second_ project-specific flag in `rush-project.json` that is more broad -- `disableIncrementalBuildForProject` (and `disableIncrementalBuildForOperation`).

With Option 3, you could get either desired behavior:

 - Use `disableIncrementalBuildForProject` to force a project to always rebuild, always paying the cost of Project X but allowing its downstream projects to read and write the build cache properly.
 - Use `disableBuildCacheForProject` to allow a project to skip, but which disables downstream projects from writing to cache.  (Probably best used for expensive, oddball, leaf node projects without significant downstream projects.)

> I think I'm leaning towards Option 3, but it is yet more flag creep.  Open to any opinions on this interaction!

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.64.0 |
| `rushVersion` from rush.json? | 5.64.0 |
| Operating system? | Mac |
| Node.js version (`node -v`)? | 14.18.0 |

dmichon-msft:
## Summary
Fixes #3311 

## Details
Disables the legacy skip behavior if the build cache is enabled.

## How it was tested
Local builds.

### DIFF
--- a/common/changes/@microsoft/rush/disable-skip-if-cache_2022-04-27-19-57.json
+++ b/common/changes/@microsoft/rush/disable-skip-if-cache_2022-04-27-19-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Disable legacy skip logic when build cache is enabled.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
+++ b/libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
@@ -409,6 +409,9 @@ export class ShellOperationRunner implements IOperationRunner {
       this._projectBuildCache = undefined;
 
       if (this._buildCacheConfiguration && this._buildCacheConfiguration.buildCacheEnabled) {
+        // Disable legacy skip logic if the build cache is in play
+        this.isSkipAllowed = false;
+
         const projectConfiguration: RushProjectConfiguration | undefined =
           await RushProjectConfiguration.tryLoadForProjectAsync(this._rushProject, terminal);
         if (projectConfiguration) {


elliot-nelson:
:+1:

I wonder if while we're looking at this, we should also look at issue https://github.com/microsoft/rushstack/issues/3378.

Would this PR fix this issue, or would it still behave incorrectly? (This might be due to a separate, unrelated check early up in the logic that says "if an operation's script is empty or missing, immediately mark it as from cache" -- what I'm unsure of is if A has changes, if that check still allows the changes to show up as a change for B.)

iclanton (approved):

PR #3383, opened 2022-05-02
[rush-lib] Allow phased commands to install prior to execution if requested

dmichon-msft:
## Summary
Allows phased commands to opt into performing a `rush install` prior to execution.

## Details
Adds a new section `installOptions` to phased command definitions. The section contains one option `alwaysInstall: boolean`. If set to `true`, the command will always invoke a full `rush install` (which may detect that nothing needs to be done) prior to execution. If set to `false`, then the command provides an `--install` CLI parameter to perform the installation but does not do so by default.

## How it was tested
Locally added the option to `rush start`, tested with `alwaysInstall: false` and verified that `--install` was available and worked as intended. Tested with `alwaysInstall: true` and verified that `--install` was not available and the installation occurred automatically.

### DIFF
--- a/libraries/rush-lib/src/api/CommandLineConfiguration.ts
+++ b/libraries/rush-lib/src/api/CommandLineConfiguration.ts
@@ -74,7 +74,7 @@ export interface ICommandWithParameters {
 
 export interface IPhasedCommandConfig extends IPhasedCommandWithoutPhasesJson, ICommandWithParameters {
   /**
-   * If set to "true," then this phased command was generated from a bulk command, and
+   * If set to `true`, then this phased command was generated from a bulk command, and
    * was not explicitly defined in the command-line.json file.
    */
   isSynthetic: boolean;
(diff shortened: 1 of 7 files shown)


libraries/rush-lib/src/api/CommandLineConfiguration.ts
```diff
@@ -83,13 +83,18 @@ export interface IPhasedCommandConfig extends IPhasedCommandWithoutPhasesJson, I
...
+   * If set to `true`, this phased command will always run in watch mode, regardless of CLI flags.
    */
   alwaysWatch: boolean;
   /**
    * The set of phases to execute when running this phased command in watch mode.
    */
   watchPhases: Set<IPhase>;
+  /**
+   * If set to `true`, then this phased command will always perform an install before executing, regardless of CLI flags.
+   * If set to `false`, then this phased command will support the "--install" CLI flag.
```
iclanton, line 95:
```suggestion
   * If set to `false`, then the "--install" CLI flag is automatically defined for this phased command.
   * If unset, an install is not automatically performed and the `--install` flag is not automatically defined.
```

libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -113,6 +118,46 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
...
+          maxInstallAttempts: 1,
+          networkConcurrency: undefined
+        }
+      );
+
+      try {
+        await installManager.doInstallAsync();
+      } finally {
+        purgeManager.deleteAll();
+      }
```
iclanton, line 158: It'd be useful to extract this all to a single place.

libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -397,6 +442,16 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
       });
     }
 
+    // If undefined, the parameter is not supported.
```
iclanton, line 445:
```suggestion
    // If undefined, the parameter is not automatically defined.
```

iclanton (approved):

apostolisms (approved):

PR #3395, opened 2022-05-05
[rush] Fix `--strict-peer-dependencies` for PNPM >= 7.0.0

gawlitza:
## Summary

In version 7.0.0, PNPM changed default value for parameter `--strict-peer-dependencies` from `false` to `true`.

Problem: Rush does not behave correctly with `pnpmOptions.strictPeerDependencies` set to false in `rush.json` as then argument `--no-strict-peer-dependencies` is not passed to the call of PNPM if PNPM is used in version >= 7.0.0. As a result, our monorepo could not been built anymore even though we have set `pnpmOptions.strictPeerDependencies` to `false`.

Solution: Ensure that Rush adds argument `--no-strict-peer-dependencies` to the call of PNPM if `pnpmOptions.strictPeerDependencies` is set to false and PNPM >= 7.0.0 is used. Argument `--strict-peer-dependencies` is added if `pnpmOptions.strictPeerDependencies` is set to true and PNPM < 7.0.0 is used.

## How it was tested

I tested it with the same repo in which we have `pnpmOptions.strictPeerDependencies` set to `false`. It was tested once with pnpm version 6.32.10 and once with pnpm version 7.0.0. The command line argument `--strict-peer-dependencies` and `--no-strict-peer-dependencies` are set correctly.

### DIFF
--- a/common/changes/@microsoft/rush/fix-pnpmNoStrictPeerDependenciesOption_2022-05-07-09-46.json
+++ b/common/changes/@microsoft/rush/fix-pnpmNoStrictPeerDependenciesOption_2022-05-07-09-46.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Incorporate the fact that --strict-peer-dependencies is now true by default in PNPM >= 7.0.0.",
+      "comment": "Fix handing of the `strictPeerDependencies` option when using PNPM >= 7.0.0.",
       "type": "none"
     }
   ],
--- a/common/changes/@microsoft/rush/fix-pnpmNoStrictPeerDependenciesOption_2022-05-07-09-46.json
+++ b/common/changes/@microsoft/rush/fix-pnpmNoStrictPeerDependenciesOption_2022-05-07-09-46.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Incorporate the fact that --strict-peer-dependencies is now true by default in PNPM >= 7.0.0.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/libraries/rush-lib/src/api/RushConfiguration.ts
+++ b/libraries/rush-lib/src/api/RushConfiguration.ts
@@ -484,7 +484,6 @@ export class RushConfiguration {
   private _repositoryDefaultRemote: string;
 
   private _npmOptions: NpmOptionsConfiguration;
-  private _pnpmVersion: string;
   private _pnpmOptions: PnpmOptionsConfiguration;
   private _yarnOptions: YarnOptionsConfiguration;
   private _packageManagerConfigurationOptions!: PackageManagerOptionsConfigurationBase;
@@ -582,7 +581,6 @@ export class RushConfiguration {
     this.__rushPluginsConfiguration = new RushPluginsConfiguration(rushPluginsConfigFilename);
 
     this._npmOptions = new NpmOptionsConfiguration(rushConfigurationJson.npmOptions || {});
-    this._pnpmVersion = rushConfigurationJson.pnpmVersion || '';
     this._pnpmOptions = new PnpmOptionsConfiguration(
       rushConfigurationJson.pnpmOptions || {},
       this._commonTempFolder
@@ -1471,13 +1469,6 @@ export class RushConfiguration {
    */
   public get npmOptions(): NpmOptionsConfiguration {
     return this._npmOptions;
-  }
-
-  /**
-   * TODO
-   */
-  public get pnpmVersion(): string {
-    return this._pnpmVersion;
   }
 
   /**
--- a/libraries/rush-lib/src/api/RushConfiguration.ts
+++ b/libraries/rush-lib/src/api/RushConfiguration.ts
@@ -484,6 +484,7 @@ export class RushConfiguration {
   private _repositoryDefaultRemote: string;
 
   private _npmOptions: NpmOptionsConfiguration;
+  private _pnpmVersion: string;
   private _pnpmOptions: PnpmOptionsConfiguration;
   private _yarnOptions: YarnOptionsConfiguration;
   private _packageManagerConfigurationOptions!: PackageManagerOptionsConfigurationBase;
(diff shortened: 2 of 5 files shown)


libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -581,6 +582,7 @@ export class RushConfiguration {
     this.__rushPluginsConfiguration = new RushPluginsConfiguration(rushPluginsConfigFilename);
 
     this._npmOptions = new NpmOptionsConfiguration(rushConfigurationJson.npmOptions || {});
+    this._pnpmVersion = rushConfigurationJson.pnpmVersion || '';
```
D4N14L, line 585: Unnecessary

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -1471,6 +1473,13 @@ export class RushConfiguration {
     return this._npmOptions;
   }
 
+  /**
```
D4N14L, line 1476: Unnecessary

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -574,8 +574,16 @@ export abstract class BaseInstallManager {
         args.push('--network-concurrency', options.networkConcurrency.toString());
       }
 
-      if (this._rushConfiguration.pnpmOptions.strictPeerDependencies) {
-        args.push('--strict-peer-dependencies');
+      if (semver.gte(this._rushConfiguration.pnpmVersion, '7.0.0')) {
```
D4N14L, line 577: use `this._rushConfiguration.packageManagerToolVersion` since at this point you already know that the tool is PNPM.

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -574,8 +574,16 @@ export abstract class BaseInstallManager {
...
+        // pnpm >= 7.0.0 handles peer dependencies strict by default
+        if (this._rushConfiguration.pnpmOptions.strictPeerDependencies === false) {
+          args.push('--no-strict-peer-dependencies');
```
D4N14L, line 580: Update the `rush.schema.json` to specify that the default behaviour is the package manager default. Currently it says that the default behaviour is to set it to `false`.

D4N14L (changes requested):

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
gawlitza: I see. Ok. I was not aware of this. Will do so

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
gawlitza: Got it. Will do so. Thanks.

gawlitza: `rush.schema.json` updated accordingly

D4N14L:
The more I think about it, the more I wonder if we should be specifying `false` by default within Rush, regardless of the version of PNPM used. It seems like a pretty jarring change, especially since many monorepos do not have strict peer dependencies enabled and would be met with a large number of errors on bumping PNPM. @octogonz thoughts?

gawlitza:
I agree. Maybe we should collect the pros and cons of both options. I actually first have implemented the option where the `--no-strict-peer-dependencies` stays the default. 

- Option 1: `--no-strict-peer-dependencies` stays the default.
  - Pro: For people who do not have set `pnpmOptions.strictPeerDependency` explicitly, it continuous to work as this then translates to `--no-strict-peer-dependencies.

- Option 2: Default aligns with PNPM's default.
  - Cons: For people who do not have set `pnpmOptions.strictPeerDependency` explicitly, it builds will fail if they have missing or invalid peer dependencies. 
  - Pro: It is maybe more intuitive to have the default behaviour of pnpm here.
  - Pro: `--strict-peer-dependencies` is the safer default compared to `--no-strict-peer-dependencies`
  - The behaviour of rush with pnpm < 7.0.0 does not change and it is probably also not too big of a deal if the default value changes with pnpm >= 7.0.0. At the end of the day the problem only occurs if you change the major version of pnpm. Additionally, you can resolve the problem by simply setting `pnpmOptions.strictPeerDependency` explicitly to `false`. In our repo we had this set to false explicitly anyway. May this is even not uncommon.

What one could also think about is too make sure that there is an instructive error message so that users understand that they now need to set `pnpmOptions.strictPeerDependency` explicitly to `false` to restore the behaviour they experienced before PNPM < 7.0.0.

libraries/rush-lib/src/schemas/rush.schema.json
```diff
@@ -92,7 +92,7 @@
           "enum": ["local", "global"]
         },
         "strictPeerDependencies": {
-          "description": "If true, then Rush will add the \"--strict-peer-dependencies\" option when invoking PNPM. This causes \"rush install\" to fail if there are unsatisfied peer dependencies, which is an invalid state that can cause build failures or incompatible dependency versions. (For historical reasons, JavaScript package managers generally do not treat this invalid state as an error.) The default value is false.",
+          "description": "If true, then the installation will fail if there is a missing or invalid peer dependency in the tree. The default is the default of PNPM. Note that the default of PNPM changed in version 7 of PNPM. For PNPM version < 7.0.0, the default was false and for PNPM >= 7.0.0 the default is true. If strictPeerDependency is set to true, then Rush will add the \"--strict-peer-dependencies\" option when invoking PNPM version < 7.0.0. If strictPeerDependency is set to false, then Rush will add the \"--no-strict-peer-dependencies\" option when invoking PNPM version < 7.0.0.",
```
iclanton, line 95:
```suggestion
          "description": "If true, then the installation will fail if there is a missing or invalid peer dependency in the tree, which is an invalid state that can cause build failures or incompatible dependency versions. (For historical reasons, JavaScript package managers generally do not treat this invalid state as an error.) This is done via the \"--strict-peer-dependencies\" flag in PNPM version < 7.0.0 and via the \"--no-strict-peer-dependencies\" flag in PNPM >= 7.0.0. The default value is false.",
```

common/changes/@microsoft/rush/fix-pnpmNoStrictPeerDependenciesOption_2022-05-07-09-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Incorporate the fact that --strict-peer-dependencies is now true by default in PNPM >= 7.0.0.",
```
iclanton, line 5:
```suggestion
      "comment": "Fix handing of the `strictPeerDependencies` option when using PNPM >= 7.0.0.",
```

iclanton:
@gawlitza - We should make that change during the next major release of Rush. Otherwise, it'll be a breaking change.

iclanton (approved):

D4N14L (approved):

PR #3401, opened 2022-05-06
[rush] always run `postRushInstall`, and run after writing telemetry

### LINKED ISSUE #3400: [rush] `postRushInstall` hook runs before telemetry file is written
## Summary

The `postRushInstall` hook runs (and is awaited for completion) before telemetry file is written.

This makes it impossible to read the latest telemetry for that install until the next time the script runs.

The `postRushBuild` hook _does_ run after the telemetry has been written.

The [docs](https://rushjs.io/pages/configs/rush_json/) say

```
  /**
   * Rush can collect anonymous telemetry about everyday developer activity such as
   * success/failure of installs, builds, and other operations.  You can use this to identify
   * problems with your toolchain or Rush itself.  THIS TELEMETRY IS NOT SHARED WITH MICROSOFT.
   * It is written into JSON files in the common/temp folder.  It's up to you to write scripts
   * that read these JSON files and do something with them.  These scripts are typically registered
   * in the "eventHooks" section.
   */
```

## Repro steps
Configure a `postRushInstall` hook that runs

```js
'use strict';
const fs = require('fs');
const path = require('path');
const telemetryDir = path.resolve(__dirname, '..', 'temp', 'telemetry');

try {
  console.log('telemetry files', fs.readdirSync(telemetryDir));
} catch(e) {
  console.error(e);
  process.exit(1);
}
```

- Remove the "common/temp/telemetry" folder
- Set `telemetryEnabled` to `true` in the rush.json config
- Run `rush --debug update`.

  **Expected result:** It will log the filename of the telemetry that was just written

  **Actual result:** `Error: ENOENT: no such file or directory, scandir '<snip>/common/temp/telemetry'`

## Details

Should we add a new hook that's used specifically for collecting telemetry? Or should we move the `postRushInstall` to run after the telemetry has been written?

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.68.1 |
| `rushVersion` from rush.json? | 5.68.1 |
| Operating system? | Mac |
| Node.js version (`node -v`)? | v16.14.2 |

tjenkinson:
## Summary

This moves executing the `postRushInstall` hook to after telemetry is written and flushed.

It also changes the logic to run the hook even on a failure. If this would be a breaking change maybe we should add a new hook that is specifically used for reading telemetry files?

fixes #3400

## How it was tested

Tested locally in another project using `node <snip>/rushstack/libraries/rush-lib/lib/start.js --debug update`, and it worked.

### DIFF
--- a/common/changes/@microsoft/rush/fix-post-install-hook_2022-05-06-14-56.json
+++ b/common/changes/@microsoft/rush/fix-post-install-hook_2022-05-06-14-56.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "[rush] always run `postRushInstall`, and run _after_ writing telemetry",
+      "comment": "Update the `postRushInstall` hook to always run, and move its execution to after telemetry is written.",
       "type": "none"
     }
   ],
--- a/common/changes/@microsoft/rush/fix-post-install-hook_2022-05-06-14-56.json
+++ b/common/changes/@microsoft/rush/fix-post-install-hook_2022-05-06-14-56.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "[rush] always run `postRushInstall`, and run _after_ writing telemetry",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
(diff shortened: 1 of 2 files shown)


common/changes/@microsoft/rush/fix-post-install-hook_2022-05-06-14-56.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "[rush] always run `postRushInstall`, and run _after_ writing telemetry",
```
iclanton, line 5:
```suggestion
      "comment": "Update the `postRushInstall` hook to always run, and move its execution to after telemetry is written.",
```

iclanton (approved):

iclanton:
This has been released with Rush v5.69.0

PR #3406, opened 2022-05-09
[rush] Rework the way we get the Git status and update the "rush change" warning to only refer to unstaged changes.

iclanton:
## Summary

The way we currently detect if there are any uncommitted changes is awkward and requires two git invocations. This PR changes that to use `git status --porcelain=2 --null`.

The semantics of `rush change` were also somewhat recently changed to consider a project "changed" if it has staged, but uncommitted, changes. This updates a warning in `rush change` to only be concerned with unstaged changes.

## How it was tested

Tested in a real repo, and added a test case for parsing the git status.


libraries/rush-lib/src/logic/Git.ts
```diff
@@ -368,26 +425,314 @@ export class Git {
...
+
+    return false;
+  }
+
+  public hasUnstagedChanges(): boolean {
+    const gitStatusEntries: ReadonlyArray<IGitStatusEntry> = this.getGitStatus();
+    for (const gitStatusEntry of gitStatusEntries) {
+      if (
+        gitStatusEntry.kind === 'untracked' ||
+        (gitStatusEntry as IChangedGitStatusEntry).unstagedChangeType
```
iclanton, line 443:
```suggestion
        (gitStatusEntry as IChangedGitStatusEntry).unstagedChangeType !== undefined
```

libraries/rush-lib/src/logic/Git.ts
```diff
@@ -368,26 +425,307 @@ export class Git {
...
+    const output: string = this._executeGitCommandAndCaptureOutput(gitPath, [
+      'status',
+      '--porcelain=2',
+      '--null'
+    ]);
+
+    // State machine for parsing a git status entry
+    // See reference https://git-scm.com/docs/git-status?msclkid=1cff552bcdce11ecadf77a086eded66c#_porcelain_format_version_2
+
+    const enum GitStatusEntryState {
```
iclanton, line 483:
```suggestion
    enum GitStatusEntryState {
```

libraries/rush-lib/src/logic/Git.ts
```diff
@@ -368,26 +425,307 @@ export class Git {
...
   public getTagSeparator(): string {
     return this._rushConfiguration.gitTagSeparator || DEFAULT_GIT_TAG_SEPARATOR;
   }
 
+  public *getGitStatus(): Iterable<IGitStatusEntry> {
+    const gitPath: string = this.getGitPathOrThrow();
+    // See Git.test.ts for example output
+    const output: string = this._executeGitCommandAndCaptureOutput(gitPath, [
+      'status',
+      '--porcelain=2',
```
dmichon-msft, line 476: I'd recommend using `--no-renames` to simplify parsing, and expressly pass `--untracked-files=normal` even if it is the default.

dmichon-msft: Also probably `--ignored=no`

libraries/rush-lib/src/logic/Git.ts
```diff
@@ -368,26 +425,307 @@ export class Git {
...
+          break;
+        }
+
+        case GitStatusEntryState.stage3ObjectName: {
+          (currentObject as IUnmergedGitStatusEntry).stage3ObjectName = getFieldAndAdvancePos(' ');
+          state = GitStatusEntryState.path;
+          break;
+        }
+
+        case GitStatusEntryState.renameOrCopyScore: {
```
dmichon-msft, line 686: With `--no-renames` this field will never show up.

libraries/rush-lib/src/logic/Git.ts
```diff
@@ -544,7 +865,46 @@ export class Git {
...
+      }
+
+      case 'R': {
+        return 'renamed';
+      }
+
+      case 'C': {
+        return 'copied';
+      }
+
```
dmichon-msft, line 893: With `--no-renames` these won't happen.

libraries/rush-lib/src/logic/test/Git.test.ts
```diff
@@ -32,4 +33,144 @@ describe(Git.name, () => {
...
+      expect(
+        getGitStatusEntriesForCommandOutput([
+          // Staged add
+          '1 A. N... 000000 100644 100644 0000000000000000000000000000000000000000 a171a25d2c978ba071959f39dbeaa339fe84f768 path/a.ts',
+          // Modifications, some staged and some unstaged
+          '1 MM N... 100644 100644 100644 d20c7e41acf4295db610f395f50a554145b4ece7 8299b2a7d657ec1211649f14c85737d68a920d9e path/b.ts',
+          // Unstaged deletion
+          '1 .D N... 100644 100644 000000 3fcb58810c113c90c366dd81d16443425c7b95fa 3fcb58810c113c90c366dd81d16443425c7b95fa path/c.ts',
+          // Ignored file
+          '! ignored.log',
```
dmichon-msft, line 62: We can exclude these via CLI flag

libraries/rush-lib/src/logic/test/Git.test.ts
```diff
@@ -32,4 +33,144 @@ describe(Git.name, () => {
...
+      return Array.from(gitInstance.getGitStatus());
+    }
+
+    it('parses a git status', () => {
+      expect(
+        getGitStatusEntriesForCommandOutput([
+          // Staged add
+          '1 A. N... 000000 100644 100644 0000000000000000000000000000000000000000 a171a25d2c978ba071959f39dbeaa339fe84f768 path/a.ts',
+          // Modifications, some staged and some unstaged
+          '1 MM N... 100644 100644 100644 d20c7e41acf4295db610f395f50a554145b4ece7 8299b2a7d657ec1211649f14c85737d68a920d9e path/b.ts',
```
dmichon-msft, line 58: Could you throw in a file with AM status as well?

dmichon-msft (commented):
Some opportunities to simplify

libraries/rush-lib/src/logic/Git.ts
```diff
@@ -368,26 +425,307 @@ export class Git {
...
+      renameOrCopyScore,
+      path
+    }
+
+    let pos: number = 0;
+    let state: GitStatusEntryState | undefined = GitStatusEntryState.mode;
+
+    let isRenamedOrCopied: boolean = false;
+    let isUnmerged: boolean = false;
+    let currentObject: Partial<IGitStatusEntry> = {};
```
dmichon-msft, line 507: For large diffs the pattern of creating an empty object and adding fields to it one-by-one is going to degrade performance significantly vs. switching on the mode field and running a parser that creates an object literal directly.

dmichon-msft: Also, 4 different functions:
- `parseUntrackedEntry`
- `parseAddModifyOrDeleteEntry`
- `parseCopyOrRenameEntry`
- `parseUnmergedEntry`
is a lot easier to read than the state machine

libraries/rush-lib/src/logic/Git.ts
iclanton: I like leaving the renames here, in case this gets used for something more complex in the future.

dmichon-msft (approved):

PR #3408, opened 2022-05-09
[rush] After execute operations hook

dmichon-msft:
## Summary
Adds a new `afterExecuteOperations` hook to phased command execution. Uses the new hook to power the `--timeline` flag and to print the default result summary.

## Details
Adds a new `afterExecuteOperations` hook that provides the results of all executed commands and the overall execution result.
Refactors support for `--timeline` into a `ConsoleTimelinePlugin` that uses the new hook.
Refactors the default results summary into an `OperationResultSummaryPlugin` that uses the new hook.

Note that the timeline rendering has been migrated to be between the individual operation logs and the overall summary, so that failures are still the most immediate visible result at the end of command execution.

Fixes a bug in `Async` that allowed unbounded parallelism in async iterables.
Modifies `MockWritable` in `@rushstack/terminal` to implement the `ITerminalProvider` interface for ease of unit testing.

## How it was tested
Extensive use of local Rush commands.
Snapshot tests.

### DIFF
--- a/common/reviews/api/rush-lib.api.md
+++ b/common/reviews/api/rush-lib.api.md
@@ -287,6 +287,12 @@ export interface IEnvironmentConfigurationInitializeOptions {
     doNotNormalizePaths?: boolean;
 }
 
+// @alpha
+export interface IExecutionResult {
+    readonly operationResults: ReadonlyMap<Operation, IOperationExecutionResult>;
+    readonly status: OperationStatus;
+}
+
 // @beta
 export interface IExperimentsJson {
     buildCacheWithAllowWarningsInSuccessfulBuild?: boolean;
@@ -358,6 +364,14 @@ export class IndividualVersionPolicy extends VersionPolicy {
 
 // @internal
 export interface _INpmOptionsJson extends IPackageManagerOptionsJsonBase {
+}
+
+// @alpha
+export interface IOperationExecutionResult {
+    readonly error: Error | undefined;
+    readonly status: OperationStatus;
+    readonly stdioSummarizer: StdioSummarizer;
+    readonly stopwatch: IStopwatchResult;
 }
 
 // @alpha
@@ -444,6 +458,14 @@ export interface IRushSessionOptions {
     getIsDebugMode: () => boolean;
     // (undocumented)
     terminalProvider: ITerminalProvider;
+}
+
+// @alpha
+export interface IStopwatchResult {
+    get duration(): number;
+    get endTime(): number | undefined;
+    get startTime(): number | undefined;
+    toString(): string;
 }
 
 // @beta (undocumented)
@@ -605,6 +627,7 @@ export abstract class PackageManagerOptionsConfigurationBase implements IPackage
 
 // @alpha
 export class PhasedCommandHooks {
+    readonly afterExecuteOperations: AsyncSeriesHook<[IExecutionResult, ICreateOperationsContext]>;
     readonly createOperations: AsyncSeriesWaterfallHook<[Set<Operation>, ICreateOperationsContext]>;
     readonly waitingForChanges: SyncHook<void>;
 }
--- a/libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
+++ b/libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
@@ -33,6 +33,9 @@ import { PhasedOperationPlugin } from '../../logic/operations/PhasedOperationPlu
 import { ShellOperationRunnerPlugin } from '../../logic/operations/ShellOperationRunnerPlugin';
 import { Event } from '../../api/EventHooks';
 import { ProjectChangeAnalyzer } from '../../logic/ProjectChangeAnalyzer';
+import { OperationStatus } from '../../logic/operations/OperationStatus';
+import { IExecutionResult } from '../../logic/operations/IOperationExecutionResult';
+import { OperationResultSummarizerPlugin } from '../../logic/operations/OperationResultSummarizerPlugin';
 
 /**
  * Constructor parameters for BulkScriptAction.
@@ -58,11 +61,11 @@ interface IRunPhasesOptions {
 }
 
 interface IExecutionOperationsOptions {
+  createOperationsContext: ICreateOperationsContext;
   executionManagerOptions: IOperationExecutionManagerOptions;
   ignoreHooks: boolean;
   operations: Set<Operation>;
   stopwatch: Stopwatch;
-  isWatch: boolean;
   terminal: Terminal;
 }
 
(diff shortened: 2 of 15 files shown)


libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -167,6 +171,13 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
     const changedProjectsOnly: boolean = this._isIncrementalBuildAllowed && this._changedProjectsOnly.value;
 
     const terminal: Terminal = new Terminal(this.rushSession.terminalProvider);
+    if (showTimeline) {
+      const { ConsoleTimelinePlugin } = await import('../../logic/operations/ConsoleTimelinePlugin');
+      new ConsoleTimelinePlugin(terminal).apply(this.hooks);
+    }
+    // Do the standard summary
+    new OperationResultSummarizerPlugin(terminal).apply(this.hooks);
```
dmichon-msft, line 179: This should move to before the action hooks in case a plugin wants to schedule its taps relative to these plugins.

libraries/rush-lib/src/logic/operations/ConsoleTimelinePlugin.ts
```diff
@@ -0,0 +1,244 @@
...
+};
+
+interface ITimelineRecord {
+  startTime: number;
+  endTime: number;
+  durationString: string;
+  name: string;
+  status: OperationStatus;
+}
+
```
ThomasMichon, line 79: It's worth having a comment with some example output text. Given that it's ASCII it should be easy to put an inline sample.

libraries/rush-lib/src/logic/operations/ConsoleTimelinePlugin.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  // Start with a newline
+  terminal.writeLine('');
+  terminal.writeLine('='.repeat(maxWidth));
+
+  for (const { startTime, endTime, durationString, name, status } of data) {
+    // Track busy CPUs
+    const openCpu: number = getOpenCPU(startTime);
+    busyCpus[openCpu] = Math.max(startTime, endTime - 10);
```
ThomasMichon, line 183: Explain the 10-ms fudge factor.

libraries/rush-lib/src/logic/operations/ConsoleTimelinePlugin.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  terminal.writeLine('='.repeat(maxWidth));
+
+  //
+  // Format legend and summary areas
+  //
+
+  const usedCpus: number = busyCpus.length;
+
+  const legend: string[] = ['LEGEND:', '  [#] Success  [!] Failed/warnings  [%] Skipped/cached/no-op'];
```
ThomasMichon, line 209: Refer to the constants above?

libraries/rush-lib/src/logic/operations/ConsoleTimelinePlugin.ts
dmichon-msft: Don't actually need this anymore. Turns out there was a bug in Async.forEachAsync (which I fixed both here and in another PR) that was causing the concurrency to exceed the expected value.

dmichon-msft, line 183:
```suggestion
    busyCpus[openCpu] = endTime;
```

D4N14L (approved):

common/reviews/api/rush-lib.api.md
```diff
@@ -419,6 +433,14 @@ export interface _IPnpmOptionsJson extends IPackageManagerOptionsJsonBase {
     useWorkspaces?: boolean;
 }
 
+// @alpha
+export interface IReadonlyStopwatch {
```
iclanton, line 437:
```suggestion
export interface IStopwatchResult {
```

libraries/terminal/src/MockWritable.ts
```diff
@@ -1,17 +1,29 @@
...
-import { AnsiEscape } from '@rushstack/node-core-library';
+import { AnsiEscape, ITerminalProvider, TerminalProviderSeverity } from '@rushstack/node-core-library';
 
 /**
  * A {@link TerminalWritable} subclass for use by unit tests.
  *
  * @beta
  */
-export class MockWritable extends TerminalWritable {
+export class MockWritable extends TerminalWritable implements ITerminalProvider {
```
iclanton, line 13: This is mixing together two currently unrelated APIs.

The `StringBufferTerminalProvider` from `node-core-library` already does most of this for the `ITermalProvider` interface.

common/reviews/api/rush-lib.api.md
iclanton: Or `StopwatchStatus`

libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import type { StdioSummarizer } from '@rushstack/terminal';
+import type { OperationStatus } from './OperationStatus';
+import type { IReadonlyStopwatch } from '../../utilities/Stopwatch';
+import type { Operation } from './Operation';
+
+/**
+ * The `IOperationExecutionResult` interface represents the results of executing an `Operation`.
```
iclanton, line 10:
```suggestion
 * The `IOperationExecutionResult` interface represents the results of executing an {@link Operation}.
```

libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
```diff
@@ -0,0 +1,49 @@
...
+   */
+  readonly stopwatch: IReadonlyStopwatch;
+  /**
+   * Object used to report a summary at the end of the Rush invocation.
+   */
+  readonly stdioSummarizer: StdioSummarizer;
+}
+
+/**
+ * The `IExecutionResult` interface represents the results of executing a set of `Operation`s.
```
iclanton, line 37:
```suggestion
 * The `IExecutionResult` interface represents the results of executing a set of {@link Operation}s.
```

iclanton (approved):

libraries/terminal/src/MockWritable.ts
dmichon-msft: I need a way to route the output from a `Terminal` and a `TerminalWritable` to the same backing store for evaluating the correctness of the output. If you have another avenue to suggest for doing so, I'm open to ideas.

iclanton (approved):

PR #3410, opened 2022-05-10
[node-core-library] Fix Async.forEachAsync concurrency

dmichon-msft:
## Summary
Fix a bug where `Async.forEachAsync` operating on an async iterable would exceed the max concurrency specified in the options.

## Details
The state of "waiting for iterator to return" was not tracked as a concurrent operation, so more instances of the waiter function could be paused at the iterator than the maximum concurrency specified in the original call.

## How it was tested
Affects concurrency of `rush build`. Validated that after the fix was applied that max concurrency of a highly-cached build was capped at the number of CPU cores.

### DIFF
--- a/common/changes/@rushstack/node-core-library/fix-async-concurrency_2022-05-10-00-46.json
+++ b/common/changes/@rushstack/node-core-library/fix-async-concurrency_2022-05-10-00-46.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/node-core-library",
+      "comment": "Fix Async.forEachAsync with an async iterator overflowing the max concurrency.",
+      "type": "patch"
+    }
+  ],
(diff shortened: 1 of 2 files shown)


common/changes/@rushstack/node-core-library/fix-async-concurrency_2022-05-10-00-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/node-core-library",
+      "comment": "Fix Async.forEachAsync with an async iterator overflowing the max concurrency.",
```
iclanton, line 5:
```suggestion
      "comment": "Fix and issue where Async.forEachAsync with an async iterator can overflow the max concurrency",
```

iclanton (approved):

PR #3419, opened 2022-05-16
[rush] Add basic CLI localization to Rush.

iclanton:
This PR introduces an unpublished (for now) Heft plugin that handles localization files the same way our Webpack localization plugin does, and it extracts a few strings from Rush's CLI to an `en.loc.json` file. It also introduces "error codes" for localized error messages and a `RUSH_LOCALE` environment variable.


libraries/rush-lib/src/utilities/RushErrors.ts
```diff
@@ -0,0 +1,84 @@
+import { getFilledCompositeString, strings } from '../loc';
```
dmichon-msft, line 1:
```suggestion
import { getFilledCompositeString, strings } from '../loc/index';
```

dmichon-msft, line 1:
```suggestion
import { getFilledCompositeString, strings } from '../loc/index';
```

libraries/rush-lib/src/loc/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * @remarks
+ * This object should not be exported outside rush-lib.
+ *
+ * @internal
+ */
+export const strings: RushLibStrings = {
+  ...translations,
+  ...enStrings
```
dmichon-msft, line 37:
```suggestion
  ...enStrings,
  ...translations
```
The untranslated strings should be the default, then overwrite with the translations.

dmichon-msft, line 37:
```suggestion
  ...enStrings,
  ...translations
```
The untranslated strings should be the default, then overwrite with the translations.

heft-plugins/heft-node-loc-plugin/src/utilities/Async.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import { ScopedLogger } from '@rushstack/heft';
+
+export class Async {
```
dmichon-msft, line 6: Move to a file called `runWatcherWithErrorHandling` and `export function runWatcherWithErrorHandling`

dmichon-msft, line 6: Move to a file called `runWatcherWithErrorHandling` and `export function runWatcherWithErrorHandling`

heft-plugins/heft-node-loc-plugin/src/schemas/node-loc-plugin.schema.json
```diff
@@ -0,0 +1,43 @@
...
+      "type": "string",
+      "description": "Source code root directory."
+    },
+
+    "generatedTsFolder": {
+      "type": "string",
+      "description": "Output directory for generated typings."
+    },
+
+    "stringsOutputFolders": {
```
dmichon-msft, line 30: Phasing would be simpler if this plugin operated out of a folder other than `src/` and had a single output folder that was referenced via extra `../` in the import path. Emitting multiple `.json` files being emitted just because we have multiple module formats is bad for package sizes. Something to revisit with regards to static assets and the like, as well.

dmichon-msft, line 30: Phasing would be simpler if this plugin operated out of a folder other than `src/` and had a single output folder that was referenced via extra `../` in the import path. Emitting multiple `.json` files being emitted just because we have multiple module formats is bad for package sizes. Something to revisit with regards to static assets and the like, as well.

heft-plugins/heft-node-loc-plugin/src/NodeLocPlugin.ts
```diff
@@ -0,0 +1,132 @@
...
+          for (const [stringName, { value }] of Object.entries(locFile)) {
+            locFileStrings[stringName] = value;
+          }
+
+          const sourceFileRelativePath: string = path.relative(
+            locTypingsConfiguration.srcFolder,
+            locFilePath
+          );
+          const locFileJson: string = JsonFile.stringify(locFileStrings);
+          await Promise.all(
```
dmichon-msft, line 83: There's a `getAdditionalOutputFiles` option in the typings generator to tell it about extra output files to be cleaned up. Should ideally set this to ensure that watch mode handles deleting the loc file.

dmichon-msft, line 83: There's a `getAdditionalOutputFiles` option in the typings generator to tell it about extra output files to be cleaned up. Should ideally set this to ensure that watch mode handles deleting the loc file.

dmichon-msft (changes requested):
Needs a fix in the Rush strings file; order of precedence is backwards.

PR #3427, opened 2022-05-19
[api-extractor] Add support for showing whether properties are readonly

mrshllstock:
### Summary
Fixes https://github.com/microsoft/rushstack/issues/3265

### Details
Adds a new ApiReadonlyMixin mixin similar to the existing ApiStaticMixin mixin. 

This mixin is applied to api-extractor-model as to whether a class member is readonly. This can be via modifier, only having a getter, or TSDoc

### How it was tested
Created a new API extractor scenario that should cover the different ways that something can be readonly and verified outputs with rush rebuild.


apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -653,14 +653,16 @@ export class ApiModelGenerator {
       const apiItemMetadata: ApiItemMetadata = this._collector.fetchApiItemMetadata(astDeclaration);
       const docComment: tsdoc.DocComment | undefined = apiItemMetadata.tsdocComment;
       const releaseTag: ReleaseTag = apiItemMetadata.effectiveReleaseTag;
+      const isReadonly: boolean = this._determineReadonly(astDeclaration);
```
zelliott, line 656: My understanding is that only the `processApiVariable`, `processApiProperty`, `processApiMethod`, `processApiPropertySignature`, and `processApiMethodSignature` methods should need to call `_determineReadonly`, because those are the only entities that can possibly `isReadonly` set to true. What am I missing?

apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -669,6 +671,15 @@ export class ApiModelGenerator {
     this._processChildDeclarations(astDeclaration, exportedName, apiInterface);
   }
 
+  private _determineReadonly(astDeclaration: AstDeclaration): boolean {
+    const apiItemMetadata: ApiItemMetadata = this._collector.fetchApiItemMetadata(astDeclaration);
+    const docComment: tsdoc.DocComment | undefined = apiItemMetadata.tsdocComment;
+    const declarationMetadata: DeclarationMetadata = this._collector.fetchDeclarationMetadata(astDeclaration);
+    return (astDeclaration.modifierFlags & ts.ModifierFlags.Readonly) !== 0
+      || (docComment !== undefined &&  docComment.modifierTagSet.hasTagName('@readonly'))
```
zelliott, line 679: Extra space in `&&  docComment.`

apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -669,6 +671,15 @@ export class ApiModelGenerator {
     this._processChildDeclarations(astDeclaration, exportedName, apiInterface);
   }
 
+  private _determineReadonly(astDeclaration: AstDeclaration): boolean {
+    const apiItemMetadata: ApiItemMetadata = this._collector.fetchApiItemMetadata(astDeclaration);
+    const docComment: tsdoc.DocComment | undefined = apiItemMetadata.tsdocComment;
+    const declarationMetadata: DeclarationMetadata = this._collector.fetchDeclarationMetadata(astDeclaration);
+    return (astDeclaration.modifierFlags & ts.ModifierFlags.Readonly) !== 0
```
zelliott, line 678: I think this logic is complicated enough that it might be worth splitting up into three separate checks, with each check getting a nice doc comment above it explaining the case it's checking.

apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -669,6 +671,15 @@ export class ApiModelGenerator {
     this._processChildDeclarations(astDeclaration, exportedName, apiInterface);
   }
 
+  private _determineReadonly(astDeclaration: AstDeclaration): boolean {
```
zelliott, line 674: Are we handling the `const SOMETHING = 5` case yet?

apps/api-extractor/src/generators/ApiModelGenerator.ts
zelliott, line 679: Suppose someone puts a `@readonly` tag on an item that cannot be readonly (i.e. they use the tag incorrectly). For example:

```
class Example {
  /** @readonly */
  set foo(value: string) { throw new Error('Cannot set foo'); }
}
```

Suppose `set foo` has no corresponding getter. It's obviously not readonly, because there's only a setter. But it has the `@readonly` tag, which means the developer has indicated that it should be displayed as readonly. What should we do in this case?

The logic as we have it will display it as readonly. I think that's _probably_ fine... to defer to the developer's understanding and just let the doc comment tag win. Just thought I'd call out this edge case.

common/changes/@microsoft/api-documenter/api-readonly-mixin_2022-05-19-20-18.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-documenter",
+      "comment": "Adding readonlyMixin for Api Extractor",
```
zelliott, line 5: I think this is minor instead of a patch because it's adding new functionality, not patching a bug fix. WDYT? Also, consider updating the change comments to be more specific/meaningful per package. For example, for api-documenter: "Add support for rendering readonly information next to API items".

libraries/api-extractor-model/src/mixins/ApiReadonlyMixin.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+export interface IApiReadonlyMixinJson extends IApiItemJson {
+  isReadonly: boolean;
+}
+
+const _isReadonly: unique symbol = Symbol('ApiReadonlyMixin._isReadonly');
+
+/**
+ * The mixin base class for API items that can have type parameters.
```
zelliott, line 22: This comment needs to be updated. The "type parameters" piece is incorrect.

libraries/api-extractor-model/src/model/ApiInterface.ts
```diff
@@ -69,7 +71,7 @@ export interface IApiInterfaceJson
  * @public
  */
 export class ApiInterface extends ApiItemContainerMixin(
-  ApiNameMixin(ApiTypeParameterListMixin(ApiReleaseTagMixin(ApiDeclaredItem)))
+  ApiNameMixin(ApiTypeParameterListMixin(ApiReleaseTagMixin(ApiReadonlyMixin(ApiDeclaredItem))))
```
zelliott, line 74: How can an interface be readonly? I think you maybe meant to add this to `ApiPropertySignature` instead?

zelliott (changes requested):
Looking good!

apps/api-extractor/src/generators/ApiModelGenerator.ts
mrshllstock: I'd think the inclusion of the comment that is specifically for documentation would over-ride poor implementation, especially in the context of a documentation generation tool, but in reality the comment has no meaning.

My 2c is that we would still evaluate the comment as isReadonly.

apps/api-extractor/src/generators/ApiModelGenerator.ts
mrshllstock: Reference below on the mixup

apps/api-extractor/src/generators/ApiModelGenerator.ts
mrshllstock: No, will look into adding now.

apps/api-extractor/src/generators/ApiModelGenerator.ts
mrshllstock: Comments have been added. Let me know if more verbosity is desired.

apps/api-extractor/src/generators/ApiModelGenerator.ts
mrshllstock: Addressed

common/changes/@microsoft/api-documenter/api-readonly-mixin_2022-05-19-20-18.json
mrshllstock: Definitely agree it would be minor, I tend to view functionality as a new pathway of using existing code, but should be more liberal with that definition. Let me know if changed verbiage should be reworded.

libraries/api-extractor-model/src/model/ApiInterface.ts
mrshllstock: Yes, when I was reading through the different pieces measured I read through ApiPropertySignature and mistakenly added functionality to interfaces post-fact. Addressed

apps/api-documenter/src/documenters/MarkdownDocumenter.ts
```diff
@@ -997,6 +998,15 @@ export class MarkdownDocumenter {
         new DocPlainText({ configuration, text: ' ' })
       ]);
     }
+    //Should this be under modifer or descriptions?
```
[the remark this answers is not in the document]
mrshllstock: Moved under modifiers

apps/api-extractor/src/generators/ApiModelGenerator.ts
mrshllstock: Added case

apps/api-extractor/src/analyzer/TypeScriptInternals.ts
```diff
@@ -124,4 +124,12 @@ export class TypeScriptInternals {
     }
     return resolver;
   }
+
+  /**
+   * Returns whether a variable has the const declaration
```
zelliott, line 129: nit: "Returns whether a variable is declared with the const keyword."

apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -471,8 +473,9 @@ export class ApiModelGenerator {
       const apiItemMetadata: ApiItemMetadata = this._collector.fetchApiItemMetadata(astDeclaration);
       const docComment: tsdoc.DocComment | undefined = apiItemMetadata.tsdocComment;
       const releaseTag: ReleaseTag = apiItemMetadata.effectiveReleaseTag;
+      const isReadonly: boolean = this._determineReadonly(astDeclaration);
 
-      apiEnum = new ApiEnum({ name, docComment, releaseTag, excerptTokens });
+      apiEnum = new ApiEnum({ name, docComment, releaseTag, excerptTokens, isReadonly });
```
zelliott, line 478: I don't believe enums can be readonly or not, so remove this? The `const enum` doesn't mean that enum is "readonly" (https://www.typescriptlang.org/docs/handbook/enums.html#const-enums). Note that we'll also then want to remove the mixin from the `ApiEnum` item.

apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -709,6 +712,7 @@ export class ApiModelGenerator {
       }
       const isOptional: boolean =
         (astDeclaration.astSymbol.followedSymbol.flags & ts.SymbolFlags.Optional) !== 0;
+      const isReadonly: boolean = this._determineReadonly(astDeclaration);
```
zelliott, line 715: Method declarations cannot be `readonly`, remove? https://www.typescriptlang.org/play?#code/MYGwhgzhAECiAeYC2AHEBTaBvAUNf0ATumACYD2AdiAJ7QBm55AFAJQBc0AbuQJanYAvjmE5elAC7pC9MMEwJkadACZseAsTJVaDJm049+AbhFA. And update the mixin usage accordingly.

apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -765,6 +770,7 @@ export class ApiModelGenerator {
       const releaseTag: ReleaseTag = apiItemMetadata.effectiveReleaseTag;
       const isOptional: boolean =
         (astDeclaration.astSymbol.followedSymbol.flags & ts.SymbolFlags.Optional) !== 0;
+      const isReadonly: boolean = this._determineReadonly(astDeclaration);
```
zelliott, line 773: Same as above, I don't believe method signatures can be readonly. https://www.typescriptlang.org/play?#code/MYGwhgzhAECiAeYC2AHEBTaBvAUNf0ATumACYD2AdiAJ7QBm55AFAJQBc0AbuQJanYAvjmE5elAC7pC9MMEwJkadACZseAsTJVaDJm049+AbhFA

apps/api-extractor/src/generators/ApiModelGenerator.ts
```diff
@@ -1042,4 +1054,18 @@ export class ApiModelGenerator {
     }
     return parameters;
   }
+
+  private _determineReadonly(astDeclaration: AstDeclaration): boolean {
```
zelliott, line 1058: If this `isReadonly` flag only applies to properties, property signatures, and variables, does it make sense to inline this logic within each specific "process" method instead of pulling it out into a helper? I don't feel super strongly either way.

mrshllstock: I'm slightly partial towards a helper since the TSDoc comment and getter logic would apply across the board, but I can separate out if greatly desired.

zelliott (approved):
LGTM, some minor outstanding comments, I think ready for @octogonz to take a look.

apps/api-documenter/src/documenters/MarkdownDocumenter.ts
```diff
@@ -1018,6 +1019,12 @@ export class MarkdownDocumenter {
       }
     }
 
+    if (ApiReadonlyMixin.isBaseClassOf(apiItem)) {
+      if (apiItem.isReadonly) {
+        section.appendNodeInParagraph(new DocCodeSpan({ configuration, code: 'readonly' }));
```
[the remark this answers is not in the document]
mrshllstock: I threw a static readonly property into docclass1 in api-documenter-tests and got this output:

```
|  [staticReadonlyThing](./api-documenter-test.docclass1.staticreadonlything.md) | <code>static</code><code>readonly</code> | boolean |  |
```

I believe this would be satisfactory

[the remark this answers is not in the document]
zelliott: As we discussed, let's pull out the api-documenter changes and I'll include them as part of https://github.com/zelliott/rushstack/pull/2.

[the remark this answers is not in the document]
mrshllstock: Believe this has been resolved

common/changes/@microsoft/api-documenter/api-readonly-mixin_2022-05-19-20-18.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-documenter",
+      "comment": "Add support for rendering readonly information next to API items",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@microsoft/api-documenter"
+}
```
zelliott, line 10: Remove

zelliott (approved):
LGTM

zelliott:
@octogonz This PR should be ready for you to take a look.

common/changes/@microsoft/api-documenter/api-readonly-mixin_2022-05-19-20-18.json
mrshllstock: Sorry for missing before, addressed

octogonz (approved):

octogonz:
@mrshllstock Thanks for contributing this feature!  I made a few minor changes to your branch:

- Resolve merge conflict with #3387
- Include your data change in ApiJsonSchemaVersion.V_1006
- I removed `ApiReadonlyMixin` from `ApiProperty` and `ApiPropertySignature`, because this mixin is already inherited from their base class `ApiPropertyItem`
- Add some API docs for `isReadonly`. The criteria is pretty interesting

PR #3428, opened 2022-05-20
Add webpack5-module-minifier-plugin

dmichon-msft:
## Summary
Adds a new package `@rushstack/webpack5-module-minifier-plugin` to provide equivalent functionality to `@rushstack/module-minifier-plugin` but support `webpack@5`.

## Details
A range of implementation details are subtly different than in the webpack 4 version of the plugin, since webpack 5 has a concept of "inline modules" and the chunk content of the runtime chunk shares a scope with the runtime implementation.

To address this, "inline modules" and "external modules" are both left in the asset during the minification process instead of being minified independently. External modules are trivial, so the impact on minification latency should be minimal, and the inline modules are usually the entry point, to which adding the overhead of the runtime should have a relatively small perf impact.

Unfortunately, this means that module size information is not directly available for inline or external modules, they merely accrue to the chunk size. Detailed accounting would require post-facto analysis via source maps (which is expensive due to source maps being written in terms of lines and columns, rather than byte offsets).

The one feature present in the webpack 4 version but not the webpack 5 version is an alteration to the runtime to make async imports more compact; this may instead be offered to webpack as an optimization for general usage.

## How it was tested
Snapshot unit tests.
Integrated with `heft-webpack5-everything-test` build test project and validated that the output executes in the browser.


webpack/module-minifier-plugin-5/src/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  IModuleMinifierFunction,
+  IWorkerPoolMinifierOptions
+} from '@rushstack/module-minifier';
+export {
+  getIdentifier,
+  LocalMinifier,
+  MessagePortMinifier,
+  NoopMinifier,
+  WorkerPoolMinifier
+} from '@rushstack/module-minifier';
```
iclanton, line 26: I get exporting these from the v4 index for backwards compatibility. In V5, should we instead just have consumers get these from the packages directly?

common/config/rush/common-versions.json
```diff
@@ -101,6 +101,7 @@
     "style-loader": ["~2.0.0"],
     "terser-webpack-plugin": ["~3.0.8"],
     "terser": ["~4.8.0"],
-    "webpack": ["~4.44.2"]
+    "webpack": ["~4.44.2"],
+    "webpack-sources": ["~1.4.3"]
```
TheLarkInn, line 105: Do we need this still?

webpack/module-minifier-plugin-5/README.md
```diff
@@ -0,0 +1,60 @@
...
+- Handing smaller code chunks to the minifier at a time (AST analysis is superlinear in size of the AST)
+- Even single asset builds will likely still contain multiple modules in the final output, which can be split across available CPU cores
+
+## Use with `[hash]` and `[contenthash]` tokens
+The plugin will do its best to update webpack hashes if changing the direct inputs (`useSourceMap`) to the plugin, but if altering the `minifier` property itself, you may need to use the `output.hashSalt` property to force a change to the hashes, especially if leveraging the `MessagePortMinifier` or similar, since it has no direct access to the configuration of the minifier.
+
+## Parallel execution
+
+```js
+const { ModuleMinifierPlugin, WorkerPoolMinifier } = require('@rushstack/module-minifier-plugin');
```
TheLarkInn, line 22: rename to webpack5-module-minifier-plugin / rework examples

webpack/module-minifier-plugin-5/README.md
```diff
@@ -0,0 +1,60 @@
...
+  ]
+}
+```
+
+## Single-threaded execution
+You can also run the ModuleMinifierPlugin in a single-threaded configuration.
+
+```js
+// webpack.config.js
+const { ModuleMinifierPlugin, LocalMinifier } = require('@rushstack/module-minifier-plugin');
```
TheLarkInn, line 42: rename here too

webpack/module-minifier-plugin-5/src/ModuleMinifierPlugin.ts
```diff
@@ -0,0 +1,523 @@
...
+function isMinificationResultError(
+  result: IModuleMinificationResult
+): result is IModuleMinificationErrorResult {
+  return !!result.error;
+}
+
+// Matche behavior of terser's "some" option
+function isLicenseComment(comment: Comment): boolean {
+  // https://github.com/terser/terser/blob/d3d924fa9e4c57bbe286b811c6068bcc7026e902/lib/output.js#L175
+  return /@preserve|@lic|@cc_on|^\**!/i.test(comment.value);
```
TheLarkInn, line 113: Consider hoisting

webpack/module-minifier-plugin-5/src/ModuleMinifierPlugin.ts
```diff
@@ -0,0 +1,523 @@
...
+  public constructor(options: IModuleMinifierPluginOptions) {
+    this.hooks = {
+      rehydrateAssets: new AsyncSeriesWaterfallHook(['dehydratedContent', 'compilation']),
+
+      postProcessCodeFragment: new SyncWaterfallHook(['code', 'context'])
+    };
+
+    const { minifier, sourceMap } = options;
+
+    this._optionsForHash = {
```
TheLarkInn, line 140: webpack 5 compilation hash should be reliable

webpack/module-minifier-plugin-5/src/ModuleMinifierPlugin.ts
```diff
@@ -0,0 +1,523 @@
...
+          if (!allChunkModules) {
+            // This chunk does not contain javascript modules
+            continue;
+          }
+
+          for (const assetName of chunk.files) {
+            const asset: sources.Source = compilation.assets[assetName];
+
+            // Verify that this is a JS asset
+            if (/\.m?js(\?.+)?$/.test(assetName)) {
```
TheLarkInn, line 422: I think cjs needs to be included. Better way to identify JS file?

TheLarkInn (commented):
Looks good some small nits to look at, otherwise LGTM

iclanton (approved):

webpack/module-minifier-plugin-5/src/ModuleMinifierPlugin.ts
```diff
@@ -0,0 +1,533 @@
...
+            submittedModules.add(hash);
+
+            ++pendingMinificationRequests;
+
+            minifier.minify(
+              {
+                hash,
+                code: wrappedCode,
+                nameForMap: useSourceMaps ? nameForMap : undefined,
+                externals: undefined
```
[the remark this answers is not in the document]
dmichon-msft: Apparently I left the type as required in the interface.

PR #3434, opened 2022-05-25
[API Extractor] Add missing required parameter for getResolvedModule to ensure that modules can resolve accurately with Node16/nodenext support

### LINKED ISSUE #3433: [api-extractor] crash when used with TS 4.7.2 moduleResolution Node16
## Summary

It seems that adding `moduleResolution` with a setting of `Node16` to the `tsconfig.json` file causes `@microsoft/api-extractor` to fail every time. The error is as follows:

```
ERROR: Internal Error: getResolvedModule() could not resolve module name "./foo/bar.js"
/Users/johndoe/Documents/code/myrepo/my-package/dist/dts/index.d.ts:1:1
error Command failed with exit code 1.
```

**This is blocking us from shipping our major new library versions.**

## Repro steps

I've got a branch that demonstrates the issue here: 

https://github.com/microsoft/fast/tree/users/eisenbergeffect/fast-element-utilities

Simply pull that down and run `yarn` in the root. You will see an error similar to the above.

## Details

I've tried many combinations of settings, but the only thing that seems to make the error go away is not using the `Node16` setting. Unfortunately, this is required for some of our packages since they leverage `package.json` `exports`. You can switch the `fast-element` package to `Node` to get it working but then you will observe the problem in `fast-foundation` (which requires `Node16` because it leverages package exports from `fast-element`).

I have discussed this briefly with a member of the TS team and they suspect that api-extractor may be using some undocumented APIs, specifically `getResolvedModule`.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/api-extractor` version? | 7.24.1 |
| Operating system? | Mac |
| API Extractor scenario? | rollups (.d.ts)  |
| TypeScript compiler version? | 4.7.2 |
| Node.js version (`node -v`)? | 16.14.0 |

chrisdholt:
…ypescript can resolve accurately

## Summary
With the addition of `Node16` and `nodenext`, Typescript now requires the mode to be passed in order to accurately resolve modules. This PR adds support for `mode` thanks to the help of @DanielRosenwasser and does so in a way that returns undefined gracefully in case the fn isn't supported in older versions.

This resolves #3433.

## Details
1. Add a third param to getResolvedModule for `mode`
2. Create a new function `getModeForUsageLocation` which maps to the TS equivalent
3. Update all callers to gracefully look for the right mode and return undefined if necessary

## How it was tested
Tested locally with Typescript 4.7.2 and the error is gone, API Extractor succeeds.

### DIFF
--- a/apps/api-extractor/src/analyzer/ExportAnalyzer.ts
+++ b/apps/api-extractor/src/analyzer/ExportAnalyzer.ts
@@ -259,9 +259,18 @@ export class ExportAnalyzer {
     importOrExportDeclaration: ts.ImportDeclaration | ts.ExportDeclaration | ts.ImportTypeNode,
     moduleSpecifier: string
   ): boolean {
+    const specifier: ts.TypeNode | ts.Expression | undefined = ts.isImportTypeNode(importOrExportDeclaration)
+      ? importOrExportDeclaration.argument
+      : importOrExportDeclaration.moduleSpecifier;
+    const mode: ts.StringLiteralLike =
+      specifier && ts.isStringLiteralLike(specifier)
+        ? TypeScriptInternals.getModeForUsageLocation(importOrExportDeclaration.getSourceFile(), specifier)
+        : undefined;
+
     const resolvedModule: ts.ResolvedModuleFull | undefined = TypeScriptInternals.getResolvedModule(
       importOrExportDeclaration.getSourceFile(),
-      moduleSpecifier
+      moduleSpecifier,
+      mode
     );
 
     if (resolvedModule === undefined) {
@@ -863,9 +872,18 @@ export class ExportAnalyzer {
     exportSymbol: ts.Symbol
   ): AstModule {
     const moduleSpecifier: string = this._getModuleSpecifier(importOrExportDeclaration);
+    const mode: ts.StringLiteralLike | undefined =
+      importOrExportDeclaration.moduleSpecifier &&
+      ts.isStringLiteralLike(importOrExportDeclaration.moduleSpecifier)
+        ? TypeScriptInternals.getModeForUsageLocation(
+            importOrExportDeclaration.getSourceFile(),
+            importOrExportDeclaration.moduleSpecifier
+          )
+        : undefined;
     const resolvedModule: ts.ResolvedModuleFull | undefined = TypeScriptInternals.getResolvedModule(
       importOrExportDeclaration.getSourceFile(),
-      moduleSpecifier
+      moduleSpecifier,
+      mode
     );
 
     if (resolvedModule === undefined) {
--- a/apps/api-extractor/src/analyzer/ExportAnalyzer.ts
+++ b/apps/api-extractor/src/analyzer/ExportAnalyzer.ts
@@ -262,7 +262,7 @@ export class ExportAnalyzer {
     const specifier: ts.TypeNode | ts.Expression | undefined = ts.isImportTypeNode(importOrExportDeclaration)
       ? importOrExportDeclaration.argument
       : importOrExportDeclaration.moduleSpecifier;
-    const mode: ts.StringLiteralLike =
+    const mode: ts.ModuleKind.CommonJS | ts.ModuleKind.ESNext | undefined =
       specifier && ts.isStringLiteralLike(specifier)
         ? TypeScriptInternals.getModeForUsageLocation(importOrExportDeclaration.getSourceFile(), specifier)
         : undefined;
@@ -872,7 +872,7 @@ export class ExportAnalyzer {
     exportSymbol: ts.Symbol
   ): AstModule {
     const moduleSpecifier: string = this._getModuleSpecifier(importOrExportDeclaration);
-    const mode: ts.StringLiteralLike | undefined =
+    const mode: ts.ModuleKind.CommonJS | ts.ModuleKind.ESNext | undefined =
       importOrExportDeclaration.moduleSpecifier &&
       ts.isStringLiteralLike(importOrExportDeclaration.moduleSpecifier)
         ? TypeScriptInternals.getModeForUsageLocation(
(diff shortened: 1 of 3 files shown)


EisenbergEffect (approved):
Woohoo! Let's get this in 😄 This is a critical fix needed for us to ship our FAST vNext packages.

apps/api-extractor/src/analyzer/TypeScriptInternals.ts
```diff
@@ -83,12 +83,26 @@ export class TypeScriptInternals {
...
+   * Gets the mode required for module resolution required with the addition of Node16/nodenext
+   */
+  public static getModeForUsageLocation(
+    file: { impliedNodeFormat?: ts.SourceFile['impliedNodeFormat'] },
+    usage: ts.StringLiteralLike | undefined
+  ) {
+    // Compiler internal:
+    // https://github.com/microsoft/TypeScript/blob/v4.7.2/src/compiler/program.ts#L568
+
+    return (ts as any).getModeForUsageLocation(file, usage);
```
DanielRosenwasser, line 105: This function might not exist in older versions. It should just return `undefined` if it doesn't have a value.

```suggestion
    return (ts as any).getModeForUsageLocation?.(file, usage);
```

apps/api-extractor/src/analyzer/TypeScriptInternals.ts
```diff
@@ -83,12 +83,26 @@ export class TypeScriptInternals {
    */
   public static getResolvedModule(
     sourceFile: ts.SourceFile,
-    moduleNameText: string
+    moduleNameText: string,
+    mode: ts.StringLiteralLike | undefined
   ): ts.ResolvedModuleFull | undefined {
     // Compiler internal:
     // https://github.com/microsoft/TypeScript/blob/v3.2.2/src/compiler/utilities.ts#L218
```
DanielRosenwasser, line 90:
```suggestion
    // https://github.com/microsoft/TypeScript/blob/v4.7.2/src/compiler/utilities.ts#L161
```

apps/api-extractor/src/analyzer/TypeScriptInternals.ts
```diff
@@ -83,12 +83,26 @@ export class TypeScriptInternals {
    */
   public static getResolvedModule(
     sourceFile: ts.SourceFile,
-    moduleNameText: string
+    moduleNameText: string,
+    mode: ts.StringLiteralLike | undefined
```
DanielRosenwasser, line 87:
```suggestion
    mode: ts.ModuleKind.CommonJS | ts.ModuleKind.ESNext | undefined
```

apps/api-extractor/src/analyzer/TypeScriptInternals.ts
```diff
@@ -83,12 +83,26 @@ export class TypeScriptInternals {
...
+    return (ts as any).getResolvedModule(sourceFile, moduleNameText, mode);
+  }
+
+  /**
+   * Gets the mode required for module resolution required with the addition of Node16/nodenext
+   */
+  public static getModeForUsageLocation(
+    file: { impliedNodeFormat?: ts.SourceFile['impliedNodeFormat'] },
+    usage: ts.StringLiteralLike | undefined
+  ) {
```
DanielRosenwasser, line 101:
```suggestion
  ): ts.ModuleKind.CommonJS | ts.ModuleKind.ESNext | undefined {
```

DanielRosenwasser (changes requested):

apps/api-extractor/src/analyzer/ExportAnalyzer.ts
```diff
@@ -863,9 +872,18 @@ export class ExportAnalyzer {
...
   ): AstModule {
     const moduleSpecifier: string = this._getModuleSpecifier(importOrExportDeclaration);
+    const mode =
+      importOrExportDeclaration.moduleSpecifier &&
+      ts.isStringLiteralLike(importOrExportDeclaration.moduleSpecifier)
+        ? TypeScriptInternals.getModeForUsageLocation(
+            importOrExportDeclaration.getSourceFile(),
+            importOrExportDeclaration.moduleSpecifier
+          )
+        : undefined;
```
DanielRosenwasser, line 882: Ideally, you could leverage `_getModuleSpecifier` to return the *node* and get the string out of that. It's not critical, I wouldn't block this PR on it.

chrisdholt:
Thanks for the review @DanielRosenwasser - build is passing with updated types! @iclanton, @octogonz, @D4N14L, @apostolisms - Apologies for the ping, but curious if anyone is available to work with us on getting this out the door to unblock our release? Thank you!

apps/api-extractor/src/analyzer/ExportAnalyzer.ts
chrisdholt: Leaving this open to get feedback from maintainers here as well.

DanielRosenwasser (approved):

common/changes/@microsoft/api-extractor/users-chhol-fix-getResolvedModule-params_2022-05-25-17-20.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor",
+      "comment": "add missing required parameter for getResolvedModule to ensure typescript can accurately resolve module kind",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue where API Extractor would fail to run on a project where `\"moduleResolution\"` is set to `\"Node16\"` in `tsconfig.json`",
```

common/changes/@microsoft/api-extractor/users-chhol-fix-getResolvedModule-params_2022-05-25-17-20.json
```diff
@@ -0,0 +1,10 @@
...
+      "type": "minor"
```
iclanton, line 6:
```suggestion
      "type": "patch"
```

iclanton (approved):

DanielRosenwasser (approved):

Issue #3413, opened 2022-05-10
[rush] Publish leaving shrinkwrap in stale state


MichaelSitter:
## Summary

For publish runs using the apply & branch options which release workspace packages, the version bump edits to package.json files are not reflected in the shrinkwrap state. This will cause `rush install` to fail before running `rush update`. The workaround we are using is to omit the branch option to delay pushing the version bump commit and run `rush update`. but I'm a little hazy if this is safe in practice.

Please see zulip thread for original post: 
https://rushstack.zulipchat.com/#narrow/stream/262513-general/topic/publishing.20workspace.20dependencies.20create.20shrinkwrap.20conflict

## Repro steps

1. Setup a repo with two packages, packages A & B. Add package B as a workspace dependency in package A: `"B": "workspace:~1.0.0"`
2. Create a change for package B.
3. Run publish with the branch flag to commit the bump commit generated by apply: `node common/scripts/install-run-rush.js publish --apply -b master --publish`
4. Re-clone or purge the repo and run `rush install`

### Expected

Rush install should succeed after publish.

### Actual

Install fails due to the lockfile not reflecting package.json version bumps.

```
The shrinkwrap file contains the following issues:
  Dependencies of project "<package>" do not match the current shrinkwrap.


The shrinkwrap file is out of date. You need to run "rush update".
```

## Details

It's unclear to me from the publishing documentation if this expected behavior for publish. Our original release pipelines avoided this problem by publishing outside of the default branch, and required developers to get approvals to merge the branch with the publish bump commits. This is when they would run `rush update` and merge the results back to the default branch. But this introduced a bunch of ugly edge cases which proved difficult to address. Should publish just transparently include update? 

## Standard questions

Please answer these questions to help us investigate your issue more quickly:

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.62.3 |
| `rushVersion` from rush.json? | 5.62.3 |
| `useWorkspaces` from rush.json? | yes |
| Operating system? | mac |
| Would you consider contributing a PR? | yes |
| Node.js version (`node -v`)? | v14.17.6

dmichon-msft:
The workaround you are using is, in fact, safe in practice, except insofar as the added latency allows more opportunities for merge conflicts in your branch.

MichaelSitter:
Thanks for getting back on this, we are getting someone from the team to start working on a patch to address this. Our current thinking was to make the behavior opt-in with a new publish flag, `--update`.

[...]

PR #3441, opened 2022-06-02
feat(publish): add update flag to run update pre-publish

ethan-burrell:
## Summary

This PR is proposed as a solution to #3413, where a publish causes the the shrinkwrap in a stale state. 

#### Design considerations:

This causes a problem where our shrinkwrap was out of date. By running an update before `publish` runs, this would keep the shrinkwrap in the correct state.

If user's would like to opt into this feature they can run `rush publish .... --update` to include the update to the shrinkwrap as a part of their publish pipeline.

## Details

- 

## How it was tested

- [ ] Unit tests
- [ ] Integration tests

Integration tests will consist of me attempting to reproduce the problem with the stale shrinkwraps, and replay to the commit before the publish happened and checking that the `--update` flag fixes the publish.


libraries/rush-lib/src/cli/actions/PublishAction.ts
```diff
@@ -418,6 +428,40 @@ export class PublishAction extends BaseRushAction {
...
+        pnpmFilterArguments: [],
+        maxInstallAttempts: 1,
+        networkConcurrency: undefined
+      }
+    );
+
+    try {
+      await installManager.doInstallAsync();
+    } catch {
+      throw new Error(`"update" failed to run before "publish" was run`);
```
iclanton, line 461: This should include the inner error in the error message.

iclanton:
Just confirming - this will commit the updated shrinkwrap, right?

PR #3455, opened 2022-06-08
[localization-utilities] Expose parsers directly

dmichon-msft:
## Summary
Removes the dependency on `decache` in favor of leveraging `vm.runInNewContext` to run each pseudolocalizer in its own context.
Exposes the resjson and loc.json parsers directly, in addition to the current code path through a highly opinionated dispatcher. The motivation being to allow webpack loaders to use the parsers on arbitrary content or on output of other custom loaders.

## How it was tested
New unit tests
Usage in upcoming webpack 5 localization plugin

### DIFF
--- a/libraries/localization-utilities/src/LocFileParser.ts
+++ b/libraries/localization-utilities/src/LocFileParser.ts
@@ -1,12 +1,17 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-import { ITerminal, NewlineKind, JsonFile, JsonSchema } from '@rushstack/node-core-library';
+import { ITerminal, NewlineKind } from '@rushstack/node-core-library';
 
 import { ILocalizationFile } from './interfaces';
+import { parseLocJson } from './parsers/parseLocJson';
+import { parseResJson } from './parsers/parseResJson';
 import { readResxAsLocFile } from './ResxReader';
 
-const LOC_JSON_SCHEMA: JsonSchema = JsonSchema.fromFile(`${__dirname}/schemas/locJson.schema.json`);
+/**
+ * @public
+ */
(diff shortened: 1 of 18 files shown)


D4N14L (approved):

libraries/localization-utilities/src/Pseudolocalization.ts
```diff
@@ -1,23 +1,38 @@
...
  */
 export function getPseudolocalizer(options: IPseudolocaleOptions): (str: string) => string {
-  // pseudolocale maintains static state, so we need to load it as isolated modules
-  decache('pseudolocale');
-  const pseudolocale = require('pseudolocale'); // eslint-disable-line
-
-  pseudolocale.option = {
-    ...pseudolocale.option,
-    ...options
+  // eslint-disable-next-line @typescript-eslint/typedef
```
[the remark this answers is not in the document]
dmichon-msft: Was a relic from the old code. Fixed.

D4N14L (approved):

apostolisms (approved):

libraries/localization-utilities/src/LocFileParser.ts
```diff
@@ -15,6 +20,7 @@ export interface IParseLocFileOptions {
   terminal: ITerminal;
   filePath: string;
   content: string;
+  parser?: ParserKind;
```
iclanton, line 23: Why would someone want to use a parser that's different from what's implied by the file extension?

dmichon-msft: Now that I have exposed the parsers more directly, this feature here is less necessary. I had done this first before making each parser individually callable.
Easy example: the file is actually a different format altogether, that then gets converted by a separate loader into standard `.resjson` format.
I can revert this option, though, since if you know the format you should just call the individual parser directly and save the overhead of the dynamic dispatch.

PR #3465, opened 2022-06-11
fix(template): fix wrongly documented default path for reportFolder

alirezamirian:
.api.md file is generated in `temp` folder by default, not `etc`

## Summary
`.api.md` file is generated in `temp` folder by default, not `etc`

## Details
N/A

## How it was tested
I just ran api-extractor with default configuration and it created the report file in `temp` folder.

### DIFF
--- a/common/changes/@microsoft/api-extractor/patch-1_2022-06-13-19-22.json
+++ b/common/changes/@microsoft/api-extractor/patch-1_2022-06-13-19-22.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor",
+      "comment": "fixed wrongly documented default value for reportFolder",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/api-extractor"
+}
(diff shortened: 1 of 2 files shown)


iclanton:
Can you run `rush change` to generate a changelog file and commit that?

common/changes/@microsoft/api-extractor/patch-1_2022-06-13-19-22.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor",
+      "comment": "fixed wrongly documented default value for reportFolder",
```
iclanton, line 5:
```suggestion
      "comment": "Fix a mistake in the config/api-extractor.json template's default.",
```

iclanton (approved):

mcdurdin:
This change seems incorrect. The default for reportFolder is `etc`:

https://github.com/microsoft/rushstack/blob/8a33edff8d980dabb4c3dd2f56c515c12979051f/apps/api-extractor/src/schemas/api-extractor-defaults.json#L20

It makes the template comments confusing, because the `temp` folder should not be committed to git.

The report file is written to _both_ temp and etc on first run.

I think this change should be rolled back.

PR #3467, opened 2022-06-14
[localization] Include the ability to strip a string from parsing entirely.

iclanton:
## Summary

Currently, there is a way to ignore a string from loc file typings generation. This PR expands that to strip the string from the parsed loc file entirely.

## How it was tested

Included a test case in `localization-plugin-test-02`.


libraries/localization-utilities/src/parsers/test/parseLocJson.test.ts
```diff
@@ -1,23 +1,36 @@
...
-      })
-    ).toMatchSnapshot();
+      serializeLocFile(
+        parseLocJson({
+          content,
+          filePath: 'test.loc.json'
+        })
+      )
+    ).toMatchInlineSnapshot(`
+      "key | value | comment
```
dmichon-msft, line 29: Can we stick with the external snapshots? I'm more comfortable with keeping data and logic separate and, importantly, not having build steps modify code files.

dmichon-msft: Also, sorry to say, but I find this table harder to read than the JSON, since it makes whitespace (which is a rather significant element of the parsers) hard to compare. It is completely valid for strings to contain tabs, newlines, etc.

iclanton: Fair enough.

PR #3469, opened 2022-06-14
[api-extractor-model] Add new ApiItemContainerMixin.findMembersWithInheritance method

zelliott:
## Summary

This PR includes the api-extractor-model changes necessary for addressing https://github.com/microsoft/rushstack/issues/3429. All api-documenter changes for actually showing the inherited members as well as testing this change are in the child PR https://github.com/zelliott/rushstack/pull/3.

### Changes include

* Added a new method `ApiItemContainerMixin. findMembersWithInheritance ` which is the meat of this PR. This method walks an item's inheritance tree and finds all immediate and inherited members. It has comprehensive comments explaining its behavior and various edge cases. I'll also go into various scenarios later in this PR description.
* Added a new `IFindApiItemsResult` type that can be used as a generic return type for future "find operations" performed by api-extractor-model.
* Added api-extractor-scenarios tests for mixin patterns and merged declarations. These tests aren't strictly needed for this PR, and they're not particularly supported by API Extractor as-is, but I found them useful in validating API Extractor's current behavior. Happy to remove them or move them to a different PR.

### Support levels for various types of inheritance

There are _many_ different flavors of inheritance in TypeScript and this PR does not support all of them. Here is this PR's support level by inheritance scenario:

#### Full support

1. Basic class/interface inheritance (e.g. `class A extends B` or `interface A extends B`)
1. Multiple interface inheritance (e.g. `interface A extends B, C`)

#### Partial support

1. Class inheritance with type arguments (e.g. `class A extends B<number>`). Types of inherited members are relative to their defining class, not inherited class.
1. Class/interface inheriting from merged declaration. Members inherited from one of the merged declarations will be missing.

#### No support

1. Class/interface inheriting from unexported class/interface. Not supported as unexported items aren't present in the .api.json. **Will be supported in https://github.com/microsoft/rushstack/issues/3430.**
1. Class/interface inheriting from class/interface not present in .api.json (e.g. due to being in an external, third-party package).
1. Class inheriting from class-like const variable. Not supported as variables don't have members.
1. Class inheriting from anonymous class. Not supported as anonymous classes are (1) unexported and (2) represented in the d.ts as const variables.
1. Interface inheriting from interface-like type alias. Not supported as type aliases don't have members.

### Open questions

I'd expect these questions to be resolved before transitioning this PR from draft --> ready:

1. Are we okay with the behavior as documented above in the various scenarios (i.e. full/partial/no support)? **Answer: Yes.**
2. Is `findMembersWithInheritance` the right approach here or are we essentially "re-implementing inheritance" and we're better off using the type checker in api-extractor to encode the appropriate inheritance information to the API model **Answer: @octogonz and I decided this is the right approach for now and can be extended to support more types of inheritance in the future**
3. What's the best way of testing this change? It's not possible to test the change using the api-documenter-test project as-is, as the `showInheritedMembers` config setting is only available via the `api-documenter generate` workflow, and the test project uses the `api-documenter markdown` workflow. It's non trivial to change from `generate` to `markdown`. I've manually tested the change on my end by manually and it looks good, but I'd like actual checked in artifacts. **Answer: Will add a new api-documenter-scenarios test project in a child PR.**
4. Should we also make any changes to `YamlDocumenter`? **Answer: No.**
5. Note that this PR is also blocked by the bug fix at https://github.com/microsoft/rushstack/pull/3483. **Answer: No longer blocked.**

## How it was tested

Tested in the child PR https://github.com/zelliott/rushstack/pull/3.


octogonz:
(Let's focus on getting PRs 3483, 3484, 3435 merged first, then remind me and I'll take a look at this.)

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
```diff
@@ -0,0 +1,316 @@
...
+        }
+      } else {
+        throw new InternalError('ApiExtendsMixin expects a base class that inherits from ApiDeclaredItem');
+      }
+    }
+
+    /**
+     * Resolves all of the ApiItem's immediate and inherited members by walking down the
+     * inheritance tree. Be aware that his method returns both immediate and inherited members,
+     * which might not be evident from its name.
```
octogonz, line 111: > by walking down the inheritance tree

I've mostly seen class hierarchies portrayed with the base classes (root nodes of the tree) at the top of the diagram. In that representation, the visitor is walking up the tree, not down.

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
```diff
@@ -0,0 +1,316 @@
...
+     *
```
octogonz, line 112: > might not be evident from its name

Maybe we can improve the name?  The "resolving" word seems like an implementation detail, not the main idea of this API.  A word like "find" or "search" might help convey that nontrivial computational work is happening, and that we might not find every item.  (`findMembersWithInheritance()`?)

🤔 On the other hand, it seems like found inherited members are importantly different from the "actual" direct members.  For example, https://github.com/microsoft/rushstack/issues/3429#issuecomment-1133819235 suggested a presentation like:

image

So maybe it would actually be preferable for `findInheritedMembers()` to return ONLY the inherited members that do coincide with actual members, and then it is the documenter's job to mix the two sets together?

common/reviews/api/api-extractor-model.api.md
```diff
@@ -934,6 +954,12 @@ export interface IResolveDeclarationReferenceResult {
     resolvedApiItem: ApiItem | undefined;
 }
 
+// @public
+export interface IResolveInheritedMembersResult {
+    errorMessage?: string;
+    resolvedMembers: ApiItem[];
```
octogonz, line 960: It seems a bit limiting to only show one error message, with no way to provide additional such as the associated `ApiItem`.  A more flexible approach might be like: (?)

```ts
let incompleteResult: boolean = false;
const result: ApiItem[] = findInheritedMembers({
  messageCallback: (message) => {
    console.log(message.text);
    incompleteResult = true;
  }
});
```

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
zelliott: For a bit more context, I chose "resolve" to match the `resolveDeclarationReference` method. The two methods were similar in that they both are resolving references under the hood and return similar shaped result objects. I don't feel strongly though, so I'm fine with `resolveMembersWithInheritance` or `findMembersWithInheritance`.

> So maybe it would actually be preferable for findInheritedMembers() to return ONLY the inherited members that do coincide with actual members, and then it is the documenter's job to mix the two sets together?

The downside to be aware of here is that the documenter will need to sort the combined sets. Currently, the documenter doesn't have to do any sorting on its own.

octogonz:
> * Are we okay with the behavior as documented above in the various scenarios (i.e. full/partial/no support)?

It sounds okay to me. The most interesting case is:

> > Class/interface inheriting from class/interface not present in .api.json (e.g. due to being in an external, third-party package).

This design choice focuses on having **documentation pages** (API items) for inherited items. 

We might contrast with a design that focuses on showing **every possible item** even if we can't document it:  For each API item, API Extractor might use the compiler engine to collect every inherited member, and add supplementary trivia fields to `.api.json` with the (package, container, member, type).  This way, even if you don't have an .api.json file for the referenced library, API Documenter could still show very basic information like:

| property | type |  description |
|---|---|---|
| `colors` | <code>Map<string,  <a href="#">Color</a>\></code> | Regular member of this class |
| `title` | `string` | (inherited from <code><a href="#">Component</a></code>) Docs for this normal member  |
| `context` | `IContext \| undefined` | (inherited from `BaseThing` in `external-lib`) |

`package="external-lib", container="BaseThing", member="context", type="IContext | undefined"`

If `external-lib.api.json` is available, then API Documenter would use that to generate a normal row. But if not, then it could at least fill in the this basic trivia.  The main downside would be adding a lot of bloat to `.api.json` since potentially every API Item would need to include the full set of trivia items.  

🤔 And I suppose there's a counterargument that, if an API is actually important to end users, then we should find a way to define an `.api.json` file for it.  In fact maybe it is actually a better approach to provide a way to generate `.api.json` files for external libraries. (?)

In any case, such a feature could be a complementary to what you're implementing in this PR.

> * Is `resolveInheritedMembers` the right approach here or are we essentially "re-implementing inheritance" and we're better off using the type checker in api-extractor to encode the appropriate inheritance information to the API model?

I think I agree with `ApiExtendsMixin` and the idea that `.api.json` provides sufficient information to work out the hierarchy.  If there is something brittle about traversing `ExcerptToken`'s and resolving canonical references, I think we should try to fix that in the model.  

The most interesting gap would be something like:

**package-a**
```ts
export interface A { }
```

**package-b**
```ts
export interface B extends A { }
```

**package-c**
```ts
export interface C extends B { }
```

...where we generate `a.api.json` and `c.api.json` but NOT `b.api.json`.  (But from the above reflections, maybe generating `b.api.json` is the right way to deal with that.)

> * What's the best way of testing this change? It's not possible to test the change using the api-documenter-test project as-is, as the `showInheritedMembers` config setting is only available via the `api-documenter generate` workflow, and the test project uses the `api-documenter markdown` workflow. It's non trivial to change from `generate` to `markdown`. I've manually tested the change on my end by manually and it looks good, but I'd like actual checked in artifacts.

Ideally we should make an `api-documenter-scenarios` project with `runScenarios.ts` driven by the`api-documenter.json` config files, one file for each test.  If that's too much work, maybe instead fork `api-documenter-test` into `api-documenter-test-00` and `api-documenter-test-01`.

> * Should we also make any changes to `YamlDocumenter`?

I seem to remember that DocFX itself already has a way to include inherited members.  (?)

It's best not to modify `YamlDocumenter` unless you have a pipeline available to test it.  Microsoft's pipelines are unfortunately internal only, so we have to ask one of their product teams to test it.  @rbuckton found a way to run the open source DocFX to generate his [esfx](https://esfx.js.org/esfx/api/async-autoresetevent.html?tabs=ts) docs, but last time I looked I wasn't able to figure out what command to run to regenerate the ESFX docs. (?)

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
octogonz: > The downside to be aware of here is that the documenter will need to sort the combined sets. Currently, the documenter doesn't have to do any sorting on its own.

😅

octogonz:
> (speaking of which, when will this config-based invocation be out of beta and officially documented?).

As soon as anyone is available to do that work.  The `api-documenter.json` config file seems like the right way to go. The CLI front end for API Documenter is pretty basic, so it doesn't seem like a lot of work to make the config file the primary/recommended way to invoke API Documenter.

common/reviews/api/api-extractor-model.api.md
octogonz: The messages will have IDs like `message.id === ExtendsFromTypeAlias` or `message.id === MissingApiJson`.

Maybe we should also have `message.incompleteMembers`

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
octogonz: 🤔 I think I prefer `findMembersWithInheritance()` to `resolveInheritedMembers()`. I think of "resolving" as trying to locate the thing that an identifier is reference. That is part of this operation, but it seems like a less significant part than crawling a graph and searching for things, possibly with some query parameters, and possibly failing to find some things (that might be found by a better implementation).  In some of the more esoteric cases that we discussed, such as deducing the "members" from type algebra, `resolveDeclarationReference()` may not be involved at all.

The `WithInheritance` suffix was supposed to clarify that we find both inherited and non-inherited members, whereas `resolveInheritedMembers()` sounded like it excludes non-inherited members.

octogonz: (Looks like C# has both [`GetMembers()`](https://docs.microsoft.com/en-us/dotnet/api/system.type.getmembers?view=net-6.0#system-type-getmembers) and [`FindMembers()`](https://docs.microsoft.com/en-us/dotnet/api/system.type.findmembers?view=net-6.0#system-type-findmembers(system-reflection-membertypes-system-reflection-bindingflags-system-reflection-memberfilter-system-object)) -- maybe that influenced my thinking somehow heheh.)

zelliott: Done.

apps/api-documenter/src/documenters/YamlDocumenter.ts
```diff
@@ -559,29 +560,27 @@ export class YamlDocumenter {
...
       }
+    }
 
-      const typeParameters: IYamlParameter[] = this._populateYamlTypeParameters(uid, apiItem);
-      if (typeParameters.length) {
-        yamlItem.syntax = { typeParameters };
-      }
+    const typeParameters: IYamlParameter[] = this._populateYamlTypeParameters(uid, apiItem);
+    if (typeParameters.length) {
+      yamlItem.syntax = { typeParameters };
```
zelliott, line 583: This code was previously only called for `ApiInterface` items, I think this was a bug? Now it's called for both `ApiClass` and `ApiInterface` items.

build-tests/api-documenter-test/etc/api-documenter-test.api.json
```diff
@@ -192,11 +192,11 @@
               "text": "export interface Constraint "
             }
           ],
+          "extendsTokenRanges": [],
```
zelliott, line 195: In a future PR, I'd like to never write empty token ranges to API doc models.

zelliott:
Okay @octogonz, I think I've responded to all of your open comments on this PR. I've also pulled out all api-documenter changes necessary for actually showing inherited members into a [separate child branch](https://github.com/zelliott/rushstack/compare/inheritance-docs...zelliott:rushstack:inheritance-docs-api-documenter?expand=1). This child branch includes:

* A new `showInheritedMembers` api-documenter config.
* A new api-documenter-scenarios project. 🎉 
* Significant nit cleanups to the api-documenter project.
* Some nit cleanups to the api-extractor-scenarios project.

How do we want to go about reviewing and merging these branches/PRs?

build-tests/api-documenter-test/etc/api-documenter-test.api.json
octogonz: The serializer currently takes the approach of always writing every field, even if the contents are empty. The original motivation was to make the JSON file format somewhat self-documenting, such that someone could write a correct loader for it without having to look at the source code.  It seemed more important back then, because there was no `api-extractor-model` library to validate the deserialization, just a JSON schema and some TypeScript interfaces to access the JSON objects.

We could consider a different approach of trying to make the .api.json files as small as possible.  But that would only make sense if we applied it across all fields, not just the token ranges.

octogonz:
> How do we want to go about reviewing and merging these branches/PRs?

Let's make a second PR for `inheritance-docs-api-documenter`

zelliott:
Created https://github.com/zelliott/rushstack/pull/3.

apps/api-documenter/src/documenters/YamlDocumenter.ts
octogonz: This was introduced by PR #1317.  The PR notes say that  "type parameters are supported by docfx" but doesn't give any reason why the YAML is only populated for interfaces.

@rbuckton was this simply an oversight?  If we start emitting type parameters for classes in DocFX YAML, is there any risk of a regression?

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
```diff
@@ -0,0 +1,393 @@
...
+   * inherited members.
+   */
+  members: ApiItem[];
+
+  /**
+   * Indicates whether the members are potentially incomplete due to errors or gaps in finding
+   * inherited members. If true, the `FindMembersWithInheritanceMessageCallback` callback
+   * will be called with messages explaining the errors in more detail.
+   */
+  maybeIncompleteMembers: boolean;
```
octogonz, line 30: `missingData`? `incompleteResult`?

If I saw `maybeIncompleteMembers=true` without any docs, from the name alone I'm not sure it would be obvious that a problem occurred, or that `Members` is referring to the `FindMembersWithInheritanceResult.members` output.  (Maybe I'm being obtuse though heheh.)

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * This object is used for messages reported by `ApiExtendsMixin.findMembersWithInheritance`.
+ * @public
+ */
+export interface IFindMembersWithInheritanceMessage {
+  /**
+   * Unique identifier for the message.
+   */
+  id: FindMembersWithInheritanceMessageId;
```
octogonz, line 41: Let's call it `messageId` to be a little more clear.

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
```diff
@@ -0,0 +1,393 @@
...
+      } else {
+        throw new InternalError('ApiExtendsMixin expects a base class that inherits from ApiDeclaredItem');
+      }
+    }
+
+    /**
+     * Finds all of the ApiItem's immediate and inherited members by walking up the inheritance tree.
+     *
+     * @param messageCallback - A callback that is called when this method hits some kind of error
+     * that could impact finding inherited members.
```
octogonz, line 162:
```suggestion
     * @param messageCallback - A callback that reports diagnostic information about the
     * search operation.
```

I'd like to keep open the possibility that `messageCallback()` might report informational messages that do not indicate any error.  For example, if the crawl becomes more complicated in the future (rather than simpler), perhaps we'd want to introduce a message that traces each `ApiItem` as it is visited.  Or maybe we'd want a message that reports items that were visited but NOT added to the final data set (because they are occluded by a child class's override).  This is why I suggested a boolean to indicate actual data loss.

On the other hand, if you expect this crawl to become simpler in the future, such that the ***order in which things are crawled*** is not important information, then another idea would be to move this array of messages to `IFindMembersWithInheritanceResult` and eliminate the callback entirely.  Your intuition is probably more accurate than mine.

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
```diff
@@ -0,0 +1,393 @@
...
+      const toVisit: ApiItem[] = [];
+      let next: ApiItem | undefined = this;
+      let maybeIncompleteMembers: boolean = false;
+
+      while (next) {
+        const membersToAdd: ApiItem[] = [];
+
+        // For each member, check to see if we've already seen a member with the same name
+        // previously in the inheritance tree. If so, we know we won't inherit it, and thus
+        // do not add it to our `membersToAdd` array.
```
octogonz, line 230:
```ts
interface A {
  /** This is A.X */
  x: string;
}
interface B {
  /** This is B.X */
  x: string;
}
interface C extends A, B { }
```
Which is reported as the "inherited" `x` for `C`?  `A.x` or `B.x`?

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
```diff
@@ -0,0 +1,393 @@
...
+     * @param messageCallback - A callback that is called when this method hits some kind of error
+     * that could impact finding inherited members.
+     *
+     * @remarks
+     *
+     * For example, given the following class heritage:
+     *
+     * ```
+     * export class A {
+     *   a: number|boolean;
```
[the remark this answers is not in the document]
zelliott: Done.

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
zelliott: Done.

libraries/api-extractor-model/src/mixins/ApiExtendsMixin.ts
zelliott: From our short Zulip conversation: You said:

> It probably doesn't matter too much for our case, but it would be nice if the selection was deterministic (ideally alphabetical instead of based on declaration order). And if it prefers the item with more documentation (for example if `A.x` has 3 paragraphs of text whereas `B.x` has only a few words). Not necessary to implement in this PR, though.

I updated the implementation to be deterministic based upon alphabetical order. I didn't add any logic yet to prefer the item with more documentation (I thought this was a good idea but thought it made more sense in a subsequent PR).

common/reviews/api/api-extractor-model.api.md
```diff
@@ -698,9 +723,7 @@ export interface IApiCallSignatureOptions extends IApiTypeParameterListMixinOpti
 }
 
 // @public
-export interface IApiClassOptions extends IApiItemContainerMixinOptions, IApiNameMixinOptions, IApiReleaseTagMixinOptions, IApiDeclaredItemOptions, IApiTypeParameterListMixinOptions {
-    // (undocumented)
-    extendsTokenRange: IExcerptTokenRange | undefined;
```
octogonz, line 703: I didn't realize that `ApiClass.extendsTokenRange` got renamed to `extendsTokenRanges`.  Per our chat, let's see if we can avoid that because it would be a breaking change for consumers, without a clear benefit for consumers.

Changing to an array in the .api.json file format is fine though, as long as we bump the `ApiJsonSchemaVersion`.

common/changes/@microsoft/api-documenter/inheritance-docs_2022-07-15-17-26.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-documenter",
+      "comment": "Minor change to keep api-documenter working with api-extractor-model changes.",
```
octogonz, line 5:
```suggestion
      "comment": "Update API Documenter to support the latest \"api-extractor-model\"",
```

common/changes/@microsoft/api-extractor-model/inheritance-docs_2022-06-14-22-47.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor-model",
+      "comment": "Adds a new ApiExtendsMixin and associated findMembersWithInheritance method to easily find an item's inherited members.",
+      "type": "major"
```
octogonz, line 6:
```suggestion
      "comment": "Add a new ApiExtendsMixin class and findMembersWithInheritance() method for finding an item's inherited members",
      "type": "minor"
```

common/changes/@microsoft/api-extractor/inheritance-docs_2022-06-14-22-47.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor",
+      "comment": "Small API doc model change to support changes to api-extractor-model for finding an item's inherited members.",
+      "type": "major"
```
octogonz, line 6:
```suggestion
      "comment": "Update API Extractor to support the latest \"api-extractor-model\"",
      "type": "patch"
```

apps/api-documenter/src/documenters/YamlDocumenter.ts
zelliott: I reverted this change from this PR because it's no longer really related (I'm not even touching `YamlDocumenter.ts` anymore. It'd still be nice to fix this, it looks like an oversight, I'll open a separate PR.

common/reviews/api/api-extractor-model.api.md
zelliott: Updated this PR to avoid that. The `.api.json` format is no longer changing as well.

zelliott:
Okay, I think this PR is ready for another look. Some open questions:

1. We have a new `IFindApiItemsResult` object that is currently used by `findMembersWithInheritance` and will be used for future "find operations". It'd be nice to align on the `"find"` method name prefix for all operations that return a `IFindApiItemsResult` object, but there are two existing methods today that use this prefix and just return a readonly list of API items (i.e. `findMembersByName` and `findEntryPointsByPath`). Might we want to rename these methods to `getMembersByName` and `getEntryPointsByPath` instead? It could be a separate PR.
2. Do we like `IFindApiItemsResult` or `IFindApiItemResult`? Future find operations may return multiple API items or only a single API item.
3. I'd like some special attention given to the `IFindApiItemMessage`s. I imagine that these messages need to be generic across all "find operations", so I avoided mentioning "inherited members" in any of them.

libraries/api-extractor-model/src/mixins/ApiItemContainerMixin.ts
```diff
@@ -95,6 +102,64 @@ export interface ApiItemContainerMixin extends ApiItem {
...
+   * export class A<T> {
+   *   public a: T;
+   * }
+   *
+   * export class B extends A<number> {}
+   * ```
+   *
+   * When called on `B`, this method will return `B.a` with type `T` as opposed to type
+   * `number`, although the latter is more accurate.
+   */
```
octogonz, line 160: @zelliott I moved your doc comment to `interface ApiItemContainerMixin` which is the publicly visible signature.  (The `class MixedClass` declaration is inside a local function scope.)

octogonz (approved):

libraries/api-extractor-model/src/mixins/ApiItemContainerMixin.ts
```diff
@@ -209,6 +274,173 @@ export function ApiItemContainerMixin<TBaseClass extends IApiItemConstructor>(
...
+            const apiInterface: ApiInterface = next as ApiInterface;
+            extendsTypes = apiInterface.extendsTypes;
+            break;
+          }
+        }
+
+        if (extendsTypes === undefined) {
+          messages.push({
+            messageId: FindApiItemsMessageId.UnsupportedKind,
+            text: `Item ${next.displayName} is of unsupported kind ${next.kind}.`
```
octogonz, line 339: I don't understand this logic.  The condition `extendsTypes === undefined` means that the current thing is not a class or an interface; it is some other container such as `ApiEnum`.  If I call `apiEnum.findMembersWithInheritance()`, I would expect it to return the enum members.  The fact that there aren't any "inherited" members seems unimportant.  Whereas this message seems to imply that there is some kind of problem: `"Item apiEnum is of unsupported kind Enum"`

zelliott: > The fact that there aren't any "inherited" members seems unimportant.

Hm, I suppose I was thinking that if someone calls `apiEnum.findMembersWithInheritance()` the message might be useful in communicating essentially "Hey, just so you know, you asked for inherited members, but ApiEnums do not support inheritance". But I can see your point of view, so I'm happy to add a check at the top of this method that basically just defers to `get members()` if the API item is neither an `ApiClass` or `ApiInterface`.

libraries/api-extractor-model/src/mixins/ApiItemContainerMixin.ts
```diff
@@ -209,6 +274,173 @@ export function ApiItemContainerMixin<TBaseClass extends IApiItemConstructor>(
...
+          const firstReferenceToken: ExcerptToken | undefined = extendsType.excerpt.spannedTokens.find(
+            (token: ExcerptToken) => {
+              return token.kind === ExcerptTokenKind.Reference && token.canonicalReference;
+            }
+          );
+
+          if (!firstReferenceToken) {
+            messages.push({
+              messageId: FindApiItemsMessageId.UnexpectedExcerptTokens,
+              text: `Encountered unexpected excerpt tokens in ${next.displayName}. Excerpt: ${extendsType.excerpt.text}.`
```
octogonz, line 366: `!firstReferenceToken` occurs if the excerpt does not contain any token with a `canonicalReference`.  The analyzer can generate canonical references for a very wide variety of declarations, including things that don't get an `ApiItem`, so it's interesting to consider how this situation might arise.  Here's one example:

```ts
/** @public */
export declare class MyClass extends (class { }) { 
}
```

We can imagine other edge cases that should have references, but [DeclarationReferenceGenerator](https://github.com/microsoft/rushstack/blob/15680b899a06c614c32088b02940bf676596bcf2/apps/api-extractor/src/generators/DeclarationReferenceGenerator.ts#L17) hasn't implemented support (maybe an augmented type or somesuch).

Is "Encountered unexpected excerpt tokens" the best way to describe these cases?  Here's a couple alternative ideas:

- Safe approach: "A canonical reference was not found in the extends clause" accurately reports the implementation logic that failed
- Friendly approach: "The extends clause does not refer to type that can be analyzed" tries to provide a meaningful interpretation of the situation that will hopefully be correct

libraries/api-extractor-model/src/mixins/ApiItemContainerMixin.ts
```diff
@@ -209,6 +274,173 @@ export function ApiItemContainerMixin<TBaseClass extends IApiItemConstructor>(
...
+            });
+            maybeIncompleteResult = true;
+            continue;
+          }
+
+          const apiModel: ApiModel | undefined = this.getAssociatedModel();
+          if (!apiModel) {
+            messages.push({
+              messageId: FindApiItemsMessageId.MissingApiModel,
+              text: `Unable to get the associated model of ${next.displayName}.`
```
octogonz, line 376: This might arise if the `ApiItem` object was newly constructed but not added to an `ApiModel` yet. The problem isn't that we can't get the model, it's that no model has been associated.  Maybe we could word it like:

"Unable to analyze references of API item ${next.displayName} because it is not associated with an ApiModel."

zelliott: Took this verbatim, see https://github.com/microsoft/rushstack/pull/3545/files#diff-f0516a1941ffe40f35eb325cb731d6d3af1663276d61824a4a6da636d91d278eR392-R393.

PR #3472, opened 2022-06-15
[api-extractor] Allow for loading the `ExtractorConfig` without validating existence of target typings

D4N14L:
## Summary

Allow for loading the `ExtractorConfig` without validating existence of target typings.

## Details

There are some scenarios where loading up the API-Extractor configuration is required prior to the existence of the typings files that API Extractor sources from (such as obtaining output folder paths and ensuring they are empty). This PR adds the option `ignoreMissingConfigTargets` which can be passed into the `ExtractorConfig` options at load time to avoid validating the existence of the target files.

## How it was tested

Forked off of PR https://github.com/microsoft/rushstack/pull/3468 where it was tested extensively against the rest of the Rushstack monorepo.

### DIFF
--- a/apps/api-extractor/src/api/ExtractorConfig.ts
+++ b/apps/api-extractor/src/api/ExtractorConfig.ts
@@ -139,9 +139,10 @@ export interface IExtractorConfigPrepareOptions {
   /**
    * When preparing the configuration object, folder and file paths referenced in the configuration are checked
    * for existence, and an error is reported if they are not found.  This option can be used to disable this
-   * check. This may be useful when preparing a configuration file for an un-built project.
+   * check for the main entry point module. This may be useful when preparing a configuration file for an
+   * un-built project.
    */
-  ignoreMissingConfigTargets?: boolean;
+  ignoreMissingEntryPoint?: boolean;
 }
 
 interface IExtractorConfigParameters {
@@ -739,7 +740,7 @@ export class ExtractorConfig {
           // Use the manually specified "<lookup>" value
           projectFolder = options.projectFolderLookupToken;
 
-          if (!options.ignoreMissingConfigTargets && !FileSystem.exists(options.projectFolderLookupToken)) {
+          if (!FileSystem.exists(options.projectFolderLookupToken)) {
             throw new Error(
               'The specified "projectFolderLookupToken" path does not exist: ' +
                 options.projectFolderLookupToken
@@ -778,7 +779,7 @@ export class ExtractorConfig {
       } else {
         ExtractorConfig._rejectAnyTokensInPath(configObject.projectFolder, 'projectFolder');
 
-        if (!options.ignoreMissingConfigTargets && !FileSystem.exists(configObject.projectFolder)) {
+        if (!FileSystem.exists(configObject.projectFolder)) {
           throw new Error('The specified "projectFolder" path does not exist: ' + configObject.projectFolder);
         }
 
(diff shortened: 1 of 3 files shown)


apps/api-extractor/src/api/ExtractorConfig.ts
```diff
@@ -135,6 +135,13 @@ export interface IExtractorConfigPrepareOptions {
    * `@microsoft/api-extractor/extends/tsdoc-base.json`.
    */
   tsdocConfigFile?: TSDocConfigFile;
+
+  /**
+   * When preparing the configuration object, folder and file paths referenced in the configuration are checked
+   * for existence, and an error is reported if they are not found.  This option can be used to disable this
+   * check. This may be useful when preparing a configuration file for an un-built project.
+   */
+  ignoreMissingConfigTargets?: boolean;
```
zelliott, line 144: Is it possible for a user to set this via config file (i.e. `IConfigFile`)? It doesn't look like it's possible. Should it be?

D4N14L: I don't think it makes sense to allow a config file to set this. It's somewhat of a `loadWithoutValidation` tweak, and having that value in the config file feels odd. By the time API Extractor is invoked, all the required files should be available. This change is more to allow loading and inspecting the configuration without invoking API Extractor, so it makes sense to keep it within the API interface IMO

iclanton, line 144: This only ignores missing declarations, right? This name implies that all referenced files (including the `extends` target) are ignored if they're missing.

apps/api-extractor/src/api/ExtractorConfig.ts
```diff
@@ -826,7 +833,7 @@ export class ExtractorConfig {
         if (!tsconfigFilePath) {
           throw new Error('Either the "tsconfigFilePath" or "overrideTsconfig" setting must be specified');
         }
-        if (!FileSystem.exists(tsconfigFilePath)) {
+        if (!options.ignoreMissingConfigTargets && !FileSystem.exists(tsconfigFilePath)) {
```
iclanton, line 836: It'd be weird if this one didn't exist.

apps/api-extractor/src/api/ExtractorConfig.ts
```diff
@@ -771,7 +778,7 @@ export class ExtractorConfig {
       } else {
         ExtractorConfig._rejectAnyTokensInPath(configObject.projectFolder, 'projectFolder');
 
-        if (!FileSystem.exists(configObject.projectFolder)) {
+        if (!options.ignoreMissingConfigTargets && !FileSystem.exists(configObject.projectFolder)) {
```
iclanton, line 781: It'd be weird if this one didn't exist.

apps/api-extractor/src/api/ExtractorConfig.ts
```diff
@@ -732,7 +739,7 @@ export class ExtractorConfig {
           // Use the manually specified "<lookup>" value
           projectFolder = options.projectFolderLookupToken;
 
-          if (!FileSystem.exists(options.projectFolderLookupToken)) {
+          if (!options.ignoreMissingConfigTargets && !FileSystem.exists(options.projectFolderLookupToken)) {
```
iclanton, line 742: It'd be weird if this one didn't exist.

Seems like this option really should only apply to other tools' outputs and not any file. The option name should be updated to reflect that too.

PR #3499, opened 2022-06-27
[node-core-library] Refactor SubprocessTerminator

D4N14L:
## Summary

Refactor the `SubprocessTerminator` utility into `@rushstack/node-core-library` for use outside of Heft.

## Details

Moves `SubprocessTerminator` into `@rushstack/node-core-library` and updates existing references to point to `@rushstack/node-core-library`. Adds documentation.


common/changes/@rushstack/node-core-library/user-danade-RefactorSubprocessTerminator_2022-06-27-21-33.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/node-core-library",
+      "comment": "Add SubprocessTerminator utility, which can be used to kill a process and all it's child processes on demand or on termination of the host process.",
```
iclanton, line 5:
```suggestion
      "comment": "Add SubprocessTerminator utility, which can be used to kill a process and all of its child processes on demand or on termination of the host process.",
```

libraries/node-core-library/src/SubprocessTerminator.ts
```diff
@@ -1,14 +1,24 @@
...
+ *
+ * @beta
  */
 export interface ISubprocessOptions {
+  /**
+   * Whether or not the child process is detached.
+   *
+   * @remarks
+   * On POSIX systems, detached=true is required for killing the subtree. On Windows, detached=true
+   * creates a separate console window which is generally not the desired behavior.
```
iclanton, line 20: >On Windows, detached=true creates a separate console window which is generally not the desired behavior.

Can the process be killed on Windows without `detatched = true`? Should we change this behavior to not open a new window on Windows but still be killable?

iclanton (approved):

libraries/node-core-library/src/SubprocessTerminator.ts
D4N14L: Talked offline, this is an option specifying the current state of the child process, not an arg used for creating a new process. I updated the comment to be more useful.

PR #3500, opened 2022-06-27
[localization-plugin-5] Fix unit tests on Windows

dmichon-msft:
## Summary
Ensures that the snapshot tests for the webpack 5 localization plugin work when running on Windows.

## Details
Modifies the path normalization algorithm to use POSIX path resolution if the value of the webpack `context` is a POSIX absolute path (i.e. starts with `/`).

## How it was tested
Ran unit tests on Windows.
PR build runs unit tests on Linux.

### DIFF
--- a/webpack/localization-plugin-5/src/LocalizationPlugin.ts
+++ b/webpack/localization-plugin-5/src/LocalizationPlugin.ts
@@ -583,6 +583,9 @@ export class LocalizationPlugin implements WebpackPluginInstance {
       // END options.localizedData.passthroughLocale
 
       // START options.localizedData.translatedStrings
+      const resolveRelativeToContext: (relative: string) => string = (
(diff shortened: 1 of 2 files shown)


webpack/localization-plugin-5/src/LocalizationPlugin.ts
```diff
@@ -583,6 +583,9 @@ export class LocalizationPlugin implements WebpackPluginInstance {
       // END options.localizedData.passthroughLocale
 
       // START options.localizedData.translatedStrings
+      const resolveRelativeToContext: (relative: string) => string = (
+        configuration.context?.startsWith('/') ? path.posix.resolve : path.resolve
+      ).bind(0, configuration.context!);
```
D4N14L, line 588: Why `0` instead of `undefined`?

PR #3508, opened 2022-06-29
[api-extractor] Make enumMemberOrder configuration field optional

guoyunhe:
It is a new option with a default value. So it should be optional. Otherwise, it will become a breaking change.


guoyunhe:
@D4N14L done

iclanton (approved):

PR #3524, opened 2022-07-06
Update `api-extractor-template.json` and reorder some property signatures.

zelliott:
Moved all top-level settings to the same place in `api-extractor-template.json` and associated interfaces.

See associated https://github.com/microsoft/api-extractor.com-website/pull/49.


octogonz (approved):
Looks good, thanks for fixing this!

PR #3531, opened 2022-07-13
Upgrade to webpack-dev-server@4.9.3

dmichon-msft:
## Details
Upgrades to webpack-dev-server 4.9.3 in all 3 heft plugins to use a uniform version.
Upgrades express to 4.18.1 to match the version used by webpack-dev-server.

## How it was tested
Local run of build-tests/heft-webpack4-everything-test in start mode.

### DIFF
--- a/heft-plugins/heft-dev-cert-plugin/src/webpack-dev-middleware-workaround.d.ts
+++ b/heft-plugins/heft-dev-cert-plugin/src/webpack-dev-middleware-workaround.d.ts
@@ -4,6 +4,5 @@
 // Node 12 is still LTS so we need to support it.
 // Upstream issue: https://github.com/webpack/webpack-dev-middleware/issues/1194
 declare module 'fs' {
-  // eslint-disable-next-line
-  export type StatSyncFn = any;
+  export type StatSyncFn = typeof statSync;
 }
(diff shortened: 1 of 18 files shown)


heft-plugins/heft-dev-cert-plugin/src/webpack-dev-middleware-workaround.d.ts
```diff
@@ -4,6 +4,5 @@
 // Node 12 is still LTS so we need to support it.
 // Upstream issue: https://github.com/webpack/webpack-dev-middleware/issues/1194
 declare module 'fs' {
-  // eslint-disable-next-line
-  export type StatSyncFn = any;
+  export type StatSyncFn = typeof statSync;
```
iclanton, line 7: Is this workaround still necessary? I'm pretty sure the fix was published.

iclanton (approved):

heft-plugins/heft-dev-cert-plugin/src/webpack-dev-middleware-workaround.d.ts
dmichon-msft: Might not be, but there's no downside to having it.

PR #3533, opened 2022-07-13
[heft-config-file] Some minor typings cleanup.

iclanton:
## Summary

This PR does some cleanup to heft-config-file's typings.

## How it was tested

Built and added a test case for the deprecated enum value.

### DIFF
--- a/libraries/heft-config-file/src/ConfigurationFile.ts
+++ b/libraries/heft-config-file/src/ConfigurationFile.ts
@@ -51,23 +51,32 @@ export enum PathResolutionMethod {
   /**
    * Resolve a path relative to the configuration file
    */
-  resolvePathRelativeToConfigurationFile,
+  resolvePathRelativeToConfigurationFile = 'resolvePathRelativeToConfigurationFile',
 
   /**
    * Resolve a path relative to the root of the project containing the configuration file
    */
-  resolvePathRelativeToProjectRoot,
+  resolvePathRelativeToProjectRoot = 'resolvePathRelativeToProjectRoot',
+
+  /**
+   * Treat the property as a NodeJS-style require/import reference and resolve using standard
+   * NodeJS filesystem resolution
+   *
+   * @deprecated
+   * Use {@link PathResolutionMethod.nodeResolve} instead
+   */
+  NodeResolve = 'NodeResolve',
 
   /**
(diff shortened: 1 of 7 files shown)


dmichon-msft (approved):

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -87,7 +96,7 @@ interface IConfigurationFileFieldAnnotation<TField> {
  *
  * @beta
  */
-export interface IJsonPathMetadata {
+export interface ICustomJsonPathMetadata {
   /**
    * If `IJsonPathMetadata.pathResolutionMethod` is set to `PathResolutionMethod.custom`,
```
D4N14L, line 101:
```suggestion
   * If `ICustomJsonPathMetadata.pathResolutionMethod` is set to `PathResolutionMethod.custom`,
```

libraries/heft-config-file/src/ConfigurationFile.ts
```diff
@@ -98,7 +107,24 @@ export interface IJsonPathMetadata {
...
-  pathResolutionMethod?: PathResolutionMethod;
+  pathResolutionMethod?: PathResolutionMethod.custom;
+}
+
+/**
+ * Used to specify how node(s) in a JSON object should be processed after being loaded.
+ *
+ * @beta
+ */
+export interface INonCustomJsonPathMetadata {
```
D4N14L, line 118: I get why this isn't the original name, but feels oddly verbose since "non-custom" implies "normal" or "default". Though I'm not sure what I'd rather call it...

D4N14L (approved):

libraries/heft-config-file/src/ConfigurationFile.ts
D4N14L: Ehhh, maybe not given the type is still exported below. Just a bit funky for the comment to be referencing a different interface.

libraries/heft-config-file/src/ConfigurationFile.ts
iclanton: Yeah that's where I ended up too. It's marked as `@beta` and it's unlikely someone will actually use this type by name, so we can change it if we come up with something better.

PR #3534, opened 2022-07-13
[heft-sass-plugin] Fix an issue where cssOutputFolders specified in sass.json in a rig did not correctly resolve relative to the root of the project being built.

iclanton:
## Summary

There is an issue with `sass.json` specified in a rig. The folders specified in `cssOutputFolders` are resolved relative to the root of the rig instead of the root of the project being built.

## How it was tested

Tested in a project that reproduces this behavior.


heft-plugins/heft-sass-plugin/src/SassPlugin.ts
```diff
@@ -71,38 +72,54 @@ export class SassTypingsPlugin implements IHeftPlugin {
...
     const sassConfigurationJson: ISassConfigurationJson | undefined =
-      await SassTypingsPlugin._getSassConfigurationLoader().tryLoadConfigurationFileForProjectAsync(
+      await SassPlugin._getSassConfigurationLoader().tryLoadConfigurationFileForProjectAsync(
         logger.terminal,
         buildFolder,
         heftConfiguration.rigConfig
       );
 
+    if (sassConfigurationJson) {
+      if (sassConfigurationJson.srcFolder && !path.isAbsolute(sassConfigurationJson.srcFolder)) {
```
dmichon-msft, line 82: Is `path.isAbsolute` cheaper than `path.resolve`? Calling `path.resolve(path1, someAbsolutePath)` returns `someAbsolutePath`.

iclanton: Hmm yeah good point. Probably cheaper to do it all at once.

PR #3539, opened 2022-07-14
[heft] Add .d.ts shims for secondary module kinds

dmichon-msft:
## Summary
Modifies the multi-emit logic for the Heft TypeScript plugin to generate `.d.ts` shim files for non-primary output kinds, that redirect to the primary .d.ts files.

## Details
Creates shim files containing:
```ts
export * from '../path/to/declaration';
// If a default export exists
export { default as default } from '../path/to/declaration';
```

## How it was tested
Enabled multi-emit and isolatedModules in the node rig temporarily. Verified output content.

### DIFF
--- a/apps/heft/src/plugins/TypeScriptPlugin/EmitFilesPatch.ts
+++ b/apps/heft/src/plugins/TypeScriptPlugin/EmitFilesPatch.ts
@@ -29,6 +29,8 @@ export interface ICachedEmitModuleKind {
   isPrimary: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


dmichon-msft:
I'm not entirely happy with how the opt-in currently works. I expect it may be best to have an express option in the moduleKindsToEmit collection to emit the shim.

dmichon-msft:
This isn't needed thanks to just directly splitting .d.ts emit into a disjoint folder and using `exports` and `typesVersions` to map it.

iclanton:
This is for the old Heft design.

PR #3541, opened 2022-07-18
[rush] Update the rush init template to disable hoisting of packages.

iclanton:
## Summary

Mitigates https://github.com/microsoft/rushstack/issues/3535 in new repos configured with `rush init`.

## Details

These options are documented here: https://pnpm.io/npmrc#dependency-hoisting-settings

## How it was tested

This repo is already configured with these options. Tested by running `rush init` and using npm and yarn and verified they don't error with these options specified.


octogonz (changes requested):
Pending design discussion in https://github.com/microsoft/rushstack/issues/3542

PR #3553, opened 2022-07-27
[webpack5-module-minifier-plugin] Fix typings field

dmichon-msft:
## Summary
Fixes the target of the `typings` field in `@rushstack/webpack5-module-minifier-plugin`.

## Details
Update required after the package publish target changed.

## How it was tested
Local build.

### DIFF
--- a/webpack/module-minifier-plugin-5/package.json
+++ b/webpack/module-minifier-plugin-5/package.json
@@ -3,7 +3,7 @@
   "version": "5.1.15",
   "description": "This plugin splits minification of webpack compilations into smaller units.",
   "main": "lib/index.js",
-  "typings": "dist/module-minifier-plugin.d.ts",
+  "typings": "dist/webpack5-module-minifier-plugin.d.ts",
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


webpack/module-minifier-plugin-5/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "5.1.15",
   "description": "This plugin splits minification of webpack compilations into smaller units.",
   "main": "lib/index.js",
-  "typings": "dist/module-minifier-plugin.d.ts",
+  "typings": "dist/webpack5-module-minifier-plugin.d.ts",
```
D4N14L, line 6: Should we also change the path in the repo for the package folder? Bit odd that it's `webpack/module-minifier-plugin-5` when this is `webpack5-module-minifier-plugin`

D4N14L (approved):

iclanton (approved):

webpack/module-minifier-plugin-5/package.json
iclanton: Yeah we should just rename these.

PR #3560, opened 2022-07-28
[rush-lib] components package.json JSON format validation & version adherence to semver validation

1000RR:
## Summary
This PR seeks to solve 2 issues:
- rush repo components' package.json version not in semver format fails when listed as a dependency with `workspace:*` in another project - https://github.com/microsoft/rushstack/issues/2678
- when running `rush update` and there are parsing errors in a component's package.json, pnpm does not provide stderr/stdio feedback - only exit code 1, which is not descriptive to the user. The `rush update` error message is `ERROR: Error: The command failed with exit code 1`

## Details
- The component version is manually checked for being in semver format after `package.json` is loaded.
- Package.json parsing of rush repos' subprojects' validates strict JSON format - no comments, syntactically correct.

## How it was tested
- building a project containing 2 components A and B, where A imports B's `workspace:*` version. 2 tests cases:
  - B's package.json version is set to `1.0` (new error message shown, graceful exit)
  - B's package.json version is set to `1.0.0`(runs ok, graceful exit)
- building a project where a component's `package.json` is malformed JSON - with comments or extra/missing commas.

    
- [x] 👉 STEP 7: Don't forget to run "rush change": https://rushjs.io/pages/best_practices/change_logs/

### DIFF
--- a/libraries/node-core-library/src/JsonFile.ts
+++ b/libraries/node-core-library/src/JsonFile.ts
@@ -113,10 +113,14 @@ export class JsonFile {
   /**
    * Loads a JSON file.
    */
-  public static load(jsonFilename: string): JsonObject {
+  public static load(jsonFilename: string, validateStrictJsonFormat?: boolean): JsonObject {
     try {
       const contents: string = FileSystem.readFile(jsonFilename);
-      return jju.parse(contents);
+      const parserOptions: jju.ParseOptions = {};
+      if (validateStrictJsonFormat) {
+        parserOptions.mode = 'json';
+      }
+      return jju.parse(contents, parserOptions);
     } catch (error) {
       if (FileSystem.isNotExistError(error as Error)) {
         throw error;
--- a/libraries/node-core-library/src/JsonFile.ts
+++ b/libraries/node-core-library/src/JsonFile.ts
@@ -39,7 +39,92 @@ export type JsonObject = any;
 export type JsonNull = null;
 
 /**
- * Options for JsonFile.stringify()
+ * Specifies the variant of JSON syntax to be used.
+ *
+ * @public
+ */
+export enum JsonSyntax {
+  /**
+   * Specifies the exact RFC 8259 format as implemented by the `JSON.parse()` system API.
+   * This format was designed for machine generated inputs such as an HTTP payload.
+   * It is not a recommend choice for human-authored files, because it does not support
+   * code comments.
+   *
+   * @remarks
+   *
+   * A well-known quote from Douglas Crockford, the inventor of JSON:
+   *
+   * "I removed comments from JSON because I saw people were using them to hold parsing directives,
+   * a practice which would have destroyed interoperability.  I know that the lack of comments makes
+   * some people sad, but it shouldn't.  Suppose you are using JSON to keep configuration files,
+   * which you would like to annotate.  Go ahead and insert all the comments you like.
+   * Then pipe it through JSMin before handing it to your JSON parser."
+   *
+   * @see {@link https://datatracker.ietf.org/doc/html/rfc8259 | RFC 8259}
+   */
+  Strict = 'strict',
+
+  /**
+   * `JsonSyntax.JsonWithComments` is the recommended format for human-authored config files.
+   * It is a minimal extension to `JsonSyntax.Strict` adding support for code comments
+   * using `//` and `/*`.
+   *
+   * @remarks
+   *
+   * VS Code calls this format `jsonc`, but it should not be confused with unrelated file formats
+   * and libraries that also use the name "JSONC".
+   *
+   * To fix VS Code syntax highlighting, add this setting:
+   * `"files.associations": { "*.json": "jsonc" }`
+   *
+   * To fix GitHub syntax highlighting, add this to your `.gitattributes`:
+   * `*.json linguist-language=JSON-with-Comments`
+   */
+  JsonWithComments = 'jsonWithComments',
+
+  /**
+   * JSON5 is a project that proposes a JSON-like format supplemented with ECMAScript 5.1
+   * notations for objects, numbers, comments, and more.
+   *
+   * @remarks
+   * Files using this format should use the `.json5` file extension instead of `.json`.
+   *
+   * JSON5 has substantial differences from JSON: object keys may be unquoted, trailing commas
+   * are allowed, and strings may span multiple lines.  Whereas `JsonSyntax.JsonWithComments` can
+   * be cheaply converted to standard JSON by stripping comments, parsing JSON5 requires a
+   * nontrivial algorithm that may not be easily available in some contexts or programming languages.
+   *
+   * @see {@link https://json5.org/ | JSON5 project website}
+   */
+  Json5 = 'json5'
+}
+
+/**
+ * Options for {@link JsonFile.parseString}, {@link JsonFile.load}, and {@link JsonFile.loadAsync}.
+ *
+ * @public
+ */
+export interface IJsonFileParseOptions {
+  /**
+   * Specifies the variant of JSON syntax to be used.
+   *
+   * @defaultValue
+   * `JsonSyntax.Json5`
+   *
+   * NOTE: This default will be changed to `JsonSyntax.JsonWithComments` in a future release.
+   */
+  jsonSyntax?: JsonSyntax;
+}
+
+/**
+ * Options for {@link JsonFile.loadAndValidate} and {@link JsonFile.loadAndValidateAsync}
+ *
+ * @public
+ */
+export interface IJsonFileLoadAndValidateOptions extends IJsonFileParseOptions, IJsonSchemaValidateOptions {}
+
+/**
+ * Options for {@link JsonFile.stringify}
  *
  * @public
  */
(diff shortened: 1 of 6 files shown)


libraries/node-core-library/src/JsonFile.ts
```diff
@@ -113,10 +113,14 @@ export class JsonFile {
   /**
    * Loads a JSON file.
    */
-  public static load(jsonFilename: string): JsonObject {
+  public static load(jsonFilename: string, validateStrictJsonFormat?: boolean): JsonObject {
```
1000RR, line 116: Only the `load` function was modified to show the package.json parsing validation change. If async and callback variants need to be changed as well, please comment.

libraries/node-core-library/src/JsonFile.ts
```diff
@@ -113,10 +113,14 @@ export class JsonFile {
...
     try {
       const contents: string = FileSystem.readFile(jsonFilename);
-      return jju.parse(contents);
+      const parserOptions: jju.ParseOptions = {};
+      if (validateStrictJsonFormat) {
+        parserOptions.mode = 'json';
```
1000RR, line 121: see https://www.npmjs.com/package/jju#jjuparse-function

libraries/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -106,7 +106,7 @@ export class RushConfigurationProject {
     const packageJsonFilename: string = path.join(this._projectFolder, FileConstants.PackageJson);
 
     try {
-      this._packageJson = JsonFile.load(packageJsonFilename);
+      this._packageJson = JsonFile.load(packageJsonFilename, true);
```
1000RR, line 109: Enable JSON validation during repo's components' `package.json` parsing.

libraries/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -149,6 +149,13 @@ export class RushConfigurationProject {
       );
     }
 
+    if (!semver.valid(this._packageJson.version)) {
```
1000RR, line 152: Ensure that `package.json`'s `version` field is in semver format.

octogonz:
> * when running `rush update` and there are parsing errors in a component's package.json, pnpm does not provide stderr/stdio feedback - only exit code 1, which is not descriptive to the user. The `rush update` error message is `ERROR: Error: The command failed with exit code 1`

🥳 yes, let's finally fix this! 🙏

libraries/rush-lib/src/api/RushConfigurationProject.ts
octogonz: @elliot-nelson @D4N14L could you test this in your largest monorepos to make sure there aren't any edge cases we're not considering?

libraries/rush-lib/src/api/RushConfigurationProject.ts
octogonz: We normally would introduce an `options` parameter, since the meaning of `, true` is unclear to readers, and we may want to add additional optional parameters in the future.  I can fix this.

octogonz:
Your new API can also fix this issue:  https://github.com/microsoft/rushstack/issues/996

libraries/rush-lib/src/api/RushConfigurationProject.ts
octogonz: @1000RR I pushed an update to your branch -- let me know what you think

libraries/node-core-library/src/JsonFile.ts
```diff
@@ -39,7 +39,92 @@ export type JsonObject = any;
...
+ * Options for {@link JsonFile.parseString}, {@link JsonFile.load}, and {@link JsonFile.loadAsync}.
+ *
+ * @public
+ */
+export interface IJsonFileParseOptions {
+  /**
+   * Specifies the variant of JSON syntax to be used.
+   *
+   * @defaultValue
+   * `JsonSyntax.Json5`
```
octogonz, line 112: @elliot-nelson @iclanton @D4N14L @dmichon-msft 

This is a longstanding oversight in the `JsonFile` API:

- `JsonFile.save()` always writes strict-JSON or JSON-with-comments (if we are updating a file that already contains comments)
- `JsonFile.load()` will silently accept full JSON5 syntax, for example `{ key: 0x1234, }`  (unquoted key, hex literal, trailing comma)

In practice, we don't actually use JSON5 syntax in our config files, because `JsonFile.save()` would reset it, and because every syntax highlighter would display it as an error.  In my opinion, accepting it was more of a bug than an undocumented feature. 😅

I feel strongly that if config files want to use JSON5, then they must use the `.json5` file extension.

(A similar case can be made that JSON files with comments should use a file extension such as `.jsonc` or `.cjson`, however that argument is weaker because (1) nobody seems to agree about what this file extension should be and (2) for web projects, most `.json` files are human-authored and it's ***plainly misguided*** to forbid comments in human-authored files.)

### Proposal

In a separate PR, I propose to change the default for `IJsonFileParseOptions.jsonSyntax` to be `JsonSyntax.JsonWithComments`.  It would be a MAJOR version bump for `node-core-library`, however my suspicion is that it's unlikely to break any real-world projects using this library.

Thoughts?

libraries/rush-lib/src/api/RushConfigurationProject.ts
elliot-nelson: Tested the new semver error and the "print rush update" warning, both look good.  Also tried some commands like "rush deploy" using 1000RR's branch and everything worked as normal. 👍

libraries/rush-lib/src/api/RushConfigurationProject.ts
1000RR: I think this works.  Please see my other note in a lower thread.

libraries/node-core-library/src/JsonFile.ts
1000RR: The following thoughts are likely out of scope of this bug, but discuss consistency issues amongst files with the `.json` extension in a given RushJS repo.

The disparity between the `rush.json` file being written in `cjson` syntax and every other `package.json` necessarily being written in strict `json` mode (largely due to the package managers' (at least `pnpm`'s) inability to parse plain `cjson`/`json5`) is a bit confusing in that it sets inconsistent format expectations for files with the same extension across different `.json` files in a given repo.

Anecdotally, I have observed our devs at Adobe add `cjson` syntax to `rush.json`. I imagine that it's not far-fetched to say that other projects may have made `json5` modifications to their respective `rush.json` since its parsing has been open `json5` by default. At this stage restricting `rush.json` to `cjson` would fail parsing in certain repos that have taken liberties due to suddenly stricter parsing rules of what may have effectively become a JSON5-formatted file. 

One potential avenue is to rename the `rush.json` file currently written in `cjson` format in the RushStack repo to `.json5` extension, explicitly opening the door to JSON5 format. I imagine there would be a deprecation warning when building a repo using the the older `rush.json` file, and a loading precedence or error necessary in case both a `rush.json` and `rush.json5` are found in a project.

Caveat: my suggestion would have been to stick with `cjson` format as opposed to `json5` for `rush.json` if a respective disparate file extension could be agreed upon for that file. In that case, an error for any repos with `json5` modifications to their `rush.json` would be shown.

Just my 2c.

octogonz: > Anecdotally, I have observed our devs at Adobe add single line comment syntax to `rush.json`.

This is valid `cjson` and heavily used in Rush's template problems.

> I imagine that it's not far-fetched to say that other projects may have made `json5` modifications to their respective `rush.json` since its parsing has been open `json5` by default. At this stage restricting `rush.json` to `cjson` would fail parsing in certain repos that have taken liberties due to suddenly stricter parsing rules of what may have effectively become a JSON5-formatted file.

It's not far-fetched, but my hypothesis is that it is very rare.  You can test it easily by modifying your branch to change this line to say `cjson`, then running `rush install` [using your dev build](https://rushjs.io/pages/contributing/#testing-rush-builds):

https://github.com/microsoft/rushstack/blob/8a497e11561d03b44a73675c50318f37c4e0db7c/libraries/node-core-library/src/JsonFile.ts#L590-L595

I tried it with several if our repos and there were no problems.  It makes sense, because nothing about the syntax highlighting or existing file contents would suggest that JSON5 syntaxes are allowable.  If there are JSON5 syntaxes in a few places, fixing them should be pretty easy.

Also keep in mind that Rush always parses config files using the engine version specified by `rushVersion` from **rush.json**.  Thus, upgrading to the stricter Rush would be an explicit action, and thereafter there won't be any concern about users invoking the old version of Rush.

libraries/rush-lib/src/api/RushConfigurationProject.ts
octogonz: We need your to sign the CLA in order to get this PR merged.

1000RR:
Legal sign-off is pending.

1000RR:
@octogonz Legal has cleared the CLA - good to go.

octogonz:
@1000RR if you're good with these changes, could you promote this to a non-draft PR?

1000RR:
@octogonz Looks like it's ready to go.

octogonz:
> @octogonz may I ask for a final review and merge?

Thanks for the reminder. And thanks for your patience -- things have been very busy lately. 😄

octogonz (approved):

PR #3568, opened 2022-08-03
use pnpm 7.x

TheLarkInn:
Addresses #3567

### DIFF
--- a/common/config/rush/.npmrc
+++ b/common/config/rush/.npmrc
@@ -21,5 +21,8 @@
 registry=https://registry.npmjs.org/
 always-auth=false
 # No phantom dependencies allowed in this repository
+# Don't hoist in common/temp/node_modules
+public-hoist-pattern=
 # Don't hoist in common/temp/node_modules/.pnpm/node_modules
 hoist=false
+hoist-pattern=
--- a/common/config/rush/.npmrc
+++ b/common/config/rush/.npmrc
@@ -21,8 +21,5 @@
 registry=https://registry.npmjs.org/
 always-auth=false
 # No phantom dependencies allowed in this repository
-# Don't hoist in common/temp/node_modules
-public-hoist-pattern=
 # Don't hoist in common/temp/node_modules/.pnpm/node_modules
 hoist=false
-hoist-pattern=
--- a/eslint/eslint-plugin-security/src/no-unsafe-regexp.test.ts
+++ b/eslint/eslint-plugin-security/src/no-unsafe-regexp.test.ts
@@ -6,7 +6,12 @@ import { noUnsafeRegExp } from './no-unsafe-regexp';
 
 const { RuleTester } = ESLintUtils;
 const ruleTester = new RuleTester({
-  // In PNPN 7.0 using '@typescript-eslint/parser' unexpectedly fails, se we are resolving the absolute path of the parser
+  /*
+   * The underlying API requires an absolute path. `@typescript-eslint/utils` calls `require.resolve()` on the input
+   * and forces it to be of type '@typescript-eslint/parser' but does not have a dependency on `@typescript-eslint/parser`
+   * This means that it will always fail to resolve in a strict environment.
+   * Fortunately `require.resolve(absolutePath)` returns `absolutePath`, so we can resolve it first and cast.
+   */
   parser: require.resolve('@typescript-eslint/parser') as '@typescript-eslint/parser'
 });
 
--- a/eslint/eslint-plugin-security/src/no-unsafe-regexp.test.ts
+++ b/eslint/eslint-plugin-security/src/no-unsafe-regexp.test.ts
@@ -6,6 +6,7 @@ import { noUnsafeRegExp } from './no-unsafe-regexp';
 
 const { RuleTester } = ESLintUtils;
 const ruleTester = new RuleTester({
+  // In PNPN 7.0 using '@typescript-eslint/parser' unexpectedly fails, se we are resolving the absolute path of the parser
   parser: require.resolve('@typescript-eslint/parser') as '@typescript-eslint/parser'
 });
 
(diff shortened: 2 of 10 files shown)


common/config/rush/.npmrc
```diff
@@ -21,8 +21,5 @@
 registry=https://registry.npmjs.org/
 always-auth=false
 # No phantom dependencies allowed in this repository
-# Don't hoist in common/temp/node_modules
-public-hoist-pattern=
```
iclanton, line 25: Can we leave these on and resolve whatever issue is showing up instead of turning them off?

eslint/eslint-plugin-security/src/no-unsafe-regexp.test.ts
```diff
@@ -6,7 +6,8 @@ import { noUnsafeRegExp } from './no-unsafe-regexp';
 
 const { RuleTester } = ESLintUtils;
 const ruleTester = new RuleTester({
-  parser: '@typescript-eslint/parser'
+  // In PNPN 7.0 using '@typescript-eslint/parser' unexpectedly fails, se we are resolving the absolute path of the parser
```
D4N14L, line 9:
```suggestion
  // In PNPM 7.0 using '@typescript-eslint/parser' unexpectedly fails, so we are resolving the absolute path of the parser
```

eslint/eslint-plugin-security/src/no-unsafe-regexp.test.ts
```diff
@@ -6,7 +6,8 @@ import { noUnsafeRegExp } from './no-unsafe-regexp';
...
+  parser: require.resolve('@typescript-eslint/parser') as '@typescript-eslint/parser'
```
D4N14L, line 10: Why the type cast?

eslint/eslint-plugin-security/src/no-unsafe-regexp.test.ts
D4N14L: Same with the way it is used in eslint-plugin

D4N14L: Also, could you provide some context in the comment as to why it fails?

D4N14L (approved):

eslint/eslint-plugin-security/src/no-unsafe-regexp.test.ts
dmichon-msft, line 9:
```suggestion
  /*
   * The underlying API requires an absolute path. `@typescript-eslint/utils` calls `require.resolve()` on the input
   * and forces it to be of type '@typescript-eslint/parser' but does not have a dependency on `@typescript-eslint/parser`
   * This means that it will always fail to resolve in a strict environment.
   * Fortunately `require.resolve(absolutePath)` returns `absolutePath`, so we can resolve it first and cast.
   */
```

eslint/eslint-plugin/src/hoist-jest-mock.test.ts
```diff
@@ -6,7 +6,8 @@ import { hoistJestMock } from './hoist-jest-mock';
 
 const { RuleTester } = ESLintUtils;
 const ruleTester = new RuleTester({
-  parser: '@typescript-eslint/parser',
+  // In PNPN 7.0 using '@typescript-eslint/parser' unexpectedly fails, se we are resolving the absolute path of the parser
```
dmichon-msft, line 9:
```suggestion
  /*
   * The underlying API requires an absolute path. `@typescript-eslint/utils` calls `require.resolve()` on the input
   * and forces it to be of type '@typescript-eslint/parser' but does not have a dependency on `@typescript-eslint/parser`
   * This means that it will always fail to resolve in a strict environment.
   * Fortunately `require.resolve(absolutePath)` returns `absolutePath`, so we can resolve it first and cast.
   */
```

eslint/eslint-plugin/src/no-new-null.test.ts
```diff
@@ -6,7 +6,8 @@ import { noNewNullRule } from './no-new-null';
 
 const { RuleTester } = ESLintUtils;
 const ruleTester = new RuleTester({
-  parser: '@typescript-eslint/parser'
+  // In PNPN 7.0 using '@typescript-eslint/parser' unexpectedly fails, se we are resolving the absolute path of the parser
```
dmichon-msft, line 9:
```suggestion
  /*
   * The underlying API requires an absolute path. `@typescript-eslint/utils` calls `require.resolve()` on the input
   * and forces it to be of type '@typescript-eslint/parser' but does not have a dependency on `@typescript-eslint/parser`
   * This means that it will always fail to resolve in a strict environment.
   * Fortunately `require.resolve(absolutePath)` returns `absolutePath`, so we can resolve it first and cast.
   */
```

eslint/eslint-plugin/src/typedef-var.test.ts
```diff
@@ -6,7 +6,8 @@ import { typedefVar } from './typedef-var';
 
 const { RuleTester } = ESLintUtils;
 const ruleTester = new RuleTester({
-  parser: '@typescript-eslint/parser'
+  // In PNPN 7.0 using '@typescript-eslint/parser' unexpectedly fails, se we are resolving the absolute path of the parser
```
dmichon-msft, line 9:
```suggestion
  /*
   * The underlying API requires an absolute path. `@typescript-eslint/utils` calls `require.resolve()` on the input
   * and forces it to be of type '@typescript-eslint/parser' but does not have a dependency on `@typescript-eslint/parser`
   * This means that it will always fail to resolve in a strict environment.
   * Fortunately `require.resolve(absolutePath)` returns `absolutePath`, so we can resolve it first and cast.
   */
```

eslint/eslint-plugin-security/src/no-unsafe-regexp.test.ts
dmichon-msft: Switched to a more elaborate comment

D4N14L (approved):

dmichon-msft (approved):

iclanton (approved):

PR #3588, opened 2022-08-16
[rush-lib] Add rush remove

sherlockfeng:
Provide a "rush remove" command that undoes "rush add"
https://github.com/microsoft/rushstack/issues/1457


sherlockfeng:
pipeline failed seemed not relative to my change

TheLarkInn (changes requested):
This looks mostly okay, for one exception: 

> Scenario: A project contains (hypothetically) a dependency that is listed as both dependency of different types (peer & devDep || etc.). How should the user proceed here.

Perhaps it's good to add a case like this to your unit tests. Otherwise some changes around removing variant (unless cc @octogonz corrects me) its not relevant in removing dependencies but rather updating/adding/installing.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -78,6 +83,31 @@ export interface IPackageJsonUpdaterRushAddOptions {
...
+   */
+  skipUpdate: boolean;
+  /**
+   * If specified, "rush update" will be run in debug mode.
+   */
+  debugInstall: boolean;
+  /**
+   * The variant to consider when performing installations and validating shrinkwrap updates.
+   */
+  variant?: string | undefined;
```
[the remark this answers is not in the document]
sherlockfeng: now `remove` action will remove every type of dependency in project

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -265,6 +314,79 @@ export class PackageJsonUpdater {
     }
   }
 
+  /**
+   * Remove a dependency from a particular project. The core business logic for "rush remove".
+   */
+  public async doRushRemoveAsync(options: IPackageJsonUpdaterRushRemoveOptions): Promise<void> {
+    const { projects, packagesToRemove, skipUpdate, debugInstall, variant } = options;
```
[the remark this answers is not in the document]
sherlockfeng: done~

libraries/rush-lib/src/logic/versionMismatch/VersionMismatchFinderCommonVersions.ts
```diff
@@ -59,6 +59,10 @@ export class VersionMismatchFinderCommonVersions extends VersionMismatchFinderEn
     this._fileManager.preferredVersions.set(packageName, newVersion);
   }
 
+  public removeDependency(packageName: string): void {
+    return undefined;
```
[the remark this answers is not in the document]
sherlockfeng: now throw an error with message `should not use this function`

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -265,6 +314,79 @@ export class PackageJsonUpdater {
...
+      const installManagerOptions: IInstallManagerOptions = {
+        debug: debugInstall,
+        allowShrinkwrapUpdates: true,
+        bypassPolicy: false,
+        noLink: false,
+        fullUpgrade: false,
+        recheckShrinkwrap: false,
+        networkConcurrency: undefined,
+        collectLogFile: false,
+        variant: variant,
```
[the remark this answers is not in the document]
sherlockfeng: done

libraries/rush-lib/src/cli/actions/test/removeRepo/.gitignore
```diff
@@ -0,0 +1 @@
+common/temp
```
iclanton, line 1: This should already be ignored. All folders in the repo called `temp` are ignored.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -78,6 +83,27 @@ export interface IPackageJsonUpdaterRushAddOptions {
   variant?: string | undefined;
 }
 
+/**
+ * Options for remove a dependency to a particular project.
+ */
+export interface IPackageJsonUpdaterRushRemoveOptions {
```
iclanton, line 89: This and `IPackageJsonUpdaterRushAddOptions` should be able to extend from a common base.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -99,6 +125,19 @@ export interface IUpdateProjectOptions {
   dependencyType?: DependencyType;
 }
 
+export interface IRemoveProjectOptions {
```
iclanton, line 128: This and `IUpdateProjectOptions` should be able to extend from a common base.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -265,6 +310,90 @@ export class PackageJsonUpdater {
...
+          dependencyTypes.push(dependency.dependencyType);
+        }
+        if (devDependency) {
+          dependencyTypes.push(devDependency.dependencyType);
+        }
+        if (resolution) {
+          dependencyTypes.push(resolution.dependencyType);
+        }
+        if (!dependencyTypes.length) {
+          throw new Error(`The project "${project.packageName}" do not have ${packageName} in package.json.`);
```
iclanton, line 343: Seems like we should probably just print an informational message here instead of a hard error. The user wants the dependency removed, so a state where it's not in the dependency list at the end of the run is what's expected even if it wasn't there in the first place.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -265,6 +310,90 @@ export class PackageJsonUpdater {
     }
   }
 
+  /**
+   * Remove a dependency from a particular project. The core business logic for "rush remove".
+   */
+  public async doRushRemoveAsync(options: IPackageJsonUpdaterRushRemoveOptions): Promise<void> {
```
iclanton, line 316: There's a fair amount of code duplicated between this and `doRushAddAsync`. Can these be DRYed up?

libraries/rush-lib/src/logic/versionMismatch/VersionMismatchFinderCommonVersions.ts
```diff
@@ -59,6 +59,10 @@ export class VersionMismatchFinderCommonVersions extends VersionMismatchFinderEn
     this._fileManager.preferredVersions.set(packageName, newVersion);
   }
 
+  public removeDependency(packageName: string): void {
+    throw new Error('should not use this function');
```
iclanton, line 63:
```suggestion
    throw new Error('Not supported.');
```

libraries/rush-lib/src/cli/actions/RemoveAction.ts
```diff
@@ -0,0 +1,112 @@
...
+import * as os from 'os';
+import { CommandLineFlagParameter, CommandLineStringListParameter } from '@rushstack/ts-command-line';
+
+import { RushConfigurationProject } from '../../api/RushConfigurationProject';
+import { BaseRushAction } from './BaseRushAction';
+import { RushCommandLineParser } from '../RushCommandLineParser';
+
+import type * as PackageJsonUpdaterType from '../../logic/PackageJsonUpdater';
+
+export class RemoveAction extends BaseRushAction {
```
iclanton, line 13: This is pretty similar to `AddAction`. Can these be DRYed up with a common base?

libraries/rush-lib/src/api/PackageJsonEditor.ts
```diff
@@ -220,6 +220,10 @@ export class PackageJsonEditor {
     return this._devDependencies.get(packageName);
   }
 
+  public tryGetResolution(packageName: string): PackageJsonDependency | undefined {
```
iclanton, line 223: What needs this exposed?

common/reviews/api/rush-lib.api.md
```diff
@@ -602,6 +604,8 @@ export class PackageJsonEditor {
     // (undocumented)
     tryGetDevDependency(packageName: string): PackageJsonDependency | undefined;
     // (undocumented)
+    tryGetResolution(packageName: string): PackageJsonDependency | undefined;
```
iclanton, line 607: Is this name clear?

libraries/rush-lib/src/cli/actions/RemoveAction.ts
```diff
@@ -0,0 +1,112 @@
...
+  private _allFlag!: CommandLineFlagParameter;
+  private _skipUpdateFlag!: CommandLineFlagParameter;
+  private _packageNameList!: CommandLineStringListParameter;
+
+  public constructor(parser: RushCommandLineParser) {
+    const documentation: string[] = [
+      'Remove specified package(s) to the dependencies of the current project (as determined by the current working directory)' +
+        ' and then runs "rush update". If no package was found, there will be an Error thrown out'
```
iclanton, line 21:
```suggestion
      'Removes specified package(s) from the dependencies of the current project (as determined by the current working directory)' +
        ' and then runs "rush update".'
```

libraries/rush-lib/src/cli/actions/RemoveAction.ts
```diff
@@ -0,0 +1,112 @@
...
+    ];
+    super({
+      actionName: 'remove',
+      summary: 'Remove one or more dependencies from the package.json and runs rush update.',
```
iclanton, line 25:
```suggestion
      summary: 'Removes one or more dependencies from the package.json and runs rush update.',
```

libraries/rush-lib/src/cli/actions/RemoveAction.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  public onDefineParameters(): void {
+    this._packageNameList = this.defineStringListParameter({
+      parameterLongName: '--package',
+      parameterShortName: '-p',
+      required: true,
+      argumentName: 'PACKAGE',
+      description:
+        'The name of the package which should be removed' +
+        ' To remove multiple packages, write "rush remove --package foo --package bar".'
```
iclanton, line 40:
```suggestion
        'The name of the package which should be removed.' +
        ' To remove multiple packages, run "rush remove --package foo --package bar".'
```

libraries/rush-lib/src/cli/actions/RemoveAction.ts
```diff
@@ -0,0 +1,112 @@
...
+    });
+    this._skipUpdateFlag = this.defineFlagParameter({
+      parameterLongName: '--skip-update',
+      parameterShortName: '-s',
+      description:
+        'If specified, the "rush update" command will not be run after updating the package.json files.'
+    });
+    this._allFlag = this.defineFlagParameter({
+      parameterLongName: '--all',
+      description: 'If specified, the dependency will be removed from all projects.'
```
iclanton, line 50:
```suggestion
      description: 'If specified, the dependency will be removed from all projects that declare it.'
```

iclanton (changes requested):

common/reviews/api/rush-lib.api.md
sherlockfeng: i test remove in yarn, and yarn will not remove project in `resolutions`, so i delete this function, now rush remove will not remove anything in resolution.and i think this logic is more similar to 'rush update' or 'rush add'

libraries/rush-lib/src/cli/actions/RemoveAction.ts
sherlockfeng: done

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
sherlockfeng: done

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
sherlockfeng: done

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
sherlockfeng: done

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
sherlockfeng: done

libraries/rush-lib/src/cli/actions/test/removeRepo/.gitignore
sherlockfeng: remove this file

iclanton (approved):

sherlockfeng:
> This looks mostly okay, for one exception:
> 
> > Scenario: A project contains (hypothetically) a dependency that is listed as both dependency of different types (peer & devDep || etc.). How should the user proceed here.
> 
> Perhaps it's good to add a case like this to your unit tests. Otherwise some changes around removing variant (unless cc @octogonz corrects me) its not relevant in removing dependencies but rather updating/adding/installing.
 i add a case in my unit tests, `rush remove` will remove all type of dependency like `pnpm remove`

TheLarkInn (approved):

TheLarkInn:
Look's great now. Thank you for this contribution 🙏

PR #3591, opened 2022-08-18
[node-core-library] Remove node < 11 sort

dmichon-msft:
## Summary
Reroutes `LegacyAdapters.sortStable` to always call `Array.prototype.sort`, since node 12 and below are EOL and all currently maintained LTS versions use stable sorting.

## Details
Disables the node version check to remove the dependency on `semver` and `timsort`.

## How it was tested
Unit tests.


iclanton (approved):

iclanton:
Missing a changelog

iclanton (approved):

PR #3606, opened 2022-08-30
[heft-jest-plugin] Add a patch to update the Jest ScriptTransformer to disable cache reads/writes

D4N14L:
## Summary
Jest uses cache files to speed up execution time. However, use of cache files with the heft-jest-plugin seems to cause more problems than it solves (adds I/O overhead, encounters race conditions). We already delete the cache before execution in an attempt to avoid some of these issues. Additionally, cache files for the heft-jest-plugin typescript jest transformer only read a file from disk and feed it to Jest, which means these cache files serve little purpose. This PR patches Jest to disable cache reads/writes entirely.

## How it was tested
Tested by debugging into the execution and validating that the patched script is used.


dmichon-msft:
Regarding the description, specifically, cache files for the Heft Typescript Jest transformer are counterproductive, since all the transformer does is read a file from disk and feed it to Jest, so the cache file does nothing useful and adds an extra I/O round trip.

heft-plugins/heft-jest-plugin/src/patches/jestScriptTransformerPatch.js
```diff
@@ -0,0 +1,75 @@
...
+
+    const startIndex = match.index;
+    const endIndex = scriptContent.indexOf('};', startIndex) + 2;
+    scriptContent =
+      scriptContent.slice(0, startIndex) +
+      `const ${functionName} = () => {};` +
+      scriptContent.slice(endIndex);
+  }
+
+  fs.unlinkSync(scriptPath);
```
dmichon-msft, line 66: Add a comment that this is to break any hardlinks that might exist for this file.

heft-plugins/heft-jest-plugin/package.json
```diff
@@ -15,7 +15,8 @@
     "build": "heft build --clean",
     "start": "heft test --clean --watch",
     "_phase:build": "heft build --clean",
-    "_phase:test": "heft test --no-build"
+    "_phase:test": "heft test --no-build",
+    "postinstall": "node ./src/patches/jestScriptTransformerPatch.js"
```
dmichon-msft, line 19: Does the `src` folder ship? If not, this won't work when the package is published. Likely will need to put it in a folder that exists at both workspace install time and shipped install time, or else manually add `src/patches/jestScriptTransformerPatch.js` to the package manifest

dmichon-msft (commented):
Won't quite work as written

heft-plugins/heft-jest-plugin/package.json
D4N14L: Yeah we don't publish the `src` folder. Moved it to `includes` which does get published.

dmichon-msft (approved):

PR #3607, opened 2022-08-30
[heft-jest-plugin] Hide disabling Jest cache behind an environment variable

D4N14L:
## Summary
Hides the disabling of Jest cache read/writes behind an environment variable, to avoid touching the Jest install unless explicitly requested. Follow-on PR for #3606


iclanton:
The env variable should be in the patch code, so that it only applies when Heft is running.

iclanton (approved):

PR #3608, opened 2022-08-31
[rush-lib] Add RUSH_PNPM_VERIFY_STORE_INTEGRITY environment variable

dmichon-msft:
## Summary
Adds an environment variable `RUSH_PNPM_VERIFY_STORE_INTEGRITY` that can be used to enable or disable PNPM's store integrity verification checks during the installation process.

## Details
Reasons to disable are for performance, or if you are deliberately tampering with your local PNPM store for test purposes.

## How it was tested
Local runs in a test repository.

### DIFF
--- a/libraries/rush-lib/src/api/EnvironmentConfiguration.ts
+++ b/libraries/rush-lib/src/api/EnvironmentConfiguration.ts
@@ -80,6 +80,13 @@ export enum EnvironmentVariableNames {
   RUSH_PNPM_STORE_PATH = 'RUSH_PNPM_STORE_PATH',
 
   /**
+   * When using PNPM as the package manager, this variable can be used to control whether or not PNPM
+   * validates the integrity of the PNPM store during installation. The value of this environment variable must be
+   * `1` (for true) or `0` (for false). If not specified, defaults to the value in .npmrc.
+   */
+  RUSH_PNPM_VERIFY_STORE_INTEGRITY = 'RUSH_PNPM_VERIFY_STORE_INTEGRITY',
+
+  /**
    * This environment variable can be used to specify the `--target-folder` parameter
    * for the "rush deploy" command.
    */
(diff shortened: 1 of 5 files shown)


libraries/rush-lib/src/api/EnvironmentConfiguration.ts
```diff
@@ -373,6 +391,12 @@ export class EnvironmentConfiguration {
             break;
           }
 
+          case EnvironmentVariableNames.RUSH_PNPM_VERIFY_STORE_INTEGRITY: {
+            EnvironmentConfiguration._pnpmVerifyStoreIntegrity =
+              value === '1' ? true : value === '0' ? false : undefined;
```
D4N14L, line 396: Should we just switch on whether or not it's defined?

dmichon-msft: This is technically an override for whatever is set in .npmrc, which defaults to `true`. This means I need to be able to set it to `false`, and there is utility to setting it to `true` if your .npmrc sets it to `false` but you don't trust the environment you are running on (or know that you have tampered with the store and want to fix it).

libraries/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
```diff
@@ -428,6 +428,11 @@ export class WorkspaceInstallManager extends BaseInstallManager {
       args.push('--recursive');
       args.push('--link-workspace-packages', 'false');
 
+      const { pnpmVerifyStoreIntegrity } = EnvironmentConfiguration;
+      if (pnpmVerifyStoreIntegrity !== undefined) {
```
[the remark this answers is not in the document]
dmichon-msft: Yes.

D4N14L (approved):

iclanton (approved):

PR #3630, opened 2022-09-14
[api-extractor] fix references from computed properties

### LINKED ISSUE #3629: [api-extractor] wrong model reference from an interface to a property variable
## Summary
Having a symbol variable that is used in an interface as a property key, when I run api-extractor, the model json file contains an invalid pointer from the interface to the variable.
 
## Repro steps
It is reproducible with a simple ts file on input:
```ts
export const FLUX_VALUE = Symbol("FLUX_VALUE");
export interface FluxParameterLike {
  [FLUX_VALUE](): string;
}
```

I created a complete project so that it is easy to reproduce:

```bash
git clone git@github.com:sranka/api_extractor_defect2.git
yarn
yarn build
yarn extract
yarn document
```

**Expected result:** 
The model JSON file references `FLUX_VALUE` using a canonicalReference that is defined, i.e:
```
"canonicalReference": "api_extractor_defect!FLUX_VALUE:var "
```
**Actual result:** 
The model JSON file references
`FLUX_VALUE` (https://github.com/sranka/api_extractor_defect2/blob/master/docs/api_extractor_defect.api.json#L260)
wrongly as
```
"canonicalReference": "api_extractor_defect!FLUX_VALUE"
```
, but it was defined in https://github.com/sranka/api_extractor_defect2/blob/master/docs/api_extractor_defect.api.json#L177 as
```
"canonicalReference": "api_extractor_defect!FLUX_VALUE:var"
```

## Details
This issue actually happens in [@influxdata/influxdb-client-js](https://github.com/influxdata/influxdb-client-js), it validates  references in the api model files by a simple [node.js script](https://github.com/influxdata/influxdb-client-js/blob/master/scripts/fix-extracted-api-files.js). The same script, as a workaround, is also capable to repair wrong references.

I posted the defect just to let you know about it.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/api-extractor` version? | 7.31.0 |
| Operating system? | Linux |
| API Extractor scenario? | docs (.api.json) |
| Would you consider contributing a PR? | yes, if I would know the way of how to fix it |
| TypeScript compiler version? | 4.7 and 4.8 |
| Node.js version (`node -v`)? | 16 |

sranka:
## Summary

Fixes #3629

## Details

The code that checks whether a node is in an expression context was incomplete, an expression also appears inside object's computed property. 

## How it was tested

Tested manually, against an example repository described in #3629 , the issue is solved with the changes in this PR.

### DIFF
--- a/apps/api-extractor/src/generators/DeclarationReferenceGenerator.ts
+++ b/apps/api-extractor/src/generators/DeclarationReferenceGenerator.ts
@@ -60,6 +60,7 @@ export class DeclarationReferenceGenerator {
   private static _isInExpressionContext(node: ts.Node): boolean {
     switch (node.parent.kind) {
       case ts.SyntaxKind.TypeQuery:
+      case ts.SyntaxKind.ComputedPropertyName:
         return true;
       case ts.SyntaxKind.QualifiedName:
         return DeclarationReferenceGenerator._isInExpressionContext(node.parent);
--- a/build-tests/api-extractor-scenarios/etc/referenceTokens/api-extractor-scenarios.api.json
+++ b/build-tests/api-extractor-scenarios/etc/referenceTokens/api-extractor-scenarios.api.json
@@ -809,6 +809,42 @@
           "members": [
             {
               "kind": "PropertySignature",
+              "canonicalReference": "api-extractor-scenarios!SomeInterface1#[SomeSymbol1]:member",
+              "docComment": "",
+              "excerptTokens": [
+                {
+                  "kind": "Content",
+                  "text": "["
+                },
+                {
+                  "kind": "Reference",
+                  "text": "SomeSymbol1",
+                  "canonicalReference": "api-extractor-scenarios!SomeSymbol1:var"
+                },
+                {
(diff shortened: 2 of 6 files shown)


zelliott (approved):
Thanks for the fix!

If it's not too much work, is it possible to add a quick test case to the build-tests/api-extractor-scenarios project? It would probably make most sense to add it to either `apiItemKinds/interface.ts` or `referenceTokens/index.ts`... I think either of those files makes sense.

Then, you can run `rush rebuild -t api-extractor-scenarios`, and manually confirm that the produced `.api.json` file has the appropriate references. It's probably also a good idea to just run `rush rebuild` altogether, which will rebuild the entire monorepo, including all of the test projects in `build-tests/`.

sranka:
@zelliott test case added, results manually reviewed

zelliott:
@octogonz I think this is ready for you to take a look!

octogonz (approved):
LGTM -- thank you for fixing this!  Sorry it took us a little while to get it merged.

PR #3636, opened 2022-09-20
[rush-lib] Add commit flag to the Change action

### LINKED ISSUE #2470: [rush] Feature Request: Option to git add & commit automatically after rush change
## Summary

When running rush change, some files are created. It would be nice to have an additional flag that auto commits and adds the rush change files. that was just added. Otherwise, developers could send PRs out without the rush files actually being in the commit

## Repro steps

none, feature request

  **Expected result:** 

  **Actual result:** 

## Details

It would be more convenient to have rush commit its own files after running the `rush change` command, since rush already throws errors on uncommitted files, it would make sense to offer an option to have rush commit its own files or amend the last commit and include its change files.

## Standard questions

| Question | Answer |
| -------- | -------- |
| Operating system? |  |

a-mkljajic:
## Summary

Adds an optional flag to commit the change files, it stages the files in `rushConfiguration.changesFolder` and commits them with the message `Rush changes`.

Fixes #2470 

## Details

I solved the problem by adding a new Flag parameter to the Change action with the long name `--commit` and short name `-c`. If this flag is set, Rush invokes `git add ${rushConfiguration.changesFolder}` and commits the changes with the message `Rush changes`.
An alternative I considered but was not sure about was extending the `Git` class to be able to do staging and committing, then just invoking those functions via the `_git` instance in the `ChangeAction` class, but since there were other functions invoking git via CLI directly, I figured this was fine.
~~The problem is solved, the only question is whether committing changes that were staged already before the `rush change` action was run should be done. It does so now.~~
The problem is solved, it commits only files from the `changesFolder`, has an optional custom commit message in rush.json under `gitPolicy.changefilesCommitMessage`.
The change does not break backwards compatibility.
The change should not impact performance.

## How it was tested

Invoked `rush change` with both the `--commit` and `-c` flags and confirmed that the file was made and committed.
I tested with the custom commit message as well, and it works.


libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -690,4 +702,23 @@ export class ChangeAction extends BaseRushAction {
...
     console.log('No changes were detected to relevant packages on this branch. Nothing to do.');
   }
+
+  private _stageGitChanges(patterns: string[]): void {
+    try {
+      patterns.forEach((pattern) => {
+        child_process.execSync(`git add "${pattern}"`);
+      });
+    } catch (error) {
+      console.error(`Cannot stage git changes ${(error as Error).message}`);
```
kredenac, line 712: is this the usual way of logging in rush? Seems to be, tho I assumed they'd use custom logger wrappers

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -690,4 +702,23 @@ export class ChangeAction extends BaseRushAction {
...
+    }
+  }
+
+  private _stageAndCommitGitChanges(patterns: string[], message: string) {
+    try {
+      this._stageGitChanges(patterns);
+      child_process.execSync(`git commit -m "${message}"`);
```
kredenac, line 719: does rush invoke other processes like this too?

common/changes/@microsoft/rush/main_2022-09-20-13-40.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Add commit flag that commits the change files.",
+      "type": "none"
```
ndobryanskyy, line 6: Should this be `minor`? Since you effectively extended the API

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -130,6 +131,12 @@ export class ChangeAction extends BaseRushAction {
         `(or erroring in ${BULK_LONG_NAME} mode).`
     });
 
+    this._commitChangesFlagParameter = this.defineFlagParameter({
+      parameterLongName: '--commit',
+      parameterShortName: '-c',
+      description: `Commits the change files to git.`
```
ndobryanskyy, line 137: nit:

```suggestion
      description: `If this flag is specified generated changefiles will be commited automatically.`
```

A bit more consistent language with other flags.

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -297,6 +305,10 @@ export class ChangeAction extends BaseRushAction {
     } catch (error) {
       throw new Error(`There was an error creating a change file: ${(error as Error).toString()}`);
     }
+    if (this._commitChangesFlagParameter.value) {
+      this._stageAndCommitGitChanges([this.rushConfiguration.changesFolder], 'Rush changes');
+    }
+    return;
```
ndobryanskyy, line 311:
```suggestion
```

Is this `return` statement necessary?

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -297,6 +305,10 @@ export class ChangeAction extends BaseRushAction {
     } catch (error) {
       throw new Error(`There was an error creating a change file: ${(error as Error).toString()}`);
     }
+    if (this._commitChangesFlagParameter.value) {
+      this._stageAndCommitGitChanges([this.rushConfiguration.changesFolder], 'Rush changes');
```
ndobryanskyy, line 309: nit:

I think this commit message should be configurable from `rush.json` too.
We can extend `gitPolicy` object with new field, like `changesCommitMessage`, we just need to make sure that it is distinguishable from `changeLogUpdateCommitMessage` that is already there

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -690,4 +702,23 @@ export class ChangeAction extends BaseRushAction {
   private _logNoChangeFileRequired(): void {
     console.log('No changes were detected to relevant packages on this branch. Nothing to do.');
   }
+
+  private _stageGitChanges(patterns: string[]): void {
```
ndobryanskyy, line 706: Why does it receive an array as parameter and not a single `string`?

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -690,4 +702,23 @@ export class ChangeAction extends BaseRushAction {
...
+    try {
+      patterns.forEach((pattern) => {
+        child_process.execSync(`git add "${pattern}"`);
```
ndobryanskyy, line 709: Instead of this you can use Rush shared helper command executor from [here](https://github.com/microsoft/rushstack/blob/c7aef037ee57dbb0892b2882c9d8404cc51ce0fc/libraries/rush-lib/src/utilities/Utilities.ts#L319)

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -690,4 +702,23 @@ export class ChangeAction extends BaseRushAction {
...
+      });
+    } catch (error) {
+      console.error(`Cannot stage git changes ${(error as Error).message}`);
+    }
+  }
+
+  private _stageAndCommitGitChanges(patterns: string[], message: string) {
+    try {
+      this._stageGitChanges(patterns);
```
ndobryanskyy, line 718: Why don't we inline this method call here too? Or otherwise call two separate functions in on the call site, because otherwise this looks a bit strange with one function extracted and one function inlined. So the code has different levels of abstraction.

common/changes/@microsoft/rush/main_2022-09-20-15-16.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
```
ndobryanskyy, line 2: Are you sure we need to check in this file?

a-mkljajic: Oops, this was committed while testing, forgot to remove it 😆

libraries/rush-lib/src/cli/actions/ChangeAction.ts
a-mkljajic: Changed to that

libraries/rush-lib/src/cli/actions/ChangeAction.ts
a-mkljajic: Inlined both of them using the `Utilities.executeCommand`

ndobryanskyy:
Also @a-mkljajic for you question in the PR description you can make use of  `_warnUnstagedChanges` it already checks if when running `rush change` user has anything staged. I'd say if flag `--commit` is set then this check should just throw error, since I don't think this will be cool if we commit random things for the user.

a-mkljajic:
> Also @a-mkljajic for you question in the PR description you can make use of `_warnUnstagedChanges` it already checks if when running `rush change` user has anything staged. I'd say if flag `--commit` is set then this check should just throw error, since I don't think this will be cool if we commit random things for the user.

@ndobryanskyy My latest commit moves to just committing the staged change files, I forgot that you could do that 😄

libraries/rush-lib/src/cli/actions/ChangeAction.ts
a-mkljajic: Moved to just a single string

libraries/rush-lib/src/cli/actions/ChangeAction.ts
a-mkljajic: Added

libraries/rush-lib/src/cli/actions/ChangeAction.ts
a-mkljajic: 🤷‍♂️

libraries/rush-lib/src/schemas/rush.schema.json
```diff
@@ -181,6 +181,10 @@
           "description": "The commit message to use when committing change log files \"rush version\". Defaults to \"Update changelogs [skip ci]\"",
           "type": "string"
         },
+        "changefilesCommitMessage": {
```
iclanton, line 184: Add this property to the `rush init` template (here https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/assets/rush-init/rush.json)

libraries/rush-lib/src/schemas/rush.schema.json
```diff
@@ -181,6 +181,10 @@
...
+          "description": "The commit message to use when commiting change files made by \"rush change\". Defaults to\"Rush change\"",
```
iclanton, line 185:
```suggestion
          "description": "The commit message to use when commiting change files made by \"rush change\". Defaults to \"Rush change\"",
```

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -297,6 +306,12 @@ export class ChangeAction extends BaseRushAction {
     } catch (error) {
       throw new Error(`There was an error creating a change file: ${(error as Error).toString()}`);
     }
+    if (this._commitChangesFlagParameter.value) {
+      this._stageAndCommitGitChanges(
+        '*',
```
iclanton, line 311: It'd be better to specifically enumerate the changefiles that this command created, instead of committing every change to the `common/changes` folder.

common/changes/@microsoft/rush/main_2022-09-20-13-40.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Add commit flag that commits the change files.",
```
iclanton, line 5:
```suggestion
      "comment": "Add a `--commit` flag to `rush change` that commits the change files automatically.",
```

libraries/rush-lib/src/cli/actions/ChangeAction.ts
iclanton: This could be done fairly easily by updating `_writeChangeFiles` to return the list of file paths that it created and then passing those in to `rush add` as arguments.

iclanton: This also raises a potential issue where this flag will commit anything that the user has staged, in addition to the changefiles. That's probably not the desired behavior.

libraries/rush-lib/src/cli/actions/ChangeAction.ts
iclanton: Use `this._terminal.writeErrorLine`

common/changes/@microsoft/rush/main_2022-09-20-13-40.json
iclanton: Because Rush is lockstep versioned, bump-types aren't passed in here.

libraries/rush-lib/src/schemas/rush.schema.json
a-mkljajic: Done, didn't know about this, thanks!

libraries/rush-lib/src/cli/actions/ChangeAction.ts
a-mkljajic: We can commit only specific staged files, not all staged files by passing them to the commit command, so it's fine. 
I added the changes to `_writeChangeFiles`. Also tested them manually by running `rush change`, appending a changefile, and then afterwards running `rush change -c`. Only the latter changefile was committed, which can be seen in one of the commits. I also added a warning if `rush change -c` is run, but no changefiles were made, to warn the user that no commits were made.

By `rush add` did you mean `git add` or am I missing something?

a-mkljajic:
@iclanton Would adding an optional `--commit-message` string parameter be useful, or would that make things too cluttered?
Or is there a way to add an optional string parameter to the `--commit` flag?
I.e. you could run `rush change -c` and it would commit with the default commit message or the one defined in `rush.json` and you could run `rush change -c "Commit message"` and it would commit the changefiles with the `"Commit message"` message.
I haven't noticed anything in the `ts-command-line` package that could be used for this 2nd version, but maybe there is.

libraries/rush-lib/src/cli/actions/ChangeAction.ts
iclanton: Yeah I meant `git add`.

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -624,18 +643,26 @@ export class ChangeAction extends BaseRushAction {
...
+    const writtenFiles: string[] = [];
     await changeFileData.forEach(async (changeFile: IChangeFile) => {
-      await this._writeChangeFile(promptModule, changeFile, overwrite, interactiveMode);
+      const writtenFile: string | undefined = await this._writeChangeFile(
+        promptModule,
+        changeFile,
+        overwrite,
+        interactiveMode
+      );
+      if (writtenFile) writtenFiles.push(writtenFile);
```
iclanton, line 655:
```suggestion
      if (writtenFile) {
        writtenFiles.push(writtenFile);
      }
```

iclanton (approved):

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -297,6 +306,16 @@ export class ChangeAction extends BaseRushAction {
     } catch (error) {
       throw new Error(`There was an error creating a change file: ${(error as Error).toString()}`);
     }
+    if (this._commitChangesFlagParameter.value) {
+      if (changefiles && changefiles.length !== 0) {
+        this._stageAndCommitGitChanges(
+          changefiles,
+          this.rushConfiguration.gitChangefilesCommitMessage || 'Rush change'
```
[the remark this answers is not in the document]
a-mkljajic: I can't find any other default message in consts in the actions, the only consts I found were in `rush-lib/src/logic/RushConstants.ts`, but I'm not sure if that's the appropriate place to put this?

PR #3647, opened 2022-09-25
[rush-lib] remove gender from git config example

joelrbrandt:
## Summary

Removes an unnecessary use of the male gender (and of a male-gendered honorific) in the git config example.

## Details

Solved the problem by changing the example to a gender neutral `"Example Name <name@example.com>"`

## How it was tested

Manual testing.

Set my local git config to something that did not match the regex in rush.json (with `git config --local user.email "does-not-match@foo.com"`) and then ran `rush install` to see that the new message was correct.

### DIFF
--- a/common/changes/@microsoft/rush/remove-git-example-gender_2022-09-25-13-46.json
+++ b/common/changes/@microsoft/rush/remove-git-example-gender_2022-09-25-13-46.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Remove gender from the git config example in rush.json",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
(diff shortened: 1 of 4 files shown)


common/changes/@microsoft/rush/remove-git-example-gender_2022-09-25-13-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "[rush-lib] remove gender from git config example",
```
iclanton, line 5:
```suggestion
      "comment": "Remove gender from the git config example in rush.json",
```

iclanton:
You probably want to change the `rush init` template file too: https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/assets/rush-init/rush.json#L178

iclanton (approved):

joelrbrandt:
@iclanton 

> You probably want to change the `rush init` template file too: https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/assets/rush-init/rush.json#L178

Thanks! Changed. Back at you for review.

Love the Rush stack. Thanks for your hard work!

PR #3649, opened 2022-09-26
[rush] restore the duration of the same operation without cache hit as telemetry data

chengcyber:
## Summary

This PR makes Rush.js restores the duration of the same operation without cache hit and log `durationInSecondsWithoutCache` in operationResult

## Details

1. When operation doesn't hit cache, Rush.js creates a `.rush/temp/operation/<identifier>/state.json` which contains `durationInSecondsWithoutCache`.
2. Set the state.json in Step 1 as an additional output file when create project build cache tarball.
3. When operation hit cache, read the restored state.json and set `durationInSecondsWithoutCache` as one of the property of operationResult

## How it was tested

- [x] Tested it with local build cache in rushstack repo
- [x] Tested it with cloud build cache in private repo

Test with local build cache in rushstack repo steps:
1. Enable telemetry by setting up `"telemetryEnabled": true` in rush.json
2. Run `rm -rf common/temp/build-cache`
3. Run `node rush-lib/lib/start.js build -v --to tree-pattern`
  a.  `libraries/tree-pattern/.rush/temp/operation/_phase_build/state.json` created
  b. `libraries/tree-pattern/.rush/temp/rushstack+tree-pattern-_phase_build-<identifier>.log` shows `.rush/temp/operation/_phase_build/state.json` is included when creating build archive.
4. Remove the state.json by running `rm rf libraries/tree-pattern/.rush/temp/operation/_phase_build/state.json`
5. Run `node rush-lib/lib/start.js build -v --to tree-pattern` the second time
  a. `libraries/tree-pattern/.rush/temp/operation/_phase_build/state.json` restored
  b. `common/temp/telemetry/<...>.json` records `nonCachedDurationMs` in `operationResult` of `tree-pattern`

### DIFF
--- a/common/changes/@microsoft/rush/feat-operation-metadata_2022-09-26-12-16.json
+++ b/common/changes/@microsoft/rush/feat-operation-metadata_2022-09-26-12-16.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "restore the duration of the same operation without cache hit as telemetry data",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/common/changes/@microsoft/rush/feat-operation-metadata_2022-09-26-12-16.json
+++ b/common/changes/@microsoft/rush/feat-operation-metadata_2022-09-26-12-16.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "restore the duration of the same operation without cache hit as telemetry data",
+      "comment": "Include the operation duration in the telemetry data that was recorded during the non-cached run when a cache hit occurs.",
       "type": "none"
     }
   ],
--- a/libraries/rush-lib/src/logic/Telemetry.ts
+++ b/libraries/rush-lib/src/logic/Telemetry.ts
@@ -68,7 +68,7 @@ export interface ITelemetryOperationResult {
   endTimestampMs?: number;
 
   /**
-   * Duration in seconds when the operation does not hit cache
+   * Duration in milliseconds when the operation does not hit cache
    */
   nonCachedDurationMs?: number;
 }
--- a/libraries/rush-lib/src/logic/Telemetry.ts
+++ b/libraries/rush-lib/src/logic/Telemetry.ts
@@ -66,6 +66,11 @@ export interface ITelemetryOperationResult {
    * If the operation was blocked, will be `undefined`.
    */
   endTimestampMs?: number;
+
+  /**
+   * Duration in seconds when the operation does not hit cache
+   */
+  durationInSecondsWithoutCache?: number;
 }
 
 /**
--- a/libraries/rush-lib/src/logic/Telemetry.ts
+++ b/libraries/rush-lib/src/logic/Telemetry.ts
@@ -70,7 +70,7 @@ export interface ITelemetryOperationResult {
   /**
    * Duration in seconds when the operation does not hit cache
    */
-  durationInSecondsWithoutCache?: number;
+  nonCachedDurationMs?: number;
 }
 
 /**
--- a/libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
+++ b/libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
@@ -31,6 +31,11 @@ export interface IOperationExecutionResult {
    * Object used to report a summary at the end of the Rush invocation.
    */
   readonly stdioSummarizer: StdioSummarizer;
+  /**
+   * The value indicates the duration of the same operation without cache hit.
+   * It is specified if current operation hits cache
+   */
+  readonly durationInSecondsWithoutCache: number | undefined;
 }
 
 /**
--- a/libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
+++ b/libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
@@ -33,7 +33,6 @@ export interface IOperationExecutionResult {
   readonly stdioSummarizer: StdioSummarizer;
   /**
    * The value indicates the duration of the same operation without cache hit.
-   * It is specified if current operation hits cache
    */
   readonly nonCachedDurationMs: number | undefined;
 }
--- a/libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
+++ b/libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
@@ -35,7 +35,7 @@ export interface IOperationExecutionResult {
    * The value indicates the duration of the same operation without cache hit.
    * It is specified if current operation hits cache
    */
-  readonly durationInSecondsWithoutCache: number | undefined;
+  readonly nonCachedDurationMs: number | undefined;
 }
 
 /**
--- a/libraries/rush-lib/src/logic/operations/OperationStateFile.ts
+++ b/libraries/rush-lib/src/logic/operations/OperationStateFile.ts
@@ -49,7 +49,7 @@ export class OperationStateFile {
    * @internal
    */
   public static getFilenameRelativeToProjectRoot(phase: IPhase): string {
-    const identifier: string = `${phase.logFilenameIdentifier}`;
+    const identifier: string = phase.logFilenameIdentifier;
     return `${RushConstants.projectRushFolderName}/${RushConstants.rushTempFolderName}/operation/${identifier}/state.json`;
   }
 
--- a/libraries/rush-lib/src/logic/operations/OperationStateFile.ts
+++ b/libraries/rush-lib/src/logic/operations/OperationStateFile.ts
@@ -1,0 +1,71 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import * as path from 'path';
+import { FileSystem, InternalError, JsonFile } from '@rushstack/node-core-library';
+import { RushConstants } from '../RushConstants';
+
+import type { IPhase } from '../../api/CommandLineConfiguration';
+import type { RushConfigurationProject } from '../../api/RushConfigurationProject';
+
+export interface IOperationStateFileOptions {
+  rushProject: RushConfigurationProject;
+  phase: IPhase;
+}
+
+export interface IOperationStateJson {
+  durationInSecondsWithoutCache: number;
+}
+
+export class OperationStateFile {
+  private readonly _rushProject: RushConfigurationProject;
+  private _filename: string;
+
+  public constructor(options: IOperationStateFileOptions) {
+    const { rushProject, phase } = options;
+    this._rushProject = rushProject;
+    this._filename = OperationStateFile.getFilename(phase, rushProject);
+  }
+
+  public static getFilename(phase: IPhase, project: RushConfigurationProject): string {
+    const relativeFilename: string = OperationStateFile.getFilenameRelativeToProjectRoot(phase);
+    return path.join(project.projectFolder, relativeFilename);
+  }
+
+  public static getFilenameRelativeToProjectRoot(phase: IPhase): string {
+    const identifier: string = `${phase.logFilenameIdentifier}`;
+    return path.join(
+      RushConstants.projectRushFolderName,
+      RushConstants.rushTempFolderName,
+      'operation',
+      identifier,
+      'state.json'
+    );
+  }
+
+  /**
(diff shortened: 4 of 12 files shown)


libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
```diff
@@ -31,6 +31,11 @@ export interface IOperationExecutionResult {
    * Object used to report a summary at the end of the Rush invocation.
    */
   readonly stdioSummarizer: StdioSummarizer;
+  /**
+   * The value indicates the duration of the same operation without cache hit.
+   * It is specified if current operation hits cache
+   */
+  readonly durationInSecondsWithoutCache: number | undefined;
```
dmichon-msft, line 38: I'd call this `originalDurationInSeconds` or similar, to signify that it is the duration from the original run of an operation that was skipped/replayed

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
```diff
@@ -442,6 +461,9 @@ export class ShellOperationRunner implements IOperationRunner {
                 projectChangeAnalyzer: this._projectChangeAnalyzer,
                 phaseName: this._phase.name
               });
+              this._projectBuildCache?.addAdditionalOutputFilePaths(
```
dmichon-msft, line 464: We'll be best served by adding an entire folder to all operations that can be used for rush to dump stuff, e.g. the build log, the new state file, etc.

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
```diff
@@ -370,6 +377,18 @@ export class ShellOperationRunner implements IOperationRunner {
           ensureFolderExists: true
         });
 
+        // If the operation without cache was successful, we can save the state to disk
+        if (context instanceof OperationExecutionRecord) {
```
dmichon-msft, line 381: The point of the input context being an interface is that runners should only be interacting with that which is on the interface. The concrete class is purely meant for use by OperationExecutionManager.

dmichon-msft: It sounds like the new `originalDurationInSeconds` field should be promoted to the interface

libraries/rush-lib/src/logic/operations/OperationStateFile.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  public static getFilenameRelativeToProjectRoot(phase: IPhase): string {
+    const identifier: string = `${phase.logFilenameIdentifier}`;
+    return path.join(
+      RushConstants.projectRushFolderName,
+      RushConstants.rushTempFolderName,
+      'operation',
+      identifier,
+      'state.json'
+    );
```
iclanton, line 43:
```suggestion
    return `${RushConstants.projectRushFolderName}/${RushConstants.rushTempFolderName}/operation/${identifier}/state.json`;
```

libraries/rush-lib/src/logic/operations/OperationStateFile.ts
```diff
@@ -0,0 +1,71 @@
...
+  phase: IPhase;
+}
+
+export interface IOperationStateJson {
+  durationInSecondsWithoutCache: number;
+}
+
+export class OperationStateFile {
+  private readonly _rushProject: RushConfigurationProject;
+  private _filename: string;
```
iclanton, line 22:
```suggestion
  private readonly _filename: string;
```

libraries/rush-lib/src/logic/operations/OperationStateFile.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { IPhase } from '../../api/CommandLineConfiguration';
+import type { RushConfigurationProject } from '../../api/RushConfigurationProject';
+
+export interface IOperationStateFileOptions {
+  rushProject: RushConfigurationProject;
+  phase: IPhase;
+}
+
+export interface IOperationStateJson {
+  durationInSecondsWithoutCache: number;
```
iclanton, line 17:
```suggestion
  nonCachedDurationMs: number;
```

libraries/rush-lib/src/logic/operations/OperationStateFile.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Returns the filename of the metadata file.
+   */
+  public get filename(): string {
+    return this._filename;
+  }
+
+  public write(json: IOperationStateJson): void {
+    JsonFile.save(json, this._filename, { ensureFolderExists: true, updateExistingFile: true });
```
iclanton, line 54:
```suggestion
  public async writeAsync(json: IOperationStateJson): Promise<void> {
    await JsonFile.saveAsync(json, this._filename, { ensureFolderExists: true, updateExistingFile: true });
```

libraries/rush-lib/src/logic/operations/OperationStateFile.ts
```diff
@@ -0,0 +1,71 @@
...
+    this._rushProject = rushProject;
+    this._filename = OperationStateFile.getFilename(phase, rushProject);
+  }
+
+  public static getFilename(phase: IPhase, project: RushConfigurationProject): string {
+    const relativeFilename: string = OperationStateFile.getFilenameRelativeToProjectRoot(phase);
+    return path.join(project.projectFolder, relativeFilename);
+  }
+
+  public static getFilenameRelativeToProjectRoot(phase: IPhase): string {
```
iclanton, line 35: Is it useful to have this as a separate function?

libraries/rush-lib/src/logic/operations/OperationStateFile.ts
```diff
@@ -0,0 +1,71 @@
...
+  private readonly _rushProject: RushConfigurationProject;
+  private _filename: string;
+
+  public constructor(options: IOperationStateFileOptions) {
+    const { rushProject, phase } = options;
+    this._rushProject = rushProject;
+    this._filename = OperationStateFile.getFilename(phase, rushProject);
+  }
+
+  public static getFilename(phase: IPhase, project: RushConfigurationProject): string {
```
iclanton, line 30: Does this need to be public?

libraries/rush-lib/src/logic/operations/OperationStateFile.ts
```diff
@@ -0,0 +1,71 @@
...
+    } catch (error) {
+      if (FileSystem.isFileDoesNotExistError(error as Error)) {
+        json = undefined;
+      } else {
+        // This should not happen
+        throw new InternalError(error);
+      }
+    }
+    return json;
+  }
```
iclanton, line 70:
```suggestion
  public async tryReadAsync(): Promise<IOperationStateJson | undefined> {
    try {
      return await JsonFile.loadAsync(this._filename);
    } catch (error) {
      if (FileSystem.isNotExistError(error as Error)) {
        return undefined;
      } else {
        // This should not happen
        throw new InternalError(error);
      }
    }
  }
```

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
```diff
@@ -273,6 +275,11 @@ export class ShellOperationRunner implements IOperationRunner {
           await projectBuildCache?.tryRestoreFromCacheAsync(terminal);
 
         if (restoreFromCacheSuccess) {
+          // Restore the original state of the operation without cache if hit cache
+          if (context instanceof OperationExecutionRecord) {
```
iclanton, line 279: This check is somewhat weird. When wouldn't this be the case?

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
iclanton, line 381: Same here.

common/changes/@microsoft/rush/feat-operation-metadata_2022-09-26-12-16.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "restore the duration of the same operation without cache hit as telemetry data",
```
iclanton, line 5:
```suggestion
      "comment": "Include the operation duration in the telemetry data that was recorded during the non-cached run when a cache hit occurs.",
```

libraries/rush-lib/src/logic/Telemetry.ts
```diff
@@ -66,6 +66,11 @@ export interface ITelemetryOperationResult {
    * If the operation was blocked, will be `undefined`.
    */
   endTimestampMs?: number;
+
+  /**
+   * Duration in seconds when the operation does not hit cache
+   */
+  durationInSecondsWithoutCache?: number;
```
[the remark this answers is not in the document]
dmichon-msft: originalDurationMs

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
chengcyber: That would be great! But i think it involves more design and maybe in another MR?

Because now if Rush sees `phase.isSynthetic`, the logout is `xxx.build.log`(not under an individual folder ❌), otherwise the logouts are under `rush-log`, which is also not a good name to cover this new state file.

On the other hand, the state file in this MR is new. It can be well designed. Now I chose to save the file under `.rush/temp/operation/<logIdentifier>` folder. I am open to relocate/rename the folder to be used for Rush to dump all the stuff.

libraries/rush-lib/src/logic/operations/IOperationExecutionResult.ts
```diff
@@ -31,6 +31,11 @@ export interface IOperationExecutionResult {
    * Object used to report a summary at the end of the Rush invocation.
    */
   readonly stdioSummarizer: StdioSummarizer;
+  /**
+   * The value indicates the duration of the same operation without cache hit.
+   * It is specified if current operation hits cache
```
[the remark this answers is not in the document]
chengcyber: I checked the code, and yes.

libraries/rush-lib/src/logic/Telemetry.ts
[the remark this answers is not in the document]
chengcyber: First, the original idea of the property is more like `originalDurationMs`, since it is only specified when the operation hits cache. but, Ian suggests the property should be always specified in "operationResult". So, given these two names a second thought:

1. Let's say `originalDurationMs` wins.

It somewhat weird if it's always be specified. For instance, w/o cache

```
operationResults: {
    "<package_name>": {
        "startTimestampMs": 1300.00,
        "endTimestampMs": 1400.00,
        "originalDurationMs": 100.00,
        "result": "SUCCESS",
        "dependencies": []
    }
}
```

It kind of confuses user why it calls `originalDurationMs` other than `durationMs` 😅

___

2. Let's say `nonCachedDurationMs` wins.

The data in the above case becomes

```
operationResults: {
    "<package_name>": {
        "startTimestampMs": 1300.00,
        "endTimestampMs": 1400.00,
        "nonCachedDurationMs": 100.00,
        "result": "SUCCESS",
        "dependencies": []
    }
}
```

It seems easier for user to understand this property is for `nonCached`, even though the expected `durationMs` is missing.


In conclusion, i would like to choose `nonCachedDurationMs`.

libraries/rush-lib/src/logic/operations/OperationStateFile.ts
chengcyber: The reason is `ProjectBuildCache` expects relative paths. For example, `.rush/temp/operation/build/state.json` is given. The <package_name>-<phase_name>-<identifier>.log would be

```
======= BEGIN PROCESS INPUT ======
.rush/temp/operation/_phase_build/state.json  <-- relative
dist/index.js
dist/index.d.ts
```

If an absolute path is given, it works. but the log would be

```
======= BEGIN PROCESS INPUT ======
/Users/<user_name>/<repo_root>/<project_folder>/.rush/temp/operation/_phase_build/state.json  <-- absolute
dist/index.js
dist/index.d.ts
```

No big deal tho

chengcyber:
Thanks @iclanton and @dmichon-msft for reviewing this. 

Batch of code changes according to your suggestions 🤞:

- renamed the property to `nonCachedDurationMs`
- always specifying `nonCachedDurationMs` in `operationResult`
- made `ProjectBuildCache#_additionalProjectOutputFiles` readonly
- promoted `operationStateFile` and `stopwatch` to `IOperationRunnerContext`

Could you take another look?

iclanton (approved):

iclanton (approved):

PR #3653, opened 2022-09-26
[rush] Remove node tar fallback

### LINKED ISSUE #2827: [rush] Corrupted build cache entries produce wrong (incomplete) builds
## Summary

I recently had a situation where I was building normally, with build cache enabled locally.  I started seeing several projects generate the following error:

```console
"tar" exited with code 1 while attempting to restore cache entry. Rush will attempt to extract from the cache entry with a JavaScript implementation of tar. See "[[path]]/.rush/temp/build-cache-tar.log" for logs from the tar process.
```

The log contains the following information:

```console
Start time: Mon Jul 26 2021 11:49:13 GMT-0400 (Eastern Daylight Time)
Invoking "/usr/bin/tar -x -m -f /[[path]]/common/temp/build-cache/4ab89cf5bc634f3196593aba1f209e3d32f9b072"

======= BEGIN PROCESS OUTPUT =======
[stderr] lib/styles/fonts/[[filename]].ttf: truncated gzip input
[stderr] tar: Error exit delayed from previous errors.
======== END PROCESS OUTPUT ========
```

Even though rush is building successfully with this warning, the tar file is indeed corrupted... only half of the built JavaScript files are present in the `lib/` folder.  It seems to me that the JavaScript implementation of `tar` is ignoring the fact that the archive is incomplete.

## Repro steps

Still TBD.  It's very unclear how the _original_ corrupted cache entries were created.  It might be possible to intentionally create a truncated archive file to test with.

I think the ideal situation here would be a harder error for a corrupted cache entry; or even better, automatic deletion followed by a normal build to repopulate the cache.

## Details

I have initially tried to see if rush was just incorrectly throwing away warnings, by trying the following out in a separate console:

```javascript
let tarStream = require('tar').extract({})
tarStream.on('warn', console.error);
tarStream.on('drain', console.log);
tarStream.write(require('fs').readFileSync('/[[path]]/temp/build-cache/4ab89cf5bc634f3196593aba1f209e3d32f9b072'))
```

This test was unsuccessful... it extracted all the files it could from the archive, without producing any kind of warning, just like rush does.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.48.0 |
| `rushVersion` from rush.json? | 5.48.0 |
| Operating system? | Mac |
| Would you consider contributing a PR? | Yes, although this may not be a rush problem (it may be a `tar` module problem) |
| Node.js version (`node -v`)? | 12.17.0 |

### LINKED ISSUE #3454: [rush] rush build fails and doesn't recover if a cache entry is invalid
## Summary

If you're using the rush build cache and an invalid entry gets written, it completely derails all future builds until the offending entry is manually removed.

## Repro steps

1. Enable the local build cache
2. Connect the cache directory to an external volume (e.g. NFS mount)
3. Fill up the volume (maybe leave a tiny bit of room)
4. Run `rush build`. An empty or incomplete entry will be written, and the build will fail
5. Run `rush build` again

At this point, all future builds will fail with this error:

```
13:02:02  "tar" exited with code 2 while attempting to restore cache entry. Rush will attempt to extract from the cache entry with a JavaScript implementation of tar. See "/home/jenkins/workspace/Geiger/geiger-review/packages/cv-tools/.rush/temp/b61388c431880616de380c9fd2040232dcb886b8.log" for logs from the tar process.
```

The log file referenced above confirms that the cache entry isn't a valid tarball:

```
Start time: Fri Jun 03 2022 20:02:02 GMT+0000 (Coordinated Universal Time)
Invoking "/bin/tar -x -m -f /home/jenkins/workspace/Geiger/geiger-review/common/temp/build-cache/b61388c431880616de380c9fd2040232dcb886b8"

======= BEGIN PROCESS INPUT ======

======== END PROCESS INPUT =======
======= BEGIN PROCESS OUTPUT =======
[stderr] /bin/tar: This does not look like a tar archive
[stderr] /bin/tar: Exiting with failure status due to previous errors
======== END PROCESS OUTPUT ========

Exited with code "2"
```

## Details

Rush is correct in saying that the cache entry is invalid, but it seems like a pretty bad bug that it gives up on the entire build. I'm currently trying to implement a global network build cache, so this error is causing all CI tasks to fail, because every task encounters the same corrupt cache entry.

If Rush encounters an invalid entry, it should remove the entry and then follow the steps it normally would: run the full build for that package and then store a new entry in the cache. Otherwise, this error scenario requires human intervention, which is annoying, since it could be easily automated with smarter logic.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.68.2 |
| `rushVersion` from rush.json? | 5.68.2 |
| Operating system? | Linux |
| Node.js version (`node -v`)? | 16.13.0 |

dmichon-msft:
## Summary
Fallback to the node implementation of `tar` in practice seems to only occur when the tarball is invalid, unless the `tar` is not on `PATH`. This results in restoring from corrupt cache entries and producing an invalid build result.

This PR removes the fallback to node `tar`.

Fixes #3454
Fixes #2827

## Details
Completely disables fallback path to node implementation of `tar`. Updates logged warning to suggest ensuring that `tar` is on `PATH` or setting `RUSH_TAR_EXECUTABLE_PATH` to point to `tar` in the event that the binary cannot be found.

## How it was tested
This is the normal state for most builds.

### DIFF
--- a/common/changes/@microsoft/rush/node-tar-cleanup_2022-09-26-20-33.json
+++ b/common/changes/@microsoft/rush/node-tar-cleanup_2022-09-26-20-33.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Remove fallback from tar binary to npm 'tar' package. The npm 'tar' package would sometimes produce invalid output if the cache entry was corrupt.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/libraries/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
+++ b/libraries/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
@@ -2,11 +2,8 @@
 // See LICENSE in the project root for license information.
 
 import * as path from 'path';
-import events from 'events';
 import * as crypto from 'crypto';
-import type * as stream from 'stream';
-import * as tar from 'tar';
-import { FileSystem, Path, ITerminal, FolderItem } from '@rushstack/node-core-library';
+import { FileSystem, Path, ITerminal, FolderItem, InternalError } from '@rushstack/node-core-library';
 
 import { RushConfigurationProject } from '../../api/RushConfigurationProject';
 import { ProjectChangeAnalyzer } from '../ProjectChangeAnalyzer';
@@ -16,7 +13,7 @@ import { BuildCacheConfiguration } from '../../api/BuildCacheConfiguration';
 import { ICloudBuildCacheProvider } from './ICloudBuildCacheProvider';
 import { FileSystemBuildCacheProvider } from './FileSystemBuildCacheProvider';
 import { TarExecutable } from '../../utilities/TarExecutable';
-import { Utilities } from '../../utilities/Utilities';
+import { EnvironmentVariableNames } from '../../api/EnvironmentConfiguration';
 
 export interface IProjectBuildCacheOptions {
   buildCacheConfiguration: BuildCacheConfiguration;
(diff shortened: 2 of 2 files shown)


libraries/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
```diff
@@ -326,12 +272,10 @@ export class ProjectBuildCache {
...
-      if (!cacheEntryBuffer) {
-        if (localCacheEntryPath) {
-          cacheEntryBuffer = await FileSystem.readFileToBufferAsync(localCacheEntryPath);
-        } else {
-          throw new Error('Expected the local cache entry path to be set.');
-        }
+      if (localCacheEntryPath) {
+        cacheEntryBuffer = await FileSystem.readFileToBufferAsync(localCacheEntryPath);
+      } else {
+        throw new Error('Expected the local cache entry path to be set.');
```
iclanton, line 278:
```suggestion
        throw new InternalError('Expected the local cache entry path to be set.');
```

common/changes/@microsoft/rush/node-tar-cleanup_2022-09-26-20-33.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Remove fallback from tar binary to npm 'tar' package.",
```
iclanton, line 5:
```suggestion
      "comment": "Remove fallback from tar binary to npm 'tar' package. The npm 'tar' package would sometimes produce invalid output if the cache entry was corrupt.",
```

iclanton (approved):

PR #3656, opened 2022-09-28
[loader-load-themed-styles] Make @microsoft/load-themed-styles a peerDependency of @microsoft/loader-load-themed-styles.

iclanton:


D4N14L (approved):
Just a note, the peer dep here won't be automatically updated when the version of the package gets bumped. That may be okay, just letting you know.

iclanton:
I intentionally made the peerdep version specifier broad.

PR #3669, opened 2022-10-03
[rush-serve-plugin] Allow serving single files

dmichon-msft:
## Summary
Extends `rush-project-serve.json` configuration to allow serving a specific file, not just entire directories.

## Details
Use the `projectRelativeFile` field instead of `projectRelativeFolder` to serve a file instead of a directory. A primary use case is to configure some entry point page to be served from `/` (root).

## How it was tested
Used a local repository and configured a file to be served from root.

### DIFF
--- a/rush-plugins/rush-serve-plugin/src/RushProjectServeConfigFile.ts
+++ b/rush-plugins/rush-serve-plugin/src/RushProjectServeConfigFile.ts
@@ -12,18 +12,19 @@ export interface IRushProjectServeJson {
   routing: IRoutingRuleJson[];
 }
 
-export interface IRoutingFolderRuleJson {
-  projectRelativeFile: undefined;
-  projectRelativeFolder: string;
+export interface IBaseRoutingRuleJson {
   servePath: string;
   immutable?: boolean;
 }
 
-export interface IRoutingFileRuleJson {
+export interface IRoutingFolderRuleJson extends IBaseRoutingRuleJson {
+  projectRelativeFile: undefined;
+  projectRelativeFolder: string;
+}
+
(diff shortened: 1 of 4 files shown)


rush-plugins/rush-serve-plugin/src/RushProjectServeConfigFile.ts
```diff
@@ -12,13 +12,24 @@ export interface IRushProjectServeJson {
   routing: IRoutingRuleJson[];
 }
 
-export interface IRoutingRuleJson {
+export interface IRoutingFolderRuleJson {
```
iclanton, line 15: These can extend from a common base.

iclanton (approved):

rush-plugins/rush-serve-plugin/src/RushProjectServeConfigFile.ts
```diff
@@ -12,13 +12,24 @@ export interface IRushProjectServeJson {
...
   servePath: string;
   immutable?: boolean;
 }
 
+export interface IRoutingFileRuleJson {
+  projectRelativeFile: string;
+  projectRelativeFolder: undefined;
+  servePath: string;
+  immutable?: boolean;
+}
```
iclanton, line 27:
```suggestion
export interface IRoutingRuleJsonBase {
  servePath: string;
  immutable?: boolean;
}

export interface IRoutingFolderRuleJson extends IRoutingRuleJsonBase {
  projectRelativeFile: undefined;
  projectRelativeFolder: string;
}

export interface IRoutingFileRuleJson extends IRoutingRuleJsonBase {
  projectRelativeFile: string;
  projectRelativeFolder: undefined;
}
```

PR #3673, opened 2022-10-05
fixed loading issue for sass files starting with _

atingmicrosoft:
## Summary

Current rushstack is causing sass files to load incorrectly in odsp-web. Sass files with title names starting with _ (ex. _Animation.scss) are being excluded because some scss files don't end in '/'. Need to revert to previous version. 

## Details

Solved problem by reverting the code that checks scss files to previous build. 

## How it was tested

Tested in odsp-web manually and then came here for fix.

### DIFF
--- a/common/changes/@rushstack/heft-sass-plugin/atingmicrosoft-fix-scss-issues_2022-10-05-00-02.json
+++ b/common/changes/@rushstack/heft-sass-plugin/atingmicrosoft-fix-scss-issues_2022-10-05-00-02.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-sass-plugin",
+      "comment": "Reverted reverted _isSassPartial checker to previous version",
+      "type": "patch"
+    }
+  ],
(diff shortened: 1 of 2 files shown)


dmichon-msft:
Strictly speaking, the issue is that the recent update assumed that input paths have been normalized to use `/` as the directory separator. However, in the current version of Heft this is not the case, so the detector for SASS partials broke on Windows.

common/changes/@rushstack/heft-sass-plugin/atingmicrosoft-fix-scss-issues_2022-10-05-00-02.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-sass-plugin",
+      "comment": "Reverted reverted _isSassPartial checker to previous version",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue with identifying partial sass files on Windows.",
```

iclanton (approved):

Issue #3388, opened 2022-05-03
[rush] Feature Request - rush `upgrade-interactive`


TheLarkInn:
## Summary
Updating packages in large scale projects today involves having to fuss with package.json files, reference approved versions, and manually making entries and running rush update. This can be quite cumbersome, and this high barrier to management seems to create atrophy in staying up to date with dependencies across a repository. 

We can lower this barrier to management by leveraging prior work from tools like `yarn upgrade-interactive` and `npm-check` to provide visual terminal prompts helping users manage their project dependencies. 

[npm-check](https://github.com/dylang/npm-check) has a UI like this: 
![Example from npm-check repository](https://user-images.githubusercontent.com/3408176/166559499-9a197a13-a317-4e6f-99e2-c6620f51f1b7.png)

[yarn upgrade-interactive](https://classic.yarnpkg.com/lang/en/docs/cli/upgrade-interactive/) looks like this: 
![image](https://user-images.githubusercontent.com/3408176/166559721-caec3bd7-7643-426d-b319-68184e105c6a.png)

## Design Outline

`rush upgrade-interactive`

- List all project packages
  - User selects one (or many?) packages
- `npm-check` style prompt performed on that package.json for that project
  - Prompt for changes provided
- Tool updates the package.json file for that project
  - Rush check (?)
  - Prompt/Tool for updating approved/optional versions json file
- Prompt for additional projects to update packages
  - If yes, repeat steps
  - If no, `rush update`

TheLarkInn:
I have not reviewed the API surface for @dylang 's npm-check but I very much like the UI and it would be great to reuse a lot of its capabilities instead of rewriting our own visualizer, picker, prompter, etc. I will research this.

iclanton:
The would make a great plugin.

TheLarkInn:
I've been able to generate a prompt for packages and updating, but I'd like some design discussion around how we go about updates. 

Questions to discuss: 

- When a package is chosen to be upgraded on one package how should we upgrade the dependency: update the dep across all packages or just the one package? (I'm leaning towards all packages with an optional flag for one package)
- If we update a dependency on just one package, do we automatically resolve "optional/approved versions" or prompt to automatically resolve it?

h-parker-higgins:
[...]

PR #3693, opened 2022-10-11
[rush] Feature/upgrade interactive

TheLarkInn:
This PR Introduces upgrade-interactive feature for rush-lib. Partially addresses #3388. Separate PR's will be made for subsequent features. 

This allows a user to select one project across their monorepo and then interactively upgrade one or many of their dependencies.


common/config/rush/browser-approved-packages.json
```diff
@@ -2,6 +2,10 @@
 {
   "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/approved-packages.schema.json",
   "packages": [
+    {
+      "name": "npm-check",
```
TheLarkInn, line 6: Move to non-browser-approved packages

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -151,6 +181,196 @@ export class PackageJsonUpdater {
...
+        // TODO: Reenable to support mismatch checks for both devDeps and regularDeps
+        // we need to do a mismatch check
+        const mismatchFinder: VersionMismatchFinder = VersionMismatchFinder.getMismatches(
+          this._rushConfiguration,
+          {
+            variant: variant
+          }
+        );
+        const mismatches: string[] = mismatchFinder.getMismatches().filter((mismatch) => {
+          console.log(mismatch);
```
dmichon-msft, line 278: `this._terminal.writeLine`

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -151,6 +181,196 @@ export class PackageJsonUpdater {
...
+          return !projects.find((proj) => proj.packageName === mismatch);
```
dmichon-msft, line 279: I realize performance here is not a huge deal, but since `packageName` is required to be unique within the workspace, you should create a filtered `Map<string, RushConfigurationProject>` first and do `.has()` checks on it for O(1) instead of O(n)

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -151,6 +181,196 @@ export class PackageJsonUpdater {
...
+                      project: consumer,
+                      dependenciesToAddOrUpdateOrRemove: {
+                        [packageName]: version
+                      }
+                    });
+                  }
+                }
+              }
+            }
+          }
```
dmichon-msft, line 302: Factor this into a helper function that takes in `dependenciesToUpdate` so you can use it for `devDependenciesToUpdate` as well.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -151,6 +181,196 @@ export class PackageJsonUpdater {
...
+        implicitlyPreferredVersion,
+        explicitlyPreferredVersion,
+        inferredRangeStyle
+      );
+
+      if (devDependency) {
+        devDependenciesToUpdate[moduleName] = version;
+      } else {
+        dependenciesToUpdate[moduleName] = version;
+      }
```
dmichon-msft, line 224: Whether a given external package should be a `devDependency` or a `dependency` really depends on the nature of the consuming project. This logic is outside of the `projects` loop, which suggests that you are setting the type globally.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -151,6 +181,196 @@ export class PackageJsonUpdater {
...
+                }
+              }
+            }
+          }
+        }
+      }
+
+      this.updateProjects(otherPackageUpdates);
+
+      allPackageUpdates.push(currentProjectDepUpdate, currentProjectDevDepUpdate, ...otherPackageUpdates);
```
dmichon-msft, line 330: While not strictly a common scenario, if `otherPackageUpdates.length` ever exceeds around 10000, this will throw an Error due to stack overflow. Best not to use `...` in a function argument position for a collection of arbitrary size.

libraries/rush-lib/src/utilities/InteractiveUpgradeUI.ts
```diff
@@ -0,0 +1,187 @@
+import colors from 'colors/safe';
+import inquirer from 'inquirer';
+import type * as inquirerTypes from 'inquirer';
+import table from 'text-table';
+import _ from 'lodash';
```
dmichon-msft, line 5: I have a "kill it with fire" relationship with `lodash` dependencies since most things it does are natively supported in ECMAScript nowadays.

libraries/rush-lib/src/utilities/InteractiveUpgradeUI.ts
```diff
@@ -0,0 +1,187 @@
...
+  value: NpmCheck.INpmCheckPackage;
+  name: string | string[];
+  short: string;
+}
+
+type ChoiceTable = (Separator | IUpgradeInteractiveDepChoice | boolean | undefined)[] | undefined;
+
+// Taken from ansi-regex node module
+// todo: find out how to use the ansi-regex esmodule without node failures
+const ansiRegex = ({ onlyFirst = false } = {}): RegExp => {
```
dmichon-msft, line 28: Just because ECMAScript now has lambdas doesn't mean good old `function` doesn't still have its place for things like this. The scope hoisting magic of `function` also helps avoid definition ordering issues, and when dealing with `typedef` lint rules, avoids repeating yourself.

libraries/rush-lib/src/utilities/InteractiveUpgradeUI.ts
```diff
@@ -0,0 +1,187 @@
...
+    '[\\u001B\\u009B][[\\]()#;?]*(?:(?:(?:(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]+)*|[a-zA-Z\\d]+(?:;[-a-zA-Z\\d\\/#&.:=?%@~_]*)*)?\\u0007)',
+    '(?:(?:\\d{1,4}(?:;\\d{0,4})*)?[\\dA-PR-TZcf-nq-uy=><~]))'
+  ].join('|');
+
+  return new RegExp(pattern, onlyFirst ? undefined : 'g');
+};
+
+// Taken from strip-ansi node module
+// todo: find out how to use the strip-ansi esmodule without node failures
+const stripAnsi = (string: string): string => {
```
dmichon-msft, line 39: We have a function for this in `@rushstack/node-core-library` already.

common/config/rush/nonbrowser-approved-packages.json
```diff
@@ -554,6 +554,11 @@
       "name": "minimatch",
       "allowedCategories": [ "libraries" ]
     },
+    {
```
iclanton, line 557: Move these to `nonbrowser-approved-packages.json`

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
```diff
@@ -0,0 +1,58 @@
...
+import type * as PackageJsonUpdaterType from '../../logic/PackageJsonUpdater';
+
+export class UpgradeInteractiveAction extends BaseRushAction {
+  private _allFlag!: CommandLineFlagParameter;
+  private _makeConsistentFlag!: CommandLineFlagParameter;
+  private _installFlag!: CommandLineFlagParameter;
+  private _interactiveUpgrader!: InteractiveUpgrader;
+
+  public constructor(parser: RushCommandLineParser) {
+    const documentation: string = '';
```
iclanton, line 14: Populate this

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
```diff
@@ -0,0 +1,58 @@
...
+      actionName: 'upgrade-interactive',
+      summary: 'Provides interactive prompt for upgrading package dependencies per project',
+      safeForSimultaneousRushProcesses: false,
+      documentation,
+      parser
+    });
+    this._interactiveUpgrader = new InteractiveUpgrader(this.rushConfiguration);
+  }
+
+  public onDefineParameters(): void {
```
iclanton, line 25: These can get stuck in the constructor, and this function can be a no-op

libraries/rush-lib/src/cli/actions/npm-check-typings.d.ts
```diff
@@ -0,0 +1,50 @@
+declare module NpmCheck {
```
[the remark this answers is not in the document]
iclanton: Are these typings for a NPM package? If they are, we should stick this in a typeRoots folder and use `declare module 'npm-check'`

libraries/rush-lib/src/logic/InteractiveUpgrader.ts
```diff
@@ -0,0 +1,77 @@
...
+  packages: NpmCheck.INpmCheckPackage[];
+}
+
+interface IUpgradeInteractiveDeps {
+  projects: RushConfigurationProject[];
+  depsToUpgrade: IDepsToUpgradeAnswers;
+}
+
+export class InteractiveUpgrader {
+  private _rushConfiguration: RushConfiguration;
```
iclanton, line 21:
```suggestion
  private readonly _rushConfiguration: RushConfiguration;
```

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
```diff
@@ -0,0 +1,58 @@
...
+import { RushCommandLineParser } from '../RushCommandLineParser';
+import { BaseRushAction } from './BaseRushAction';
+import { InteractiveUpgrader } from '../../logic/InteractiveUpgrader';
+import type * as PackageJsonUpdaterType from '../../logic/PackageJsonUpdater';
+
+export class UpgradeInteractiveAction extends BaseRushAction {
+  private _allFlag!: CommandLineFlagParameter;
+  private _makeConsistentFlag!: CommandLineFlagParameter;
+  private _installFlag!: CommandLineFlagParameter;
+  private _interactiveUpgrader!: InteractiveUpgrader;
```
iclanton, line 11:
```suggestion
  private _allFlag: CommandLineFlagParameter;
  private _makeConsistentFlag: CommandLineFlagParameter;
  private _installFlag: CommandLineFlagParameter;
  private _interactiveUpgrader: InteractiveUpgrader;
```

common/config/rush/browser-approved-packages.json
```diff
@@ -18,6 +18,10 @@
       "name": "react-dom",
       "allowedCategories": [ "libraries", "tests" ]
     },
+    {
+      "name": "run-async",
```
TheLarkInn, line 22: Remove.

TheLarkInn:
If a user chooses a dependency that is specified in AllowedAltVersions, notify the user that if they want all projects to update that dependency, to remove that dependency [and version] from AllowedAltVersions.

libraries/rush-lib/src/cli/actions/npm-check-typings.d.ts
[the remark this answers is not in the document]
TheLarkInn: No other package that I could find across rushstack does this today, I think this would be a great optimization. I moved the typings to root of project for better visibility (which other projects do across rushstack).

TheLarkInn:
Build failures unblocked by #3711

octogonz (approved):

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -151,6 +181,158 @@ export class PackageJsonUpdater {
...
+
+    if (this._rushConfiguration.ensureConsistentVersions || updateOtherPackages) {
+      // TODO: Reenable to support mismatch checks for both devDeps and regularDeps
+      // we need to do a mismatch check
+      const mismatchFinder: VersionMismatchFinder = VersionMismatchFinder.getMismatches(
+        this._rushConfiguration,
+        {
+          variant: variant
+        }
+      );
```
dmichon-msft, line 285: Move this into the `if` below.
At which point, there's really only one `if`, since the outer one is redundant.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
```diff
@@ -151,6 +181,158 @@ export class PackageJsonUpdater {
...
+
+      if (updateOtherPackages) {
+        for (const update of this._getUpdates(mismatchFinder, allDependenciesToUpdate)) {
+          this.updateProject(update);
+          allPackageUpdates.set(update.project.filePath, update.project);
+        }
+      }
+    }
+
+    for (const project of allPackageUpdates.values()) {
```
dmichon-msft, line 295: Could also do
```
for (const [filePath, project] of allPackageUpdates)
```
and skip the `project.filePath` lookup.

dmichon-msft (approved):
Few minor notes, otherwise looks good

common/changes/@microsoft/rush/feature-upgrade-interactive_2022-10-11-22-51.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Introduce rush upgrade-interactive",
```
iclanton, line 5:
```suggestion
      "comment": "Introduce a `rush upgrade-interactive` action that provides an interactive way to upgrade outdated packages.",
```

libraries/rush-lib/package.json
```diff
@@ -31,6 +31,7 @@
     "@rushstack/terminal": "workspace:*",
     "@rushstack/ts-command-line": "workspace:*",
     "@types/node-fetch": "1.6.9",
+    "@types/text-table": "~0.2.2",
```
iclanton, line 34: Move to `devDependencies`?

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
```diff
@@ -0,0 +1,75 @@
...
+        ' If you are using ensureConsistentVersions policy, upgrade-interactive will update all packages which use the' +
+        ' dependencies that you are upgrading and match their SemVer range if provided. If ensureConsistentVersions' +
+        ' is not enabled, upgrade-interactive will only update the dependency in the package you specify.' +
+        ' This can be overriden by using the --make-consistent flag.'
+    ];
+    super({
+      actionName: 'upgrade-interactive',
+      summary: 'Provides interactive prompt for upgrading package dependencies per project',
+      safeForSimultaneousRushProcesses: false,
+      documentation: documentation.join(os.EOL),
```
iclanton, line 28:
```suggestion
      documentation: documentation.join(''),
```

`ts-command-line` will handle word wrapping.

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
```diff
@@ -0,0 +1,75 @@
...
+    this._skipUpdateFlag = this.defineFlagParameter({
+      parameterLongName: '--skip-update',
+      parameterShortName: '-s',
+      description:
+        'If specified, the "rush update" command will not be run after updating the package.json files.'
+    });
+  }
+
+  // TODO: Remove this after rebasing main as it is not required anymore
+  protected onDefineParameters(): void {}
```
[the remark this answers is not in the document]
iclanton: Yeah it is.

iclanton, line 47:
```suggestion
```

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  public async runAsync(): Promise<void> {
+    const packageJsonUpdater: typeof PackageJsonUpdaterType = await import('../../logic/PackageJsonUpdater');
+    const interactiveUpgrader: typeof InteractiveUpgraderType = await import(
+      '../../logic/InteractiveUpgrader'
+    );
```
iclanton, line 53:
```suggestion
    const [packageJsonUpdater, interactiveUpgrader] = await Promise.all([
      import('../../logic/PackageJsonUpdater'),
      import('../../logic/InteractiveUpgrader')
    ]);
```

Then you can remove the `import types` at the top of the file.

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
```diff
@@ -0,0 +1,75 @@
...
+    const updater: PackageJsonUpdaterType.PackageJsonUpdater = new packageJsonUpdater.PackageJsonUpdater(
+      this.rushConfiguration,
+      this.rushGlobalFolder
+    );
+    const upgrader: InteractiveUpgraderType.InteractiveUpgrader = new interactiveUpgrader.InteractiveUpgrader(
+      this.rushConfiguration
+    );
+
```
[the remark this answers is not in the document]
iclanton: Yeah maybe rename them to `packageJsonUpdater` and `ineractivveUpgrader`.

libraries/rush-lib/package.json
```diff
@@ -54,7 +55,9 @@
     "strict-uri-encode": "~2.0.0",
     "tapable": "2.2.1",
     "tar": "~6.1.11",
-    "true-case-path": "~2.2.1"
+    "true-case-path": "~2.2.1",
+    "npm-check": "~5.9.2",
+    "text-table": "~0.2.0"
```
iclanton, line 60: We already have "print in table" code [here](https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/src/cli/actions/ListAction.ts#L181-L258). Can you refactor that and reuse it instead of pulling in a new dependency.

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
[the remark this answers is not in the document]
TheLarkInn: How does this sound: 

```ts 
    const [{ PackageJsonUpdater }, { InteractiveUpgrader }] = await Promise.all([
      import('../../logic/PackageJsonUpdater'),
      import('../../logic/InteractiveUpgrader')
    ]);

    const packageJsonUpdater: PackageJsonUpdaterType.PackageJsonUpdater = new PackageJsonUpdater(
      this.rushConfiguration,
      this.rushGlobalFolder
    );
    const interactiveUpgrader: InteractiveUpgraderType.InteractiveUpgrader = new InteractiveUpgrader(
      this.rushConfiguration
    );
```

libraries/rush-lib/package.json
TheLarkInn: For MVP I'd like to keep text-table, however I would definitely refactor our print in table code, or extend it to be colorized/useful in this scenario.

libraries/rush-lib/package.json
```diff
@@ -54,7 +55,9 @@
     "strict-uri-encode": "~2.0.0",
     "tapable": "2.2.1",
     "tar": "~6.1.11",
-    "true-case-path": "~2.2.1"
+    "true-case-path": "~2.2.1",
+    "npm-check": "~5.9.2",
```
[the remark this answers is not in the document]
TheLarkInn: We are using about 2/3rd of the code. I've been in discussion with the owner of the package and he is open to a archetectural changes not only to support upgrade-interactive at an --all --only multipackage scope, but also to lessening the footprint.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
TheLarkInn: We need to check both cases from the outer `if` so just dropping the inner `if` should be fine right?

libraries/rush-lib/src/logic/InteractiveUpgrader.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  public async upgrade(): Promise<IUpgradeInteractiveDeps> {
+    const { selectProject: rushProject } = await this._getUserSelectedProjectForUpgrade();
+
+    const dependenciesState: NpmCheck.INpmCheckPackage[] = await this._getPackageDependenciesStatus(
+      rushProject
+    );
+
+    // cast inquirer.Answer type for more type strictness
```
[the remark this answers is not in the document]
TheLarkInn: Removed

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
TheLarkInn: Will you recheck change in latest commit before I resolve to verify.

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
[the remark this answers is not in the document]
TheLarkInn: Please recheck latest changes to verify if this looks readable and I'm not making glaring mistakes from this method of grabbing the names without too much duplication.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
dmichon-msft: The outer if is the one that can be dropped. Since the `getMismatches()` call doesn't have side effects, it can be moved into the inner `if`.

dmichon-msft (approved):

libraries/rush-lib/src/cli/actions/UpgradeInteractiveAction.ts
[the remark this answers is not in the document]
iclanton: LGTM

libraries/rush-lib/package.json
iclanton: There's also `cli-table`. Can that be reused?

TheLarkInn: Alright I've switched to cli-table. It's pretty identical results besides some tweaking with column widths.

libraries/rush-lib/src/logic/PackageJsonUpdater.ts
TheLarkInn: Done and done. Removed outer `if`, moved getMismatches to inner `if` at both callsites.

PR #3694, opened 2022-10-11
[heft-sass-plugin] Support non-module SCSS files via "nonModuleFileExtensions"

dmichon-msft:
## Summary
Adds a new property to the `heft-sass-plugin` configuration called `nonModuleFileExtensions`. Input files that end in these extensions (and do not match a more specific extension from `fileExtensions`, which are the file extensions for SCSS modules) will be treated as non-module SCSS files: the CSS modules transform will be skipped and the typings will not contain any names.

## Details
Also fixed a bug encountered in `Sort.sortMapKeys` wherein the comparer was run between the first element of the collection and `undefined`.

## How it was tested
Added test case to `heft-sass-test` for a file with `.global.scss` as the file extension. Validated that the typings for it don't have any names.

### DIFF
--- a/heft-plugins/heft-sass-plugin/src/SassProcessor.ts
+++ b/heft-plugins/heft-sass-plugin/src/SassProcessor.ts
@@ -5,7 +5,7 @@ import * as path from 'path';
 import { render, Result, SassError } from 'node-sass';
 import * as postcss from 'postcss';
 import cssModules from 'postcss-modules';
-import { FileSystem, LegacyAdapters } from '@rushstack/node-core-library';
+import { FileSystem, LegacyAdapters, Sort } from '@rushstack/node-core-library';
 import { IStringValueTypings, StringValuesTypingsGenerator } from '@rushstack/typings-generator';
 
 /**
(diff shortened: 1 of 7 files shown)


heft-plugins/heft-sass-plugin/src/SassProcessor.ts
```diff
@@ -213,3 +229,69 @@ export class SassProcessor extends StringValuesTypingsGenerator {
...
+
+  const isFileModule: (relativePath: string) => boolean = (relativePath: string) => {
+    // Naive comparison algorithm. O(E), where E is the number of extensions
+    // If performance becomes an issue, switch to using LookupByPath with a reverse iteration order using `.` as the delimiter
+    for (const [extension, isExtensionModule] of extensionClassifier) {
+      if (relativePath.endsWith(extension)) {
+        return isExtensionModule;
+      }
+    }
+    throw new Error(`Could not classify ${relativePath} as a SCSS module / not an SCSS module`);
```
D4N14L, line 290: If no matches are found, this will throw. Is that expected? Will it only ever be provided SCSS files?

D4N14L (approved):

heft-plugins/heft-sass-plugin/src/SassProcessor.ts
dmichon-msft: Strictly speaking this error should be unreachable code, since the file extension lookup table here has the same set of extensions that are used to select the inputs to the typings generator. Yes, the function will only ever be provides SCSS (or CSS) files.

PR #3717, opened 2022-10-21
[heft-sass-plugin] Replace sass implementation to dart sass

csobj:
## Summary
This PR updates sass transpiler to dart-sass.

## Details
- Update SassProcessor to use compileStringAsync of dart sass to transpile sass files.
- Update heft-sass-tests to test new `@use` syntax of sass.

## How it was tested
- Checked that heft-sass-test builds and renders normally.
- Changed `@import` to `@use` in styles.sass file and checked that the sass code is converted to css well.

### DIFF
--- a/build-tests/heft-sass-test/src/styles.sass
+++ b/build-tests/heft-sass-test/src/styles.sass
@@ -4,10 +4,10 @@
  */
 
 // Testing Sass imports
-@use 'stylesImport'
+@import 'stylesImport'
 
 // Testing node_modules imports
-@use '~buttono/buttono'
+@import '~buttono/buttono'
 
 // Testing root styles
 html, body
@@ -24,5 +24,5 @@ html, body
   width: 400px
 
 .exampleButton
-  @include buttono.buttono-block()
-  @include buttono.buttono-style-modifier($background-color: mediumorchid)
+  @include buttono-block()
+  @include buttono-style-modifier($background-color: mediumorchid)
--- a/build-tests/heft-sass-test/src/styles.sass
+++ b/build-tests/heft-sass-test/src/styles.sass
@@ -4,10 +4,10 @@
  */
 
 // Testing Sass imports
-@import 'stylesImport'
+@use 'stylesImport'
 
 // Testing node_modules imports
-@import '~buttono/buttono'
+@use '~buttono/buttono'
 
 // Testing root styles
 html, body
@@ -24,5 +24,5 @@ html, body
   width: 400px
 
 .exampleButton
-  @include buttono-block()
-  @include buttono-style-modifier($background-color: mediumorchid)
+  @include buttono.buttono-block()
+  @include buttono.buttono-style-modifier($background-color: mediumorchid)
--- a/heft-plugins/heft-sass-plugin/src/SassProcessor.ts
+++ b/heft-plugins/heft-sass-plugin/src/SassProcessor.ts
@@ -1,11 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
+/// <reference lib="dom" />
 
 import * as path from 'path';
-import { render, Result, SassError } from 'node-sass';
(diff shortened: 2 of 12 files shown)


csobj:
@microsoft-github-policy-service agree [company="default"]

csobj:

heft-plugins/heft-sass-plugin/src/SassProcessor.ts
```diff
@@ -1,11 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
+/// <reference lib="dom" />
```
iclanton, line 3: What needs this?

csobj: I added this because type declaration of dart-sass refers to the dom URL. At first, I considered adding it directly to tsconfig, but I think it's worse to affect the whole because the project itself is for the node environment. welcome your suggestion if you have any other comments or better ideas

build-tests/heft-sass-test/src/styles.sass
```diff
@@ -4,10 +4,10 @@
  */
 
 // Testing Sass imports
-@import 'stylesImport'
+@use 'stylesImport'
```
TheLarkInn, line 7: I'm assuming based on lower lines of code that both `@use` and `@import` are supported? I think it would be prudent for us to have a test supporting both formats so that we cover both scenarios (especially since the change file depicts this change as SemVer minor)

build-tests/heft-sass-test/src/styles.sass
```diff
@@ -24,5 +24,5 @@ html, body
   width: 400px
 
 .exampleButton
-  @include buttono-block()
-  @include buttono-style-modifier($background-color: mediumorchid)
+  @include buttono.buttono-block()
```
TheLarkInn, line 27: Also in a similar vein, the separate test which supports the @use syntax we can include this variant as well.

TheLarkInn (changes requested):

build-tests/heft-sass-test/src/styles.sass
csobj: Thank you to review and I agree with this. Would you please review the last 2 commits?

heft-plugins/heft-sass-plugin/src/SassProcessor.ts
```diff
@@ -194,39 +196,46 @@ export class SassProcessor extends StringValuesTypingsGenerator {
...
-    let result: Result;
+    let result: CompileResult;
+    const nodeModulesUrl: URL = pathToFileURL(`${buildFolder}/node_modules/`);
     try {
-      result = await LegacyAdapters.convertCallbackToPromise(render, {
-        data: fileContents,
-        file: filePath,
-        importer: (url: string) => ({ file: this._patchSassUrl(url) }),
-        includePaths: importIncludePaths
+      result = compileString(fileContents, {
```
iclanton, line 202: Does this have to be synchronous? I assume this is calling some native code and waiting for a result, so it'd be nice to not have to block the JS thread.

iclanton: Nevermind. Looks like that API is slower.

csobj: Sass API documentation says that the synchronous API is almost twice as fast as the asynchronous one.
(ref: https://sass-lang.com/documentation/js-api/modules#compileStringAsync)

To be honest, I haven't compared this directly, but the compilation job is a CPU intensive operation in general, I think the benefit of completing the compilation faster is greater than the benefit of calling a kernel function asynchronously I think.

Nevertheless, if you still consider this to work asynchronously, I'd be happy to fix it.

iclanton (approved):
LGTM. @TheLarkInn, look good to you?

TheLarkInn (approved):
Thanks for adding these tests! LGTM. ❤❤

PR #3726, opened 2022-10-30
[rush] Turn on build cache for heft-jest-plugin

elliot-nelson:
## Summary

Enable build caching for heft-jest-plugin (via rush-project).

## Details

This project wasn't using the rig, because it would cause a circular dependency, but we can still use the same build cache settings.

## How it was tested

 - Tested locally for build & test.

### DIFF
--- a/common/changes/@rushstack/heft-jest-plugin/enelson-heft-jest-cache_2022-10-30-12-13.json
+++ b/common/changes/@rushstack/heft-jest-plugin/enelson-heft-jest-cache_2022-10-30-12-13.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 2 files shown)


common/changes/@rushstack/heft-jest-plugin/enelson-heft-jest-cache_2022-10-30-12-13.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-jest-plugin",
+      "comment": "Enable build caching for heft-jest-plugin",
```
iclanton, line 5:
```suggestion
      "comment": "",
```

iclanton (approved):

PR #3740, opened 2022-11-10
[rush] Rush setup can use "_authToken" instead of "_password"

elliot-nelson:
## Summary

Add a new option, `credentialType`, to the artifactory setup schema for the `rush setup` command.

## Details

Today, the default behavior of `rush setup` is to ask for the jFrog user's API token, and then trade it in for an NPM password. This is a good default, as the API token can have all kinds of unrelated scopes and privileges on many different jFrog registries, whereas the password that we get after "trading it in" is scoped just to the NPM registry we are talking with.

However, in some cases, a monorepo maintainer might prefer that all users use their _API tokens_ instead of their NPM password (especially if the developers are also using that token for access to other jFrog registries as well).

## How it was tested

 - Tested new option locally with both the `password`, `authToken`, and missing (not defined) options.

### DIFF
--- a/common/changes/@microsoft/rush/enelson-credentialtype_2022-11-10-03-49.json
+++ b/common/changes/@microsoft/rush/enelson-credentialtype_2022-11-10-03-49.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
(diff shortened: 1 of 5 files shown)


octogonz (approved):
LGTM

PR #3741, opened 2022-11-10
Will/lockfile explorer step3

william2958:
## Summary
This MR adds the following changes:

**Quality of life improvements**
- Bookmarks + filter text is now persisted across page refreshes
- Better layout css (hopefully)
- Added jump forward functionality (works after clicking "Back")

**Fixes**
- Determinants + Transitive Referencers are now calculated correctly and calculations are slimmed down by skipping packages that don't have peer dependencies.
- Fixed a bug where the websocket would not connect and thus Hot-Reloading did not work.

**UI Changes**
- Influencers are shown as a clickable list
- You can click to go view a listed Determinant or Transitive Referrer
- Added dependency info panel (Above Influencers panel)

### DIFF
--- a/apps/lockfile-explorer-web/src/containers/LockfileEntryDetailsView/index.tsx
+++ b/apps/lockfile-explorer-web/src/containers/LockfileEntryDetailsView/index.tsx
@@ -12,14 +12,14 @@ import { LockfileEntry } from '../../parsing/LockfileEntry';
 import { logDiagnosticInfo } from '../../helpers/logDiagnosticInfo';
 import { displaySpecChanges } from '../../helpers/displaySpecChanges';
 
-enum InfluencerTypes {
+enum DependencyType {
   Determinant,
   TransitiveReferrer
 }
 
 interface IInfluencerType {
   entry: LockfileEntry;
-  type: InfluencerTypes;
+  type: DependencyType;
 }
 
 export const LockfileEntryDetailsView = (): JSX.Element | ReactNull => {
@@ -103,13 +103,13 @@ export const LockfileEntryDetailsView = (): JSX.Element | ReactNull => {
         for (const determinant of determinants.values()) {
           influencers.push({
             entry: determinant,
-            type: InfluencerTypes.Determinant
+            type: DependencyType.Determinant
           });
         }
         for (const referrer of transitiveReferrers.values()) {
           influencers.push({
             entry: referrer,
-            type: InfluencerTypes.TransitiveReferrer
+            type: DependencyType.TransitiveReferrer
           });
         }
         setInfluencers(influencers);
@@ -185,7 +185,7 @@ export const LockfileEntryDetailsView = (): JSX.Element | ReactNull => {
       <div className={`${appStyles.ContainerCard} ${styles.InfluencerList}`}>
         <h5>Determinants:</h5>
         {influencers
-          .filter((inf) => inf.type === InfluencerTypes.Determinant)
+          .filter((inf) => inf.type === DependencyType.Determinant)
           .map(({ entry }) => (
             <a
               className={styles.InfluencerEntry}
@@ -197,7 +197,7 @@ export const LockfileEntryDetailsView = (): JSX.Element | ReactNull => {
           ))}
         <h5>Transitive Referencers:</h5>
         {influencers
-          .filter((inf) => inf.type === InfluencerTypes.TransitiveReferrer)
+          .filter((inf) => inf.type === DependencyType.TransitiveReferrer)
           .map(({ entry }) => (
             <a
               className={styles.InfluencerEntry}
--- a/apps/lockfile-explorer-web/src/containers/LockfileEntryDetailsView/index.tsx
+++ b/apps/lockfile-explorer-web/src/containers/LockfileEntryDetailsView/index.tsx
@@ -10,6 +10,7 @@ import { pushToStack, selectCurrentEntry } from '../../store/slices/entrySlice';
 import { ReactNull } from '../../types/ReactNull';
 import { LockfileEntry } from '../../parsing/LockfileEntry';
 import { logDiagnosticInfo } from '../../helpers/logDiagnosticInfo';
+import { displaySpecChanges } from '../../helpers/displaySpecChanges';
 
 enum InfluencerTypes {
   Determinant,
@@ -23,6 +24,7 @@ interface IInfluencerType {
 
 export const LockfileEntryDetailsView = (): JSX.Element | ReactNull => {
   const selectedEntry = useAppSelector(selectCurrentEntry);
+  const specChanges = useAppSelector((state) => state.workspace.specChanges);
   const dispatch = useAppDispatch();
 
   const [inspectDependency, setInspectDependency] = useState<LockfileDependency | null>(null);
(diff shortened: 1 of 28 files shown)


apps/lockfile-explorer-web/src/containers/LockfileEntryDetailsView/index.tsx
```diff
@@ -36,14 +38,22 @@ export const LockfileEntryDetailsView = (): JSX.Element | ReactNull => {
...
         } else {
           logDiagnosticInfo('No resolved entry for dependency:', dependencyToTrace);
         }
       } else if (selectedEntry) {
+        console.log('dependency to trace: ', dependencyToTrace);
         setInspectDependency(dependencyToTrace);
+
+        // Check if we need to calculate influencers.
+        // If the current dependencyToTrace is a peer dependency then we do
+        if (dependencyToTrace.dependencyType !== IDependencyType.PEER_DEPENDENCY) {
```
octogonz, line 53: Since `IDependencyType` is an enum, it should be called `DependencyType`.  The "I" prefix is only used for `interface` types

apps/lockfile-explorer-web/src/containers/LogoPanel/index.tsx
```diff
@@ -15,14 +15,15 @@ export const LogoPanel = (): JSX.Element => {
...
-        <div className={styles.Title1}>
-          <img className={styles.Image} src={require('./lockfile-explorer-title-1.svg')} />
+        <div>
+          <div className={styles.Title1}>
+            <img className={styles.Image} src={require('./lockfile-explorer-title-1.svg')} />
+          </div>
+          <div className={styles.Title2}>
+            <img className={styles.Image} src={require('./lockfile-explorer-title-2.svg')} />
+          </div>
+          <div className={styles.Detail}>{appPackageVersion}&nbsp;rushstack.io</div>
```
octogonz, line 25: The old `LogoPanel` looked like this:
image

In this PR, the layout is changed to be:
image

The `0.0.3 rushstack.io` seems a bit confusing, since `0.0.3` is the version of Lockfile Explorer, not Rush Stack.

Also in the original design, I had envisioned that clicking on the words/icon would take you to the NPM page (or https://lfx.rushstack.io ?), whereas clicking on `rushstack.io` would take you to the other website https://rushstack.io

What do you think?  Another consideration is that the Projects/Packages box is the centerpiece of the UI, so we want to provide as much vertical scroll area as possible.

common/changes/@rushstack/lockfile-explorer/will-lockfile-explorer-step3_2022-11-11-20-39.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/lockfile-explorer",
+      "comment": "Fixed issues related to influencer resolution in lockfile explorer; Cleaned up UI & improved UX.",
+      "type": "patch"
```
octogonz, line 6: Let's make this a `"minor"` bump since you added new features

apps/lockfile-explorer-web/src/containers/LockfileViewer/index.tsx
```diff
@@ -155,11 +165,16 @@ export const LockfileViewer = (): JSX.Element | ReactNull => {
...
-          <h5>filter:</h5>
+          <h5>Filter:</h5>
           <input type="text" value={filter} onChange={updateFilter} />
+          <button disabled={entryStack.length > 1} onClick={pop}>
+            Back
+          </button>
+          <button disabled={entryForwardStack.length > 0} onClick={forward}>
+            Forward
+          </button>
         </div>
```
octogonz, line 178: image

Can we make it so that these checkboxes can be toggled by clicking the label ("Must have...")?  Currently it only works if you click inside the square glyph.

apps/lockfile-explorer-web/src/containers/LockfileViewer/index.tsx
```diff
@@ -71,22 +73,30 @@ const multipleVersions = (entries: LockfileEntry[]): boolean => {
...
+  const dispatch = useAppDispatch();
   const [filter, setFilter] = useState('');
   const entries = useAppSelector(selectFilteredEntries);
   const activeFilters = useAppSelector((state) => state.entry.filters);
-  const updateFilter = useCallback((e) => setFilter(e.target.value), []);
+  const updateFilter = useCallback((e) => {
+    setFilter(e.target.value);
+    saveFilterToLocalStorage(e.target.value);
+  }, []);
+  const pop = useCallback(() => {
```
octogonz, line 84: The `Back`/`Forward` buttons don't seem to be working correctly for me... let me know if you need more specific repro steps.

apps/lockfile-explorer-web/src/helpers/displaySpecChanges.ts
```diff
@@ -0,0 +1,14 @@
+import { ISpecChange } from '../parsing/compareSpec';
+
+export const displaySpecChanges = (specChanges: Map<string, ISpecChange>, dep: string): string => {
```
octogonz, line 3: In the [Adobe XD mockup](https://xd.adobe.com/view/13b05d87-8967-4485-b11a-fdb0e4fdf93d-46e3/?fullscreen&hints=off), I used a fixed-width font to make columns:

image

I like your approach of using colors:
image
...but maybe there is some way to line up the versions so they are easier to read?

apps/lockfile-explorer-web/src/helpers/displaySpecChanges.ts
```diff
@@ -0,0 +1,14 @@
...
+  switch (specChanges.get(dep)?.type) {
+    case 'add':
```
octogonz, line 5: Also, it's possible for the diff to be scrolled out of view, making it not obvious that there is any diff for this entry:
image

Maybe we could provide a visual indicator, for example show `* package spec` instead of `package spec` if anything was modified?  Maybe we could even show a `*` in the listview as well:

image

I think these transformations are fairly important information when studying lockfiles, because they are the customizations that we have to fiddle with to fix problems.

apps/lockfile-explorer-web/src/containers/SelectedEntryPreview/styles.scss
```diff
@@ -1,11 +1,12 @@
 @import 'office-ui-fabric-core/dist/sass/References';
```
octogonz, line 1: In my web browser, the `Add Bookmark` button gets scrolled off the bottom of the pane.
image

apps/lockfile-explorer-web/src/containers/LogoPanel/index.tsx
william2958: That makes sense, I've moved the version next to the lockfile explorer text logo to make it a bit more clear, also I moved the logo to above the bookmarks sidebar as that one is less important than the left panel. The appropriate links have also been added. I moved it away from having it's own column because i felt that it was taking up a lot of horizontal real-estate for a pretty small logo, this way I think gives the app overall more room:

image

apps/lockfile-explorer-web/src/containers/LockfileViewer/index.tsx
william2958: Changed

apps/lockfile-explorer-web/src/containers/LockfileViewer/index.tsx
william2958: Fixed!

apps/lockfile-explorer-web/src/helpers/displaySpecChanges.ts
william2958: fixed!

apps/lockfile-explorer-web/src/helpers/displaySpecChanges.ts
william2958: Added an asterisk to the "package spec *" header tab to indicate that there is a diff for the entry, I don't think we should show an asterisk in the list view because we need to load the package's package.json to be able to calculate if there is a spec change, (currently we just load the selected entry's package.json & calculate the spec changes then)

apps/lockfile-explorer-web/src/containers/LockfileViewer/index.tsx
```diff
@@ -155,11 +165,16 @@ export const LockfileViewer = (): JSX.Element | ReactNull => {
           }
         ]}
       />
-      <div className={`${appStyles.ContainerCard} ${styles.ViewWrapper}`}>
+      <div className={`${styles.ViewContents} ${appStyles.ContainerCard}`}>
         <div className={styles.LockfileFilterBar}>
-          {entryStack.length > 1 ? <button onClick={pop}>back</button> : null}
-          <h5>filter:</h5>
+          <h5>Filter:</h5>
```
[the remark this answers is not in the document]
william2958: Yeah I think that makes sense for the Filter to be applicable to each tab individually, so I changed it so that it saves and uses different filter text depending on which tab your in, so they can search for things independently.

As for the checkboxes, I feel that it would make sense that they affect both tabs as the user would probably want to look for the same "hits" across both tabs, but to make it a bit more clear that it's a separate entity I've added a border to separate the list and the filter checkbox area, let me know what you think:

image

apps/lockfile-explorer-web/src/containers/SelectedEntryPreview/styles.scss
william2958: Moved the button to the top right of the panel

octogonz (approved):

PR #3751, opened 2022-11-15
Pull in changes from user/danade/MultiPhaseHeft.

iclanton:


heft-plugins/heft-webpack5-plugin/src/shared.ts
```diff
@@ -84,6 +84,16 @@ export interface IWebpackPluginAccessorHooks {
...
 }
 
+/**
+ * @public
+ */
+export interface IWebpackPluginAccessorParameters {
+  /**
+   * Whether or not serve mode was enabled by passing the `--serve` flag.
+   */
+  serve: boolean;
```
iclanton, line 94:
```suggestion
  isServeMode: boolean;
```

heft-plugins/heft-webpack4-plugin/src/shared.ts
```diff
@@ -95,6 +95,16 @@ export interface IWebpackPluginAccessorHooks {
...
 }
 
+/**
+ * @public
+ */
+export interface IWebpackPluginAccessorParameters {
+  /**
+   * Whether or not serve mode was enabled by passing the `--serve` flag.
+   */
+  serve: boolean;
```
iclanton, line 105:
```suggestion
  isServeMode: boolean;
```

PR #3754, opened 2022-11-16
[rush-lib] Fix phase command to respect upstream

### LINKED ISSUE #3753: [rush] phased command not work correct with upstream
## Summary

I want to have a command for run upstream project build and self project lint without self project build.
But when I try to set it, I only get a command for for run upstream project build and self project lint and self project build.

## Repro steps

Clone this project: https://github.com/yoyo930021/rush-upstream-phase-bug
```bash
rush install
rush lint
```

  **Expected result:** 
  only run 3 operations
  project (lint)
  upstream (build)
  upstream (lint)

  **Actual result:** 
  run 4 operations
  project (lint)
  upstream (build)
  upstream (lint)
  project (build)

## Details

In this line, project will add any phase with selection project.
https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/src/logic/operations/PhasedOperationPlugin.ts#L40

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.83.3 |
| `rushVersion` from rush.json? | 5.83.3 |
| Operating system? | Mac |
| Node.js version (`node -v`)? | 16.16.0 |

yoyo930021:
## Summary
Fixes #3753 

## Details
I will add originalPhase property to know original phase.
The property name may have better one.

## How it was tested
Use https://github.com/microsoft/rushstack/issues/3753 repro steps

### DIFF
--- a/common/changes/@microsoft/rush/fix-upstream-phase_2022-11-16-08-22.json
+++ b/common/changes/@microsoft/rush/fix-upstream-phase_2022-11-16-08-22.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Fix phase command to respect upstream",
+      "comment": "Modify the scheduling behavior of phased commands to schedule only the expressly enumerated phases in all selected projects, adding additional phases only where needed to satisfy dependencies.",
       "type": "none"
     }
   ],
--- a/common/changes/@microsoft/rush/fix-upstream-phase_2022-11-16-08-22.json
+++ b/common/changes/@microsoft/rush/fix-upstream-phase_2022-11-16-08-22.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix phase command to respect upstream",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/libraries/rush-lib/src/logic/operations/PhasedOperationPlugin.ts
+++ b/libraries/rush-lib/src/logic/operations/PhasedOperationPlugin.ts
@@ -29,13 +29,18 @@ function createOperations(
   existingOperations: Set<Operation>,
   context: ICreateOperationsContext
 ): Set<Operation> {
-  const { projectsInUnknownState: changedProjects, phaseSelection, projectSelection } = context;
+  const {
+    projectsInUnknownState: changedProjects,
+    phaseOriginal,
+    phaseSelection,
+    projectSelection
+  } = context;
   const operationsWithWork: Set<Operation> = new Set();
 
   const operations: Map<string, Operation> = new Map();
 
   // Create tasks for selected phases and projects
-  for (const phase of phaseSelection) {
+  for (const phase of phaseOriginal) {
     for (const project of projectSelection) {
       getOrCreateOperation(phase, project);
     }
--- a/common/reviews/api/rush-lib.api.md
+++ b/common/reviews/api/rush-lib.api.md
@@ -263,6 +263,7 @@ export interface ICreateOperationsContext {
     readonly isIncrementalBuildAllowed: boolean;
     readonly isInitial: boolean;
     readonly isWatch: boolean;
+    readonly phaseOriginal: ReadonlySet<IPhase>;
     readonly phaseSelection: ReadonlySet<IPhase>;
     readonly projectChangeAnalyzer: ProjectChangeAnalyzer;
     readonly projectSelection: ReadonlySet<RushConfigurationProject>;
--- a/libraries/rush-lib/src/api/CommandLineConfiguration.ts
+++ b/libraries/rush-lib/src/api/CommandLineConfiguration.ts
@@ -96,6 +96,10 @@ export interface IPhasedCommandConfig extends IPhasedCommandWithoutPhasesJson, I
   isSynthetic: boolean;
   disableBuildCache?: boolean;
 
+  originalPhases: Set<IPhase>;
+  /**
+   * Include upstream and self phases.
+   */
   phases: Set<IPhase>;
 
   /**
@@ -297,6 +301,7 @@ export class CommandLineConfiguration {
 
     const commandsJson: ICommandLineJson['commands'] = commandLineJson?.commands;
     let buildCommandPhases: IPhasedCommandConfig['phases'] | undefined;
+    let buildCommandOriginalPhases: IPhasedCommandConfig['phases'] | undefined;
     if (commandsJson) {
       for (const command of commandsJson) {
         if (this.commands.has(command.name)) {
@@ -309,6 +314,7 @@ export class CommandLineConfiguration {
         let normalizedCommand: Command;
         switch (command.commandKind) {
           case RushConstants.phasedCommandKind: {
+            const originalPhases: Set<IPhase> = new Set();
             const commandPhases: Set<IPhase> = new Set();
             const watchPhases: Set<IPhase> = new Set();
 
@@ -316,6 +322,7 @@ export class CommandLineConfiguration {
               ...command,
               isSynthetic: false,
               associatedParameters: new Set<IParameterJson>(),
+              originalPhases,
               phases: commandPhases,
               watchPhases,
               alwaysWatch: false,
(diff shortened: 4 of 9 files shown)


yoyo930021:

dmichon-msft:
Fundamentally the ask here is to delete this:
https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/src/api/CommandLineConfiguration.ts#L326-L338
and this:
https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/src/logic/operations/PhasedOperationPlugin.ts#L75-L81

With those two changes, the set of phases listed in a command would be the target phases, and then you would specify on the command line the target projects, and Rush would then run the specified phases in the specified projects, plus all supporting work necessary to make that happen.

I'm not opposed to that change (though it is a breaking design change that would need to be documented), but the change in the current iteration of the PR makes the specifications rather confusing.

dmichon-msft:
Rethinking my previous comment, my main concern here is how this interacts with `--only`, `--impacted-by`, and `--watch`. There's a lot of complexity that comes up when projects or phases get skipped by the scheduler that needs to be accounted for when redesigning how the command operates.

dmichon-msft:
I've spent a while trying to figure out exactly what my main problem with this is, and I think it comes down to what the `--only` command ends up doing.

In your example in https://github.com/yoyo930021/rush-upstream-phase-bug/blob/main/common/config/rush/command-line.json:
1. `rush lint --only project` runs:
- `project (lint)`

2. `rush lint --only upstream` runs:
- `upstream (lint)`

3. But `rush lint --only project --only upstream` runs:
- `upstream (build)`
- `upstream (lint)`
- `project (lint)`

Including both `project` and `upstream` in the project selection introduces execution of a build phase in a project that was not included if running either project in isolation, which feels unexpected. At least with the current model, since phase evaluation happens first, the set of operations in (3) is exactly the union of the operations in (1) and (2), so you can safely slice the graph apart via the `--only` parameter.

Ultimately the problem comes down to that the original design specification didn't include any examples of phases that contain only one dependency, which is an upstream dependency on a strictly different phase, so there's some design consideration about what developer ergonomics we want to optimize for.

yoyo930021:
I have some suggest in issue. https://github.com/microsoft/rushstack/issues/3753#issuecomment-1317973331

yoyo930021:
> I've spent a while trying to figure out exactly what my main problem with this is, and I think it comes down to what the `--only` command ends up doing.
> 
> In your example in https://github.com/yoyo930021/rush-upstream-phase-bug/blob/main/common/config/rush/command-line.json:
> 
>     1. `rush lint --only project` runs:
> 
> 
>     * `project (lint)`
> 
> 
>     2. `rush lint --only upstream` runs:
> 
> 
>     * `upstream (lint)`
> 
> 
>     3. But `rush lint --only project --only upstream` runs:
> 
> 
>     * `upstream (build)`
> 
>     * `upstream (lint)`
> 
>     * `project (lint)`
> 
> 
> Including both `project` and `upstream` in the project selection introduces execution of a build phase in a project that was not included if running either project in isolation, which feels unexpected. At least with the current model, since phase evaluation happens first, the set of operations in (3) is exactly the union of the operations in (1) and (2), so you can safely slice the graph apart via the `--only` parameter.
> 
> Ultimately the problem comes down to that the original design specification didn't include any examples of phases that contain only one dependency, which is an upstream dependency on a strictly different phase, so there's some design consideration about what developer ergonomics we want to optimize for.

I think you have a new question.

In rush 5.83.3, 
1. `rush lint --only project` runs:
- project (lint)
- project (build)
2. `rush lint --only upstream` runs:
- upstream (lint)
- upstream (build)
3. But `rush lint --only project --only upstream` runs:
- project (lint)
- upstream (build)
- upstream (lint)
- project (build)

The problem already exists.

dmichon-msft:
> In rush 5.83.3,
> 
> 1. `rush lint --only project` runs:
> 
> * project (lint)
> * project (build)
> 
> 2. `rush lint --only upstream` runs:
> 
> * upstream (lint)
> * upstream (build)
> 
> 3. But `rush lint --only project --only upstream` runs:
> 
> * project (lint)
> * upstream (build)
> * upstream (lint)
> * project (build)

The above is what I expect to happen. `rush lint --only project` followed by `rush lint --only upstream` ends up performing all work that would execute if I invoked `rush lint --only project --only upstream`.

Your suggestion of having:
`rush lint --only project` execute
- `upstream (build)`
- `project (lint)`

will obey the same semantic requirement (and in general is probably more aligned with typical usage), but then the set of CLI parameters get confusing, because then:

`rush build --only project` executes
- `upstream (build)` // unexpected, supposed to be skipped
- `project (build)`

, which is very unintuitive when that's what `rush build --to project` means, and the point of `rush build --only project` is to explicitly _skip_ `upstream (build)`, e.g. after making local changes to rebuild just the one project.

elliot-nelson:
I'll throw my two cents in...

In @dmichon-msft 's original post he mentioned that a side effect of this change would be:

```console
rush lint --only project
# => project (lint)

rush lint --only upstream
# => upstream (lint)

rush lint --only project --only upstream
# => upstream (lint)
# => upstream (build)
# => project (lint)
```

To me this makes total sense, so if it's easy enough to get this behavior, I'd support it. My reasoning is that project selection is an operation-agnostic operation -- it happens before we ever consider what operations will be run -- so this is a _strict boundary_. Running `rush lint --only project` should never run any operation on any project outside the selection.

But, when you expand the selection, you open up the possibility of additional operations as well (because now the upstream project is available to the runner).

This corner case only really exists if you define a phased command with a list of phases that does not include all of its dependent phases, which I think is a little unusual.

libraries/rush-lib/src/logic/operations/PhasedOperationPlugin.ts
```diff
@@ -29,13 +29,18 @@ function createOperations(
...
+    phaseOriginal,
+    phaseSelection,
+    projectSelection
+  } = context;
   const operationsWithWork: Set<Operation> = new Set();
 
   const operations: Map<string, Operation> = new Map();
 
   // Create tasks for selected phases and projects
-  for (const phase of phaseSelection) {
```
dmichon-msft, line 38: This breaks how watch mode currently works. If we don't want to expand the set of phases, I'm fine with simply not expanding the set of selected phases and only tracking the dependencies when building out the graph.

libraries/rush-lib/src/logic/operations/PhasedOperationPlugin.ts
```diff
@@ -29,13 +29,18 @@ function createOperations(
   existingOperations: Set<Operation>,
   context: ICreateOperationsContext
 ): Set<Operation> {
-  const { projectsInUnknownState: changedProjects, phaseSelection, projectSelection } = context;
+  const {
+    projectsInUnknownState: changedProjects,
+    phaseOriginal,
```
dmichon-msft, line 34: `watchPhases` and `initialPhases` are not required to have any overlap whatsoever, so in the watch mode branch `phaseOriginal` should probably be set to `watchPhases`.

yoyo930021: Do you mean this?
https://github.com/microsoft/rushstack/pull/3754/commits/575f8f0d076c1c6d862166f4a672f51ced6bcb7a

dmichon-msft: Yes. That would work.

yoyo930021: Thanks

dmichon-msft (approved):

common/changes/@microsoft/rush/fix-upstream-phase_2022-11-16-08-22.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix phase command to respect upstream",
```
dmichon-msft, line 5:
```suggestion
      "comment": "Modify the scheduling behavior of phased commands to schedule only the expressly enumerated phases in all selected projects, with dependencies only scheduled as needed for execution.",
```

Issue #3750, opened 2022-11-15
[rush] Performance regression in ProjectChangeAnalyzer


dmichon-msft:
## Summary
Comparing the time spent in "Analyzing repo state" between @microsoft/rush 5.80.0 (version currently installed in the rushstack repo) and @microsoft/rush 5.83.3 (current main) shows over a second of performance regression, which in the repo is 200%.

```diff
-PS D:\rushstack> rush build -o heft-web-rig
+PS D:\rushstack> node --cpu-prof ./apps/rush/lib/start-dev.js build -o heft-web-rig

-Rush Multi-Project Build Tool 5.80.0 - https://rushjs.io
+Rush Multi-Project Build Tool 5.83.3 (unmanaged) - https://rushjs.io
Node.js version is 14.20.0 (LTS)


Starting "rush build"

-Analyzing repo state... DONE (0.56 seconds)
+Analyzing repo state... DONE (1.76 seconds)

Executing a maximum of 7 simultaneous processes...

==[ @rushstack/heft-web-rig (build) ]==============================[ 1 of 1 ]==
"@rushstack/heft-web-rig (build)" did not define any work.



==[ NO OP: 1 operation ]=======================================================

These operations did not define any work:
  @rushstack/heft-web-rig (build)


-rush build (0.68 seconds)
+rush build (1.86 seconds)
```

## Repro steps
Compare the "Analyzing repo state" stage of `node ./apps/rush/lib/start-dev.js build -o heft-web-rig` vs. `rush build -o heft-web-rig`.

## Details
![image](https://user-images.githubusercontent.com/26827560/202013686-b11f3cbe-8511-44ad-a69c-66082cbd90bd.png)
CPU Profiling shows the time is all spent in the `applySubmodulesTreeState` function, despite the rushstack repository not having any submodules.

Related #3749.

Recommended fix would be to make submodules handling an opt-in setting in rush.json, or find an alternate code path to detect that submodules are present, since the check for submodule status takes longer than the entire rest of the process twice over.

Also worth investigating is switching `_getRepoDeps` in `ProjectChangeAnalyzer` to be async, and creating `getRepoStateAsync` that does as many of the Git operations as possible in parallel. The `git status` call and resulting `git hash-object` still have to be serial, but the `git ls-tree` and the `ProjectChangeAnalyzer`'s own `git hash-object` call can be parallelized with the rest to reduce the latency of operations.

## Standard questions

Please answer these questions to help us investigate your issue more quickly:

[...]

PR #3756, opened 2022-11-16
[package-deps-hash] Parallelize getRepoState

dmichon-msft:
## Summary
Prototypes parallelizing as many of the Git calls involved in calculating the state used by `ProjectChangeAnalyzer` as possible.

## Details
Parallelizes the following operations:
- `git ls-tree HEAD` for the state of the committed tree
- `git status` for the working tree state
- `git hash-object` on the project shrinkwrap files

The following operation still needs to run serially:
- `git hash-object` for changed files in the working tree, though it can use the same process invocation and delay passing the filenames on stdin

Reduces time for initializing `ProjectChangeAnalyzer` on my Windows dev box for the rushstack repository from 0.5 seconds to 0.3 seconds.
Reduces time for initializing `ProjectChangeAnalyzer` on my Windows dev box for an internal repository from 2.2 seconds to 1.5 seconds.

Related #3750

## How it was tested
Comparing invocations of `rush build -o heft-web-rig` to `node ./apps/rush/lib/start-dev.js build -o heft-web-rig` with respect to the timing of the `Analyzing repo state...` section.

### DIFF
--- a/common/reviews/api/package-deps-hash.api.md
+++ b/common/reviews/api/package-deps-hash.api.md
@@ -20,9 +20,6 @@ export function getRepoChanges(currentWorkingDirectory: string, revision?: strin
 export function getRepoRoot(currentWorkingDirectory: string, gitPath?: string): string;
 
 // @beta
-export function getRepoState(currentWorkingDirectory: string, gitPath?: string): Map<string, string>;
-
-// @beta
 export function getRepoStateAsync(rootDirectory: string, additionalRelativePathsToHash?: string[], gitPath?: string): Promise<Map<string, string>>;
 
 // @beta
--- a/common/reviews/api/package-deps-hash.api.md
+++ b/common/reviews/api/package-deps-hash.api.md
@@ -23,6 +23,9 @@ export function getRepoRoot(currentWorkingDirectory: string, gitPath?: string): 
 export function getRepoState(currentWorkingDirectory: string, gitPath?: string): Map<string, string>;
 
 // @beta
+export function getRepoStateAsync(rootDirectory: string, additionalRelativePathsToHash?: string[], gitPath?: string): Promise<Map<string, string>>;
+
+// @beta
 export interface IFileDiffStatus {
     // (undocumented)
     mode: string;
--- a/libraries/package-deps-hash/src/getRepoState.ts
+++ b/libraries/package-deps-hash/src/getRepoState.ts
@@ -2,7 +2,10 @@
 // See LICENSE in the project root for license information.
 
 import type * as child_process from 'child_process';
-import { Executable, FileSystem } from '@rushstack/node-core-library';
+import { once } from 'events';
+import { Readable } from 'stream';
+
+import { Executable, FileSystem, IExecutableSpawnOptions } from '@rushstack/node-core-library';
 
 export interface IGitVersion {
   major: number;
@@ -190,7 +193,11 @@ export function getRepoRoot(currentWorkingDirectory: string, gitPath?: string): 
       throw new Error(`git rev-parse exited with status ${result.status}: ${result.stderr}`);
     }
 
-    repoRootCache.set(currentWorkingDirectory, (cachedResult = result.stdout.trim()));
+    cachedResult = result.stdout.trim();
+
+    repoRootCache.set(currentWorkingDirectory, cachedResult);
+    // To ensure that calling getRepoRoot on the result is a no-op.
+    repoRootCache.set(cachedResult, cachedResult);
   }
 
   return cachedResult;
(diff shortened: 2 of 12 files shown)


libraries/package-deps-hash/src/getRepoState.ts
```diff
@@ -270,6 +277,149 @@ export function applyWorkingTreeState(
...
+
+  let stdout: string = '';
+  let stderr: string = '';
+
+  const proc: child_process.ChildProcess = Executable.spawn(gitPath || 'git', args, spawnOptions);
+  proc.stdout!.setEncoding('utf-8');
+  proc.stderr!.setEncoding('utf-8');
+
+  proc.stdout!.on('data', (chunk: string) => {
+    stdout += chunk.toString();
```
iclanton, line 306: Would it be more performant to append these chunks to an array and then `.join('')` then at the end?

libraries/package-deps-hash/src/getRepoState.ts
```diff
@@ -270,6 +277,149 @@ export function applyWorkingTreeState(
...
+    stderr += chunk.toString();
+  });
+
+  if (stdin) {
+    stdin.pipe(proc.stdin!);
+  }
+
+  const [status] = await once(proc, 'exit');
+  if (status !== 0) {
+    throw new Error(`git ${args[0]} exited with code ${status}: ${stderr}`);
```
iclanton, line 318: We don't care if it wrote to `stderr` and exited with a 0?

libraries/package-deps-hash/src/getRepoState.ts
```diff
@@ -270,6 +277,149 @@ export function applyWorkingTreeState(
...
+
+  const hashObjectPromise: Promise<string> = spawnGitAsync(
+    gitPath,
+    ['--no-optional-locks', 'hash-object', '--stdin-paths'],
+    rootDirectory,
+    Readable.from(getFilesToHash(), {
+      encoding: 'utf-8',
+      objectMode: false,
+      autoDestroy: true
+    })
```
iclanton, line 377: Didn't @D4N14L find that using stdin with Git was slower than an alternative during his Heft work?

libraries/package-deps-hash/src/getRepoState.ts
```diff
@@ -270,6 +277,149 @@ export function applyWorkingTreeState(
...
+  );
+
+  const [{ files, submodules }, hashObject] = await Promise.all([
+    statePromise,
+    hashObjectPromise,
+    locallyModifiedPromise
```
iclanton, line 383:
```suggestion
    hashObjectPromise
```

You're not looking at the result of `locallyModifiedPromise`, and it's already awaited in the `getFilesToHash` call.

libraries/package-deps-hash/src/getRepoState.ts
dmichon-msft: It's a memory vs. CPU trade-off. In V8 string concatenation by default just allocates a "cons string" object which is a head/tail pointer pair, so all computation of the completed string is deferred until the end.
Ultimately the performance impact of this part of the process is negligible; all our time is spent in the `git` process itself.

common/reviews/api/package-deps-hash.api.md
```diff
@@ -22,6 +22,9 @@ export function getRepoRoot(currentWorkingDirectory: string, gitPath?: string):
 // @beta
 export function getRepoState(currentWorkingDirectory: string, gitPath?: string): Map<string, string>;
```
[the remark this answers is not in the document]
dmichon-msft: Not locally, but it is part of the API, so even though it is currently `@beta` I wouldn't want to delete it without a major version bump.

libraries/package-deps-hash/src/getRepoState.ts
dmichon-msft: It was found to be slower than performing filtering in-proc, but mostly because the performance of `git check-ignore` is apparently rather abysmal.

common/reviews/api/package-deps-hash.api.md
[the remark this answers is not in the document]
iclanton: A major bump may be fine.

libraries/package-deps-hash/src/getRepoState.ts
dmichon-msft: Technically this means `locallyModifiedPromise` is a hard dependency of `hashObjectPromise`, but if it throws, it is easier to surface the error directly here.

common/reviews/api/package-deps-hash.api.md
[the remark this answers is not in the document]
dmichon-msft: Updated to a major bump and removed synchronous `getRepoState`.

libraries/package-deps-hash/src/test/getRepoDeps.test.ts
```diff
@@ -268,4 +274,37 @@ describe(getRepoState.name, () => {
...
+          'testProject/file1.txt': 'c7b2f707ac99ca522f965210a7b6b0b109863f34',
+          'testProject/file  2.txt': 'a385f754ec4fede884a4864d090064d9aeef8ccb',
+          'testProject/file蝴蝶.txt': 'ae814af81e16cb2ae8c57503c77e2cab6b5462ba',
+          'testProject/log.log': '2e65efe2a145dda7ee51d1741299f848e5bf752e',
+          [`testProject/${FileConstants.PackageJson}`]: '18a1e415e56220fa5122428a4ef8eb8874756576'
+        })
+      );
+
+      for (const [filePath, hash] of expectedFiles) {
+        expect(filteredResults.get(filePath)).toEqual(hash);
```
iclanton, line 303: Use a snapshot instead of hardcoding these?

iclanton (approved):

iclanton (approved):

PR #3757, opened 2022-11-16
[certificate-manager] Shorter validity period

dmichon-msft:
## Summary
Shortens the default validity period for certificates generated by `@rushstack/debug-certificate-manager` to ~~30~~ 365 days to avoid browsers complaining about unacceptably long validity periods. Adds validation during `ensureCertificateAsync` that the existing certificate's validity period includes now.

## Details
Validates the `notBefore` and `notAfter` fields in addition to all the other current checks.

## How it was tested
Local testing in `heft-webpack5-everything-test`.

Edit: Revised validity period from 30 days to 365 days.

### DIFF
--- a/common/changes/@rushstack/debug-certificate-manager/cert-validity-period_2022-11-16-21-58.json
+++ b/common/changes/@rushstack/debug-certificate-manager/cert-validity-period_2022-11-16-21-58.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/debug-certificate-manager",
+      "comment": "Reduce default certificate validity period to 365 days. Check certificate validity period as part of validating the existing certificate.",
+      "type": "patch"
+    }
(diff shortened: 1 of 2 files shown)


iclanton:
Going from 3 years to 30 days is a huge change. What is the specific complaint that browsers have?

dmichon-msft:
Apparently Chrome caps at 397 days (https://chromium.googlesource.com/chromium/src/+/HEAD/net/docs/certificate_lifetimes.md), so could do 1 year, but frankly most security teams would prefer if we reup the certificate more often rather than leaving it around. The flow is pretty streamlined on Windows and MacOS, but I could see it being a chore on Linux.

iclanton:
Maybe 90 days? That's the maximum token validity period on AzDO.

iclanton (approved):

D4N14L (approved):

PR #3758, opened 2022-11-17
[rush] fix performance regression from supporting git submodules

chengcyber:
## Summary

Fixes https://github.com/microsoft/rushstack/issues/3750

## Details

Refactor the logic in `getRepoState`:
1. Detecting git submodule by checking the existence of `.gitmodule` file
2. Recursively invoking `getRepoState` instead of running `git submodule foreach` command

## How it was tested

Run `rush build -o heft-web-rig` on MacOS 12 and Windows 10
- MacOS: `Analyzing repo state... DONE (0.31 seconds)`
- Windows: `Analyzing repo state... DONE (0.46 seconds)`

### DIFF
--- a/libraries/package-deps-hash/src/getRepoState.ts
+++ b/libraries/package-deps-hash/src/getRepoState.ts
@@ -2,7 +2,7 @@
 // See LICENSE in the project root for license information.
 
 import type * as child_process from 'child_process';
-import { Executable } from '@rushstack/node-core-library';
+import { Executable, FileSystem } from '@rushstack/node-core-library';
 
 export interface IGitVersion {
   major: number;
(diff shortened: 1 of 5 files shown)


libraries/package-deps-hash/src/getRepoState.ts
```diff
@@ -394,11 +293,23 @@ export function getRepoState(currentWorkingDirectory: string, gitPath?: string):
...
+  const { files, submodules } = parseGitLsTree(lsTreeResult.stdout);
 
-  applySubmodulesTreeState(rootDirectory, state, gitPath);
+  applyWorkingTreeState(rootDirectory, files, gitPath);
 
-  applyWorkingTreeState(rootDirectory, state, gitPath);
+  const state: Map<string, string> = files;
+
+  // Existence check for the .gitmodules file
+  const hasSubmodules: boolean = FileSystem.exists(`${rootDirectory}/.gitmodules`);
```
dmichon-msft, line 303:
```suggestion
  const hasSubmodules: boolean = submodules.size > 0 && FileSystem.exists(`${rootDirectory}/.gitmodules`);
```
That way we can skip the file system access if we don't think we have submodules. That said, the FileSystem check might be redundant unless we think we can have objects of type `commit` that aren't submodules.

dmichon-msft (approved):
I like this approach; it'll scale well to being converted to an async parallel operation, too.

dmichon-msft (approved):

iclanton (approved):

PR #3759, opened 2022-11-17
[heft-typescript-plugin] Fix incremental watch mode

dmichon-msft:
## Summary
Adds use of the `runIncremental` hook in `@rushstack/heft-typescript-plugin`.
Uses `oldProgram` and source file caching to get incremental rebuild, whether the `incremental` tsconfig setting is used or not.
Builds are faster if `incremental` is set, since less type-checking happens.

## Details
Single file changes (if not deep in the graph) can execute in <100ms for the TypeScript portion now.

## How it was tested
Running heft in watch mode in `rush-lib` and in `heft-node-everything-test`. Changed file, monitored performance and output files.

### DIFF
--- a/heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
+++ b/heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
@@ -15,7 +15,7 @@ import {
   Async,
   FileError
 } from '@rushstack/node-core-library';
-import type { IScopedLogger } from '@rushstack/heft';
+import type { IChangedFileState, IScopedLogger } from '@rushstack/heft';
 
 import type { ExtendedTypeScript } from './internalTypings/TypeScriptInternals';
 import { EmitFilesPatch, type ICachedEmitModuleKind } from './EmitFilesPatch';
@@ -111,6 +111,17 @@ const OLDEST_SUPPORTED_TS_MINOR_VERSION: number = 9;
 const NEWEST_SUPPORTED_TS_MAJOR_VERSION: number = 4;
 const NEWEST_SUPPORTED_TS_MINOR_VERSION: number = 8;
 
+interface ITypeScriptTool {
+  ts: ExtendedTypeScript;
+  measureSync: PerformanceMeasurer;
+  measureAsync: PerformanceMeasurerAsync;
+
+  sourceFileCache: Map<string, TTypescript.SourceFile>;
+
+  oldBuilderProgram: TTypescript.EmitAndSemanticDiagnosticsBuilderProgram | undefined;
+  oldProgram: TTypescript.Program | undefined;
+}
+
 export class TypeScriptBuilder {
   private readonly _configuration: ITypeScriptBuilderConfiguration;
   private readonly _typescriptLogger: IScopedLogger;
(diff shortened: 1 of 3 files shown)


common/changes/@rushstack/heft-typescript-plugin/typescript-incremental_2022-11-17-23-26.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-typescript-plugin",
+      "comment": "Restore incrementality of watch builds.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/heft-typescript-plugin"
+}
```
iclanton, line 10: Just delete this.

heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
```diff
@@ -159,117 +172,141 @@ export class TypeScriptBuilder {
...
+
+        oldBuilderProgram: undefined,
+        oldProgram: undefined
       };
-    };
+    }
+
+    const { performance } = this._tool.ts;
+    performance.disable();
+    performance.enable();
```
iclanton, line 293:
```suggestion
    // Reset performance counters
    performance.disable();
    performance.enable();
```

heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
```diff
@@ -159,117 +172,141 @@ export class TypeScriptBuilder {
...
 
-    this._typescriptTerminal.writeLine(`Using TypeScript version ${ts.version}`);
+    if (changedFiles) {
+      const { sourceFileCache } = this._tool;
+      for (const file of changedFiles.keys()) {
+        // There might need to be a transform to the format of `file` here.
+        sourceFileCache.delete(file);
```
iclanton, line 299: Make sure these paths are canonicalized on Windows/mac.

heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
```diff
@@ -874,14 +923,50 @@ export class TypeScriptBuilder {
...
+
+      const result: TTypescript.SourceFile | undefined = innerGetSourceFile(
+        fileName,
+        languageVersionOrOptions,
+        onError,
+        shouldCreateNewSourceFile
+      );
+      if (result) {
+        sourceFileCache.set(fileName, result);
+      }
```
iclanton, line 963:
```suggestion
      if (result) {
        sourceFileCache.set(fileName, result);
      } else {
        sourceFileCache.delete(fileName);
      }
```

heft-plugins/heft-typescript-plugin/src/TypeScriptPlugin.ts
```diff
@@ -308,7 +336,7 @@ export default class TypeScriptPlugin implements IHeftTaskPlugin {
 
     if (!partialTsconfigFile) {
```
iclanton, line 337: Consider dropping this check.

iclanton: Let's add tsconfig to the "break watch mode" with a comment saying that we should revise that in the future to allow during-runtime modifications to tsconfig.

iclanton (approved):

heft-plugins/heft-typescript-plugin/src/TypeScriptPlugin.ts
dmichon-msft: Unfortunately, the name is dynamic, so adding it to the config universally is non-trivial.

heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
dmichon-msft: File counts correct on Windows.

heft-plugins/heft-typescript-plugin/src/TypeScriptPlugin.ts
dmichon-msft: Leaving a comment, not fixing for now, because the tool doesn't like if you run it without a tsconfig.json. I could detect it being in the list of change files and fail out inside of the TypeScriptPlugin, will consider when fixing solution mode watch mode.

PR #3778, opened 2022-11-22
[rush] Add a feature (behind the "cleanInstallAfterNpmrcChanges" experiment) that will cause a clean install to be performed if the common/temp/.npmrc file has changed since the last install.

iclanton:
## Summary

Right now, Rush does not clear out the `common/temp/node_modules` folder if the `.npmrc` file changed since the last install. This can produce unexpected behavior, for instance if the `hoist` option is changed in under PNPM.

This PR adds a new feature (under a new "cleanInstallAfterNpmrcChanges" experiment) that triggers a clean install if the hash of the `.npmrc` file has changed since the last install.

## How it was tested

Included a unit test for the `LastInstallFlag` class and manually confirmed that changing the `.npmrc ` does not trigger a clean install with the experiment off and does with the experiment turned on.

### DIFF
--- a/libraries/rush-lib/src/api/LastInstallFlag.ts
+++ b/libraries/rush-lib/src/api/LastInstallFlag.ts
@@ -13,6 +13,14 @@ const lodash: typeof import('lodash') = Import.lazy('lodash', require);
 export const LAST_INSTALL_FLAG_FILE_NAME: string = 'last-install.flag';
 
 /**
+ * @internal
+ */
+export interface ICheckNpmrcOptions {
+  statePropertiesToIgnore?: string[];
+  rushVerb?: string;
+}
+
+/**
  * A helper class for managing last-install flags, which are persistent and
  * indicate that something installed in the folder was successfully completed.
  * It also compares state, so that if something like the Node.js version has changed,
@@ -36,8 +44,8 @@ export class LastInstallFlag {
   /**
    * Returns true if the file exists and the contents match the current state.
    */
-  public isValid(): boolean {
-    return this._isValid(false);
+  public isValid(options?: ICheckNpmrcOptions): boolean {
+    return this._isValid(false, options);
   }
 
   /**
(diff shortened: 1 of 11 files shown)


TheLarkInn (approved):

libraries/rush-lib/src/api/LastInstallFlag.ts
```diff
@@ -12,6 +12,14 @@ const lodash: typeof import('lodash') = Import.lazy('lodash', require);
 
 export const LAST_INSTALL_FLAG_FILE_NAME: string = 'last-install.flag';
 
+/**
+ * @internal
+ */
+export interface ICheckNpmrcOptions {
```
D4N14L, line 18: Hmm, given that this is being passed into the lockfile `isValid` methods, it feels a bit out of place to reference NPMRC.... maybe there should be a new `ILockfileValidityOptions` or something like that, which contains an `npmrcOptions` field? Also, the `rushVerb` property feels out of place in this interface but wouldn't in `ILockfileValidityOptions`

iclanton: Oops. I had npmrc on the mind when I wrote that interface name. Lemme change that.

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -126,15 +127,14 @@ export interface IInstallManagerOptions {
  * This class implements common logic between "rush install" and "rush update".
  */
 export abstract class BaseInstallManager {
-  private _rushConfiguration: RushConfiguration;
-  private _rushGlobalFolder: RushGlobalFolder;
-  private _commonTempInstallFlag: LastInstallFlag;
-  private _commonTempLinkFlag: LastLinkFlag;
```
D4N14L, line 132: Was the link flag not getting used at all? IIRC I left it in-place for legacy rush installs. But if it's no longer cared about, I have no issue removing it.

iclanton: It's getting used. I'm just marking it as `readonly` here.

Issue #3432, opened 2022-05-24
[api-extractor] Better support for type imports of the form `import { type X }`


D4N14L:
## Summary

Use of `import { type X }` appears to be ignored by API-Extractor for roll-up/documentation scenarios.

## Details

Currently, if you (for example) export an interface that takes only a type dependency on a module from some package using `import { type X }`, that type import becomes a full module import in the generated API-Extractor typings. Type imports of the form `import type { X }` do not encounter this issue.

### MyInterface.ts
```typescript
import { type X } from 'somePackage'

export interface MyInterface {
  myField: X
}
```

### index.ts
```typescript
export { MyInterface } from './MyInterface.ts';
```

### API-Extractor generated .d.ts
```typescript
import { X } from 'somePackage';

export declare interface MyInterface {
  myField: X
}
```

## Standard questions

Please answer these questions to help us investigate your issue more quickly:

| Question | Answer |
| -------- | -------- |
| `@microsoft/api-extractor` version? | 7.23.1 |
| Operating system? | Windows |
| API Extractor scenario? | rollups (.d.ts) |
| Would you consider contributing a PR? | Yes |
| TypeScript compiler version? | 4.6.3 |
| Node.js version (`node -v`)? | 14.19.1 |

PR #3784, opened 2022-11-23
[rush] Add a credentialMetadata field.

iclanton:
## Summary

Adds a metadata field to credentials.

## Details

This is a breaking API change, but this API is likely not used by partners yet.

## How it was tested

Added unit tests.

### DIFF
--- a/rush-plugins/rush-azure-storage-build-cache-plugin/src/AzureAuthenticationBase.ts
+++ b/rush-plugins/rush-azure-storage-build-cache-plugin/src/AzureAuthenticationBase.ts
@@ -3,7 +3,8 @@
 
 import { DeviceCodeCredential, type DeviceCodeInfo, AzureAuthorityHosts } from '@azure/identity';
 import type { ITerminal } from '@rushstack/node-core-library';
-import { CredentialCache, type ICredentialCacheEntry } from '@rushstack/rush-sdk';
+import { CredentialCache } from '@rushstack/rush-sdk';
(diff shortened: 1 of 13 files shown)


rush-plugins/rush-azure-storage-build-cache-plugin/src/AzureAuthenticationBase.ts
```diff
@@ -3,7 +3,8 @@
 
 import { DeviceCodeCredential, type DeviceCodeInfo, AzureAuthorityHosts } from '@azure/identity';
 import type { ITerminal } from '@rushstack/node-core-library';
-import { CredentialCache, type ICredentialCacheEntry } from '@rushstack/rush-sdk';
+import { CredentialCache } from '@rushstack/rush-sdk';
+import type { ICredentialCacheEntry } from '@rushstack/rush-sdk'; // Use a separate import line so the .d.ts file ends up with an `import type { ... }`
```
D4N14L, line 7: Maybe link the bug? (#3432)

D4N14L (approved):

rush-plugins/rush-azure-storage-build-cache-plugin/src/AzureAuthenticationBase.ts
iclanton:
```suggestion
// Use a separate import line so the .d.ts file ends up with an `import type { ... }`
// See https://github.com/microsoft/rushstack/issues/3432
import type { ICredentialCacheEntry } from '@rushstack/rush-sdk';
```

PR #3786, opened 2022-11-26
[rush] Tweak an incorrect changelog message

elliot-nelson:
## Summary

This changelog message says that rush _change_ user prompts are customizable, but actually it is rush _setup_ user prompts.

(Oops.)

## Details

 - Patching changelog source, which should automatically update the markdown during the next publish.


octogonz (approved):
Good catch 👍

PR #3789, opened 2022-11-28
[rush] Autoinstaller "all" functionality

elliot-nelson:
## Summary

A small pile of quality-of-life improvements for dealing with autoinstallers:

 - `rush update-autoinstaller --name folder1 --name folder2`
 - `rush update-autoinstaller --all`
 - `rush update --autoinstallers`
 - `rush install --autoinstallers`
 - All autoinstaller human-readable output now refers to _lockfiles_ instead of _shrinkwrap files_.

### Updating multiple autoinstallers

Following in the vein of `rush add`, you can now provide the `--name` parameter multiple times when updating an autoinstaller. Useful for specifically targeting 2 or more autoinstaller folders.

The output of the `update-autoinstaller` command has been updated to provide a summary of the autoinstallers that were updated and what new lockfiles should be committed to source control.

```console
rush update-autoinstaller --name rush-plugins --name rush-prettier
```

### Updating all autoinstallers

When your monorepo reaches 4-5 autoinstallers, it becomes easier to simply "mass update" them and evaluate the _result_ of the command (ensuring it matches your expectations) rather than typing out specific folder names.

```console
rush update-autoinstaller --all
```

### Convenience method: update + autoinstaller update

After a maintenance update frenzy, with multiple package.json changes spread across project and autoinstaller folders alike, a single command to make sure _everything_ is updated is a nice time saver.

```console
rush update --autoinstallers

# Roughly equivalent to:
rush update && rush update-autoinstaller --all
```

### New functionality: prepare all autoinstallers

A sibling command for `rush update --autoinstaller`, `rush install --autoinstaller` is unique in that it provides a feature that doesn't exist in Rush today: a way to _prepare all of your autoinstallers for use_ (without waiting until someone actually calls a related rush global command defined in `command-line.json`).

This is very useful for situations where processes outside of Rush's control expect autoinstallers to be up-to-date. A great example is the common `common/autoinstallers/rush-prettier` autoinstaller, which might also be used by VSCode, WebStorm, etc. If you've upgraded prettier or added new prettier plugins to your autoinstaller, `rush install --autoinstallers` can immediately fix "Missing plugin" errors for your VSCode developers.

## How it was tested

 - Tested installs and updates, with and without autoinstallers, in the rushstack repo (and local monorepo)
 - Tested update-autoinstaller with all and with multiple names in local monorepo (and rushstack repo)

### DIFF
--- a/libraries/rush-lib/src/cli/actions/BaseInstallAction.ts
+++ b/libraries/rush-lib/src/cli/actions/BaseInstallAction.ts
@@ -22,6 +22,7 @@ import { VersionMismatchFinder } from '../../logic/versionMismatch/VersionMismat
 import { Variants } from '../../api/Variants';
 import { RushConstants } from '../../logic/RushConstants';
 import { SelectionParameterSet } from '../parsing/SelectionParameterSet';
+import { Autoinstaller } from '../../logic/Autoinstaller';
 
 const installManagerFactoryModule: typeof import('../../logic/InstallManagerFactory') = Import.lazy(
   '../../logic/InstallManagerFactory',
@@ -40,6 +41,7 @@ export abstract class BaseInstallAction extends BaseRushAction {
   protected readonly _debugPackageManagerParameter: CommandLineFlagParameter;
   protected readonly _maxInstallAttempts: CommandLineIntegerParameter;
   protected readonly _ignoreHooksParameter: CommandLineFlagParameter;
+  protected readonly _autoinstallersParameter: CommandLineFlagParameter;
   /*
    * Subclasses can initialize the _selectionParameters property in order for
    * the parameters to be written to the telemetry file
@@ -89,6 +91,10 @@ export abstract class BaseInstallAction extends BaseRushAction {
     this._ignoreHooksParameter = this.defineFlagParameter({
       parameterLongName: '--ignore-hooks',
       description: `Skips execution of the "eventHooks" scripts defined in rush.json. Make sure you know what you are skipping.`
+    });
+    this._autoinstallersParameter = this.defineFlagParameter({
+      parameterLongName: '--autoinstallers',
+      description: `After installing all packages, also refresh the install for all existing autoinstallers.`
     });
     this._variant = this.defineStringParameter(Variants.VARIANT_PARAMETER);
   }
@@ -152,6 +158,14 @@ export abstract class BaseInstallAction extends BaseRushAction {
     let installSuccessful: boolean = true;
     try {
       await installManager.doInstallAsync();
+
+      if (this._autoinstallersParameter.value) {
+        if (installManagerOptions.allowShrinkwrapUpdates) {
+          Autoinstaller.updateAutoinstallers(this.rushConfiguration);
+        } else {
+          await Autoinstaller.prepareAutoinstallersAsync(this.rushConfiguration);
+        }
+      }
 
       if (warnAboutScriptUpdate) {
         console.log(
--- a/libraries/rush-lib/src/logic/Autoinstaller.ts
+++ b/libraries/rush-lib/src/logic/Autoinstaller.ts
@@ -4,7 +4,14 @@
 import colors from 'colors/safe';
 import * as path from 'path';
 
-import { FileSystem, IPackageJson, JsonFile, LockFile, NewlineKind } from '@rushstack/node-core-library';
+import {
+  FileSystem,
+  FolderItem,
+  IPackageJson,
+  JsonFile,
+  LockFile,
+  NewlineKind
+} from '@rushstack/node-core-library';
 import { Utilities } from '../utilities/Utilities';
 
 import { PackageName, IParsedPackageNameOrError } from '@rushstack/node-core-library';
@@ -155,7 +162,14 @@ export class Autoinstaller {
     }
   }
 
-  public update(): void {
+  /**
+   * Update this autoinstaller, installing the latest packages defined in the package.json file
+   * and regenerating the lockfile.
+   *
+   * Returns `true` if the lockfile was updated and should be committed to git, or `false` if
+   * the autoinstaller folder is already up-to-date.
+   */
+  public update(): boolean {
     const autoinstallerPackageJsonPath: string = path.join(this.folderFullPath, 'package.json');
 
     if (!FileSystem.exists(autoinstallerPackageJsonPath)) {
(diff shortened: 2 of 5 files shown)


isc30:
Hey, the PR looks good. I would go 1 step further tho, I would:

- remove `update-autoinstaller` in favour of plain `rush update`
- remove `--autoinstallers` flag from `rush update`
- keep the `--autoinstallers` flag from `rush install`

With this, the only options will be:
- run `rush update` to update all the *changed packages and autoinstallers*
- run `rush install` to install the dependencies of your packages (and let autoinstallers do a lazy install)
- run `rush install --autoinstallers` to install all the dependencies of your packages and autoinstallers (not lazy)

elliot-nelson:
@isc30 I want to be careful about `rush update-autoinstaller` because this does have an important use: the way Rush loads _Rush plugins_ is through an autoinstaller, which means you have to have a way to update autoinstallers that is outside the normal bounds of `install/update/build` (because plugins can actually modify the behavior of those commands). You could end up in a state where a bad autoinstall breaks your ability to fix your autoinstaller.

I also think `rush update` and `rush update-autoinstaller` tend to have different audiences... typically only your monorepo maintainers, plus a small set of advanced users, ever touch autoinstallers (usually to set up custom global commands). Whereas application developers in your repo often run `rush update`, because they are constantly updating the dependencies of their own projects, merging changes from main that have updates and fixing their lockfile, etc. I wouldn't want the _application developer_ workflow to get too muddied up with the _maintainer workflow_, if I can avoid it.

> Hey, the PR looks good. I would go 1 step further tho, I would:
> 
> * remove `update-autoinstaller` in favour of plain `rush update`
> * remove `--autoinstallers` flag from `rush update`
> * keep the `--autoinstallers` flag from `rush install`
> 
> With this, the only options will be:
> 
> * run `rush update` to update all the _changed packages and autoinstallers_
> * run `rush install` to install the dependencies of your packages (and let autoinstallers do a lazy install)
> * run `rush install --autoinstallers` to install all the dependencies of your packages and autoinstallers (not lazy)

isc30:
okay then it sounds good, I like having the option to do `rush update --autoinstallers` as a shortcut for all of them

isc30 (approved):

isc30:
Yeah, my case was for `prettier` specifically since I wanted it to run on file save from vscode. People that never did a commit (that downloaded and executed prettier) wouldn't get that working so my take was to force install it during normal installation.

I think `rush install --autoinstallers` will do for that, so I can replace my custom JS code with that

elliot-nelson:
@isc30 The new option `rush install --autoinstallers` would install stuff if people ran it, which you could suggest (if VSCode wasn't working). But if you add it as a post install hook I think you'll end up in a recursive loop -- that's why we might need that extra command.

isc30:
Yeah I can change our onboarding docs to run `rush install --autoinstallers` before working with the repo.

In the past we couldn't do that as there wasn't a way to install autoinstallers from CLI

libraries/rush-lib/src/cli/actions/BaseInstallAction.ts
```diff
@@ -153,6 +159,14 @@ export abstract class BaseInstallAction extends BaseRushAction {
     try {
       await installManager.doInstallAsync();
 
+      if (this._autoinstallersParameter.value) {
+        if (installManagerOptions.allowShrinkwrapUpdates) {
+          Autoinstaller.updateAutoinstallers(this.rushConfiguration);
```
iclanton, line 164: Sorta odd that this one is sync.

libraries/rush-lib/src/logic/Autoinstaller.ts
```diff
@@ -233,4 +247,100 @@ export class Autoinstaller {
...
+    }
+
+    console.log(colors.gray('='.repeat(79)));
+
+    if (autoinstallerNames.length === 0) {
+      console.log('\nNo autoinstallers to prepare.');
+      return;
+    }
+
+    console.log(colors.green(`\n${autoinstallerNames.length} autoinstaller(s) prepared.`));
```
iclanton, line 297: What does "prepare" mean in this context? I'm not sure a user is going to understand that.

libraries/rush-lib/src/logic/Autoinstaller.ts
```diff
@@ -233,4 +247,100 @@ export class Autoinstaller {
...
+    }
+
+    const updated: Autoinstaller[] = [];
+    const skipped: Autoinstaller[] = [];
+
+    for (const autoinstallerName of autoinstallerNames) {
+      const autoinstaller: Autoinstaller = new Autoinstaller({
+        autoinstallerName,
+        rushConfiguration
+      });
```
iclanton, line 316: Will this error if the named autoinstaller doesn't exist?

elliot-nelson: Yes -- it treats it essentially like a series of steps, for example:

```console
rush update-autoinstaller --name good1 --name good2 --name oops
```

Assuming `oops` is a misspelled autoinstaller name, the command would update good1, good2, and then dump out with the typical error. The user could then re-run the command (fixing the mistake), and it would end up printing: 2 autoinstallers skipped, 1 autoinstaller updated (since good1 and good2 had just been updated by the last run).

It would probably be a smidge nicer of an experience if we completely wrapped up (skips, successes, failures) in a bow and presented them, but failing install on an autoinstaller is unusual enough that it seemed unnecessary for this PR.

libraries/rush-lib/src/cli/actions/BaseInstallAction.ts
elliot-nelson: Hmm... I'll see what happens if I make it async, might be worth it it to safeguard against future need for async methods.

Faithfinder:
May I suggest `rush add -m --autoinstallers`?
My case is for Prettier. We have Prettier installed in most projects simply for VScode's extension discovery benefit, so upgrading it in one command would be nice

elliot-nelson:
> May I suggest `rush add -m --autoinstallers`? My case is for Prettier. We have Prettier installed in most projects simply for VScode's extension discovery benefit, so upgrading it in one command would be nice

@Faithfinder Ah so, `rush add -m` being used to upgrade a dependency (possibly), and the `--autoinstallers` flag means "if this is in any of my autoinstallers, update it there too".

I'm beginning to wonder if "--autoinstallers" is too brief/vague and it should be "--include-autoinstallers", to make it clear that the command in each situation does what it normally does, just _also_ does it for autoinstallers.

Faithfinder:
> > May I suggest `rush add -m --autoinstallers`? My case is for Prettier. We have Prettier installed in most projects simply for VScode's extension discovery benefit, so upgrading it in one command would be nice
> 
> @Faithfinder Ah so, `rush add -m` being used to upgrade a dependency (possibly), and the `--autoinstallers` flag means "if this is in any of my autoinstallers, update it there too".
> 
> I'm beginning to wonder if "--autoinstallers" is too brief/vague and it should be "--include-autoinstallers", to make it clear that the command in each situation does what it normally does, just _also_ does it for autoinstallers.

Yep! `--include-autoinstallers` is indeed more clear, but I don't care too much because there's no way I'm keeping it in my head anyway - will need to check the docs most of the time.

PR #3790, opened 2022-11-29
[jest-heft-plugin] Remove the Jest postinstall step.

iclanton:
This PR removes the Jest postinstall step. PNPM's patch feature is a better solution to this problem.

The patch should look like this `@jest__transform@27.4.5.patch`:

```
diff --git a/build/ScriptTransformer.js b/build/ScriptTransformer.js
index fb7b610e9559e11b2e58dd7a39f565c98ce6ed1d..cdb923705cd18aca87e134e93b7c14c7a48cd4a1 100644
--- a/build/ScriptTransformer.js
+++ b/build/ScriptTransformer.js
@@ -1002,24 +1002,7 @@ function readCodeCacheFile(cachePath) {
  */
 
 const writeCacheFile = (cachePath, fileData) => {
-  try {
-    (0, _writeFileAtomic().sync)(cachePath, fileData, {
-      encoding: 'utf8',
-      fsync: false
-    });
-  } catch (e) {
-    if (cacheWriteErrorSafeToIgnore(e, cachePath)) {
-      return;
-    }
-
-    e.message =
-      'jest: failed to cache transform results in: ' +
-      cachePath +
-      '\nFailure message: ' +
-      e.message;
-    removeFile(cachePath);
-    throw e;
-  }
+  return;
 };
 /**
  * On Windows, renames are not atomic, leading to EPERM exceptions when two
@@ -1034,31 +1017,7 @@ const cacheWriteErrorSafeToIgnore = (e, cachePath) =>
   fs().existsSync(cachePath);
 
 const readCacheFile = cachePath => {
-  if (!fs().existsSync(cachePath)) {
-    return null;
-  }
-
-  let fileData;
-
-  try {
-    fileData = fs().readFileSync(cachePath, 'utf8');
-  } catch (e) {
-    e.message =
-      'jest: failed to read cache file: ' +
-      cachePath +
-      '\nFailure message: ' +
-      e.message;
-    removeFile(cachePath);
-    throw e;
-  }
-
-  if (fileData == null) {
-    // We must have somehow created the file but failed to write to it,
-    // let's delete it and retry.
-    removeFile(cachePath);
-  }
-
-  return fileData;
+  return;
 };
 
 const getScriptCacheKey = (filename, instrument) => {
```


common/changes/@rushstack/heft-jest-plugin/remove-postinstall_2022-11-28-23-57.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-jest-plugin",
+      "comment": "Remove a postinstall step that patches Jest in-place. This is better achieved with a PNPM patch.",
```
iclanton, line 5:
```suggestion
      "comment": "Remove a postinstall step that patches Jest in-place. This is better achieved with a PNPM patch. See https://github.com/microsoft/rushstack/pull/3790 for more information.",
```

D4N14L (approved):

PR #3793, opened 2022-11-29
[rush-lib] Upgrade `npm-check` dependency to `6.0.1`

xirzec:
## Summary

A simple version bump. From what I saw in the dependency's repo they didn't actually make a breaking change to their public surface for `6.x`

## How it was tested

Ran the full `rush test` suite. I don't think there's any new tests that need to be added since no logic was changed.

### DIFF
--- a/common/changes/@microsoft/rush/updateNpmCheck_2022-11-29-19-46.json
+++ b/common/changes/@microsoft/rush/updateNpmCheck_2022-11-29-19-46.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Upgraded `npm-check` from `5.9.2` to `6.0.1`",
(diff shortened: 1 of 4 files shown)


common/changes/@microsoft/rush/updateNpmCheck_2022-11-29-19-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Upgraded `npm-check` from `5.9.2` to `6.0.1`",
```
iclanton, line 5:
```suggestion
      "comment": "Upgrade `npm-check` dependency from `~5.9.2` to `~6.0.1`",
```

iclanton (approved):

PR #3795, opened 2022-11-30
Enelson/buildcachetemp

### LINKED ISSUE #3783: [rush] Cache often fails when multiple builds are happening simultaneously
## Summary

This is a follow-up from issue #3454, which is better, but not quite fixed.

## Repro steps

Use a network volume as a Rush cache and have multiple jobs running builds at the same time.

## Details

My comment from the other issue:

I'm not convinced this is completely fixed. We've upgraded to a version of Rush with this fix (5.81.0) and though the problem is better, we still see this a lot:

```
12:11:43  ==[ @arista/cv-components ]======================================[ 32 of 33 ]==
12:14:04  File does not exist: /home/jenkins/workspace/Geiger/geiger-merge/common/temp/build-cache/2e6ce6b868ae4a277db20f881fc977d34ccd6ad0.temp
12:14:04  ENOENT: no such file or directory, rename '/home/jenkins/workspace/Geiger/geiger-merge/common/temp/build-cache/2e6ce6b868ae4a277db20f881fc977d34ccd6ad0.temp' -> '/home/jenkins/workspace/Geiger/geiger-merge/common/temp/build-cache/2e6ce6b868ae4a277db20f881fc977d34ccd6ad0'
12:14:04  "@arista/cv-components" failed to build.
```

Re-running the build always passes, so it's clearly some temporary cache issue. My guess is that multiple builds are running simultaneously and trying to add cache entries at the same time. If this particular rename fails, Rush could check if the desired cache entry already exists, and if so, just move along without failing.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.81.0 |
| `rushVersion` from rush.json? | 5.81.0 |
| Operating system? | Linux |
| Node.js version (`node -v`)? | 16.17.0 |

elliot-nelson:
## Summary

When writing a local build cache entry, we tar up the build output folders into a temp file based on the computed final output file for the cache entry, and once tar is successful, we move that file on top of the target file (in an attempt to prevent corrupted entries if there are issues).

But, if multiple rush builds are attempting to write the same cache entry at the same time, these tar processes will end up writing to the same temp file. This PR fixes that by adding a random sequence to the end of the temp file name.

Potentially fixes #3783.

## How it was tested

 - Runs locally on rushstack and our local monorepo
 - I don't have a way to recreate the symptom, so this is guesswork.

### DIFF
--- a/common/changes/@microsoft/rush/enelson-buildcachetemp_2022-11-30-12-56.json
+++ b/common/changes/@microsoft/rush/enelson-buildcachetemp_2022-11-30-12-56.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Writing local build cache is more robust on a network drive.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/libraries/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
+++ b/libraries/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
@@ -239,8 +239,13 @@ export class ProjectBuildCache {
     const tarUtility: TarExecutable | undefined = await ProjectBuildCache._tryGetTarUtility(terminal);
     if (tarUtility) {
       const finalLocalCacheEntryPath: string = this._localBuildCacheProvider.getCacheEntryPath(cacheId);
+
       // Derive the temp file from the destination path to ensure they are on the same volume
-      const tempLocalCacheEntryPath: string = `${finalLocalCacheEntryPath}.temp`;
+      // In the case of a shared network drive containing the build cache, we also need to make
+      // sure the the temp path won't be shared by two parallel rush builds.
+      const randomSuffix: string = crypto.randomBytes(8).toString('hex');
+      const tempLocalCacheEntryPath: string = `${finalLocalCacheEntryPath}-${randomSuffix}.temp`;
+
       const logFilePath: string = this._getTarLogFilePath();
       const tarExitCode: number = await tarUtility.tryCreateArchiveFromProjectPathsAsync({
         archivePath: tempLocalCacheEntryPath,


iclanton:
I'm not sure this will solve the issue described in https://github.com/microsoft/rushstack/issues/3783. @kaiyoma had an idea to check if the move destination path already exists if the move fails. We could do that in addition to this.

elliot-nelson:
> I'm not sure this will solve the issue described in #3783. @kaiyoma had an idea to check if the move destination path already exists if the move fails. We could do that in addition to this.

@iclanton That makes sense, we could also do some sleeps and retries on a failed move. But based on the error message, the move is not "failing" -- the `.temp` file does not exist. The only explanation for that I can think of is that two different processes were writing to the `.temp` file, one moved it, and the other tried to move it and failed (because it no longer existed).

If the move fails due to conflict at the destination, I'd expect an EACCESS or similar, not the error message described.

iclanton:
>If the move fails due to conflict at the destination, I'd expect an EACCESS or similar, not the error message described.

@elliot-nelson - That's a good point. So it sounds like there's probably a race condition, where two machines are writing to the same cache entry and one renames the temp folder before the other.

So what if we do this:
- Include some random stuff in the temp filename
- Before we rename, check if the target file exists
  - If it does, delete the temp file without renaming
  - If it doesn't, rename
    - If this rename fails, check if the target file exists again, and if it does, delete. If it doesn't, bubble the error up.

elliot-nelson:
@iclanton I like that plan, except for one detail - I don't think the first check if file exists should be in there.

Imagine for example (in this case) 2 Jenkins builds run, and there's a 10 minute build. The scenario likely is a follow-up commit push -- so Build A has built most of it and reached the expensive project, Build B has caught up fast because it's already cached so they are building the same thing. The behavior I expect is for Build B to overwrite whatever exists, even if Build A just wrote it 3 seconds ago.

(maybe a more compelling argument -- "rush rebuild" is expected to ignore entries and overwrite them, not build everything and then discard it)

iclanton:
@elliot-nelson - Good point. Agreed.

elliot-nelson:
Updates:

 - Use the Async wrapper to do a simple 3-try, 500ms delay move, and then error if none of them worked
 - No need to delete anything -- if we can't overwrite it with a move, we won't be able to delete it either

iclanton:
>No need to delete anything -- if we can't overwrite it with a move, we won't be able to delete it either

Not necessarily. Someone might have a lock on the destination file. I feel like we should still try to delete so we don't end up with temp entries hanging around.

Do we bubble the error up if we fail to rename?

elliot-nelson:
Ohhhh... delete the temp file? That makes sense for sure, I will add that at least.

iclanton (approved):

PR #3802, opened 2022-12-02
[heft-next] Various Updates

D4N14L:
## Details

Addresses various issues with the new Heft package, including:
- adds a `heft clean` action, which is scopeable using `--to`/`--to-except`/`--only`
  - default functionality will run all clean specs
- updates the watcher to initialize state using `git ls-files` to avoid additional overhead of ingesting non-source files at the beginning of a build
- updates the watcher to check for gitignored files using `ignore` library
- updates the sass typings plugin to implement fixes for incremental mode
- updates the lint plugin to not run during watch mode
- updates the `addCopyOperations` API to support incremental copying

## How it was tested

Tested using internal repo + added tests


apps/heft/src/pluginFramework/HeftTaskSession.ts
```diff
@@ -164,55 +165,33 @@ export interface IChangedFileState {
...
+ * Used to specify a selection of files to copy from a specific source folder to one
+ * or more destination folders.
  *
  * @public
  */
-export interface IGlobChangedFilesOptions {
+export interface IIncrementalCopyOperation extends ICopyOperation {
   /**
-   * Current working directory that the glob pattern will be applied to.
+   * If true, the file will be copied only if the source file has changed since it was last copied.
```
dmichon-msft, line 175: If true, the copy operation will only consider files that have changed since the last time the current task executed.

apps/heft/src/operations/runners/TaskOperationRunner.ts
```diff
@@ -77,92 +96,140 @@ export class TaskOperationRunner implements IOperationRunner {
...
+          if (!taskOperationLockFile) {
+            throw new InternalError(
+              `Failed to acquire lock file ${JSON.stringify(lockFileName)}. Are multiple instances of ` +
+                'Heft running?'
+            );
+          }
+          await lockfileChangePromise;
+
+          // We can save some time by avoiding deleting the lockfile
+          taskOperationLockFile.release(/*deleteFile:*/ false);
```
dmichon-msft, line 224: This seems like it ought to be a helper function

apps/heft/src/cli/HeftActionRunner.ts
```diff
@@ -494,7 +569,15 @@ export class HeftActionRunner {
...
         fileEventListener
-      ).then(() => false);
+      ).then(
+        () => false,
+        (error: unknown) => {
+          // Re-throw the error so that it doesn't get swallowed. This way we can ensure that the list of
+          // changed files is not cleared, since we will want to resurface the error on future incremental
+          // builds.
+          throw error;
+        }
```
dmichon-msft, line 579: This should technically do the same thing as not passing this parameter at all. Curious what behavior you are seeing?

dmichon-msft (commented):
Noted a few things. Still looking more at the ignore side of things.

iclanton:
So at this point, Heft has a hard requirement on Git, right? Do we semi-gracefully handle the case when Git isn't present, or at least provide a useful error message instead of just crashing?

apps/heft/src/utilities/GitUtilities.ts
```diff
@@ -230,6 +293,131 @@ export class GitUtilities {
...
+          // Add the pattern to the list of resolved patterns in the order they are read, since the order
+          // of declaration of patterns in a .gitignore file matters for negations
+          resolvedGitIgnorePatterns.push(gitIgnorePattern);
+        }
+
+        // Add the patterns to the ignore matcher patterns. Since we are crawling up the directory tree to
+        // the root of the Git repo we need to prepend the patterns, since the order of declaration of
+        // patterns in a .gitignore file matters for negations
+        ignoreMatcherPatterns.unshift(...resolvedGitIgnorePatterns);
+        currentPath = path.dirname(currentPath);
```
iclanton, line 385:
```suggestion
        currentPath = currentPath.substring(0, currentPath.lastIndexOf('/'));
```

apps/heft/src/utilities/GitUtilities.ts
```diff
@@ -230,6 +293,131 @@ export class GitUtilities {
     }
   }
 
+  private async _getIgnoreMatchersAsync(gitRepoRootPath: string): Promise<Map<string, IIgnoreMatcher>> {
```
iclanton, line 296: Unit tests?

common/reviews/api/heft.api.md
```diff
@@ -54,7 +54,10 @@ export { CommandLineStringListParameter }
 export { CommandLineStringParameter }
 
 // @public
-export type GlobChangedFilesFn = (patterns: string | string[], options?: IGlobChangedFilesOptions) => string[];
+export type GlobFn = (pattern: string | string[], options?: IPartialGlobOptions | undefined) => Promise<string[]>;
+
+// @public
+export type GlobSyncFn = (pattern: string | string[], options?: IPartialGlobOptions | undefined) => string[];
```
iclanton, line 60:
```suggestion
export type GlobFn = (pattern: string | string[], options?: IPartialGlobOptions | undefined) => Promise<string[]> | string[];
```

Do we actually care if it's async or not? Aren't we going just await this?

common/reviews/api/heft.api.md
```diff
@@ -214,10 +209,11 @@ export interface IHeftTaskRunHookOptions {
 
 // @public
 export interface IHeftTaskRunIncrementalHookOptions extends IHeftTaskRunHookOptions {
+    readonly addCopyOperations: (...copyOperations: IIncrementalCopyOperation[]) => void;
     // @beta
     readonly cancellationToken: CancellationToken;
     readonly changedFiles: ReadonlyMap<string, IChangedFileState>;
-    readonly globChangedFiles: GlobChangedFilesFn;
+    readonly globChangedFiles: GlobSyncFn;
```
iclanton, line 216: Should this type be `GlobSyncFn | GlobFn`?

apps/heft/src/utilities/GitUtilities.ts
```diff
@@ -109,8 +112,68 @@ export class GitUtilities {
...
+      if (!path.isAbsolute(filePath)) {
+        throw new Error(`The filePath must be an absolute path: "${filePath}"`);
+      }
+
+      // Find the best matcher for the file path by finding the longest matcher path that is a prefix to
+      // the file path
+      let longestMatcherPath: string | undefined;
+      let foundMatcher: IIgnoreMatcher | undefined;
+      for (const [matcherPath, matcher] of ignoreMatcherMap) {
+        if (filePath.startsWith(matcherPath) && matcherPath.length > (longestMatcherPath?.length || 0)) {
```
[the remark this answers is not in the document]
D4N14L: Talked offline, while I don't disagree, going to avoid for 3 reasons:
- It's currently in rush-lib and that would complicate publishing release versions
- It doesn't support leaf node traversal yet
- The `startsWith` check checks against every ignorer, but the ignorers are only created per-gitignore under a project. Because there are generally 0-1 .gitignores in any given project (with a few exceptions), the general case is just not affected all that much by doing a `startsWith` check for all ignorers

Will add a comment

apps/heft/src/plugins/CopyFilesPlugin.ts
```diff
@@ -2,10 +2,11 @@
 // See LICENSE in the project root for license information.
 
 import * as path from 'path';
+import glob from 'fast-glob';
```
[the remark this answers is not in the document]
D4N14L: In the longer term, yeah probably. fast-glob is a good bit more performant than the normal globber. I think there are some subtle differences in supported features, but only more esoteric scenarios.

apps/heft/src/plugins/FileGlobSpecifier.ts
```diff
@@ -35,13 +35,69 @@ export interface IFileSelectionSpecifier {
   includeGlobs?: string[];
 }
 
-export async function getFilePathsAsync(fileGlobSpecifier: IFileSelectionSpecifier): Promise<Set<string>> {
+/**
+ * A supported subset of options used when globbing files.
+ *
+ * @public
+ */
+export interface IPartialGlobOptions {
```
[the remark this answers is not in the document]
D4N14L: Yes. Because we're using a `FileSystemAdapter` passed into the globber, and the  FileSystemAdapter is only partially implemented (doesn't do stuff regarding symlinks, etc.), I only allow passing a specific subset of args that would be useful.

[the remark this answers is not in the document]
D4N14L: I'll update it to be a `Pick<...>` type instead, since all those properties are already optional.

apps/heft/src/utilities/GitUtilities.ts
```diff
@@ -109,8 +112,68 @@ export class GitUtilities {
...
+
+    // Return the set of unignored files. The output is relative to the working directory, so join
+    // them to make an absolute path.
+    return new Set(result.outputLines.map((line: string) => path.join(this._workingDirectory, line)));
+  }
+
+  public async checkIgnoreAsync(filePaths: Iterable<string>): Promise<Set<string>> {
+    this._ensurePathIsUnderGitWorkingTree();
+
+    const gitInfo: IGitRepoInfo = this.getGitInfo()!;
```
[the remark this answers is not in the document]
D4N14L: Yeah it is. Though, this is addressed by tweaking how we check if we're under the working tree (addressing David's comment).

apps/heft/src/utilities/GitUtilities.ts
D4N14L: Oops, forgot to include those. They were staged but not comitted.

common/reviews/api/heft.api.md
D4N14L: I made both types because I use the `getFilePathsAsync` method in FileGlobSpecifier.ts both for async and sync contexts. Sometimes the actual glob method is passed in (ex. in `CopyFilesPlugin.copyFilesAsync`), and sometimes it uses the sync method (ex. in `CopyFilesPlugin.copyIncrementalFilesAsync`, which eventually does async file copies)

D4N14L: Though I see what you're saying... it can just be the async type and not do any async work. I'll remove.

apps/heft/src/utilities/GitUtilities.ts
```diff
@@ -109,8 +112,68 @@ export class GitUtilities {
...
+
+    if (result.exitCode !== 0) {
+      throw new Error(
+        `The "git ls-files" command failed with status ${result.exitCode}: ` + result.errorLines.join('\n')
+      );
+    }
+
+    // Return the set of unignored files. The output is relative to the working directory, so join
+    // them to make an absolute path.
+    return new Set(result.outputLines.map((line: string) => path.join(this._workingDirectory, line)));
```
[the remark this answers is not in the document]
D4N14L: I've updated the call to `git ls-files` to take a pattern that matches only gitignore files, so now we will only have relevant files. The other call to `git ls-files` was removed for the sake of non-git compatibility, and because the watcher initial crawl was going to take a long time regardless of whether or not the `ignoreInitial` var was true.

D4N14L:
> So at this point, Heft has a hard requirement on Git, right? Do we semi-gracefully handle the case when Git isn't present, or at least provide a useful error message instead of just crashing?

I've updated it so that we can gracefully handle the no-git scenario. In this case, devs would need to manually specify ignore globs in the heft.json file. Seems like a fair compromise.

iclanton (approved):

apps/heft/src/plugins/FileGlobSpecifier.ts
```diff
@@ -35,13 +35,56 @@ export interface IFileSelectionSpecifier {
...
+
+/**
+ * Glob a set of files and return a list of paths that match the provided patterns.
+ *
+ * @param patterns - Glob patterns to match against.
+ * @param options - Options that are used when globbing the set of files.
+ *
+ * @public
+ */
+export type GlobFn = (pattern: string | string[], options?: IGlobOptions | undefined) => Promise<string[]>;
```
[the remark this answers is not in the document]
D4N14L: I'm going to leave as-is to reduce confusion when using the API, since we make this type available on the taskSession

PR #3810, opened 2022-12-05
[rush] Simplify JSON Schema loading.

iclanton:
This PR replaces all of the `JsonSchema.fromFile` instances with `JsonSchema.fromLoadedObject` in `rush-lib`. This improves bundleability of `rush-lib`.


octogonz (approved):
I was going to complain about the `fromLoadedObject()` API name, but apparently I created that API 😆

PR #3813, opened 2022-12-05
[rush] Support CI build log folding using env vars

### LINKED ISSUE #2198: [rush] Support custom formatting of merged logs
## [rush] Support custom formatting of merged logs ##

- [x] Feature
- [ ] Bug

**Summary.**
Depending on the consumer of the logs, there is value to inserting additional text at the collation stage to take advantage of the capabilities of the intended viewer. For example, the Azure DevOps log viewer supports collapsible regions via the `##[group]<summary text>` and `##[endgroup]` tags. This could be implemented as simply as a configuration option for text to prepend and append to the logs for each project.

elliot-nelson:
## Summary

Allow monorepo maintainers to fold long build logs in CI using simple header and footer environment variables.

Fixes #2198.

## Details

Since this behavior typically doesn't make sense on local machines, going with environment variables makes the most sense.

Although this behavior is called "grouping" by both GitHub and Azure, I think this word is rather ambiguous; log _folding_ is much more specific and better describes the actual functionality you get. I've proposed `RUSH_BUILD_FOLDING_HEADER` and `RUSH_BUILD_FOLDING_FOOTER` but am open to alternatives!

## How it was tested

 - Tested locally to verify output with env vars on and off, verbose on and off.

### DIFF
--- a/libraries/rush-lib/src/api/EnvironmentConfiguration.ts
+++ b/libraries/rush-lib/src/api/EnvironmentConfiguration.ts
@@ -154,6 +154,44 @@ export enum EnvironmentVariableNames {
   RUSH_TAR_BINARY_PATH = 'RUSH_TAR_BINARY_PATH',
 
   /**
+   * If defined, when Rush runs any bulk or phased command in verbose mode, this line will be
+   * printed above the build output. The following variables are provided as interpolated values
+   * if desired:
+   *
+   *   - `${taskName}` - the name of the task being executed (as printed in the header)
+   *
+   * This value is useful for folding long build output in CI environments. Here are some suggested
+   * values for this environment variable:
+   *
+   *   - GitHub Actions:
+   *     - `::group::${taskName}`
+   *   - Azure Devops:
+   *     - `##[group]${taskName}`
+   *   - Travis CI:
+   *     - `travis_fold:start:${taskName}`
+   */
+  RUSH_BUILD_FOLDING_HEADER = 'RUSH_BUILD_FOLDING_HEADER',
+
+  /*
+   * If defined, when Rush runs any bulk or phased command in verbose mode, this line will be
+   * printed below the build output. The following variables are provided as interpolated values
+   * if desired:
+   *
+   *   - `${taskName}` - the name of the task being executed (as printed in the header)
+   *
+   * This value is useful for folding long build output in CI environments. Here are some suggested
+   * values for this environment variable:
+   *
+   *   - GitHub Actions:
+   *     - `::endgroup::`
+   *   - Azure Devops:
+   *     - `##[endgroup]`
+   *   - Travis CI:
+   *     - `travis_fold:end:${taskName}`
+   */
+  RUSH_BUILD_FOLDING_FOOTER = 'RUSH_BUILD_FOLDING_FOOTER',
+
+  /**
    * When Rush executes shell scripts, it sometimes changes the working directory to be a project folder or
    * the repository root folder.  The original working directory (where the Rush command was invoked) is assigned
    * to the the child process's `RUSH_INVOKED_FOLDER` environment variable, in case it is needed by the script.
@@ -199,6 +237,10 @@ export class EnvironmentConfiguration {
   private static _gitBinaryPath: string | undefined;
 
   private static _tarBinaryPath: string | undefined;
+
+  private static _buildFoldingHeader: string | undefined;
+
+  private static _buildFoldingFooter: string | undefined;
 
   /**
    * An override for the common/temp folder path.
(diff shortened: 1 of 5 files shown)


libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -137,11 +144,23 @@ export class OperationExecutionManager {
 
       const middlePart: string = colors.gray(']' + '='.repeat(middlePartLengthMinusTwoBrackets) + '[');
 
+      if (!this._quietMode) {
+        if (this._projectLogFooter && this._lastTask) {
+          this._terminal.writeStdoutLine(this._projectLogFooter.replace(/\$\{taskName\}/g, this._lastTask));
```
iclanton, line 149: Extract the token replacement into a function?

libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -201,6 +220,12 @@ export class OperationExecutionManager {
       }
     );
 
+    if (!this._quietMode) {
+      if (this._projectLogFooter && this._lastTask) {
+        this._terminal.writeStdoutLine(this._projectLogFooter.replace(/\$\{taskName\}/g, this._lastTask));
```
iclanton, line 225: Seems like there should be a matching header here.

libraries/rush-lib/src/api/EnvironmentConfiguration.ts
```diff
@@ -153,6 +153,44 @@ export enum EnvironmentVariableNames {
...
+   * values for this environment variable:
+   *
+   *   - GitHub Actions:
+   *     - `::group::${taskName}`
+   *   - Azure Devops:
+   *     - `##[group]${taskName}`
+   *   - Travis CI:
+   *     - `travis_fold:start:${taskName}`
+   */
+  RUSH_BUILD_FOLDING_HEADER = 'RUSH_BUILD_FOLDING_HEADER',
```
iclanton, line 173: We've started moving away from the word "build" in favor of "operation" since commands and tasks aren't necessarily builds.

libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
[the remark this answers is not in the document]
elliot-nelson: The "footer" here is being written before we write the next project separator line (to close out the previous build log section).

libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
elliot-nelson: In the "stream started" callback for each operation, we write the footer for last section, the separator, and then the header for the next section, and wait until all operations have executed the stream started callback.

At this point, all the operations are done, but we're still "inside" the last folding group that was started, so we need one final line to close it out (if defined).

libraries/rush-lib/src/api/EnvironmentConfiguration.ts
elliot-nelson: :+1: Open to changing this, my reasoning is:

(1) It matches all existing `RUSH_BUILD_` settings like for cache, etc.
(2) Even for people very familiar with Rush, who can parse a paragraph about phases, tasks, and operations, they still think colloquially of "build" (as opposed to update, or install, or check, or publish, etc.). I guess my inclination is to use names like `RUSH_BUILD_` to take advantage of that natural shortcut, and explain in documentation that it applies to any bulk or phased command, rather than being "correct" and calling it `RUSH_OPERATION_`, leaving many casual readers totally adrift.

(Let me know what you think -- I suppose one rebuttal is that the only way to get people used to "operations" is by using it all over, but I don't know that we're really making a concerted effort to do that right now.)

libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
iclanton: Does this mean that we don't fold the first operation that runs?

It seems like we should write the header, log stuff for the operation, and then write the footer (and it should appear that way in the code). Shouldn't stream-collator be able to handle that?

libraries/rush-lib/src/api/EnvironmentConfiguration.ts
iclanton: I can't stay I have a particularly strong opinion. @octogonz and @dmichon-msft came up with the name IIRC, so they might have some thoughts.

libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
elliot-nelson: Well, StreamCollator _is_ handling it -- it provides a callback called `onWriterActive` that triggers whenever a new stream starts writing to the collator (including the first stream). We pass this function as that callback, then we just pile all the build log streams into the collator, letting its `onWriterActive` hook write out the fancy header lines as appropriate.

The first stream to write to the collator will trigger this hook, `_lastTask` will be empty so it will skip writing a folding footer, then it will write the header separator, then it will write the folding header.

elliot-nelson:
Incoming changes based on PR feedback:

 - I've changed the `${taskName}` interpolation to be `[taskName]`, to match our existing interpolated values
 - Refactored interpolation to happen in a dedicated method
 - There is now a single env var that controls log folding style (which defaults to off)
 - Your supported log folding styles are configured in `logging.json`

Should be ready for another round.

elliot-nelson:
Thanks for that @octogonz!

So I wanted to test out exactly how it looks and feels to collapse all the insides of each operation, and I am worried that perhaps this change is not worth it.

Check out this log output for an operation that is failing:
Screen Shot 2022-12-12 at 10 46 07 PM

I can see that this operation is failing in the bottom of the build log, with its summary output; but if I scroll up to find it and view the full log, I basically have to search by operation name, because nothing about the output indicates a problem.

🤔 I'm not sure what improvement would be required. Perhaps if the log is folded, and the task ended in warning or error, an additional line _below_ the fold could print the result of the operation so it's visible when quickly scanning the log output?

PR #3836, opened 2022-12-12
[heft-storybook-plugin] support storybook with static build

RongqiZ:
We want to use storybook static build files with the heft.
So I added this plugin into the build actions, and we could pass the static build module path.

This is the heft config after change:
```
{
      "plugin": "@rushstack/heft-storybook-plugin",
      "options": {
        "storykitPackageName": "@storybook/react",
        "startupModulePath": "@storybook/react/bin/index.js",
        "staticBuildModulePath": "@storybook/react/bin/build.js"
      }
    }
```

And this is the command in the package.json
```
"storybook": "heft start --storybook",
```

So that in the local environment, developers could use ```storybook``` command to start the server to see the result.
In the CI, we could use the ``` build-storybook``` command to build static files for visual regressions.

Test scenarios:
- only configured startupModulePath, and run with ``` heft start --storybook ```
- startupModulePath & staticBuildModulePath configured, run with ``` heft start --storybook ```
- startupModulePath & staticBuildModulePath configured, run with ``` heft build --storybook ```

## Impacted documentation
No impact

### DIFF
--- a/apps/rush/src/start-dev.ts
+++ b/apps/rush/src/start-dev.ts
@@ -4,7 +4,7 @@
 // This file is used during development to load the built-in plugins and to bypass
 // some other checks
 
-import * as rushLib from '@microsoft/rush-lib';
+import * as rushLib from '@microsoft/rush-lib/lib/index';
 import { PackageJsonLookup, Import } from '@rushstack/node-core-library';
 
 import { RushCommandSelector } from './RushCommandSelector';
--- a/common/changes/@microsoft/rush/bundle-in-lib_2022-12-13-02-11.json
+++ b/common/changes/@microsoft/rush/bundle-in-lib_2022-12-13-02-11.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "",
+      "type": "none",
+      "packageName": "@microsoft/rush"
+    }
+  ],
+  "packageName": "@microsoft/rush",
+  "email": "iclanton@users.noreply.github.com"
+}
--- a/libraries/rush-lib/.npmignore
+++ b/libraries/rush-lib/.npmignore
@@ -30,6 +30,5 @@
 # (Add your project-specific overrides here)
 !/assets/**
 /lib/__mocks__/**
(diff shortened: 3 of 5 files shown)


heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -117,14 +127,25 @@ export class StorybookPlugin implements IHeftPlugin<IStorybookPluginOptions> {
...
           'The command line does not include "--storybook", so bundling will proceed without Storybook'
         );
         return;
       }
 
+      this._modulePath = storybookStaticBuildParameters.actionAssociated ? options.staticBuildModulePath: options.startupModulePath;
+
       this._logger.terminal.writeVerboseLine(
         'The command line includes "--storybook", redirecting Webpack to Storybook'
       );
```
D4N14L, line 151: The webpack bundle is still disabled when storybook is running in non-start mode. I get the rationale behind it when in start mode (since you'd have a different start mode experience that you're targeting), but is it expected to be skipped when building the static assets?

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -177,17 +198,17 @@ export class StorybookPlugin implements IHeftPlugin<IStorybookPluginOptions> {
...
     }
 
-    this._logger.terminal.writeVerboseLine(`Resolving startupModulePath "${this._startupModulePath}"`);
+    this._logger.terminal.writeVerboseLine(`Resolving startupModulePath "${this._modulePath}"`);
```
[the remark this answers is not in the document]
RongqiZ: updated

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
RongqiZ: Make sense, updated

RongqiZ:
> 

@microsoft-github-policy-service agree [company="Microsoft"]

PR #3843, opened 2022-12-14
[heft-storybook] Support storybook static build

RongqiZ:
We want to use storybook static build files with the heft.
So I added this plugin into the build actions, and we could pass the static build module path.

This is the heft config after change:
```
{
      "plugin": "@rushstack/heft-storybook-plugin",
      "options": {
        "storykitPackageName": "@storybook/react",
        "startupModulePath": "@storybook/react/bin/index.js",
        "staticBuildModulePath": "@storybook/react/bin/build.js"
      }
    }
```

And this is the command in the package.json
```
"storybook": "heft start --storybook",
```

So that in the local environment, developers could use ```storybook``` command to start the server to see the result.
In the CI, we could use the ``` build-storybook``` command to build static files for visual regressions.

Test scenarios:
- only configured startupModulePath, and run with ``` heft start --storybook ```
- startupModulePath & staticBuildModulePath configured, run with ``` heft start --storybook ```
- startupModulePath & staticBuildModulePath configured, run with ``` heft build --storybook ```


heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -139,8 +163,8 @@ export class StorybookPlugin implements IHeftPlugin<IStorybookPluginOptions> {
         bundle.hooks.configureWebpack.tap(
           { name: PLUGIN_NAME, stage: Number.MAX_SAFE_INTEGER },
           (webpackConfiguration: unknown) => {
-            // Discard Webpack's configuration to prevent Webpack from running
-            return null;
+            // Discard Webpack's configuration to prevent Webpack from running only when starting a storybook server
+            return storybookStaticBuildParameters.actionAssociated ? webpackConfiguration : null;
```
D4N14L, line 167: Alternately, I would return `null` if `storybookParameters.actionAssociated && storybookParameters.value` is true, since we would only want to take over from the normal bundler if in start mode AND `--storybook` is specified.

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -117,14 +127,28 @@ export class StorybookPlugin implements IHeftPlugin<IStorybookPluginOptions> {
...
+        (!storybookParameters.actionAssociated || !storybookParameters.value) &&
+        (!storybookStaticBuildParameters.actionAssociated || !storybookStaticBuildParameters.value)
+      ) {
         this._logger.terminal.writeVerboseLine(
           'The command line does not include "--storybook", so bundling will proceed without Storybook'
         );
         return;
       }
 
+      this._modulePath = storybookStaticBuildParameters.actionAssociated
```
D4N14L, line 148: If these options are made "optional", then we should throw if still undefined here, since the modulePath would be needed to run.

D4N14L (changes requested):
@octogonz does this make sense to you? The idea is to be able to produce storybook assets alongside the webpack assets during a build, but not during start mode.

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -117,14 +127,28 @@ export class StorybookPlugin implements IHeftPlugin<IStorybookPluginOptions> {
         '(EXPERIMENTAL) Used by the "@rushstack/heft-storybook-plugin" package to launch Storybook.'
     });
 
+    const storybookStaticBuildParameters: IHeftFlagParameter = heftSession.commandLine.registerFlagParameter({
```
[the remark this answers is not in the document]
RongqiZ: updated, but the tutorial itself has some issue
https://github.com/microsoft/rushstack/issues/3676

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -55,15 +55,26 @@ export interface IStorybookPluginOptions {
...
+  /**
+   * The module entry point that Heft `build` command should use to launch the Storybook toolchain.
+   * Typically it is the path loaded the `build-storybook` shell script.
+   *
+   * @example
+   * If you are using `@storybook/react`, then the startup path would be:
+   *
+   * `"startupModulePath": "@storybook/react/bin/build.js"`
+   */
+  staticBuildModulePath: string;
```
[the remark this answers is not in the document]
RongqiZ: Updated, I make the two fields optional, but cannot be empty together.

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
RongqiZ: Updated, thanks!

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
RongqiZ: Good catch, updated

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -23,6 +24,7 @@ import { StorybookRunner } from './StorybookRunner';
 
 const PLUGIN_NAME: string = 'StorybookPlugin';
 const TASK_NAME: string = 'heft-storybook';
+const PLUGIN_SCHEMA_PATH: string = path.resolve(__dirname, 'schemas', 'heft-storybook-plugin.schema.json');
```
iclanton, line 27:
```suggestion
const PLUGIN_SCHEMA_PATH: string = `${__dirname}/schemas/heft-storybook-plugin.schema.json`;
```

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -117,14 +130,35 @@ export class StorybookPlugin implements IHeftPlugin<IStorybookPluginOptions> {
         '(EXPERIMENTAL) Used by the "@rushstack/heft-storybook-plugin" package to launch Storybook.'
     });
 
+    const storybookStaticBuildParameters: IHeftFlagParameter = heftSession.commandLine.registerFlagParameter({
```
iclanton, line 133: Is this separate from `storybookParameters` just because the documentation is different between the two? Seems like we could update the phrasing in the docs and deduplicate these.

D4N14L: Yeah I like that approach better than having separate parameters

heft-plugins/heft-storybook-plugin/src/schemas/heft-storybook-plugin.schema.json
```diff
@@ -0,0 +1,25 @@
...
+      "title": "Specifies an NPM package that will provide the Storybook dependencies for the project.",
+      "description": "Storybook's conventional approach is for your app project to have direct dependencies\non NPM packages such as `@storybook/react` and `@storybook/addon-essentials`.  These packages have\nheavyweight dependencies such as Babel, Webpack, and the associated loaders and plugins needed to\nbuild the Storybook app (which is bundled completely independently from Heft).  Naively adding these\ndependencies to your app's package.json muddies the waters of two radically different toolchains,\nand is likely to lead to dependency conflicts, for example if Heft installs Webpack 5 but\n`@storybook/react` installs Webpack 4. To solve this problem, `heft-storybook-plugin` introduces the concept of a separate\n\"storykit package\".  All of your Storybook NPM packages are moved to be dependencies of the\nstorykit.  Storybook's browser API unfortunately isn't separated into dedicated NPM packages,\nbut instead is exported by the Node.js toolchain packages such as `@storybook/react`.  For\nan even cleaner separation the storykit package can simply reexport such APIs.",
```
D4N14L, line 11: Trim out the `\n` chars. They aren't needed in the description

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
D4N14L: Actually, are action names available in legacy Heft? If they aren't, then the only way to differentiate between the two scenarios would be the separate parameters.

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -117,14 +130,35 @@ export class StorybookPlugin implements IHeftPlugin<IStorybookPluginOptions> {
...
         this._logger.terminal.writeVerboseLine(
           'The command line does not include "--storybook", so bundling will proceed without Storybook'
         );
         return;
       }
 
+      const modulePath: string | undefined = storybookStaticBuildParameters.actionAssociated
+        ? options.staticBuildModulePath
+        : options.startupModulePath;
+      if (!modulePath) {
```
D4N14L, line 154: This check is now duplicated, since you are already checking on line 119

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
RongqiZ: Do you know where to get the action name? I searched in the heftSession, but I did not find one about the actionName, so I added another parameter.

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
RongqiZ: This check for if the user only specify `startupModulePath` in the heft.json, but run it with `heft build --storybook` command. Then it will pass the check on line 119, but no static build path.

heft-plugins/heft-storybook-plugin/src/schemas/heft-storybook-plugin.schema.json
RongqiZ: Updated, BTW will the schema json generated to some docs?

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
iclanton: You just need to add `'build'` to the `associatedActionNames` array in the other parameter, and rephrase the documentation.

D4N14L: Yeah, but the behaviour is different depending on the action. How would you differentiate between which action was called?

heft-plugins/heft-storybook-plugin/src/schemas/heft-storybook-plugin.schema.json
D4N14L: No these will not. I believe @iclanton investigated converting schemas to docs in the past, it's a non-trivial task. Usually schemas can provide VSCode tooltips when hovering over properties, however that's only when using the root level `$schema` field to link to the schema, which isn't possible in Heft plugin options. The documentation here is mostly for developer reference.

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
D4N14L: Ah yes, you're right

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
iclanton: You can look at the `build.properties.serveMode` property to switch on behavior.

heft-plugins/heft-storybook-plugin/src/schemas/heft-storybook-plugin.schema.json
iclanton: You can also use it to validate the options object. This PR isn't doing that, but I figured that could just get integrated into the Heft vNext version.

iclanton:
@RongqiZ - Can you sign the CLA? https://github.com/microsoft/rushstack/pull/3843#issuecomment-1350776602

RongqiZ:
>

RongqiZ:
> @RongqiZ please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> 

@microsoft-github-policy-service agree [company="Microsoft"]

RongqiZ:
> @RongqiZ - Can you sign the CLA? [#3843 (comment)](https://github.com/microsoft/rushstack/pull/3843#issuecomment-1350776602)

Sure, updated

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
RongqiZ: updated

RongqiZ:
> @microsoft-github-policy-service agree [company="Microsoft"]

@microsoft-github-policy-service agree [company="Microsoft"]

RongqiZ:
> @microsoft-github-policy-service agree company="your company"

@microsoft-github-policy-service agree company="Microsoft"

elliot-nelson:
Very interesting! Where exactly does the static site get built in this model?

The way we have been structuring this internally is usually something like:

 - apps/dynamite depends on libraries/dynamite-ui (which is a storybook)
 - sandbox/dynamite-storybook depends on libraries/dynamite-ui
 - libraries/dynamite-ui depends on libraries/dynamite-ui-storykit (for babel etc)

In this structure, the dynamite app actually uses the storybook UI and has its own deployment scripts (terraform or whatever), and the sandbox project also has its own deployment scripts, so that the storybook site can be deployed somewhere QA and other teams can play with it.

So the "storybook build" doesn't happen in the UI library, it happens in a separate project with its own dist folder.

I'm wondering if this PR would be compatible with this approach, so we could cut out of some of our custom scripts.

common/reviews/api/heft-storybook-plugin.api.md
```diff
@@ -12,9 +12,10 @@ export default _default;
 
 // @public
 export interface IStorybookPluginOptions {
-    startupModulePath?: string;
-    staticBuildModulePath?: string;
-    storykitPackageName: string;
+  options?: string[];
```
djyde, line 15: I prefer name it `cliOptions` instead

heft-plugins/heft-storybook-plugin/src/StorybookRunner.ts
```diff
@@ -29,6 +30,10 @@ export class StorybookRunner extends SubprocessRunnerBase<IStorybookRunnerConfig
 
     terminal.writeLine('Launching ' + this._configuration.resolvedStartupModulePath);
 
+    if (this._configuration.options) {
+      process.argv.push(...this._configuration.options);
```
D4N14L, line 34: Hmm... I like the idea behind this, but I'm a bit worried about modifying the process.argv since that may be used by other plugins/functionality while running. Since it uses the subprocess runner, it's not really a concern in this version of Heft since it only modifies the subprocess's argv. However, the subprocess runner was removed in the in-development version of Heft and that means that storybook would run in-process, and any modifications to the argv could affect other plugins/functions that directly access the `process.argv`. This is also true when `heft --debug` is used, since the subprocess runner requires the module in-process there as well.

The fact that this module is blindly looking at the argv for this parameter (even though the command that is being run isn't applicable) is an indication that it's not designed very well to begin with. Ideally, we would require the module, then call some exported function to execute the main functionality.

It may be best to leave the path as-is for now.

djyde: what about we do a teardown after requiring the startupModule?

```
// cache the original process.argv
const originalArgv = [...process.argv]

process.argv.push(...this._configuration. Options);
require(this._configuration.resolvedStartupModulePath);
terminal.writeVerboseLine('Completed synchronous portion of launching startupModulePath');

// teardown
process.argv = originalArgv
```

D4N14L: Yes, good point. It's a synchronous require call, no other processes should be affected.

common/reviews/api/heft-storybook-plugin.api.md
D4N14L: I think making the options provided to Heft as explicit as possible is the best route IMO... we could use an option named `outputPath` and only append the applicable parameter to the `process.argv`

heft-plugins/heft-storybook-plugin/src/StorybookRunner.ts
```diff
@@ -29,6 +30,12 @@ export class StorybookRunner extends SubprocessRunnerBase<IStorybookRunnerConfig
 
     terminal.writeLine('Launching ' + this._configuration.resolvedStartupModulePath);
 
+    const originalArgv: string[] = process.argv;
+    if (this._configuration.staticBuildOutputDir) {
+      process.argv.push(`--output-dir=${this._configuration.staticBuildOutputDir}`);
```
D4N14L, line 35: It looks like this will get passed regardless of whether or not the static build module was specified, since the options coalesce into the `resolvedStartupModulePath` option. Does this impact the output folder when specifying `startupModulePath` and running in `--serve` mode? If so, the option is a bit oddly named, and should probably be generalized to `outputDir` or `outputPath`.

D4N14L: I pulled your branch and tried it out. Using the setting actually breaks the `--serve` mode flow:
```
---- Bundle started ---- 
[heft-storybook] Starting Storybook runner...
[heft-storybook-runnner] Launching D:\Work\rushstack2\common\temp\node_modules\.pnpm\@storybook+react@6.4.22_6pxnn2xtr5yph4wackblrq76km\node_modules\@storybook\react\bin\index.js
error: unknown option '--output-dir=dist-storybook'
 ---- Bundle continuing with errors (1648ms) ----
-------------------- Finished (7.461s) --------------------
Project: heft-storybook-react-tutorial@1.0.0
Heft version: 0.48.9
Node version: v14.19.1
Error: Subprocess exited before sending "exit" message.
```

You will need to only add this option when running the static build flow.

RongqiZ: updated, thanks for pointing it out.

D4N14L (approved):

heft-plugins/heft-storybook-plugin/src/StorybookPlugin.ts
```diff
@@ -55,25 +57,49 @@ export interface IStorybookPluginOptions {
...
+  /**
+   * The customized output dir for storybook static build.
+   * If this is empty, then it will use the storybook default output dir.
+   *
+   * @example
+   * If you want to change the static build output dir to staticBuildDir, then the static build output dir would be:
+   *
+   * `"staticBuildOutputDir": "newStaticBuildDir"`
+   */
+  staticBuildOutputDir?: string;
```
iclanton, line 90: Can you change this property name to `staticBuildOutputFolder`? We generally prefer the term "folder" and try to avoid using abbreviations in APIs.

RongqiZ: renamed to staticBuildOutputFolder

iclanton (approved):

PR #3853, opened 2022-12-16
[lockfile-explorer] Cleanup and refactor in lockfile-explorer.

iclanton:
## Summary

- Cleaned up the state API a bit
- Converted the HTTP calls to async to not block the server's main thread
- Replaced `fs` with `FileSystem` and `JsonFile`
- Cleaned up some duplicated work in `init()`
- Removed unnecessary path resolutions

## How it was tested

Tested by running the application locally. It behaves the same as it did before.


elliot-nelson:
Pulled and tested branch on local monorepo ✅

william2958 (approved):

PR #3864, opened 2022-12-21
[rush] Improve the experience during a rush operation when the cached credentials have expired.

iclanton:
## Summary

Currently, if a user has expired build cache credentials and attempts run a cached operation in a Rush repo, the operation will error. This change instead prints a warning and attempts to continue without credentials.

## Details

See the example screenshots below from user with expired credentials for a repo with cache backed by Azure Storage that allows anonymous access.

Before:
image

After:
image

## How it was tested

Tested in the above mentioned repo.


TheLarkInn (approved):
LGTM

PR #3865, opened 2022-12-22
[rush] rush add --peer

### LINKED ISSUE #1456: [rush] Feature: Enable "rush add" to install peer dependencies
Hello!

Thank you for this great tool and your hard work!

However, I thought that it would be nice to be able to install a peer dependency using `rush add` command. It could be implemented as a `--peer` flag for example.

Thanks!

elliot-nelson:
## Summary

Add `--peer` flag for `rush add`, for adding peer dependencies.

Fixes #1456.

## Details

 - Extend existing functionality to also include peer dependencies.

## How it was tested

In local monorepo:

 - Confirmed 'rush add --peer' can add a peer dependency
 - Confirmed specifying '--dev --peer' is not allowed
 - Confirmed 'rush add --peer' on a dep used in other packages correctly selects existing version
 - Confirmed 'rush remove' correctly removes a peer dep (and deletes peerDependencies section if appropriate)

## Impacted documentation

 - https://github.com/microsoft/rushstack-websites/blob/main/websites/rushjs.io/docs/pages/commands/rush_add.md


common/changes/@microsoft/rush/enelson-peer_2022-12-22-13-56.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "The 'rush add' command now supports '--peer' flag",
```
iclanton, line 5:
```suggestion
      "comment": "Add a `--peer` flag to the `rush add` command to add the specified dependency as a `peerDependency`.",
```

libraries/rush-lib/src/cli/actions/AddAction.ts
```diff
@@ -90,6 +96,12 @@ export class AddAction extends BaseAddAndRemoveAction {
       );
     }
 
+    if (this._devDependencyFlag.value && this._peerDependencyFlag.value) {
+      throw new Error(
```
iclanton, line 100: Usually you add a dependency as both a `peerDependency` and as a `devDependency` so it's available during development. Seems natural to support both flags.

libraries/rush-lib/src/logic/PackageJsonUpdaterTypes.ts
```diff
@@ -64,9 +64,13 @@ export interface IPackageJsonUpdaterRushBaseUpdateOptions {
  */
 export interface IPackageJsonUpdaterRushAddOptions extends IPackageJsonUpdaterRushBaseUpdateOptions {
   /**
-   * Whether or not this dependency should be added as a devDependency or a regular dependency.
+   * Whether or not this dependency should be added as a devDependency, instead of a regular dependency.
```
iclanton, line 67:
```suggestion
   * Whether or not this dependency should be added as a devDependency instead of as a regular dependency.
```

libraries/rush-lib/src/logic/PackageJsonUpdaterTypes.ts
```diff
@@ -64,9 +64,13 @@ export interface IPackageJsonUpdaterRushBaseUpdateOptions {
...
    */
   devDependency: boolean;
+  /**
+   * Whether or not this dependency should be added as a peerDependency, instead of a regular dependency.
```
iclanton, line 71:
```suggestion
   * Whether or not this dependency should be added as a peerDependency instead of as a regular dependency.
```

iclanton (approved):

libraries/rush-lib/src/cli/actions/AddAction.ts
elliot-nelson: Now that you say it, that does make sense... either flag turns off "regular" dependency, but you can specify both.

Faithfinder: There is a small case that peer and dev dependency often would have different version specified - e.g. peer is `^17 || ^18`, but dev is `~18.0.3`, but in the context of monorepo I'd take the convenience of adding in one command.

PR #3873, opened 2023-01-05
[rush] Create custom git hooks instead of copy

### LINKED ISSUE #3816: [rush] rush git-hooks edge case
## Summary

Background: our team is migrating husky to rush git-hooks, and we found an edge case with current rush git hooks implementation. So, this PR reports the edge case and proposes a alternative git hooks implementation.

## Repro steps

Assuming the repo maintainer wants to add a "rush prettier" to "pre-commit" git hooks. So the maintainer follows the official doc:
1. Add a `rush prettier` custom command
2. Use `rush prettier` in `common/git-hooks/pre-commit`.

Ok. Push and merge it to the main branch. Everything seems fine now.

Actually, users may run into an issue that the following issue throws:

```
rush: error: argument "<command>": Invalid choice: prettier (choose from ....
```

## Details

Let's say we got the following git branches with their commit history:

1. branch `main`: A -> B(rush prettier + pre-commit)
2. branch `feat-1`: B -> C (checkout from main when the latest commit is B)
3. branch `feat-2`: A -> D (checkout from main when the latest commit is A)

And B is the commit for `rush prettier` in pre-commit.

Now. Users checkout to `feat-1`, rush update to sync the git hooks to `.git` folder and correctly run `pre-commit` with `rush prettier`
And then, checkout to `feat-2`, modify a readme file, and run `git commit -m 'xxx'`, which triggers `pre-commit` hooks. BOOM, there is no `rush prettier` defined in branch `feat-2`.

### Possible solution

Instead of syncing files from `common/git-hooks` to `.git/hooks`, how about setting `core` by running `git config core.hooksPath common/git-hooks`

Similar logic in Husky at https://github.com/typicode/husky/blob/9efb720c165ffcfec774b4584aa61f59dae774c7/src/index.ts#L49

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? |  5.86.0 |
| `rushVersion` from rush.json? | 5.86.0 |
| Operating system? | Mac |
| Node.js version (`node -v`)? | 14 |

stekycz:
## Summary

Fixes #3816

## Details

I change copy of the script to create a new script from inlined template. The template might be another file but I chose simple solution now because the script is very simple. I can move it to a template file if needed but I would need guidance to where the file should be placed and what templating should be leveraged to hardcode specific values.

I am trying to keep the script universal to work even the directory of the repo is renamed. That is why I do not use absolute paths and I detect the script directory.

## How it was tested

I am not sure how to test these changes from the branch. However, I use the pattern of executing another script in one of my private repos.

## Impacted documentation

The [Git Hooks](https://rushjs.io/pages/maintainer/git_hooks/) documentation might need an update.

### DIFF
--- a/common/changes/@microsoft/rush/stekycz-git-hooks-exec_2023-01-05-10-36.json
+++ b/common/changes/@microsoft/rush/stekycz-git-hooks-exec_2023-01-05-10-36.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "[rush] Create custom git hooks instead of copy",
+      "comment": "Generate scripts in the Git hooks folder referring to the actual hook implementations in-place in the Rush `common/git-hooks/` folder instead of copying the scripts to the Git hooks folder.",
       "type": "none"
     }
   ],
--- a/common/changes/@microsoft/rush/stekycz-git-hooks-exec_2023-01-05-10-36.json
+++ b/common/changes/@microsoft/rush/stekycz-git-hooks-exec_2023-01-05-10-36.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "[rush] Create custom git hooks instead of copy",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/libraries/rush-lib/src/logic/base/BaseInstallManager.ts
+++ b/libraries/rush-lib/src/logic/base/BaseInstallManager.ts
@@ -449,11 +449,22 @@ export abstract class BaseInstallManager {
         // Clear the currently installed git hooks and install fresh copies
         FileSystem.ensureEmptyFolder(hookDestination);
 
+        // Find the relative path from Git hooks directory to the directory storing the actual scripts.
+        const hookRelativePath: string = path.relative(hookDestination, hookSource);
+
         // Only copy files that look like Git hook names
         const filteredHookFilenames: string[] = hookFilenames.filter((x) => /^[a-z\-]+/.test(x));
         for (const filename of filteredHookFilenames) {
-          // Copy the file.  Important: For Bash scripts, the EOL must not be CRLF.
-          const hookFileContent: string = FileSystem.readFile(path.join(hookSource, filename));
+          const hookFileContent: string = `#!/bin/bash
+SCRIPT_DIR="$( cd "$( dirname "\${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+SCRIPT_IMPLEMENTATION_PATH="$SCRIPT_DIR/${hookRelativePath}/${filename}"
+if [[ -f "$SCRIPT_IMPLEMENTATION_PATH" ]]; then
+  exec "$SCRIPT_IMPLEMENTATION_PATH"
+else
+  echo "The ${filename} Git hook does not exist in your version of the repo. Either remove the ${filename} hook file (e.g. '.git/hooks/${filename}') in case this feature is not used anymore in main branch. Otherwise, updating your branch to current mainstream will solve the issue."
+fi
+`;
+          // Create the hook file.  Important: For Bash scripts, the EOL must not be CRLF.
           FileSystem.writeFile(path.join(hookDestination, filename), hookFileContent, {
             convertLineEndings: NewlineKind.Lf
           });
--- a/libraries/rush-lib/src/logic/base/BaseInstallManager.ts
+++ b/libraries/rush-lib/src/logic/base/BaseInstallManager.ts
@@ -462,7 +462,7 @@ SCRIPT_IMPLEMENTATION_PATH="$SCRIPT_DIR/${hookRelativePath}/${filename}"
 if [[ -f "$SCRIPT_IMPLEMENTATION_PATH" ]]; then
   exec "$SCRIPT_IMPLEMENTATION_PATH"
 else
-  echo "The ${filename} Git hook no longer exists in your version of the repo. Run 'rush install' or 'rush update' to refresh your installed Git hooks."
+  echo "The ${filename} Git hook no longer exists in your version of the repo. Run 'rush install' or 'rush update' to refresh your installed Git hooks." >&2
 fi
 `;
           // Create the hook file.  Important: For Bash scripts, the EOL must not be CRLF.
--- a/libraries/rush-lib/src/logic/base/BaseInstallManager.ts
+++ b/libraries/rush-lib/src/logic/base/BaseInstallManager.ts
@@ -16,7 +16,8 @@ import {
   FileSystemStats,
   ConsoleTerminalProvider,
   Terminal,
-  ITerminalProvider
+  ITerminalProvider,
+  Path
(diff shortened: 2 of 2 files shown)


dmichon-msft:
For testing, the easiest way is to, given another rush repository with a custom Git hook defined, you can invoke your local build of Rush by directly executing `node <path/to/rushstack>/apps/rush/lib/start-dev.js install`, then test the relevant hook via the appropriate Git command.

dmichon-msft:
Looks fine, just want to make sure it is validated first.

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -449,11 +449,22 @@ export abstract class BaseInstallManager {
...
-          const hookFileContent: string = FileSystem.readFile(path.join(hookSource, filename));
+          const hookFileContent: string = `#!/bin/bash
+SCRIPT_DIR="$( cd "$( dirname "\${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+SCRIPT_IMPLEMENTATION_PATH="$SCRIPT_DIR/${hookRelativePath}/${filename}"
+if [[ -f "$SCRIPT_IMPLEMENTATION_PATH" ]]; then
+  exec "$SCRIPT_IMPLEMENTATION_PATH"
+else
+  echo "The ${filename} Git hook does not exist in your version of the repo. Either remove the ${filename} hook file (e.g. '.git/hooks/${filename}') in case this feature is not used anymore in main branch. Otherwise, updating your branch to current mainstream will solve the issue."
+fi
+`;
```
Faithfinder, line 466: Is this portable to Windows?

stekycz:
> For testing, the easiest way is to, given another rush repository with a custom Git hook defined, you can invoke your local build of Rush by directly executing `node <path/to/rushstack>/apps/rush/lib/start-dev.js install`, then test the relevant hook via the appropriate Git command.

Thanks! Will do and post the result here.

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
stekycz: That is a good! Unfortunately, I do not have an opportunity to test it in Windows. Does anyone know how Git hooks are being resolved on Windows? Does it depend on Git installation? How about IDEs?

dmichon-msft: On Windows the hooks run in Git Bash; the hooks will always be bash scripts. Only consideration is that I would recommend normalizing the directory separator to `/` in the relative path (you can use `Path.convertToSlashes()` from `@rushstack/node-core-library` on the result).

stekycz: Thanks, I have just updated it. Please, take a look

stekycz:
@dmichon-msft I have just tested it locally and the installation of Git hooks works as well as they work on my Mac. However, I did not test it on Windows or Linux.

dmichon-msft (approved):
lgtm

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -449,11 +450,22 @@ export abstract class BaseInstallManager {
...
         for (const filename of filteredHookFilenames) {
-          // Copy the file.  Important: For Bash scripts, the EOL must not be CRLF.
-          const hookFileContent: string = FileSystem.readFile(path.join(hookSource, filename));
+          const hookFileContent: string = `#!/bin/bash
+SCRIPT_DIR="$( cd "$( dirname "\${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+SCRIPT_IMPLEMENTATION_PATH="$SCRIPT_DIR/${hookRelativePath}/${filename}"
+if [[ -f "$SCRIPT_IMPLEMENTATION_PATH" ]]; then
+  exec "$SCRIPT_IMPLEMENTATION_PATH"
+else
+  echo "The ${filename} Git hook does not exist in your version of the repo. Either remove the ${filename} hook file (e.g. '.git/hooks/${filename}') in case this feature is not used anymore in main branch. Otherwise, updating your branch to current mainstream will solve the issue."
```
iclanton, line 465:
```suggestion
  echo "The ${filename} Git hook no longer exists in your version of the repo. Run 'rush install' or 'rush update' to refresh your installed Git hooks."
```

iclanton: This should also probably return a nonzero exit code.

common/changes/@microsoft/rush/stekycz-git-hooks-exec_2023-01-05-10-36.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "[rush] Create custom git hooks instead of copy",
```
iclanton, line 5:
```suggestion
      "comment": "Generate scripts in the Git hooks folder referring to the actual hook implementations in-place in the Rush `common/git-hooks/` folder instead of copying the scripts to the Git hooks folder.",
```

iclanton (approved):

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
stekycz: Why would you return a nonzero exit code? 🤔

It would stop git commands for pre-* hooks but would just go away with post-* hooks. I believe it does not matter that much if these files will remain in the directory for some period of time. What are the down sides of having these files in the directory and getting this message in console?

What are the use cases to remain there for a long time? I can see the following:

1. Changing a branch without running `rush install` or `rush update`. I do not think this is much likely to happen.
2. Removing Rush.js from monorepo completely. I guess this is even less likely to happen.

Am I missing something?

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -449,11 +450,22 @@ export abstract class BaseInstallManager {
...
         for (const filename of filteredHookFilenames) {
-          // Copy the file.  Important: For Bash scripts, the EOL must not be CRLF.
-          const hookFileContent: string = FileSystem.readFile(path.join(hookSource, filename));
+          const hookFileContent: string = `#!/bin/bash
+SCRIPT_DIR="$( cd "$( dirname "\${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+SCRIPT_IMPLEMENTATION_PATH="$SCRIPT_DIR/${hookRelativePath}/${filename}"
+if [[ -f "$SCRIPT_IMPLEMENTATION_PATH" ]]; then
+  exec "$SCRIPT_IMPLEMENTATION_PATH"
+else
+  echo "The ${filename} Git hook no longer exists in your version of the repo. Run 'rush install' or 'rush update' to refresh your installed Git hooks."
```
stekycz, line 465: Redirect the warning message to stderr instead of stdout.
```suggestion
  echo "The ${filename} Git hook no longer exists in your version of the repo. Run 'rush install' or 'rush update' to refresh your installed Git hooks." >&2
```

dmichon-msft (approved):
lgtm

PR #3879, opened 2023-01-09
[api-extractor] Introduce sort modifier to api extractor and dts rollup to avoid spurious reports between builds

ajithAPER:
**Summary**
Fixed https://github.com/microsoft/rushstack/issues/1958

**Details**
API reports and DTS rollouts generated show signs of spurious inferred union types between incremental build. This implementation simply introduces a sort modifier to sort all union type in DTS rollups and API reports.

**How was it tested**
Tested against reports within the repository generated automatically.

### DIFF
--- a/common/changes/@microsoft/api-extractor/api-extractor-sort-modifier-for-union-spans-in-generators_2023-01-09-04-30.json
+++ b/common/changes/@microsoft/api-extractor/api-extractor-sort-modifier-for-union-spans-in-generators_2023-01-09-04-30.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor",
+      "comment": "Introduces sort modifier in ApiReportGenerator and DtsRollupGenerator for union types to avoid spurious outputs between builds",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@microsoft/api-extractor"
+}
--- a/apps/api-extractor/src/analyzer/Span.ts
+++ b/apps/api-extractor/src/analyzer/Span.ts
@@ -77,7 +77,7 @@ export class SpanModification {
   /**
    * Used if the parent span has Span.sortChildren=true.
    */
-  public sortKey: string | undefined;
+  public sortKey: number | string | undefined;
 
   /**
    * Optionally configures getModifiedText() to search for a "/*" doc comment and indent it.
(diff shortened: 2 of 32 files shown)


ajithAPER:
This pull request cleans up the additional changes noted in https://github.com/microsoft/rushstack/pull/3698 and reintriduces the original change.

common/changes/@microsoft/api-extractor/api-extractor-sort-modifier-for-union-spans-in-generators_2023-01-09-04-30.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor",
+      "comment": "Introduces sort modifier in ApiReportGenerator and DtsRollupGenerator for union types to avoid spurious outputs between builds",
```
iclanton, line 5:
```suggestion
      "comment": "Make the ordering of elements in union types in API reports and .D.TS rollups always be in alphabetical order.",
```

common/changes/@microsoft/api-extractor/api-extractor-sort-modifier-for-union-spans-in-generators_2023-01-09-04-30.json
```diff
@@ -0,0 +1,10 @@
...
+      "type": "patch"
```
iclanton, line 6:
```suggestion
      "type": "minor"
```

iclanton:
As mentioned in the original issue, it would be nice to preserve the order of elements in explicitly defined union types and only apply the sorting to inferred types. It looks like this change sorts all union types. How difficult would it be to only sort inferred types? If it's too difficult to do that, can this be behind a configuration option?

Imagine you have a type that looks like this in source:

```ts
type DigitAsWord = 'zero' | 'one' | 'two' | 'three' | 'four' | 'five' | 'six' | 'seven' | 'eight' | 'nine';
```

where the constituents of the type have a natural order. If they're always sorted, this type will look like this in the API report:

```ts
type DigitAsWord = 'eight' | 'five' | 'four' | 'nine' | 'one' | 'seven' | 'six' | 'two' | 'zero';
```

which is harder to review for correctness. Did you notice that `'three'` is missing from between `'six'` and `'two'` in the code snippet above? That would've been much easier to catch if the elements were in the same order as the source code.

PR #3881, opened 2023-01-10
[lockfile-explorer] Introduce components library

william2958:
## Summary
This PR adds a component library that currently the Lockfile Explorer uses, but can also be used by other rush applications in the future. This library is still under development but contains all the necessary components for use in the Lockfile Explorer.

## Details
Currently the @rushstack/components library includes the following components:
- Text
- Button
- Input
- Checkbox
- ScrollArea (This element allows for a uniform experience when handling scroll areas, providing a standard scrollbar across all systems)
- Tabs

## How it was tested
Tested manually


apps/lockfile-explorer-web/src/containers/SelectedEntryPreview/styles.scss
```diff
@@ -1,15 +1,10 @@
-@import 'office-ui-fabric-core/dist/sass/References';
```
octogonz, line 1: On a smaller responsive window, the `SelectedEntryPreview` can get squished like this:
 
image

...and also the `renderDependencyMetadata()` pane can get squished:

image

Here's a couple ways we might address that:

- Since the `SelectedEntryPreview` and `renderDependencyMetadata()` have a fixed amount of content (unlike the other panes), we could force a fixed height to them, maybe just 100% of their content height
- Maybe could also use a responsive breakpoint, so that if the browser window height is too small, then the middle column scrolls (rather than stretching to fill the window height)

Sometimes these CSS layout behaviors can be tricky to implement, so don't spend too much time on it.  But maybe there's some easy fixes.

apps/lockfile-explorer-web/src/containers/LockfileViewer/index.tsx
```diff
@@ -166,61 +167,55 @@ export const LockfileViewer = (): JSX.Element | ReactNull => {
...
           <div className={styles.filterSection}>
-            <h5>Filters</h5>
-            <div
-              className={styles.filterOption}
-              onClick={changeFilter(
+            <Text type="h5" bold>
+              Filters
+            </Text>
+            <Checkbox
+              label="Must have side-by-side versions"
```
octogonz, line 204: The sizing of the checkboxes is slightly odd if the label gets word-wrapped:
image

Maybe we can set a minimum width?

william2958: Fixed by adding min width

apps/lockfile-explorer-web/src/containers/SelectedEntryPreview/styles.scss
william2958: fixed by removing top and bottom padding as well as adding a min height

iclanton:
This is quite a bit of explicit styling. Is that something that `radix-ui` requires?

william2958:
> This is quite a bit of explicit styling. Is that something that `radix-ui` requires?

Radix UI only provides the functionality behind the components, they are meant to be styled yourself, but they provide example stylings in their documentation which is where most of this styling comes from, see https://www.radix-ui.com/docs/primitives/components/tabs for an example

octogonz (approved):

PR #3884, opened 2023-01-11
Will/lockfile explorer update detector

william2958:
## Summary
This PR adds the ability to automatically detect if the lockfile explorer package is out of date using the `update-notifier` npm package.
See details on the package here: https://www.npmjs.com/package/update-notifier


octogonz (approved):
LGTM.  I tried testing it using a local install with the **package.json** version modified to be an earlier version, but it didn't seem to print anything. Maybe it needs to be really published?  Let's give that a try.

PR #3886, opened 2023-01-11
[rush] Restore log files when cache hit

chengcyber:
## Summary

This PR add a feature: restoring operation log files when cache hit

## Details

Introduce a new "OperationMetadataManager"

1. when a task successful without cache, save the log files into `.rush/temp/operation/<identifier>/all.log` and `.rush/temp/operation/<identifier>/error.log`.
2. when a task successful with cache hit, append the `all.log` file content to current operation log and copy the error log if exists.

## How it was tested

Tested in rushstack and a repo without phase commands.

1. Save the log file
2023-01-12 at 15 19

2. Append the cached log content to current log
2023-01-12 at 15 21

## Impacted documentation

 Nope

### DIFF
--- a/common/changes/@microsoft/rush/feat-build-cache-log_2023-01-11-16-25.json
+++ b/common/changes/@microsoft/rush/feat-build-cache-log_2023-01-11-16-25.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Restore log files when cache hit",
+      "comment": "Include operation log files in the cache, and restore them during cache hits.",
       "type": "none"
     }
   ],
--- a/common/changes/@microsoft/rush/feat-build-cache-log_2023-01-11-16-25.json
+++ b/common/changes/@microsoft/rush/feat-build-cache-log_2023-01-11-16-25.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Restore log files when cache hit",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/common/reviews/api/rush-lib.api.md
+++ b/common/reviews/api/rush-lib.api.md
@@ -389,6 +389,24 @@ export interface IOperationExecutionResult {
     readonly stopwatch: IStopwatchResult;
 }
 
+// @internal (undocumented)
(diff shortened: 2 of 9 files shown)


dmichon-msft (commented):
I appreciate the consolidation into OperationMetadataManager. At some point it will be necessary to finally move a lot of logic out of ShellOperationRunner (which should only own the logic to invoke a subprocess, not any of the caching, logging, etc.), but that's not really a blocker.

chengcyber:
Hi @dmichon-msft , code changes applied. Please take another look~

elliot-nelson:
@chengcyber I was wondering, do I need to have any special config to enable the feature?

I pulled branch and test as follows:

```console
$ rush install
$ node ~/dev/rushstack/apps/rush/lib/start-dev.js test --verbose --to <someproject>

( a bunch of jest output)

$ ls -al <someproject>/.rush/temp/operation/_phase_test/
-rw-r--r--  1 elliot.nelson  staff   47 Feb  4 11:50 state.json
```

No log file seems to be created (so I couldn't test the part where it retrieves it from cache.)

dmichon-msft (approved):

chengcyber:
@elliot-nelson Thanks for the information! Actually, i broke something accidently at that time. Now it has been fixed!
image

common/changes/@microsoft/rush/feat-build-cache-log_2023-01-11-16-25.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Restore log files when cache hit",
```
iclanton, line 5:
```suggestion
      "comment": "Include operation log files in the cache, and restore them during cache hits.",
```

iclanton (approved):

elliot-nelson:
Thanks @chengcyber! It's working great :+1:

(Worth noting that this feature on its own changes nothing about build output -- `rush build --verbose` still shows _this build's_ information, not the contents of `all.log` -- which is what I expected.)

octogonz:
> (Worth noting that this feature on its own changes nothing about build output -- `rush build --verbose` still shows _this build's_ information, not the contents of `all.log` -- which is what I expected.)

@chengcyber I had the opposite expectation -- shouldn't `rush build --verbose` faithfully reproduce the log outputs from a build that was restored from the cache?

PR #3888, opened 2023-01-11
[webpack5-localization-plugin] drop _initializeAndValidateOptions into thisCompilation hook 

### LINKED ISSUE #3887: [webpack5-localization-plugin] TypeError: `Cannot destructure property 'WebpackError' of 'compiler.webpack' as it is undefined.`
## Summary
Upon using `@rushstack/webpack5-localization-plugin` in a webpack 5 project you receive the following error: 

image

`Cannot destructure property 'WebpackError' of 'compiler.webpack' as it is undefined.` It looks like the suspect code is one of two places:

https://github.com/microsoft/rushstack/blob/8387963bb4ab479395018de705d8d98987549236/webpack/webpack5-localization-plugin/src/AssetProcessor.ts#L75 

or

https://github.com/microsoft/rushstack/blob/8387963bb4ab479395018de705d8d98987549236/webpack/webpack5-localization-plugin/src/LocalizationPlugin.ts#L134

I haven't investigated too deeply here but my guess is that either WebpackError doesn't live on the `compiler.webpack` property or somewhere `compiler` or `compilation` might be undefined. To be safe, at minimal cost, we can just import `WebpackError` instead of accessing it from the `webpack` object mid plugin.

TheLarkInn:
## Summary
This PR moves the `_initializeAndValidateOptions` function into the `compiler.hooks.thisCompilation` hook (where by this time `compiler.webpack` will be set).

Fixes #3887

### DIFF
--- a/common/changes/@rushstack/webpack5-localization-plugin/issue-3887_2023-01-11-18-21.json
+++ b/common/changes/@rushstack/webpack5-localization-plugin/issue-3887_2023-01-11-18-21.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/webpack5-localization-plugin",
+      "comment": "Fix errors where `compiler.webpack` was not accessible when using the plugin.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/webpack5-localization-plugin"
+}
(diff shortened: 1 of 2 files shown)


TheLarkInn:
Closing this as the issue was discovered via user error. The code is valid and I was mistaken.

PR #3908, opened 2023-01-19
[rush-lib] react to pnpm v7 local install path breaking change

jeremymeng:
## Summary

In pnpm v7, the format of local installation paths and the hashing algorithm used to shorten paths have been changed.  It caused error described in https://github.com/microsoft/rushstack/issues/3652. This PR adds handling logic for pnpm v7.

## How it was tested

The code is tested using https://github.com/Azure/azure-sdk-for-js repository


iclanton:
Can we just use the `dependency-path` package instead of duplicating its functionality?

jeremymeng:
> Can we just use the `dependency-path` package instead of duplicating its functionality?

@iclanton Good point. I made the change.  As far as I know our repo doesn't have such cases that are impacted by the hasing change so I don't know how to test that.  Also how do I test my PR in our repo?  Previously I was just patching the js file under node_modules.

common/config/rush/browser-approved-packages.json
```diff
@@ -66,6 +70,10 @@
       "name": "redux",
       "allowedCategories": [ "libraries" ]
     },
+    {
+      "name": "rfc4648",
+      "allowedCategories": [ "libraries" ]
+    },
```
iclanton, line 76:
```suggestion
```

iclanton:
You can test this change in your repo by running `node <path-to-rush-lib-in-your-clone-of-rushstack>/lib/start <args>`, or, if you're testing something that uses the caching features, use `node <path-to-rush-in-your-clone-of-rushstack>/lib/start-dev <args>`.

common/config/rush/browser-approved-packages.json
jeremymeng: Fixed

jeremymeng:
> You can test this change in your repo by running `node <path-to-rush-lib-in-your-clone-of-rushstack>/lib/start <args>`, or, if you're testing something that uses the caching features, use `node <path-to-rush-in-your-clone-of-rushstack>/lib/start-dev <args>`.

Thanks a lot!  Tested with our repo on Linux/Windows.

iclanton (approved):

PR #3921, opened 2023-01-25
[rush] Fix an issue where 'rush add' does not update the approved packages files.

iclanton:
## Summary

Right now, if you run `rush add -p <some package that's new to the repo>`, the [`non`]`browser-approved-packages.json` file isn't updated.

## Details

This happens because after the `packageJsonEditor` saves the file, it doesn't update the `packageJson` `RushConfigurationProject` property with the new `package.json` contents. This PR fixes that.

## How it was tested

Added a unit test, and ran `rush add -p @radix-ui/react-dialog` in a repo that isn't already using `@radix-ui/react-dialog`.


octogonz (approved):
LGTM

PR #3924, opened 2023-01-26
[rush] Revert creation of custom git hooks

D4N14L:
## Summary

This reverts changes brought in by https://github.com/microsoft/rushstack/pull/3873

## Details

Reverts the change due to a few issues encountered when attempting to integrate into our own repository:
- Linux-based issues caused by not `chmod -x`-ing the scripts commited to the common/git-hooks, causing permissions errors when running hooks on Linux-based devices
- Issues with compatibility with `git lfs`, causing the error `This should be run through Git's pre-push hook.  Run 'git lfs update' to install it.` even when running through the pre-push hook

It just seems like this change requires a bit more time to bake/test, so for now, going to revert.

CC @stekycz

### DIFF
--- a/common/changes/@microsoft/rush/user-danade-RevertGitHooksCHange_2023-01-26-00-13.json
+++ b/common/changes/@microsoft/rush/user-danade-RevertGitHooksCHange_2023-01-26-00-13.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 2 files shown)


iclanton (approved):

stekycz:
@D4N14L Thank you! I believe I have addressed both issues in https://github.com/microsoft/rushstack/pull/4132

PR #3926, opened 2023-01-26
[module-minifier] Include terser version in config hash

dmichon-msft:
## Summary
Includes the version string from the installed version of `terser` in the config hash for `@rushstack/module-minifier` for the minifier implementations that use `terser`.

## Details
Added to both `LocalMinifier` and `WorkerPoolMinifier`.

## How it was tested
Added unit tests for the config hash strings.

### DIFF
--- a/common/changes/@rushstack/module-minifier/terser-version-hash_2023-01-24-19-28.json
+++ b/common/changes/@rushstack/module-minifier/terser-version-hash_2023-01-24-19-28.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/module-minifier",
+      "comment": "Include terser package version in config hash.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/module-minifier"
+}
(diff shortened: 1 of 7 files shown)


libraries/module-minifier/src/LocalMinifier.ts
```diff
@@ -44,8 +44,11 @@ export class LocalMinifier implements IModuleMinifier {
         : {}
     };
 
+    const { version: terserVersion } = require('terser/package.json');
+
     this._configHash = createHash('sha256')
       .update(LocalMinifier.name, 'utf8')
+      .update(`terser@${terserVersion}`)
       .update(serialize(terserOptions))
       .digest('base64');
```
iclanton, line 53: This is duplicated between the two. Put it in a utlity?

common/changes/@rushstack/module-minifier/terser-version-hash_2023-01-24-19-28.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/module-minifier",
+      "comment": "Include terser package version in config hash.",
```
iclanton, line 5:
```suggestion
      "comment": "Ensure updates to the version of Terser trigger a hash change.",
```

libraries/module-minifier/src/LocalMinifier.ts
dmichon-msft: Doing so doesn't really save much in terms of code footprint, and the two are slightly different. Might revisit later if there's a third one.

iclanton (approved):

PR #3928, opened 2023-01-27
[all] Update @types/node and @types/node-fetch

elliot-nelson:
## Summary

 - Update `@types/node`
 - Update `@types/node-fetch`

## Details

The rushstack repo has specified a minimum of node 14 for a while now, so it makes sense to upgrade `@types/node` to latest 14.x (this enables using node 14+ APIs like `URL`, for example).

We are using `node-fetch` 2.x, so also upgrades `@types/node-fetch` to latest 2.x (includes some additional properties on Request and Response interfaces).

A few other minor updates and tweaks were made to get things compiling again.

## How it was tested

 - rush test


iclanton:
This looks good, but I think we should also move the `@types/node` dependency to be a `peerDependency`. See https://github.com/microsoft/rushstack/pull/3937

elliot-nelson:
> This looks good, but I think we should also move the `@types/node` dependency to be a `peerDependency`. See #3937

Makes sense, I can wait for your PR to merge and follow up.

iclanton (approved):

PR #3944, opened 2023-02-01
Use an internal node build rig for internal, non-cyclic projects and fix caching across the repo.

iclanton:
This will allow us to put some stricter requirements and extra processes for our own projects that we might not yet be ready to ship to everyone.


D4N14L (approved):
Seems like the main changes that are added in this new rig are:
- uploading the `types` to the rig tsconfig
- adding Jest coverage/junit reporting and writing them to the <projectFolder>/coverage directory

Does that sound right?

iclanton:
@D4N14L - Yup, that's it so far.

rushstack-internal-rigs/heft-rushstack-internal-node-rig/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "heft-rushstack-internal-node-rig",
+  "version": "0.1.0",
+  "description": "A rig package for Node.js projects that build using Heft inside the RushStack repository.",
```
octogonz-proxy, line 4: Add a README.md file explaining that this is used locally to manage projects in the Rush Stack repo.  Doing so is a stereotypical practice that we should encourage. Even though your rig project is not published to NPM, it is an example that the community can copy and reuse in their own repos. 

As such, I don't really think of it as "internal" (hidden and subject to change at any time), but more like "local" (customized for this repo specifically).

rushstack-internal-rigs/heft-rushstack-internal-node-rig/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "heft-rushstack-internal-node-rig",
+  "version": "0.1.0",
```
octogonz-proxy, line 3: Unpublished projects don't get version bumps. The version should be something like 0.0.0 or 1.0.0 right?

rushstack-internal-rigs/heft-rushstack-internal-node-rig/profiles/default/config/rush-project.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "@rushstack/heft-node-rig/profiles/default/config/rush-project.json",
+
+  "operationSettings": [
+    {
+      "operationName": "_phase:build",
+      "outputFolderNames": [".heft", "release"]
```
octogonz-proxy, line 7: What is `release`? These projects build into the `dist` folder, right?

And what about `temp` and `lib`?

iclanton: `temp` and `lib` come from the rig that we're extending (`@rushstack/heft-node-rig`). Not sure why I included `release` in there. Removed.

PR #3948, opened 2023-02-03
[rush-serve-plugin] Add workspace-level routing

dmichon-msft:
## Summary
Adds a field `globalRouting` to the plugin config for `rush-serve-plugin`, to enable Rush workspaces to serve assets that are owned by the entire workspace (or plugins) rather than specific projects within said workspace.

## Details
Workspace level routing rules come from the plugin's options file, rather than files in individual projects. Expected use case is if a Rush plugin writes files that it wants served, that the plugin config can be used to specify a location, regardless of what projects are in scope of the Rush command.

## How it was tested
Copied into another repository and tested in place with a workspace config.

## Impacted documentation
Documented in the schema. Will need an init template added eventually.


TheLarkInn (approved):
LGTM

PR #3961, opened 2023-02-08
Feature/introduce rust repo support

TheLarkInn:


rust/hello-world/config/heft.json
```diff
@@ -0,0 +1,11 @@
+/**
```
dmichon-msft, line 1: Since this is a complete override, you'll need to define the clean eventAction yourself.

dmichon-msft: Or add an `extends` clause.

Issue #3965, opened 2023-02-08
[repo] Setup codespaces default for rushstack monorepo


TheLarkInn:
- [ ] Dev Container  
  - [x] Node 16
  - [x] heft, rush globally installed
  - [x] Auto git local config for name w/ conventions? 
  - [ ] Port Forwarding for Lockfile Explorer App?
  - [x] Rust/Cargo Stable Toolchain
  - [x] Pre Rush install (might need bypass policy if git local config not setup)
- [ ] [Readme Badge](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/setting-up-your-repository/adding-a-codespaces-badge) 
- [x] #3969

Resources: https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/setting-up-your-nodejs-project-for-codespaces

PR #3966, opened 2023-02-08
Setup default dev container for rushstack with Rust and Node (v16) toolchains.

TheLarkInn:
Add customized default .devcontainer with rust and node toolchains only. This should considerably speed up the setup of a codespace as the default from github dev container was also including python, ruby, go, java, etc. 

Partially addresses many of the boxes in #3965 

Additionally in this PR the build-test folder lockfile got bumped running a full test build, so I'll also commit that change here since it is a small and inconsequential change.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -17,7 +17,7 @@
   // "forwardPorts": [],
 
   // Use 'postCreateCommand' to run commands after the container is created.
-  "onCreateCommand": "npm install -g @microsoft/rush @rushstack/heft && rush update --bypass-policy",
+  "onCreateCommand": "/bin/bash ./.devcontainer/setup.sh",
 
   // Configure tool-specific properties.
   "customizations": {
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,42 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/typescript-node
+{
+  "name": "Node.js & TypeScript",
+  // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:0-16",
+  "features": {
+    "ghcr.io/devcontainers/features/github-cli:1": {},
+    "ghcr.io/devcontainers/features/rust:1": {},
+    "devwasm.azurecr.io/dev-wasm/dev-wasm-feature/rust-wasi:0": {}
+  },
+
+  // Features to add to the dev container. More info: https://containers.dev/features.
+  // "features": {},
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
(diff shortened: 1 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,42 @@
...
+  },
+
+  // Features to add to the dev container. More info: https://containers.dev/features.
+  // "features": {},
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
+  // "forwardPorts": [],
+
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "onCreateCommand": "npm install -g @microsoft/rush @rushstack/heft && rush update --bypass-policy",
```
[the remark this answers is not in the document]
iclanton:
```suggestion
  "onCreateCommand": "npm install -g @microsoft/rush @rushstack/heft && rush install --bypass-policy",
```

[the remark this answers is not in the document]
iclanton: Can this be updated to point to a BASH script instead of the commands chained with `&&`? Can we also see if there's a way to get at the GitHub username and display name during setup to preemptively set the git `user.name` and `user.email` during intialization?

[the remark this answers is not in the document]
TheLarkInn: I was able to find a github API endpoint which can give me a full user name based on their github username! So I made a tiny node script for doing an https call to grab that data and then programmatically set the git config settings.

.devcontainer/setGitConfigUserName.js
```diff
@@ -0,0 +1,44 @@
...
+const main = (username) => {
+  const requestOptions = buildRequestOptionsForGetUser(username);
+  const request = https.request(requestOptions, (response) => {
+    let data = '';
+    response.on('data', (chunk) => {
+      data += chunk;
+    });
+
+    response.on('end', () => {
+      const user = JSON.parse(data);
```
iclanton, line 29:
```suggestion
  fetch(requestOptions).then((data) => {
    return data.json();
  }).then((user) => {
```

.devcontainer/setGitConfigUserName.js
```diff
@@ -0,0 +1,44 @@
...
+      // Execute git config command
+      console.log(`git config --local user.email "${username}@users.noreply.github.com"`);
+      child_process.execSync(`git config --local user.email "${username}@users.noreply.github.com"`);
+
+      console.log(`git config --local user.name "${name}"`);
+      child_process.execSync(`git config --local user.name "${name}"`);
+    });
+  });
+
+  request.end();
```
iclanton, line 41:
```suggestion
```

.devcontainer/setGitConfigUserName.js
```diff
@@ -0,0 +1,44 @@
+const child_process = require('child_process');
+const https = require('https');
```
iclanton, line 2:
```suggestion
```

iclanton (approved):

.devcontainer/setGitConfigUserName.js
iclanton: Nevermind, not available until node 17.

PR #3968, opened 2023-02-09
Hristovs/got fix

samhristov:
## Summary
Package got was updated from 9.6.0 to 12.5.3

## Details
The got package dependency(9.6.0) has a security vulnerability. 
Using the globalOverrides, forcing got version to 12.5.3 to remove vulnerability

## How it was tested
Manual rush install and build 
<!--------------------------------------------------------------------------
👉 STEP 6: What test cases did you use to validate your work?
     Given the complexities of how build tools interact with the OS, we only
     require unit tests for algorithmic code (e.g. parsing a string, sorting a list).
     Manual testing is fine; you might write something like:

     "Invoked 'rush install' with useWorkspaces=true and useWorkspaces=false
     and confirmed that peer dependencies were handled correctly."

     NOTE: Manual testing should be performed on the *final* commit.
     Pushing additional commits with "small" fixes often invalidates testing.

### DIFF
--- a/common/config/rush/pnpm-config.json
+++ b/common/config/rush/pnpm-config.json
@@ -91,6 +91,7 @@
   "globalOverrides": {
     // "example1": "^1.0.0",
     // "example2": "npm:@company/example2@^1.0.0"
+    "got": "12.5.3"
   },
 
   /**
--- a/build-tests/install-test-workspace/workspace/common/pnpm-lock.yaml
+++ b/build-tests/install-test-workspace/workspace/common/pnpm-lock.yaml
@@ -4,13 +4,13 @@ importers:
 
   rush-lib-test:
     specifiers:
-      '@microsoft/rush-lib': file:microsoft-rush-lib-5.90.0.tgz
+      '@microsoft/rush-lib': file:microsoft-rush-lib-5.90.2.tgz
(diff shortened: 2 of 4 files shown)


common/config/rush/pnpm-config.json
```diff
@@ -91,6 +91,7 @@
   "globalOverrides": {
     // "example1": "^1.0.0",
     // "example2": "npm:@company/example2@^1.0.0"
+    "got": "12.5.3"
```
TheLarkInn, line 94: The current change breaks functionality. It appears that `got` is derriving only from `package-json` in the repo.

```suggestion
      "globalOverrides": {
    // "example1": "^1.0.0",
    // "example2": "npm:@company/example2@^1.0.0"
    // TODO: Remove once https://github.com/dylang/npm-check/issues/499 
    // has been closed and a new version of `npm-check` is published.
    "package-json": "^7"
  }
```

After verifying the lockfile there should be no more reference to `got` less than version 11.

TheLarkInn (changes requested):

TheLarkInn (approved):
LGTM. Thank you for this fix!

D4N14L (approved):

iclanton (approved):

PR #3975, opened 2023-02-14
[webpack] Introduce webpack-embedded-dependencies-plugin

TheLarkInn:
This plugin introduces two new features into rushstack: 

1. `webpack-embedded-dependencies-plugin`
2. A new `Testing` module inside of `webpack-plugin-utilities` that should make it easier to write tests for webpack plugins and loaders.


webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+              license,
+              copyright
+            });
+          }
+
+          const dataToStringify: IEmbeddedDependenciesFile = {
+            embeddedDependencies: packages
+          };
+
+          console.log(dataToStringify);
```
TheLarkInn, line 123: Remove

webpack/webpack-embedded-dependencies-plugin/src/test/WebpackEmbeddedDependenciesPlugin.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      );
+      const embeddedDepAsset = stats
+        ?.toJson({ all: false, assets: true })
+        .assets?.some((asset) => asset.name === 'custom-file-name.json');
+
+      expect(embeddedDepAsset).toBe(true);
+    });
+  });
+
+  switch (fixture) {
```
TheLarkInn, line 59: Fixture specific tests will go here

webpack/webpack-plugin-utilities/src/Testing.ts
```diff
@@ -0,0 +1,49 @@
...
+  const compilerOptions: Configuration = webpackMerge(_defaultWebpackConfig(entry), additionalConfig);
+  const compiler: Compiler = webpack(compilerOptions);
+
+  compiler.outputFileSystem = createFsFromVolume(new Volume());
+  compiler.outputFileSystem.join = path.join.bind(path);
+
+  return new Promise((resolve, reject) => {
+    compiler.run((err, stats) => {
+      if (err) reject(err);
+      if (stats?.hasErrors()) reject(stats?.toJson().errors);
```
dmichon-msft, line 30: If child compilations are a possibility, this will miss errors from said child compilations.

webpack/webpack-plugin-utilities/src/Testing.ts
```diff
@@ -0,0 +1,49 @@
...
+): Promise<(Stats | MultiStats) | undefined> {
+  const compilerOptions: Configuration = webpackMerge(_defaultWebpackConfig(entry), additionalConfig);
+  const compiler: Compiler = webpack(compilerOptions);
+
+  compiler.outputFileSystem = createFsFromVolume(new Volume());
+  compiler.outputFileSystem.join = path.join.bind(path);
+
+  return new Promise((resolve, reject) => {
+    compiler.run((err, stats) => {
+      if (err) reject(err);
```
dmichon-msft, line 29:
```suggestion
      if (err) {
        return reject(err);
      }
```

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+const LICENSE_FILES: string = '@(LICENSE|LICENSE-MIT.txt|LICENSE.md|LICENSE.txt|license)';
+const PLUGIN_NAME: 'EmbeddedDependenciesWebpackPlugin' = 'EmbeddedDependenciesWebpackPlugin';
+
+interface IEmbeddedDependenciesFile {
+  name?: string;
+  version?: string;
+  embeddedDependencies: IPackageData[];
+}
+
+interface IDependency {
```
dmichon-msft, line 24: `IDependencyRecord`? `IDependencyCollection`? `IDependencies`?

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+          }
+
+          return module;
+        }
+      );
+    });
+
+    // Tap into compilation so we can tap into compilation.hooks.processAssets
+    compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tapAsync(
```
dmichon-msft, line 98: Since you're using an async function anyway, prefer `tapPromise`

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  return undefined;
+}
+
+/**
+ * Given a module path, try to parse the module's copyright attribution.
+ */
+async function parseCopyright(modulePath: string): Promise<string | undefined> {
+  // Add copyright info
+  const licenseFile: string[] = await globAsync(path.join(modulePath, LICENSE_FILES));
```
dmichon-msft, line 162: Pass `{ fs: compiler.inputFileSystem }` in the options structure to `globAsync`. Either `fs` or `memfs` implement all the necessary methods.

dmichon-msft: Also, using `path.join` on a glob specifier is not semantically valid, because `\` is a reserved character in a glob. You need to use `cwd: modulePath` in the options instead.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+    if (match) {
+      return match[0];
+    }
+  }
+
+  return undefined;
+}
+
+function parsePackageAuthor(p: IPackageData): string | undefined {
+  return typeof p.author === 'string' ? p.author : p.author ? p.author.name : undefined;
```
dmichon-msft, line 177:
```suggestion
  return typeof p.author === 'string' ? p.author : p.author?.name;
```

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+
+import { FileSystem, Import, LegacyAdapters } from '@rushstack/node-core-library';
+import { Compilation, Compiler, WebpackPluginInstance, sources } from 'webpack';
+
+const glob: typeof import('glob') = Import.lazy('glob', require);
+const globAsync = (pattern: string, options: IOptions = {}): Promise<string[]> => {
+  return LegacyAdapters.convertCallbackToPromise(glob, pattern, options);
+};
+
+const COPYRIGHT_REGEX: RegExp = /^Copyright .*$/m;
```
dmichon-msft, line 14: Put a comment with an example of what this matches?

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+  descriptionFileData?: IPackageData;
+  descriptionFileRoot?: string;
+  relativePath?: string;
+}
+
+interface IWebpackModuleCreateData {
+  resourceResolveData?: IResourceResolveData;
+}
+
+interface IPackageData {
```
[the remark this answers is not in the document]
TheLarkInn: How should I reconcile licenses which appears to be a field not defined in IPackageJson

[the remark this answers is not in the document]
TheLarkInn:
```ts

interface IPackageData extends IPackageJson {
  copyright: string | undefined;
  author?: string | { name?: string };
  licenses?: { type: string; url: string }[];
}

```

I chose this to handle these additional fields. Unless you have an alternative solution.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
TheLarkInn: After running it through these tests, it appears that this RegExp is only successfully capturing matches inside of MIT-like licenses. I will need to expand it to support the other licenses I've provided in fixtures.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
TheLarkInn: > Pass `{ fs: compiler.inputFileSystem }` in the options structure to `globAsync`. Either `fs` or `memfs` implement all the necessary methods.

`glob` doesn't seem to have a `fs` property for its options in its types however I do see the API listed in the documentation. I'll give this a try and see how far I get.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,178 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+import path from 'path';
+import type { IOptions } from 'glob';
+
+import { FileSystem, Import, LegacyAdapters } from '@rushstack/node-core-library';
+import { Compilation, Compiler, WebpackPluginInstance, sources } from 'webpack';
+
+const glob: typeof import('glob') = Import.lazy('glob', require);
```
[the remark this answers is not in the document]
TheLarkInn: Removed glob in favor of `compiler.inputFileSystem.readdir`

webpack/webpack-embedded-dependencies-plugin/src/index.ts
TheLarkInn: > Another note, since you're only looking top-level in the folder, you might be better served throwing out `glob` altogether and just issuing a single call to `compiler.inputFileSystem.readdir` and sifting through the results via a Set of allowed file names.

I chose this path but will definitely want a re-review on the implementation.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
TheLarkInn: Added comments for these regular expressions. They have also been isolated to a separate file and exported so that they are not exposed from the plugin export but still testable.

webpack/webpack-embedded-dependencies-plugin/src/regexpUtils.ts
```diff
@@ -0,0 +1,26 @@
...
+ * should cover the majority of cases that we come across in the wild.
+ *
+ * @remarks The following copyright statements are covered via unit tests to be matched:
+ *
+ * - `'Copyright © 2023 FAKE-PACKAGE-MIT-LICENSE'`
+ * - `'Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>'`
+ * - `'Copyright 2023 Some Licenser Name'`
+ *
+ */
+export const COPYRIGHT_REGEX: RegExp = /Copyright\s(\([cC]\)|©)?\s?\d{4}\s.*$/m;
```
dmichon-msft, line 26: Generally favor `\s+` and `\s*` when matching whitespace in regular expressions, unless `Copyright  2024  Someone` is not supposed to match.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,242 @@
...
+      );
+    });
+
+    // Tap into compilation so we can tap into compilation.hooks.processAssets
+    compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tapAsync(
+        { name: PLUGIN_NAME, stage: Compilation.PROCESS_ASSETS_STAGE_REPORT },
+        async (assets, callback) => {
+          const rawPackages: FlattenedPackageEntry[] = Array.from(thirdPartyPackages).sort((first, next) => {
+            return first[0].localeCompare(next[0]);
```
dmichon-msft, line 95: It's usually more stable to use an ordinal comparison; it seems that the rules on comparing shorter vs. longer strings vary across geos (i.e. if `webpack-sources` is before or after `webpack`.

dmichon-msft: Also, see this regarding using `localeCompare` in sort functions: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#performance

dmichon-msft: I recommend using the `<` and `>` operators to define your sort.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  private _defaultLicenseFileGenerator(packages: IPackageData[]): string {
+    const licenseFileStrings: string[] = [];
+
+    const licenseTemplateForPackage = (pkg: IPackageData, licenseContent: string): string => {
+      return `
+        <hr />
+        ${pkg.name} - ${pkg.version}
+
+        ${licenseContent}
```
dmichon-msft, line 202: The thing about multiline template literals is that they tend to end up injecting arbitrary amounts of pointless whitespace due to pretty-printing of the source code. I'd recommend against for this reason.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,242 @@
...
+      `;
+    };
+
+    for (const pkg of packages) {
+      if (pkg.licensePath) {
+        const licenseContent: string | undefined = FileSystem.readFile(pkg.licensePath);
```
dmichon-msft, line 208: `_parseCopyright` will have already read this file. Can we read it in advance and include the raw content in the object being passed to the formatter?

dmichon-msft (commented):
Some duplicate file system operations happening that can be trimmed.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,242 @@
...
+    });
+  }
+
+  private async _getLicenseFilePath(modulePath: string, compiler: Compiler): Promise<string | undefined> {
+    type InputFileSystemReadDirResults = Parameters<
+      Parameters<typeof compiler.inputFileSystem.readdir>[1]
+    >[1];
+
+    const files: InputFileSystemReadDirResults = await LegacyAdapters.convertCallbackToPromise(
+      compiler.inputFileSystem.readdir,
```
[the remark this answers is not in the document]
TheLarkInn: I don't see this option on the `inputFileSystem.readdir` types.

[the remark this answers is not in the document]
TheLarkInn: https://github.com/webpack/webpack/blob/b67626c7b4ffed8737d195b27c8cea1e68d58134/lib/util/fs.js#L114-L127 so types are just approximations and probably don't have the overloaded methods.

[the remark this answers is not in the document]
TheLarkInn: Left TODO for now on this topic.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,242 @@
...
+              } else {
+                compilation.emitAsset(
+                  this.generatedLicenseFilename,
+                  new sources.RawSource(this._defaultLicenseFileGenerator(packages))
+                );
+              }
+            } catch (e) {
+              compilation.errors.push(
+                new WebpackError(
+                  `[embedded-dependencies-webpack-plugin]: Failed to generate license file: ${e}`
```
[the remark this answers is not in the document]
TheLarkInn: Abstracted this to a try/catch handler that can take a custom error message and context message.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,242 @@
...
+
+          const dataToStringify: IEmbeddedDependenciesFile = {
+            embeddedDependencies: packages
+          };
+
+          compilation.emitAsset(this.outputFileName, new sources.RawSource(JSON.stringify(dataToStringify)));
+
+          if (this.generateLicenseFile) {
+            if (packages.length === 0) {
+              compilation.warnings.push(
```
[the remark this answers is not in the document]
TheLarkInn: Removed. I agree.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+        }
+      );
+    });
+
+    // Tap into compilation so we can tap into compilation.hooks.processAssets
+    compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tapPromise(
+        { name: PLUGIN_NAME, stage: Compilation.PROCESS_ASSETS_STAGE_REPORT },
+        async (assets) => {
+          const rawPackages: FlattenedPackageEntry[] = Array.from(thirdPartyPackages).sort((first, next) => {
```
dmichon-msft, line 98: Per discussion, makes more sense to postpone the sorting until after the async processing happens.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+
+          const packages: IPackageData[] = [];
+
+          Async.forEachAsync(rawPackages, async ([, { dir, data }]) => {
+            const { name, version } = data;
+            let licenseSource: string | undefined;
+            const license: string | undefined = parseLicense(data);
+            const licensePath: string | undefined = await this._getLicenseFilePath(dir, compiler);
+            if (licensePath) {
+              licenseSource = this._licenseSourceCache.get(licensePath);
```
dmichon-msft, line 110: Unless the license file path points out of the folder containing the package, this will always be undefined during this loop.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    return undefined;
+  }
+
+  private _defaultLicenseFileGenerator(packages: IPackageData[]): string {
+    const licenseFileStrings: string[] = [];
+
+    const licenseTemplateForPackage = (pkg: IPackageData, licenseContent: string): string => {
+      return '<hr />' + pkg.name + ' - ' + pkg.version + '<br /><br />' + licenseContent;
```
dmichon-msft, line 243: Template strings are fine; my comment was about _multiline_ template strings.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,282 @@
...
+                license,
+                copyright
+              });
+            }
+          }).catch((error: unknown) => {
+            this._emitWebpackError(compilation, 'Failed to process embedded dependencies', error);
+          });
+
+          const dataToStringify: IEmbeddedDependenciesFile = {
+            embeddedDependencies: packages
```
dmichon-msft, line 144: Does the JSON file need the non-portable path to the license file, or would it be better to just inline the content of the license file?

TheLarkInn: I think you are right, I was trying to overengineer memory usage here and keeping the format of the document slim and I think its best to just throw the source inside the embedded dep json file. `licensePath` is now removed since non-portable like you mentioned and now just have license source included.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
TheLarkInn: Using Sort.sortBy instead!

webpack/webpack-embedded-dependencies-plugin/src/test/WebpackEmbeddedDependenciesPlugin.test.ts
```diff
@@ -0,0 +1,160 @@
+import path from 'path';
+import EmbeddedDependenciesWebpackPlugin from '../';
+
+import { LICENSE_FILES_REGEXP, COPYRIGHT_REGEX } from '../regexpUtils';
+
+import { Testing } from '@rushstack/webpack-plugin-utilities';
+import { FileSystem } from '@rushstack/node-core-library';
+
+const TESTS_FOLDER_PATH: string = path.resolve(path.join(process.cwd(), 'src', 'test'));
```
dmichon-msft, line 10: `path.resolve(path.join(` is redundant; resolve does both.

webpack/webpack-embedded-dependencies-plugin/src/regexpUtils.ts
dmichon-msft: Expressly set the case-insensitive flag to be as permssive as possible.

webpack/webpack-embedded-dependencies-plugin/src/index.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  public constructor(options?: IEmbeddedDependenciesWebpackPluginOptions) {
+    this.outputFileName = options?.outputFileName || 'embedded-dependencies.json';
+    this.generateLicenseFile = options?.generateLicenseFile || false;
+    this.generateLicenseFileFunction = options?.generateLicenseFileFunction || undefined;
+    this.generatedLicenseFilename = options?.generatedLicenseFilename || 'THIRD-PARTY-NOTICES.html';
+  }
+
+  public apply(compiler: Compiler): void {
+    const thirdPartyPackages: ThirdPartyPackageMap = new Map();
```
dmichon-msft: You can avoid having to side-channel this data between the compilation hook and the `normalModuleFactory` hook by accessing `normalModuleFactory` via the second argument on the `compilation` hook.

webpack/webpack-plugin-utilities/src/Testing.ts
```diff
@@ -0,0 +1,129 @@
...
+  memFs: IFs = createFsFromVolume(new Volume())
+): Promise<(Stats | MultiStats) | undefined> {
+  const compilerOptions: Configuration = webpackMerge(_defaultWebpackConfig(entry), additionalConfig);
+  const compiler: Compiler = webpack(compilerOptions);
+
+  compiler.outputFileSystem = memFs;
+  compiler.outputFileSystem.join = path.join.bind(path);
+
+  return new Promise((resolve, reject) => {
+    compiler.run((err, stats) => {
```
dmichon-msft, line 72: Per the v5 documentation, you're supposed to close your compiler after running it.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,363 @@
...
+    const licenseContent = (pkg: IPackageData): string =>
+      pkg.licenseSource || pkg.copyright || 'License or Copyright not found';
+
+    const licenseTemplateForPackage = (pkg: IPackageData): DefaultLicenseTemplate => {
+      return `<hr />${pkg.name} - ${pkg.version}<br /><br />${licenseContent(pkg)}`;
+    };
+
+    for (const pkg of packages) {
+      licenseFileStrings.push(licenseTemplateForPackage(pkg));
+    }
```
dmichon-msft, line 330:
```suggestion
    const licenseFileStrings = packages.map(licenseTemplateForPackage);
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,363 @@
...
+   *   FileSystem.readFile('some-file');
+   * } catch (error: unknown) {
+   *   this._emitWebpackError(compilation, 'Failed to do some operation', error);
+   * }
+   * ```
+   */
+  private _emitWebpackError(compilation: Compilation, errorMessage: string, error: unknown): void {
+    // If the error is a string, we can just emit it as is with message prefix and error message
+    if (typeof error === 'string') {
+      compilation.errors.push(new WebpackError(`${PLUGIN_ERROR_PREFIX}: ${errorMessage}: ${error}`));
```
dmichon-msft, line 270: Deduplicate the `compilation.errors.push(` call by simply assigning a variable via a few different paths, then do the push once.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,363 @@
...
+    return files
+      ?.map((file) => file.toString())
+      .filter((file) => LICENSE_FILES_REGEXP.test(file))
+      .map((file) => path.join(modulePath, file))[0]; // Grabbing the first license file if multiple are found
+  }
+
+  /**
+   * Given a module path, try to parse the module's copyright attribution.
+   */
+  private async _parseCopyright(licenseSource: string): Promise<string | undefined> {
```
dmichon-msft, line 309: This doesn't need to be async anymore

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,363 @@
...
+    });
+
+    // Tap into compilation so we can tap into compilation.hooks.processAssets
+    compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tapPromise(
+        { name: PLUGIN_NAME, stage: Compilation.PROCESS_ASSETS_STAGE_REPORT },
+        async (assets) => {
+          const packages: IPackageData[] = [];
+
+          Async.forEachAsync(Array.from(thirdPartyPackages), async ([, { dir, data }]) => {
```
dmichon-msft, line 184: You shouldn't need the `Array.from` call; Async.forEachAsync accepts a general iterable.

webpack/webpack-embedded-dependencies-plugin/README.md
```diff
@@ -0,0 +1,142 @@
+# webpack-embedded-dependencies-plugin
+
+## Installation
+
+`npm install @rushstack/webpack-embedded-dependencies-plugin --save`
+
+## Overview
+A webpack plugin for generating a list of embedded dependencies. Embedded dependencies are third-party packages which are being
+bundled into your released code and are often times subject to license, security, and other legal requirements. This plugin
```
iclanton, line 9:
```suggestion
bundled into your released code and are oftentimes subject to license, security, and other legal requirements. This plugin
```

webpack/webpack-embedded-dependencies-plugin/README.md
```diff
@@ -0,0 +1,142 @@
...
+aims to make it easier to generate a list of embedded dependencies and their associated metadata, so they can be analyzed by additional tools.
+
+The plugin also includes the ability to generate a secondary asset which contains the license text for each embedded dependency into a single file called
+THIRD-PARTY-NOTICES.html. This is a common legal requirement for large companies deploying commercial services/products containing open source code.
```
iclanton, line 13:
```suggestion
THIRD-PARTY-NOTICES.html. This is a common legal requirement when deploying services or products containing open-source code.
```

iclanton, line 13: HTML? Seems like the name and output format should be configurable.

iclanton: Oh, the name is configurable. This summary seems to imply that it isn't.

HTML seems like a weird default choice.

webpack/webpack-embedded-dependencies-plugin/README.md
```diff
@@ -0,0 +1,142 @@
...
+ * Loose string type that represents the name of the generated license file.
+ *
+ * @example
+ * ```ts
+ * const licenseFileName: LicenseFileName = 'custom-license-file-name.html';
+ * const licenseMarkdownFileName: LicenseFileName = 'custom-license-file-name.md';
+ * const licenseTextFileName: LicenseFileName = 'custom-license-file-name.txt';
+ * ```
+ */
+export declare type LicenseFileName = `${string}.${'html' | 'md' | 'txt'}`; -->
```
iclanton, line 107: ?

common/reviews/api/webpack-embedded-dependencies-plugin.api.md
```diff
@@ -0,0 +1,45 @@
...
+export interface IPackageData extends IPackageJson {
+    author?: string | {
+        name?: string;
+    };
+    copyright: string | undefined;
+    licenses?: {
+        type: string;
+        url: string;
+    }[];
+    licenseSource?: string;
```
iclanton, line 36: Should we just include these fields in `IPackageJson` in `node-core-library`? @octogonz?

webpack/webpack-embedded-dependencies-plugin/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "typings": "dist/webpack-embedded-dependencies-plugin.d.ts",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/rushstack.git",
+    "directory": "webpack/webpack-embedded-dependencies-plugin"
+  },
+  "scripts": {
+    "build": "heft build --clean",
+    "_phase:build": "heft build --clean"
```
iclanton, line 15:
```suggestion
    "_phase:build": "heft build --clean",
    "_phase:test": "heft test --no-build"
```

webpack/webpack-embedded-dependencies-plugin/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "dependencies": {
+    "@rushstack/node-core-library": "workspace:*",
+    "glob": "~7.0.5"
```
iclanton, line 19: The new Heft is using `fast-glob`. Maybe take a look into that?

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
+import path from 'path';
+import { Async, IPackageJson, Sort } from '@rushstack/node-core-library';
+import { LegacyAdapters, FileSystem } from '@rushstack/node-core-library';
+import { Compilation, Compiler, WebpackPluginInstance, sources, WebpackError } from 'webpack';
```
iclanton, line 4:
```suggestion
import type { Compilation, Compiler, WebpackPluginInstance, sources, WebpackError } from 'webpack';
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
+import path from 'path';
+import { Async, IPackageJson, Sort } from '@rushstack/node-core-library';
+import { LegacyAdapters, FileSystem } from '@rushstack/node-core-library';
```
iclanton, line 3:
```suggestion
import {
  Async,
  IPackageJson,
  Sort,
  LegacyAdapters,
  FileSystem
} from '@rushstack/node-core-library';
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+  licenses?: { type: string; url: string }[];
+
+  /**
+   * The source of the license file itself used for generating THIRD-PARTY-NOTICES.html or custom license files.
+   */
+  licenseSource?: string;
+}
+
+/**
+ * @alpha
```
iclanton, line 54: The Rushstack repo doesn't use the `@alpha` label. Use `@beta` isntead.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+   */
+  generateLicenseFileFunction?: LicenseFileGeneratorFunction;
+  /**
+   * Name of the generated license file. Defaults to THIRD-PARTY-NOTICES.html
+   */
+  generatedLicenseFilename?: LicenseFileName;
+}
+
+/**
+ * @alpha
```
iclanton, line 100:
```suggestion
 * @beta
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
iclanton:
```suggestion
 * @beta
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+ *   .map((pkg) => {
+ *    return `<h2>${pkg.name}</h2><p>${pkg.license}</p>`;
+ *  }).join('');
+ * }
+ * ```
+ */
+export type LicenseFileGeneratorFunction = (packages: IPackageData[]) => string;
+
+/**
+ * @alpha
```
iclanton, line 116:
```suggestion
 * @beta
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+ * ```
+ */
+export type LicenseFileName = `${string}.${'html' | 'md' | 'txt'}`;
+
+type PackageMapKey = `${string}@${string}`;
+type ThirdPartyPackageMap = Map<PackageMapKey, { dir: string; data: IPackageData }>;
+type DefaultLicenseTemplate = `<hr />${string}<br /><br />${string}`;
+
+/**
+ * @alpha
```
iclanton, line 133:
```suggestion
 * @beta
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+  private readonly _generateLicenseFile: boolean;
+  private readonly _generateLicenseFileFunction: LicenseFileGeneratorFunction;
+  private readonly _generatedLicenseFilename: LicenseFileName;
+
+  public constructor(options?: IEmbeddedDependenciesWebpackPluginOptions) {
+    this._outputFileName = options?.outputFileName || 'embedded-dependencies.json';
+    this._generateLicenseFile = options?.generateLicenseFile || false;
+    this._generateLicenseFileFunction =
+      options?.generateLicenseFileFunction || this._defaultLicenseFileGenerator;
+    this._generatedLicenseFilename = options?.generatedLicenseFilename || 'THIRD-PARTY-NOTICES.html';
```
iclanton, line 148: Extract the `'THIRD-PARTY-NOTICES.html'` to a const.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+  }
+
+  /**
+   * @alpha
```
iclanton, line 152:
```suggestion
```

Should be inferred from the tag on the class.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+ * ```ts
+ * const licenseFileName: LicenseFileName = 'custom-license-file-name.html';
+ * const licenseMarkdownFileName: LicenseFileName = 'custom-license-file-name.md';
+ * const licenseTextFileName: LicenseFileName = 'custom-license-file-name.txt';
+ * ```
+ */
+export type LicenseFileName = `${string}.${'html' | 'md' | 'txt'}`;
+
+type PackageMapKey = `${string}@${string}`;
+type ThirdPartyPackageMap = Map<PackageMapKey, { dir: string; data: IPackageData }>;
```
iclanton, line 129:
```suggestion
type ThirdPartyPackageMap = Map<PackageMapKey, { packageFolderPath: string; packageJsonData: IPackageData }>;
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+ * @example
+ * ```ts
+ * const licenseFileName: LicenseFileName = 'custom-license-file-name.html';
+ * const licenseMarkdownFileName: LicenseFileName = 'custom-license-file-name.md';
+ * const licenseTextFileName: LicenseFileName = 'custom-license-file-name.txt';
+ * ```
+ */
+export type LicenseFileName = `${string}.${'html' | 'md' | 'txt'}`;
+
+type PackageMapKey = `${string}@${string}`;
```
iclanton, line 128:
```suggestion
type PackageNameAndVersion = `${string}@${string}`;
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+        async (assets) => {
+          const packages: IPackageData[] = [];
+
+          Async.forEachAsync(Array.from(thirdPartyPackages), async ([, { dir, data }]) => {
+            const { name, version } = data;
+            let licenseSource: string | undefined;
+            const license: string | undefined = parseLicense(data);
+            const licensePath: string | undefined = await this._getLicenseFilePath(dir, compiler);
+            if (licensePath) {
+              licenseSource = await FileSystem.readFile(licensePath);
```
iclanton, line 190:
```suggestion
              licenseSource = await FileSystem.readFileAsync(licensePath);
```

iclanton: Gracefully handle this file not existing?

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+      compilation.errors.push(
+        new WebpackError(`${PLUGIN_ERROR_PREFIX}: ${errorMessage}: ${JSON.stringify(error || '')}`)
+      );
+    }
+  }
+
+  /**
+   * Searches a third party package directory for a license file.
+   */
+  private async _getLicenseFilePath(modulePath: string, compiler: Compiler): Promise<string | undefined> {
```
iclanton, line 287:
```suggestion
  private async _getLicenseFilePathAsync(modulePath: string, compiler: Compiler): Promise<string | undefined> {
```

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,361 @@
...
+            }
+          })
+            .catch((error: unknown) => {
+              this._emitWebpackError(compilation, 'Failed to process embedded dependencies', error);
+            })
+            .finally(() => {
+              // Sort the packages by name after we have processed all of them
+              // Note this is a mutable sort
+              Sort.sortBy(packages, (pkg) => pkg.name);
+            });
```
iclanton, line 221: Put this in a normal `try`/`catch`/`finally` and await the `Async.forEachAsync` call?

webpack/webpack-embedded-dependencies-plugin/src/regexpUtils.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @remarks The following file names are covered via unit tests to be matched:
+ *
+ * - `'LICENSE'`
+ * - `'LICENSE.txt'`
+ * - `'LICENSE.md'`
+ * - `'LICENSE-MIT.txt'`
+ * - `'license'`
+ * - `'license.txt'`
+ */
+export const LICENSE_FILES_REGEXP: RegExp = /^LICENSE(-MIT)?(\.(txt|md))?$/i;
```
iclanton, line 13:
```suggestion
export const LICENSE_FILES_REGEXP: RegExp = /^LICENSE(-[A-Z-]+)?(\.(txt|md))?$/i;
```

webpack/webpack-embedded-dependencies-plugin/README.md
TheLarkInn: Whoops! Will remove!

common/reviews/api/webpack-embedded-dependencies-plugin.api.md
TheLarkInn: I don't think `licenseSource` belongs in there personally but I'm apathetic to the others.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
TheLarkInn: Compilation and WebpackError and sources have to be values but I did separate type-only imports from value imports.

webpack/webpack-embedded-dependencies-plugin/README.md
TheLarkInn: When reviewing existing implementations, each major product was generating an HTML file. Since this is an optional generation, having the most common scenario covered felt right to me.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
TheLarkInn: With this now wrapped in a `try/catch`, the error should propagate up to the catch. If we find a licensePath but fail to read it, this could be for permissions, or FS related problems and should just float this message up the consumer. 

Or do we also need a nested try/catch here?

TheLarkInn: Resolved this conversation offline. This is good with the trycatch we have.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,383 @@
...
+   */
+  private async _getLicenseFilePathAsync(
+    modulePath: string,
+    compiler: Compiler
+  ): Promise<string | undefined> {
+    type InputFileSystemReadDirResults = Parameters<
+      Parameters<typeof compiler.inputFileSystem.readdir>[1]
+    >[1];
+
+    // TODO: Real fs.readdir can take an arguement ({ withFileTypes: true }) which will filter out directories for better performance
```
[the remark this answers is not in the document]
TheLarkInn: I added a github webpack issue to todo tracking this.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,383 @@
...
+  /**
+   * @beta
+   * Webpack plugin apply method. This method is called by the webpack compiler to apply the plugin, however it not usually
+   * needed to be invoked manually by the developer in a webpack configuration. However, if you are calling this plugin (applying it from another plugin)
+   * you can call `plugin.apply(compiler)` to apply the plugin and invoke it.
+   * @param compiler - The webpack compiler instance.
+   */
+  public apply(compiler: Compiler): void {
+    // Tap into compilation so we can tap into compilation.hooks.processAssets
+    compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (compilation, { normalModuleFactory }) => {
```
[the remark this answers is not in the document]
TheLarkInn: The only _common_ childCompilation is html-webpack-plugin. I think for our usecase we can leave this as is, and if someone requests tracking every compilation we can introduce the new complexity.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,383 @@
...
+  relativePath?: string;
+}
+
+interface IWebpackModuleCreateData {
+  resourceResolveData?: IResourceResolveData;
+}
+
+/**
+ * @beta
+ * Data type for a package.json file. This is a subset of the full package.json file and includes additional fields
```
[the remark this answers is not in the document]
TheLarkInn: Corrected this.

webpack/webpack-embedded-dependencies-plugin/src/test/node_modules/@rushstack/tree-pattern/lib/TreePattern.js
```diff
@@ -0,0 +1,177 @@
+"use strict";
```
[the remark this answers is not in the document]
TheLarkInn: I removed this and instead just added a fake package in the fake node_modules folder which a test will filter out using the filter function. 

Inevitably the goal of this is to cover the filter function.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,383 @@
...
+        (module, moduleCreateData: IWebpackModuleCreateData, resolveData) => {
+          const { resourceResolveData } = moduleCreateData;
+          const pkg: IPackageData | undefined = resourceResolveData?.descriptionFileData;
+          const filePath: string | undefined = resourceResolveData?.descriptionFileRoot;
+
+          if (
+            pkg &&
+            filePath &&
+            this._packageFilterFunction(pkg, filePath) &&
+            filePath?.includes('node_modules')
```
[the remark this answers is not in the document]
TheLarkInn: If this situation arises, we could implement an API which manually declares packages like this into the configuration (like externals etc.). What I don't want for the user: 

* Extra overhead of providing an more strict filter function because the default filter is overly loose. 
* Overly strict filter which doesn't provide flexibility

I think this strikes the balance for now, however if a need for this scenario arises we can add it.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,384 @@
...
+            this._emitWebpackError(compilation, 'Failed to process embedded dependencies', error);
+          } finally {
+            Sort.sortBy(packages, (pkg) => pkg.name);
+          }
+
+          const dataToStringify: IEmbeddedDependenciesFile = {
+            embeddedDependencies: packages
+          };
+
+          compilation.emitAsset(this._outputFileName, new sources.RawSource(JSON.stringify(dataToStringify)));
```
dmichon-msft, line 248: `sources` should come from `compiler.webpack.sources` to guarantee version synchronization. Use type-only imports for everything from webpack.

dmichon-msft (commented):
Found a potential compatibility issue.

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
```diff
@@ -0,0 +1,387 @@
+import path from 'path';
+import { Compilation } from 'webpack';
```
dmichon-msft, line 2: Consolidate as a type-only import.

dmichon-msft (approved):
Just one errant non-type import of `webpack` (even if in practice it is being used as a type import)

webpack/webpack-embedded-dependencies-plugin/src/EmbeddedDependenciesWebpackPlugin.ts
TheLarkInn: This is required for `Compilation.PROCESS_ASSETS_STAGE_REPORT`

dmichon-msft: `compiler.webpack.Compilation.PROCESS_ASSETS_STAGE_REPORT`

dmichon-msft (approved):

webpack/webpack-plugin-utilities/src/Testing.ts
```diff
@@ -0,0 +1,131 @@
+import { createFsFromVolume, Volume, IFs } from 'memfs';
```
iclanton, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
// See LICENSE in the project root for license information.

import { createFsFromVolume, Volume, IFs } from 'memfs';
```

iclanton (approved):

PR #3976, opened 2023-02-15
[rush] rush-http-build-cache-plugin

elliot-nelson:
## Summary

This PR ports our internal implementation of a "Rush HTTP Build Cache Plugin" to Rush, lining it up with the existing providers (Amazon S3 + Azure Blob Storage).

The goal of this plugin is to provide a simple baseline solution for using any generic HTTP backend as a remote build cache, as long as it meets a few basic requirements (essentially any server that can read/write cache entries as raw request and response bodies and uses appropriate HTTP status codes).

## Details

 - Basic plugin boilerplate and implementation
 - _Authentication_ today falls completely back to the monorepo maintainer, in the form of a custom node script that can return any authentication value you want.
 - Include a README, with one detailed example (configuring the plugin to talk to a new or existing Gradle Build Cache Node).
 
> A possible area for improvement before merging would be to see if the "tokenHandler" script can be the worst-case fallback, with other (simpler) options for configuration -- maybe a "userEnvironmentVariable" and "tokenEnvironmentVariable" that would handle the environment variable case, or a ".netrc" option that would read the credentials from an existing .netrc file, etc.

## How it was tested

 - A sprinkling of unit tests
 - Followed the steps described in the README to reconfigure the Rushstack repo to talk to a local Gradle Build Cache Node, and confirmed that it could read and write cache entries to the server.

## Impacted documentation

 - https://rushjs.io/pages/maintainer/build_cache/


rush-plugins/rush-http-build-cache-plugin/src/HttpBuildCacheProvider.ts
```diff
@@ -0,0 +1,384 @@
...
+  private async _tryGetCredentials(options: CredentialsOptions.Omit): Promise<undefined>;
+  private async _tryGetCredentials(options: CredentialsOptions): Promise<string | undefined>;
+  private async _tryGetCredentials(options: CredentialsOptions): Promise<string | undefined> {
+    if (options === CredentialsOptions.Omit) {
+      return;
+    }
+
+    let credentials: string | undefined = this._environmentCredential;
+
+    if (typeof credentials !== 'string') {
```
iclanton, line 259:
```suggestion
    if (credentials !== 'string') {
```

rush-plugins/rush-http-build-cache-plugin/src/HttpBuildCacheProvider.ts
```diff
@@ -0,0 +1,384 @@
...
+      credentials = await this._tryGetCredentialsFromCache();
+    }
+
+    if (typeof credentials !== 'string' && options === CredentialsOptions.Required) {
```
iclanton, line 263:
```suggestion
    if (credentials === undefined && options === CredentialsOptions.Required) {
```

rush-plugins/rush-http-build-cache-plugin/src/HttpBuildCacheProvider.ts
```diff
@@ -0,0 +1,384 @@
...
+    }
+  }
+
+  private _getFailureType(requestMethod: string, response: Response, isRedirect: boolean): FailureType {
+    if (response.ok) {
+      return FailureType.None;
+    }
+
+    switch (response.status) {
+      case 503: {
```
iclanton, line 306: Seems like this should be for any 5xx error.

rush-plugins/rush-http-build-cache-plugin/src/RushHttpBuildCachePlugin.ts
```diff
@@ -0,0 +1,75 @@
...
+            pluginName: this.pluginName,
+            rushProjectRoot: rushConfig.rushJsonFolder,
+            url: url,
+            headers: headers,
+            tokenHandler: tokenHandler,
+            cacheKeyPrefix: cacheKeyPrefix,
+            isCacheWriteAllowed: !!isCacheWriteAllowed
+          };
+
+          return new HttpBuildCacheProviderModule.HttpBuildCacheProvider(options, rushSession);
```
iclanton, line 70: Can we change the signature of [`CloudBuildCacheProviderFactory`](https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/src/pluginFramework/RushSession.ts#L21) to

```TS
export type CloudBuildCacheProviderFactory = (buildCacheJson: IBuildCacheJson) => ICloudBuildCacheProvider | Promise<ICloudBuildCacheProvider>;
```

and just use an async import `HttpBuildCacheProvider` here instead of `Import.lazy`? `Import.lazy` screws up bundling, and async import is a standard practice, and it's faster at runtime.

rush-plugins/rush-http-build-cache-plugin/src/exec.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @param args - Arguments to pass to command
+ * @returns The exit code and output of the executed command
+ */
+export function exec(cmd: string, cwd: string): Promise<IExecResult> {
+  return new Promise(function (resolve: (result: IExecResult) => void, reject: (error: unknown) => void) {
+    child_process_exec(cmd, { cwd }, function (error, stdout, stderr) {
+      resolve({ stdout, stderr, error: error || undefined });
+    });
+  });
+}
```
iclanton, line 35: Doesn't need to be in this PR, but this would be a useful utility in `node-core-library`'s `Executable`.

rush-plugins/rush-http-build-cache-plugin/src/index.ts
```diff
@@ -0,0 +1,4 @@
+import { RushHttpBuildCachePlugin } from './RushHttpBuildCachePlugin';
+
+export default RushHttpBuildCachePlugin;
+export { IHttpBuildCacheProviderOptions } from './HttpBuildCacheProvider';
```
iclanton, line 4:
```suggestion
export type { IHttpBuildCacheProviderOptions } from './HttpBuildCacheProvider';
```

rush-plugins/rush-http-build-cache-plugin/src/schemas/plugin-config.schema.json
```diff
@@ -0,0 +1,39 @@
...
+  "required": ["url"],
+  "properties": {
+    "url": {
+      "type": "string",
+      "description": "(Required) The URL of the server that stores the caches (e.g. \"https://build-caches.example.com\").",
+      "format": "uri"
+    },
+    "method": {
+      "type": "string",
+      "description": "(Required) The URL of the server that stores the caches (e.g. \"https://build-caches.example.com\").",
```
iclanton, line 14: Update.

rush-plugins/rush-http-build-cache-plugin/src/schemas/plugin-config.schema.json
```diff
@@ -0,0 +1,39 @@
...
+      "enum": ["PUT", "POST", "GET", "PATCH"],
```
iclanton, line 15:
```suggestion
      "enum": ["PUT", "POST", "PATCH"],
```

`GET` can't take a body.

rush-plugins/rush-http-build-cache-plugin/src/schemas/plugin-config.schema.json
```diff
@@ -0,0 +1,39 @@
...
+  "title": "Configuration for build cache with HTTPS server",
+  "type": "object",
+  "required": ["url"],
+  "properties": {
+    "url": {
+      "type": "string",
+      "description": "(Required) The URL of the server that stores the caches (e.g. \"https://build-caches.example.com\").",
+      "format": "uri"
+    },
+    "method": {
```
iclanton, line 12:
```suggestion
    "uploadMethod": {
```

rush-plugins/rush-http-build-cache-plugin/src/schemas/plugin-config.schema.json
```diff
@@ -0,0 +1,39 @@
...
+      "type": "object",
+      "description": "(Optional) HTTP headers to pass to the cache server",
+      "properties": {},
+      "additionalProperties": {
+        "type": "string"
+      }
+    },
+    "tokenHandler": {
+      "type": "string",
+      "description": "(Optional) Shell command that prints the authorization token needed to communicate with the HTTPS server and exits with code 0."
```
iclanton, line 28: Maybe mention that this command is executed at from the root of the Rush repo.

rush-plugins/rush-http-build-cache-plugin/src/schemas/plugin-config.schema.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "cacheKeyPrefix": {
+      "type": "string",
+      "description": "(Optional) prefix for cache keys."
+    },
+    "isCacheWriteAllowed": {
+      "type": "boolean",
+      "description": "(Optional) If set to true, allow writing to the cache. Defaults to false."
+    }
```
iclanton, line 37: Aren't these part of the standard options for all plugins? Seems like they should be.

elliot-nelson: The way it works today is that each cache provider options individually get an option called `isCacheWriteAllowed` -- but I agree, it would make sense to turn this into part of the interface for this type of Plugin Provider instead.

rush-plugins/rush-http-build-cache-plugin/src/RushHttpBuildCachePlugin.ts
elliot-nelson: This sounds interesting -- I didn't tackle this yet because I'm not 100% on the ramifications.

We could adjust all 3 cache providers to avoid Import.lazy if the Factory was inherently async, but I'd need to trace up the paths where the Factory is actually used to see what other problems it might cause. (For example, it would mean you cannot call the factory from inside a chain containing a constructor.)

elliot-nelson:
Addressed feedback posted so far, except for the `Import.lazy` issue and the potential node-core-library utility.

elliot-nelson:
@iclanton Let me know what you think of latest changes... based on some issues we've had recently in our monorepo with windows/mac/linux users, I think it's safer to force a spawn (instead of exec) by default for the external script, so I moved to just use the standard Executable handler out of node-core-library.

rush-plugins/rush-http-build-cache-plugin/src/HttpBuildCacheProvider.ts
elliot-nelson: So I ended up adding a comment instead of changing this -- our testing is all on S3, where there's a couple types of failures we are catching (Service Unavailable, Rate Limit Throttle), which we know are transient.

But even according to HTTP Status spec, 500 = Unknown Error, 501 = Not Implemented, 502 = Bad Gateway, 503 = Service Unavailable. So I think our assumption is that a generic 5xx error is actually most likely a configuration problem in the plugin and _should_ be reported (failed build), whereas a 503 is likely temporary and can just be printed.

rush-plugins/rush-http-build-cache-plugin/src/HttpBuildCacheProvider.ts
```diff
@@ -0,0 +1,395 @@
...
+      } bytes`
+    );
+
+    return result;
+  }
+
+  private async _tryGetCredentials(options: CredentialsOptions.Required): Promise<string>;
+  private async _tryGetCredentials(options: CredentialsOptions.Optional): Promise<string | undefined>;
+  private async _tryGetCredentials(options: CredentialsOptions.Omit): Promise<undefined>;
+  private async _tryGetCredentials(options: CredentialsOptions): Promise<string | undefined>;
```
iclanton, line 255:
```suggestion
```

Looks like this overload is redundant.

iclanton (approved):

rush-plugins/rush-http-build-cache-plugin/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+```json
+{
+    "buildCacheEnabled": true,
+
+    "cacheProvider": "http"
+}
+```
+
+Once enabled, configure the HTTP build cache in config file `common/config/rush-plugins/rush-http-build-cache-plugin.json`:
```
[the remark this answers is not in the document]
elliot-nelson: It does!

However, if you go that route, there's no way to specify a schema for the file anymore, as the "extra properties" are basically anything.

I think this relates to your other question: if this is a "built in" plugin, with support for HTTP backends out of the box, then I think we should just extend the build-cache.json schema with the fields from this plugin.

If it's a separate plugin you add yourself, I think it makes more sense to act like any other plugin, with its own configuration file.

rush-plugins/rush-http-build-cache-plugin/src/HttpBuildCacheProvider.ts
```diff
@@ -0,0 +1,395 @@
...
+      const result = await this._http({
+        terminal: terminal,
+        relUrl: `${this._cacheKeyPrefix}${cacheId}`,
+        method: 'GET',
+        body: undefined,
+        warningText: 'Could not get cache entry',
+        readBody: true
+      });
+
+      return Buffer.isBuffer(result) ? result : undefined;
```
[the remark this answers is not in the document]
elliot-nelson: If you try to get a cache entry by Id and it doesn't exist on the server, the return value from `_http()` will be `false` (and we'll return `undefined` to the caller).

rush-plugins/rush-http-build-cache-plugin/src/HttpBuildCacheProvider.ts
elliot-nelson: I played with it, but it seems that TypeScript requires the full list of overloads if you have any overloads, otherwise you can't call the method with a generic CredentialsOptions object.

libraries/rush-lib/package.json
```diff
@@ -88,7 +88,8 @@
   },
   "publishOnlyDependencies": {
     "@rushstack/rush-amazon-s3-build-cache-plugin": "workspace:*",
-    "@rushstack/rush-azure-storage-build-cache-plugin": "workspace:*"
+    "@rushstack/rush-azure-storage-build-cache-plugin": "workspace:*",
+    "@rushstack/rush-http-build-cache-plugin": "workspace:*"
```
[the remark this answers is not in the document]
elliot-nelson: I think my feeling is that HTTP would be a core plugin that's available out of the box, but I am open to arguments otherwise!

rush-plugins/rush-http-build-cache-plugin/src/HttpBuildCacheProvider.ts
```diff
@@ -0,0 +1,395 @@
+import { ITerminal, Executable } from '@rushstack/node-core-library';
+import {
+  ICloudBuildCacheProvider,
+  ICredentialCacheEntry,
+  CredentialCache,
+  RushSession,
+  EnvironmentConfiguration
+} from '@rushstack/rush-sdk';
+import fetch, { BodyInit, Response } from 'node-fetch';
```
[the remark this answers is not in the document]
elliot-nelson: Interesting... we use node-fetch a lot and I don't remember running into this, but simpler might be better. I could play with this this weekend.

(My hope is if we hold out long enough, we'll just use node core fetch, but I think it's a couple versions away.)

libraries/rush-lib/package.json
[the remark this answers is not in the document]
iclanton: I'm good with that.

rush-plugins/rush-http-build-cache-plugin/src/HttpBuildCacheProvider.ts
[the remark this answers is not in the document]
iclanton: We can make that change after this PR goes in.

elliot-nelson:
I've updated the plugin to be a bundled plugin. The code itself was already set up that way (just for ease of testing), so changes made were:

 - Updated the schema and rush init template for `build-cache.json`, validated that it passes muster out of the box and after enabling HTTP cache plugin.
 - Updated plugin to read its options from build-cache config instead of a separate file.
 - Updated instructions in the README for the http build cache plugin.

iclanton (approved):

libraries/rush-lib/src/schemas/build-cache.schema.json
```diff
@@ -85,6 +85,56 @@
...
+            "uploadMethod": {
+              "type": "string",
+              "description": "(Optional) The HTTP method to use when writing to the cache (defaults to PUT).",
+              "enum": ["PUT", "POST", "PATCH"],
+              "default": "PUT"
+            },
+            "headers": {
+              "type": "object",
+              "description": "(Optional) HTTP headers to pass to the cache server",
+              "properties": {},
```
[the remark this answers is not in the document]
elliot-nelson: I think it does right?

```
              "additionalProperties": {
                "type": "string"
              }
```

PR #3989, opened 2023-02-26
[heft] Remove ts.readJson and ts.emitFiles patches

jakebailey:
## Summary

TS 5.0 no longer allows its exported API to be patched at runtime, so heft ends up crashing. This PR removes the patches, replacing them with other (still unsafe) means of achieving the same thing.

## Details

The `ts.readJson` no longer needs a patch in modern TypeScript.

The `ts.emitFiles` patch was replaced with a function which replaces the `emit` method on each program, plus code to reconstruct which files are to be emitted via `getSourceFiles`.

## How it was tested

Truthfully, I don't really know how to test this. In development, I did have failing tests for certain builds, which eventually worked as I continued to refine this change. But, as noted in the template, many things are not tested, so I'd appreciate some help testing this as I don't have any rushstack projects to test on.


dmichon-msft:
A number of projects directly in this repository use the multi-emit functionality, so a clean + build in those (e.g. build-tests/heft-typescript-composite-test) will suffice for testing that the output at least works. Of more interest is the performance implications.

jakebailey:
I would expect no performance change, as this still only calls emit; the function replaced here is the one that calls `ts.emitFiles` which was previously patched.

The only cases I'm actually worried about are solutions / watch mode, given those (I think?) have multiple programs. But, I'm pretty sure that providing a createProgram function which patches the program before returning it means that it's fine.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
```diff
@@ -794,15 +805,135 @@ export class TypeScriptBuilder extends SubprocessRunnerBase<ITypeScriptBuilderCo
...
+            writeBOM,
+            onError,
+            sourceFiles
+          );
+        };
+      }
+
+      const compilerOptions: TTypescript.CompilerOptions = program.getCompilerOptions();
+
+      function runWithSaveRestoreOptions<T>(fn: () => T): T {
```
dmichon-msft, line 850: Would it be safer to override `program.getCompilerOptions()` and return a value from a closure? That way we don't have to validate the reversibility of the mutation of the options structure.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
```diff
@@ -794,15 +805,135 @@ export class TypeScriptBuilder extends SubprocessRunnerBase<ITypeScriptBuilderCo
...
+    ts: ExtendedTypeScript,
+    program: Program
+  ): Program {
+    if ((program as Program & { __heftPatched?: boolean }).__heftPatched) {
+      return program;
+    }
+    (program as Program & { __heftPatched?: boolean }).__heftPatched = true;
+
+    const originalEmit: Program['emit'] = program.emit;
+    program.emit = (...args) => {
```
dmichon-msft, line 818: I seem to recall trying to call emit() multiple times originally, and running into an issue that it doesn't work when invoked on a BuilderProgram, because calling `emit()` has side effects (namely that it clears the list of files that are pending emit).

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
jakebailey: I'm pretty sure no; we take reference to this internally so you have to hack the object.

@weswigham suggested this when we were going through the different patchers, maybe there's a better way or I misinterpreted his suggestion, but I think this is what we thought might work.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
dmichon-msft: Likely this needs to be careful to modify the inner program and not the builder program instance.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
dmichon-msft: It looks like you only bind to `program.getCompilerOptions` in the `getEmitHost` function: https://github.com/microsoft/TypeScript/blob/main/src/compiler/program.ts#L2486
So replacing said function ought to work, as `getEmitHost` is called inside of `emit`

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
jakebailey: I was also worried about inner/outer here. All of the tests pass, so, unless this is one of those untested cases, I'm not totally sure what the deal is. If you think there's a test within the repo I can run that would test this, I'd appreicate some instructions on how to do this. I've just been running `rush test` which I've assumed does the right thing.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
jakebailey: Sure, I guess I'll try it like that.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
```diff
@@ -794,15 +805,135 @@ export class TypeScriptBuilder extends SubprocessRunnerBase<ITypeScriptBuilderCo
...
+        const replacementExtension: string = `${jsExtensionOverride}$1`;
+        return (
+          fileName: string,
+          data: string,
+          writeBOM: boolean,
+          onError?: ((message: string) => void) | undefined,
+          sourceFiles?: readonly TTypescript.SourceFile[] | undefined
+        ) => {
+          return writeFile(
+            fileName.replace(/\.js(\.map)?$/g, replacementExtension),
```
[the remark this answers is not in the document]
jakebailey: Sure, though in a microbenchmark for this, the difference is very minimal.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
dmichon-msft: rush test will clean files, and `rush-lib` uses multi-emit for commonJS for Jest and ESM for webpack, so that should suffice for the non-incremental non-watch case. I have a local variant on this change to experiment with that also cleans up some other quirks to bring in improvements I did during the upcoming Heft refactor (namely passing a custom function when creating a watch host, solution host, etc. that wraps ts.createEmitAndSemanticDiagnosticsBuilderProgram instead of passing it directly).

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
dmichon-msft: Ah, found the binding to `options` in `emitWorker`; it's passed to the getEmitResolver call.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
jakebailey: > uses multi-emit for commonJS for Jest and ESM for webpack, so that should suffice for the non-incremental non-watch case.

Ah, okay. That's the test that failed until I finished off the change, so, it seems like it's good.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
jakebailey: Yeah, that's the one I remembered seeing but promptly forgot about.

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
jakebailey: Hm, your comment is gone here; do you still want me to change this?

apps/heft/src/plugins/TypeScriptPlugin/TypeScriptBuilder.ts
```diff
@@ -794,15 +805,135 @@ export class TypeScriptBuilder extends SubprocessRunnerBase<ITypeScriptBuilderCo
...
-    const changedFiles: Set<IExtendedSourceFile> = new Set<IExtendedSourceFile>();
-    EmitFilesPatch.install(ts, tsconfig, this._moduleKindsToEmit, changedFiles);
+    const sourceFileMayBeEmittedHost: ISourceFileMayBeEmittedHost = {
+      getCompilerOptions: genericProgram.getCompilerOptions,
+      getResolvedProjectReferenceToRedirect: () => undefined,
+      isSourceFileFromExternalLibrary: () => false,
+      isSourceOfProjectReferenceRedirect: () => false
+    };
+
+    const changedFiles: Set<IExtendedSourceFile> = new Set<IExtendedSourceFile>(
```
jakebailey, line 930: This whole thing is to just preserve the behavior of running ESLint only on files which had emit. I'm not actually sure that this is desirable though, so I would prefer to remove all of this and not depend on this internal API.

dmichon-msft: We don't actually need to hack this, since we can still aggregate the information from the first argument to `emit`; in a builder program it is called on each affected source file separately.

dmichon-msft:
I put together a PR #3993, borrowing heavily from this and my PR for overhauling Heft (which cleaned up the solution and watch mode interaction). Testing locally in watch and solution mode it is producing the expected results (and can run fine even with `typescript@next`). Also updated the constants that specify the maximum version Heft has been tested against.

jakebailey:
Cool, I'll close this as #3993 probably works better.

PR #3991, opened 2023-02-27
[eslint-config] allow implicitly typed exports created with a functioncall

### LINKED ISSUE #3990: [eslint-config] Proposal: allow implicitly typed constants when created from a function call
## Summary

When creating React apps it's common to use styled components.

E.g. this code will throw a warning: "Expected a type annotation. eslint[@rushstack/typedef-var](https://www.npmjs.com/package/%40rushstack/eslint-plugin)"

```js
export const StyledButton = styled('button', ...)
```

The goal of the rule as I understand it, is to prevent a the return value of a function, or a variable, to be accidentally become another type. When calling a function it doesn't really change. Also the function returns a complex type.

## Details

We can fix it by adding this to the `typedef-var` 

```ts
if (node.init?.type === 'CallExpression') {
    // The variable is declared by a function call
    return;
}
```

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@rushstack/eslint-config` version? |  |
| Operating system? | Mac  |
| TypeScript compiler version? |  |

alber70g:
## Summary

Fixes #3990

This change allows users to omit the type when it's created from a function call.

When creating React apps it's common to use styled components.

E.g. this code will throw a warning: "Expected a type annotation. eslint[@rushstack/typedef-var](https://www.npmjs.com/package/%40rushstack/eslint-plugin)"

```js
export const StyledButton = styled('button', ...)
```

### DIFF
--- a/eslint/eslint-plugin/src/typedef-var.ts
+++ b/eslint/eslint-plugin/src/typedef-var.ts
@@ -51,6 +51,11 @@ const typedefVar: TSESLint.RuleModule<MessageIds, Options> = {
           // An explicit type declaration was provided
           return;
         }
+        
+        if (node.init?.type === 'CallExpression') {
+          // The variable is declared by a function call
+          return;
+        }
 
         // These are @typescript-eslint/typedef exemptions
         if (


iclanton:
This change kinda goes against the purpose of typedef-var. In your particular example case, someone may be able to infer the type, but this rule exists to give reviewers and people unfamiliar with a codebase context when it would otherwise be missing.

This code would now be allowed, and it shouldn't be:

```
import { fn } from 'black-box-package';

export const something = fn();
```

Realistically, nobody except the author knows the type of the `something` variable without pulling the code and installing dependencies. Requiring types avoids that necessity.

PR #4002, opened 2023-03-06
[rush] Do not delete the local pnpm store after all install attempts has failed.

iclanton:
## Summary

Fixes https://github.com/microsoft/rushstack/issues/3996

## How it was tested

The PNPM store is pretty stable at this point, and `rush purge` will still delete a local store.


dmichon-msft (approved):

D4N14L (approved):

sachinjoseph:
Finally! I have always wanted us to do this https://github.com/microsoft/rushstack/pull/1844

PR #4005, opened 2023-03-08
[heft-webpack5-plugin]update WebpackConfigurationLoader to support any valid webpack config file name

### LINKED ISSUE #3988: [heft] WebpackConfigurationLoader lacks support for .cjs file extensions
## Summary
While working to update a rush monorepo to ESM I hit an issue when attempting to update webpack config files to leverage a `.cjs` file extension. For "reasons", we'd prefer to not refactor the entire webpack config system away from require, etc...and instead just append the appropriate file extension.

It seems that [this file](https://github.com/microsoft/rushstack/blob/84ad0baf6f1308549e6ebc031f0cf38a4e46d27b/heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts#L27) has hardcoded strings for valid file names which look specifically for `.js` files.

## Repro steps
I can't share the repository, but initializing the minimum rush repo with support for ESM (module resolution of `nodenext` or `Node16`, file extensions on all import/exports, packages with `"type": "module"`. Configure webpack with a .cjs extension.

## Details
This is currently blocking our migration, so while I think a broader feature for more configuration here might be helpful, ideally this could come in with a quick fix as not being able to use a `.cjs` file extension for webpack is quite prescriptive and "buggy".  

## Standard questions

| Question | Answer |
| -------- | -------- |
| `rushVersion` from rush.json? | 5.78.0 |
| Operating system? | all |
| Node.js version (`node -v`)? | 16.18.0 |

chrisdholt:
## Summary
fixes #3988

## Details
This PR adds a function to gather any valid webpack config file (.js, .mjs, .cjs) rather than using the hardcoded values. I considered integrating this with existing methods but wasn't sure if that would be preferred.

## How it was tested
I have not tested it yet, opening this draft to figure out the best method of ensuring this works without introducing breaking changes.

### DIFF
--- a/common/changes/@rushstack/heft-webpack5-plugin/users-chhol-heft-webpack-check-for-valid-files-types_2023-03-08-18-11.json
+++ b/common/changes/@rushstack/heft-webpack5-plugin/users-chhol-heft-webpack-check-for-valid-files-types_2023-03-08-18-11.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@rushstack/heft-webpack5-plugin",
-      "comment": "adds support to find any valid webpack config file type",
+      "comment": "Add support for `.cjs` and `.mjs` file extensions for webpack config files, in addition to the `.js` file extension.",
       "type": "minor"
     }
   ],
--- a/common/changes/@rushstack/heft-webpack5-plugin/users-chhol-heft-webpack-check-for-valid-files-types_2023-03-08-18-11.json
+++ b/common/changes/@rushstack/heft-webpack5-plugin/users-chhol-heft-webpack-check-for-valid-files-types_2023-03-08-18-11.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-webpack5-plugin",
+      "comment": "adds support to find any valid webpack config file type",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@rushstack/heft-webpack5-plugin"
+}
--- a/heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
+++ b/heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
@@ -106,16 +106,21 @@ async function findWebpackConfigAsync(buildFolder: string): Promise<IWebpackConf
     const prod: string[] = [];
 
     for (const folderItem of folderItems) {
-      if (folderItem.isFile() && folderItem.name.match(/^webpack.dev.config\.(cjs|js|mjs)$/)) {
-        dev.push(folderItem.name);
-      } else if (folderItem.isFile() && folderItem.name.match(/^webpack.config\.(cjs|js|mjs)$/)) {
-        prod.push(folderItem.name);
+      if (folderItem.isFile()) {
+        if (folderItem.name.match(/^webpack.dev.config\.(cjs|js|mjs)$/)) {
+          dev.push(folderItem.name);
+        } else if (folderItem.isFile() && folderItem.name.match(/^webpack.config\.(cjs|js|mjs)$/)) {
+          prod.push(folderItem.name);
+        }
       }
     }
 
-    if (dev.length > 1 || prod.length > 1) {
-      throw new Error(`Error: Found more than one matching webpack configuration files.`);
+    if (dev.length > 1) {
+      throw new Error(`Error: Found more than one dev webpack configuration file.`);
+    } else if (prod.length > 1) {
+      throw new Error(`Error: Found more than one non-dev webpack configuration file.`);
     }
+
     return {
       dev: dev[0],
       prod: prod[0]
--- a/heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
+++ b/heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
@@ -25,12 +25,9 @@ type IWebpackConfigJsExport =
 type IWebpackConfigJs = IWebpackConfigJsExport | { default: IWebpackConfigJsExport };
 
 interface IWebpackConfigFileNames {
-  dev: string;
-  prod: string;
+  dev: string | undefined;
+  prod: string | undefined;
 }
-
-const WEBPACK_CONFIG_FILENAME: string = 'webpack.config.js';
-const WEBPACK_DEV_CONFIG_FILENAME: string = 'webpack.dev.config.js';
 
 export class WebpackConfigurationLoader {
   public static async tryLoadWebpackConfigAsync(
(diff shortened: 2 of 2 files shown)


heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
```diff
@@ -92,3 +94,18 @@ export class WebpackConfigurationLoader {
     }
   }
 }
+
+function findWebpackConfig(buildFolder: string, configurationRegExp: RegExp): string {
+  try {
+    const configs = FileSystem.readFolderItemNames(buildFolder).filter((config) =>
```
iclanton, line 100:
```suggestion
async function findWebpackConfigAsync(buildFolder: string, configurationRegExp: RegExp): Promise<string> {
  try {
    const folderItems: FolderItem = await FileSystem.readFolderItemsAsync(buildFolder);
    const configs: string[] = [];
    for (folderItem of folderItems) {
      if (folderItem.isFile() && folderItem.name.match(configurationRegExp)) {
        configs.push(folderItem.name);
      }
    }

```

heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
```diff
@@ -36,6 +36,8 @@ export class WebpackConfigurationLoader {
     // TODO: Eventually replace this custom logic with a call to this utility in in webpack-cli:
     // https://github.com/webpack/webpack-cli/blob/next/packages/webpack-cli/lib/groups/ConfigGroup.js
 
+    const WEBPACK_CONFIG_FILENAME: string = findWebpackConfig(buildFolder, WEBPACK_CONFIG_FILENAME_REGEX);
+    const WEBPACK_DEV_CONFIG_FILENAME: string = findWebpackConfig(buildFolder, WEBPACK_CONFIG_FILENAME_REGEX);
```
iclanton, line 40:
```suggestion
    const WEBPACK_CONFIG_FILENAME: string = await findWebpackConfigAsync(buildFolder, WEBPACK_CONFIG_FILENAME_REGEX);
    const WEBPACK_DEV_CONFIG_FILENAME: string = await findWebpackConfigAsync(buildFolder, WEBPACK_CONFIG_FILENAME_REGEX);
```

iclanton: This will read the folder items twice. It'd be more efficient for the function to just return both the dev and non-dev filenames.

chrisdholt: All for this and considered it initially. Let me make the changes locally - there's also a couple of tests failing I can look into. Once I get it updated and address any comments, I can move this from draft.

chrisdholt: @iclanton if we want to run this _once_ for both configs, thoughts on returning an object with keys of `dev` and `prod` in this scenario rather than an array?

heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
chrisdholt: @iclanton I've incorporated this as well as an adjustment to return an object w/ dev/prod keys respectively. A bit different than the initial approach, but serves fallbacks in the case either is undefined. Let me know your thoughts when you can.

heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
iclanton: Yeah that's how I'd do it.

chrisdholt: Should be updated with an approach that follows that. Let me know any thoughts/feedback when you get a chance to review. Thanks @iclanton!

heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
```diff
@@ -92,3 +98,29 @@ export class WebpackConfigurationLoader {
...
+    const folderItems: FolderItem[] = await FileSystem.readFolderItemsAsync(buildFolder);
+    const dev: string[] = [];
+    const prod: string[] = [];
+
+    for (const folderItem of folderItems) {
+      if (folderItem.isFile() && folderItem.name.match(/^webpack.dev.config\.(cjs|js|mjs)$/)) {
+        dev.push(folderItem.name);
+      } else if (folderItem.isFile() && folderItem.name.match(/^webpack.config\.(cjs|js|mjs)$/)) {
+        prod.push(folderItem.name);
+      }
```
iclanton, line 113:
```suggestion
      if (folderItem.isFile()) {
        if (folderItem.name.match(/^webpack.dev.config\.(cjs|js|mjs)$/)) {
          dev.push(folderItem.name);
        } else if (folderItem.name.match(/^webpack.config\.(cjs|js|mjs)$/)) {
          prod.push(folderItem.name);
        }
      }
```

heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
```diff
@@ -92,3 +98,29 @@ export class WebpackConfigurationLoader {
...
+    }
+
+    if (dev.length > 1 || prod.length > 1) {
+      throw new Error(`Error: Found more than one matching webpack configuration files.`);
+    }
```
iclanton, line 118:
```suggestion
    if (dev.length > 1) {
      throw new Error(`Error: Found more than one dev webpack configuration file.`);
    } else if (prod.length > 1) {
      throw new Error(`Error: Found more than one non-dev webpack configuration file.`);
    }

```

common/changes/@rushstack/heft-webpack5-plugin/users-chhol-heft-webpack-check-for-valid-files-types_2023-03-08-18-11.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-webpack5-plugin",
+      "comment": "adds support to find any valid webpack config file type",
```
iclanton, line 5:
```suggestion
      "comment": "Add support for `.cjs` and `.mjs` file extensions for webpack config files, in addition to the `.js` file extension.",
```

iclanton (approved):

eljefe223 (approved):

heft-plugins/heft-webpack5-plugin/src/WebpackConfigurationLoader.ts
```diff
@@ -92,3 +98,34 @@ export class WebpackConfigurationLoader {
...
+  try {
+    const folderItems: FolderItem[] = await FileSystem.readFolderItemsAsync(buildFolder);
+    const dev: string[] = [];
+    const prod: string[] = [];
+
+    for (const folderItem of folderItems) {
+      if (folderItem.isFile()) {
+        if (folderItem.name.match(/^webpack.dev.config\.(cjs|js|mjs)$/)) {
+          dev.push(folderItem.name);
+        } else if (folderItem.isFile() && folderItem.name.match(/^webpack.config\.(cjs|js|mjs)$/)) {
```
iclanton, line 112:
```suggestion
        } else if (folderItem.name.match(/^webpack.config\.(cjs|js|mjs)$/)) {
```

chrisdholt: AH dangit - I missed it. @iclanton mind if we commit here?

iclanton (approved):

PR #4009, opened 2023-03-14
[rush] Fix an issue with running 'rush update-autoinstaller'.

iclanton:
## Summary

There is an issue with `rush update-autoinstaller` where it doesn't ensure the package manager is installed before trying to invoke it in the autoinstaller's folder. This PR fixes that and ensure that the autoinstaller updater uses the same `RushGlobalFolder` as other installers.

## How it was tested

Ran `rush purge` and then `rush update-autoinstaller` in a test repo with a change to an autoinstalled package.


octogonz (approved):
Lgtm

PR #4010, opened 2023-03-14
[rush] Exposing executionRecords in PhasedScriptsAction allows users to observe the execution process

sherlockfeng:
## Summary
Exposing executionRecords in PhasedScriptsAction allows users to observe the execution process

## Details
As a member of the monorepo infrastructure team, there will be many compilations in progress at the same time, 
 I hope to be able to collect the real-time status of the rush execution command. If I can get the real-time status in the hooks of rushLifeCycle, I can observe the start and end times of all project compilation in a task, and then send it to the data server to build a dashboard that can view the execution status in real time

## How to use
```
rushSession.hooks.runAnyPhasedCommand.tap(PLUGIN_NAME, (phase: ReadonlyMap<Operation, IOperationExecutionResult>) => {
    phase.hooks.beforeExecuteOperations.tapPromise(PLUGIN_NAME, async (records) => {
        // all operation records
    });
    phase.hooks.onOperationStatusChanged.tap(PLUGIN_NAME, (record: IOperationExecutionResult) => {
        // statue changed operation record
    });
});
```
### Demo use local data
image

### DIFF
--- a/common/changes/@microsoft/rush/feat-periodically-send-operation-state_2023-03-17-02-07.json
+++ b/common/changes/@microsoft/rush/feat-periodically-send-operation-state_2023-03-17-02-07.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "add hooks to PhasedCommandHooks to give user the operation records",
+      "comment": "Include some more hooks to allow plugins to monitor phased command execution in real-time.",
       "type": "none"
     }
   ],
--- a/common/changes/@microsoft/rush/feat-periodically-send-operation-state_2023-03-17-02-07.json
+++ b/common/changes/@microsoft/rush/feat-periodically-send-operation-state_2023-03-17-02-07.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "add hooks to PhasedCommandHooks to give user the operation records",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
+++ b/libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
@@ -28,6 +28,7 @@ import { BuildCacheConfiguration } from '../../api/BuildCacheConfiguration';
 import { SelectionParameterSet } from '../parsing/SelectionParameterSet';
 import type { IPhase, IPhasedCommandConfig } from '../../api/CommandLineConfiguration';
 import { Operation } from '../../logic/operations/Operation';
+import { OperationExecutionRecord } from '../../logic/operations/OperationExecutionRecord';
 import { PhasedOperationPlugin } from '../../logic/operations/PhasedOperationPlugin';
 import { ShellOperationRunnerPlugin } from '../../logic/operations/ShellOperationRunnerPlugin';
 import { Event } from '../../api/EventHooks';
@@ -340,7 +341,13 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
       quietMode: isQuietMode,
       debugMode: this.parser.isDebug,
       parallelism,
-      changedProjectsOnly
+      changedProjectsOnly,
+      beforeExecuteOperations: async (records: Map<Operation, OperationExecutionRecord>) => {
+        await this.hooks.beforeExecuteOperations.promise(records);
+      },
+      onOperationStatusChanged: (record: OperationExecutionRecord) => {
+        this.hooks.onOperationStatusChanged.call(record);
+      }
     };
 
     const internalOptions: IRunPhasesOptions = {
--- a/libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
+++ b/libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
@@ -18,6 +18,9 @@ export interface IOperationExecutionManagerOptions {
   parallelism: number;
   changedProjectsOnly: boolean;
   destination?: TerminalWritable;
+
+  onOperationStatusChanged?: (record: OperationExecutionRecord) => void;
+  beforeExecuteOperations?: (records: Map<Operation, OperationExecutionRecord>) => Promise<void>;
 }
 
 /**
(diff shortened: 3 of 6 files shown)


dmichon-msft:
I needed a way to expose this data in #3651 as well, and at the time it seemed more useful to add an additional hook to `PhasedOperationHooks` that gets invoked whenever the state of an individual operation changes, rather than having to poll the execution. See these callbacks here: https://github.com/microsoft/rushstack/pull/3651/files#diff-454747232f90e3a30fe693116f4913b6ca94cdbe7e9eb5b4a94e744b69fc6c3dR25-R26

I'd prefer not to add additional API surface to the PhasedScriptAction class and have any extensibility mechanisms be via the PhasedOperationHooks object for consistency.

libraries/rush-lib/src/pluginFramework/RushLifeCycle.ts
```diff
@@ -35,6 +36,12 @@ export interface IPhasedCommand extends IRushCommand {
    * @alpha
    */
   readonly hooks: PhasedCommandHooks;
+
+  /**
+   * Function that return current phase command result
+   * @alpha
+   */
+  readonly getCurrentExecutionResult: () => IExecutionResult['operationResults'] | undefined;
```
dmichon-msft, line 44: Would rather add hooks to `PhasedCommandHooks` to give you the records immediately after initialization, and to notify you when the state of a record changes.

sherlockfeng: yeah, i agree this is a better solution to achieve this requirement.

libraries/rush-lib/src/pluginFramework/PhasedCommandHooks.ts
```diff
@@ -87,6 +88,23 @@ export class PhasedCommandHooks {
   public readonly createOperations: AsyncSeriesWaterfallHook<[Set<Operation>, ICreateOperationsContext]> =
     new AsyncSeriesWaterfallHook(['operations', 'context'], 'createOperations');
 
+  /**
+   * Hook invoked when operation statue changed
+   * Use the context to distinguish between the initial run and phased runs.
+   * Hook is series for stable output.
+   */
+  public readonly beforeExecuteOperations: AsyncSeriesHook<[Map<Operation, OperationExecutionRecord>]> =
```
dmichon-msft, line 96: Favor the `IOperationExecutionResult` interface over the internal `OperationExecutionRecord` class; it's meant to hide some internals and avoid having to export quite as many types, as well as making explicit that some types are meant to be readonly.

sherlockfeng: Done~

libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -340,7 +341,13 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
       quietMode: isQuietMode,
       debugMode: this.parser.isDebug,
       parallelism,
-      changedProjectsOnly
+      changedProjectsOnly,
+      beforeExecuteOperations: async (records: Map<Operation, OperationExecutionRecord>) => {
+        await this.hooks.beforeExecuteOperations.promise(records);
+      },
+      onOperationStatusChanged: async (record: OperationExecutionRecord) => {
```
dmichon-msft, line 348: Since this particular hook is purely informational and we aren't waiting for this callback to complete, we should be clear about that in the hook type and make it a `SyncHook`, and use `(record: OperationExecutionRecord) => this.hooks.onOperationStatusChanged.call(record)`

libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -183,6 +200,8 @@ export class OperationExecutionManager {
       prioritySort
     );
 
+    this._beforeExecuteOperations?.(this._executionRecords);
```
dmichon-msft, line 203:
```suggestion
    await this._beforeExecuteOperations?.(this._executionRecords);
```

libraries/rush-lib/src/logic/operations/OperationExecutionManager.ts
```diff
@@ -18,6 +18,9 @@ export interface IOperationExecutionManagerOptions {
   parallelism: number;
   changedProjectsOnly: boolean;
   destination?: TerminalWritable;
+
+  onOperationStatusChanged?: (record: OperationExecutionRecord) => void;
+  beforeExecuteOperations?: (records: Map<Operation, OperationExecutionRecord>) => void;
```
dmichon-msft, line 23:
```suggestion
  beforeExecuteOperations?: (records: Map<Operation, OperationExecutionRecord>) => Promise<void>;
```

dmichon-msft (commented):
Let's avoid floating promise issues. Either await async hooks or use sync hooks, depending on if it makes sense for the given hook to potentially modify state.

Edit: otherwise, looks about ready.

libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
sherlockfeng: yeah, I should think about which one needs to wait for completion before implementing these two hooks

dmichon-msft (approved):
LGTM

libraries/rush-lib/src/pluginFramework/PhasedCommandHooks.ts
```diff
@@ -87,6 +87,19 @@ export class PhasedCommandHooks {
...
 
+  /**
+   * Hook invoked before operation start
+   * Hook is series for stable output.
+   */
+  public readonly beforeExecuteOperations: AsyncSeriesHook<[Map<Operation, IOperationExecutionResult>]> =
+    new AsyncSeriesHook(['records']);
+
+  /**
+   * Hook invoked when operation statue changed
```
dmichon-msft, line 98:
```suggestion
   * Hook invoked when operation status changed
```

common/changes/@microsoft/rush/feat-periodically-send-operation-state_2023-03-17-02-07.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "add hooks to PhasedCommandHooks to give user the operation records",
```
iclanton, line 5:
```suggestion
      "comment": "Include some more hooks to allow plugins to monitor phased command execution in real-time.",
```

iclanton (approved):

PR #4012, opened 2023-03-14
[rush-lib] Extend LookupByPath to return index of remaining path

dmichon-msft:
## Summary
Enhances the `LookupByPath` helper class with a method `findChildPathAndIndex` that returns both the matched entry and the index at which the remainder of the query path starts.

## Details
Useful for `ProjectChangeAnalyzer` when routing files to projects, to get the remaining path relative to the project.

## How it was tested
Added unit tests.

## Impacted documentation
None

### DIFF
--- a/libraries/rush-lib/src/logic/LookupByPath.ts
+++ b/libraries/rush-lib/src/logic/LookupByPath.ts
@@ -15,6 +15,16 @@ interface IPathTreeNode<TItem> {
   children: Map<string, IPathTreeNode<TItem>> | undefined;
 }
 
+interface IPrefixEntry {
+  prefix: string;
+  index: number;
+}
+
+export interface IPrefixMatch<TItem> {
+  value: TItem;
+  index: number;
+}
+
 /**
  * This class is used to associate POSIX relative paths, such as those returned by `git` commands,
  * with entities that correspond with ancestor folders, such as Rush Projects.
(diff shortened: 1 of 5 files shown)


libraries/rush-lib/src/logic/LookupByPath.ts
```diff
@@ -146,6 +170,23 @@ export class LookupByPath<TItem> {
...
+   * @returns the found item, or `undefined` if no item was found
+   *
+   * @example
+   * ```ts
+   * const tree = new LookupByPath([['foo', 1], ['foo/bar', 2]]);
+   * tree.findChildPathAndIndex('foo/baz'); // returns { item: 1, index: 4 }
+   * tree.findChildPathAndIndex('foo/bar/baz'); // returns { item: 2, index: 8 }
+   * ```
+   */
+  public findChildPathAndIndex(childPath: string): IPrefixMatch<TItem> | undefined {
```
D4N14L, line 186:
```suggestion
  public findNearestChildPathAndIndex(childPath: string): IPrefixMatch<TItem> | undefined {
```

D4N14L (approved):

libraries/rush-lib/src/logic/LookupByPath.ts
dmichon-msft: Is this a request for a general renaming of the API surface, or just this specific new method?

D4N14L: General renaming of the API surface. This is a new API, just feel that the name suggested fits what it does better than the name it has is all

dmichon-msft: Not renaming the existing API for now since some external code is known to take a dependency.

D4N14L (approved):

PR #4023, opened 2023-03-21
[api-extractor] support .d.mts and .d.cts files

ds300:
## Summary

TypeScript supports emitting and loading typings with `.d.cts` and `.d.mts` file extensions. We published a package which uses these extensions for its typings file, but it broke our api-extractor usage so we applied this patch.

## Details

This was the only place I found that specifically hard-coded the `.d.ts` requirement.

## How it was tested

Sorry I don't have time to dig into how you're doing testing in this repo.

## Impacted documentation

I don't think this affects any docs, or needs to be explicitly document itself.

### DIFF
--- a/apps/api-extractor/src/api/ExtractorConfig.ts
+++ b/apps/api-extractor/src/api/ExtractorConfig.ts
@@ -210,7 +210,7 @@ export class ExtractorConfig {
     path.join(__dirname, '../schemas/api-extractor-defaults.json')
   );
 
-  private static readonly _declarationFileExtensionRegExp: RegExp = /\.d\.ts$/i;
+  private static readonly _declarationFileExtensionRegExp: RegExp = /\.d\.(c|m)?ts$/i;
 
   /** {@inheritDoc IConfigFile.projectFolder} */
   public readonly projectFolder: string;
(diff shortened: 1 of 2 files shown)


apps/api-extractor/src/api/ExtractorConfig.ts
```diff
@@ -210,7 +210,7 @@ export class ExtractorConfig {
     path.join(__dirname, '../schemas/api-extractor-defaults.json')
   );
 
-  private static readonly _declarationFileExtensionRegExp: RegExp = /\.d\.ts$/i;
```
dmichon-msft, line 213: TypeScript 5 is also adding the ability to reference `.d.*.ts` for typings for non-js files, with `.d.js.ts` technically a valid alternative to standard `.d.ts`. We may want to add that consideration here as well.

ds300: Do you want me to do that?

JoshuaKGoldberg:
👋 friendly ping @dmichon-msft, is there anything we can do to get this merged in?

apps/api-extractor/src/api/ExtractorConfig.ts
dmichon-msft: We can deal with that separately.

dmichon-msft (approved):

JoshuaKGoldberg:
cc ... @iclanton? Is there someone we can bug about this? 🙂

JoshuaKGoldberg:
🙌 thanks!

PR #4025, opened 2023-03-22
[rush] Fix an issue with update-autoinstaller.

iclanton:
## Summary

I recently fixed another issue with `rush update-autoinstaller` where it would fail if the package manager wasn't already installed (https://github.com/microsoft/rushstack/pull/4009). This introduced a regression where Rush would attempt to install autoinstallers with `--frozen-lockfile` before updating them, which doesn't work if their dependencies change.

This PR fixes that issue by explicitly installing the package manager during the autoinstaller's update function instead.

## How it was tested

Tested in a repo that was failing to update the autoinstaller before.

### DIFF
--- a/common/changes/@microsoft/rush/main_2023-03-22-19-25.json
+++ b/common/changes/@microsoft/rush/main_2023-03-22-19-25.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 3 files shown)


dmichon-msft (approved):
LGTM

PR #4026, opened 2023-03-22
[rush] Fix an issue where rush plugin autoinstallers would fail to install because the Rush global folder had not yet been initialized.

iclanton:


dmichon-msft (approved):
LGTM

PR #4030, opened 2023-03-24
[heft] introduce shell script plugin

robertf224:
Adding a shell script plugin for Heft (relevant past discussion about this here https://github.com/microsoft/rushstack/issues/3519).  Whipped this up internally at my company a few weeks ago / figured I would port here.  Could use help on (1) how to best add tests for something like this (2) where to document.

You can see the plugin schema [here](https://github.com/robertf224/rushstack/blob/7683a19fbfc3827d70bab0d7ec7241f5e64a6dd5/heft-plugins/heft-shell-script-plugin/src/schemas/heft-shell-script-plugin.schema.json), rough idea is you give it a command you want to run, and what build stage you'd like to run the command in.  You can also give it either a separate command to run in watch mode (which i.e. might be the same as your main command with `--watch` appended, obvi depends on the tool) or a list of globs to watch if the tool you're proxying doesn't have a native watch mode.

Example `heft.json` wrapping `protoc` from within our internal repo:
```json
{
    "$schema": "https://developer.microsoft.com/json-schemas/heft/heft.schema.json",
    "extends": "../node_modules/@chapter/heft-rig/profiles/node/config/heft.json",
    "heftPlugins": [
        {
            "plugin": "@chapter/heft-shell-script-plugin",
            "options": {
                "scripts": [
                    {
                        "name": "grpc-codegen",
                        "stage": "pre-compile",
                        "command": "mkdir -p ./src/__generated__/proto && protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto --ts_proto_opt=env=node,outputServices=grpc-js,esModuleInterop=true --ts_proto_out=./src/__generated__/proto --proto_path=./src/proto $(find ./src/proto -iname '*.proto')",
                        "watchGlobs": ["./src/proto/*.proto"]
                    }
                ]
            }
        }
    ]
}

```


heft-plugins/heft-shell-script-plugin/src/ShellScriptPlugin.ts
```diff
@@ -0,0 +1,99 @@
...
+            } else {
+              logger.terminal.writeWarningLine('No watch configuration specified.');
+            }
+          }
+        } catch (error) {
+          logger.emitError(error instanceof Error ? error : new Error('Unknown error.'));
+        }
+      };
+
+      switch (shellScript.stage) {
```
robertf224, line 52: this is obvi kinda jank but wasn't sure how else to add the hooks

dmichon-msft:
This looks pretty solid for the current version of Heft; issue is we're pretty close to merging the major refactor of Heft to generalize the scheduling process and that'll require a rather different way of hooking into the pipeline, so it doesn't feel quite worth merging this only to immediately redo it for compatibility. See migration documentation here: https://github.com/microsoft/rushstack/blob/release/heft-0.50.0-rc/apps/heft/UPGRADING.md

Watch mode documentation is currently a bit out of date, but the non-watch case should be quite straightforward to migrate. For watch mode, the new version of Heft no longer supports processes just performing rebuilds willy-nilly and there will instead need to be a way for the process being invoked to communicate to the Heft host that it wants to rebuild, then be notified when it is safe to do so. This is to ensure that the build pipeline steps happen in the expected sequence when changes occur.

robertf224:
ah gotcha, I'll just wait until that merges and refactor then 👍 

for my understanding, this means when wrapping most tools you'll need to pay the cost of startup on each rebuild even if the tool has a native watch mode?  obvi not a big deal for rust-based/etc. tools which seems to be where most of the ecosystem is headed, but for node-based tools might be more of a factor

robertf224:
would you rather leave this PR open or close and I'll open a new one later for the redo?

dmichon-msft:
Feel free to just leave it open.

Faithfinder:
Let me do a "bump" here, now that heft upgrade is done

heft-plugins/heft-shell-script-plugin/LICENSE
```diff
@@ -0,0 +1,24 @@
+@rushstack/heft-dev-cert-plugin
```
iclanton, line 1: Update

heft-plugins/heft-shell-script-plugin/README.md
```diff
@@ -0,0 +1,40 @@
+# @rushstack/heft-shell-script-plugin
+
+This is a Heft plugin to run shell scripts within your project.
+
+## Usage
+
+You can see the plugin schema [here](./src/schemas/heft-shell-script-plugin.schema.json).  For each shell script you would like to run, you give the plugin a shell command you would like to run via the `command` argument, and what build stage you would like to run the command in via the `stage` argument.  You can also give the plugin either a separate command to run in watch mode via the `watchCommand` argument (which i.e. might be the same as your `command` argument with --watch appended depending on the tool you're using) or a list of globs to watch via the `watchGlobs` argument if the tool you're using doesn't have a native watch mode.
```
iclanton, line 7: I don't think this link will work on npmjs.org

heft-plugins/heft-shell-script-plugin/README.md
```diff
@@ -0,0 +1,40 @@
...
+    "extends": "../node_modules/@rushstack/heft-node-rig/profiles/default/config/heft.json",
+    "heftPlugins": [
+        {
+            "plugin": "@rushstack/heft-shell-script-plugin",
+            "options": {
+                "scripts": [
+                    {
+                        "name": "grpc-codegen",
+                        "stage": "pre-compile",
+                        "command": "mkdir -p ./src/__generated__/proto && protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto --ts_proto_opt=env=node,outputServices=grpc-js,esModuleInterop=true --ts_proto_out=./src/__generated__/proto --proto_path=./src/proto $(find ./src/proto -iname '*.proto')",
```
iclanton, line 24: Given this example, it seems like it would be useful to be able to pass in an array of commands that are executed in order unless one fails instead of needing the `&&` operator

heft-plugins/heft-shell-script-plugin/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "start": "heft test --clean --watch",
+    "_phase:build": "heft build --clean",
+    "_phase:test": "heft test --no-build"
+  },
+  "peerDependencies": {
+    "@rushstack/heft": "^0.50.0"
+  },
+  "dependencies": {
+    "@rushstack/node-core-library": "workspace:*",
+    "chokidar": "~3.4.0"
```
iclanton, line 25: Heft now has watching built-in, so you shouldn't need chokidar

heft-plugins/heft-shell-script-plugin/src/ShellScript.ts
```diff
@@ -0,0 +1,13 @@
+/**
```
iclanton, line 1: Missing copyright header.

heft-plugins/heft-shell-script-plugin/src/ShellScriptPlugin.ts
```diff
@@ -0,0 +1,102 @@
+import type { IHeftPlugin, HeftSession, HeftConfiguration, ScopedLogger } from '@rushstack/heft';
```
iclanton, line 1: Missing copyright header.

heft-plugins/heft-shell-script-plugin/src/ShellScriptPlugin.ts
```diff
@@ -0,0 +1,102 @@
...
+const PLUGIN_SCHEMA_PATH: string = `${__dirname}/schemas/heft-shell-script-plugin.schema.json`;
+
+/**
+ * @public
+ */
+export interface IShellScriptPluginOptions {
+  scripts: ShellScript[];
+}
+
+export class ShellScriptPlugin implements IHeftPlugin<IShellScriptPluginOptions> {
```
iclanton, line 17: This looks like it's using the old Heft plugin format.

heft-plugins/heft-shell-script-plugin/src/utils.ts
```diff
@@ -0,0 +1,30 @@
+import { HeftConfiguration } from '@rushstack/heft';
```
iclanton, line 1: Missing copyright header.

heft-plugins/heft-shell-script-plugin/src/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+import { SubprocessTerminator } from '@rushstack/node-core-library';
+import child_process from 'child_process';
+import path from 'path';
+
+export function runShellCommand(heftConfiguration: HeftConfiguration, command: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const shellScriptProcess: child_process.ChildProcess = child_process.spawn(command, {
```
iclanton, line 8: Look at `Executable.spawn` from `node-core-library`

heft-plugins/heft-shell-script-plugin/src/utils.ts
```diff
@@ -0,0 +1,30 @@
+import { HeftConfiguration } from '@rushstack/heft';
+import { SubprocessTerminator } from '@rushstack/node-core-library';
+import child_process from 'child_process';
+import path from 'path';
+
+export function runShellCommand(heftConfiguration: HeftConfiguration, command: string): Promise<void> {
```
iclanton, line 6:
```suggestion
export function runShellCommandAsync(heftConfiguration: HeftConfiguration, command: string): Promise<void> {
```

heft-plugins/heft-shell-script-plugin/src/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+        ...process.env,
+        PATH: `${path.join(heftConfiguration.buildFolder, 'node_modules', '.bin')}:${process.env.PATH}`
+      },
+      ...SubprocessTerminator.RECOMMENDED_OPTIONS
+    });
+    SubprocessTerminator.killProcessTreeOnExit(shellScriptProcess, SubprocessTerminator.RECOMMENDED_OPTIONS);
+    shellScriptProcess.on('error', (error) => {
+      reject(error);
+    });
+    shellScriptProcess.on('exit', (code) => {
```
iclanton, line 22: You can use `once` from `events` here and just await that.

iclanton (changes requested):

PR #4042, opened 2023-03-30
[rush] Allow ProjectChangeAnalyzer to acccept an ignore glob

elliot-nelson:
## Summary

When writing tools that use ProjectChangeAnalyzer, it would be very convenient to be able to pass a custom list of ignore globs that can augment (or override) the existing project-specific ignore globs.

Add an option that can be used by monorepo tooling (but isn't exposed in any way to the rush command-line) to support this.

## Details

Here's the scenario that spawned this PR: in pull requests, we want to run `terraform plan` on all projects that have impacted infrastructure. However, due to the way terraforming has been designed across the organization, we need to do so by using a reusable workflow that we don't control. This means it's _really important_ to us not to run plan unless we think there's a _good chance_ that infrastructure has been impacted.

Our criteria for this is as follows:

 - List all projects that have git diffs (compared to same target e.g. `main`)
 - Filter this list to _only_ projects that have touched files in the relative folder `infra/terraform`
 - From this list, calculate "impacted by" (project and all recursive dependents)
 - From this list, filter to only projects tagged `terraform`.

(We use the inherit Rush project structure to list infrastructure components as "dependencies" of consuming projects; this allows us to find all projects that might need to plan infrastructure changes, if any project they depend on has changes in the folder where infrastructure lives.)

### DIFF
--- a/libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
+++ b/libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
@@ -42,6 +42,15 @@ export interface IGetChangedProjectsOptions {
    * and exclude matched files from change detection.
    */
   enableFiltering: boolean;
+
+  /**
+   * If provided, this array of strings will be interpreted as a list of globs to ignore, for this call
+   * only. This option can be used for customized filtering of project change lists.
+   *
+   * If project filtering is enabled, the provided list will be applied AFTER the individual project's
+   * ignore globs, if any.
+   */
+  ignoreGlobs?: string[];
 }
 
 interface IGitState {
@@ -117,6 +126,7 @@ export class ProjectChangeAnalyzer {
       project,
       unfilteredProjectData,
       rootDir,
+      [],
       terminal
     );
 
@@ -182,11 +192,17 @@ export class ProjectChangeAnalyzer {
     project: RushConfigurationProject,
     unfilteredProjectData: Map<string, T>,
     rootDir: string,
+    additionalIgnoreGlobs: string[],
     terminal: ITerminal
   ): Promise<Map<string, T>> {
-    const ignoreMatcher: Ignore | undefined = await this._getIgnoreMatcherForProjectAsync(project, terminal);
+    let ignoreMatcher: Ignore | undefined = await this._getIgnoreMatcherForProjectAsync(project, terminal);
     if (!ignoreMatcher) {
-      return unfilteredProjectData;
+      if (additionalIgnoreGlobs.length === 0) {
+        return unfilteredProjectData;
+      } else {
+        ignoreMatcher = ignore();
+        ignoreMatcher.add(additionalIgnoreGlobs);
+      }
     }
 
     const projectKey: string = path.relative(rootDir, project.projectFolder);
@@ -217,6 +233,7 @@ export class ProjectChangeAnalyzer {
     const { _rushConfiguration: rushConfiguration } = this;
 
     const { targetBranchName, terminal, includeExternalDependencies, enableFiltering, shouldFetch } = options;
+    const ignoreGlobs: string[] = options.ignoreGlobs || [];
 
     const gitPath: string = this._git.getGitPathOrThrow();
     const repoRoot: string = getRepoRoot(rushConfiguration.rushJsonFolder);
@@ -292,7 +309,7 @@ export class ProjectChangeAnalyzer {
           continue;
         }
 
-        if (enableFiltering) {
+        if (enableFiltering || ignoreGlobs.length > 0) {
           let projectChanges: Map<string, IFileDiffStatus> | undefined = changesByProject.get(project);
           if (!projectChanges) {
             projectChanges = new Map();
@@ -305,7 +322,7 @@ export class ProjectChangeAnalyzer {
       }
     }
 
-    if (enableFiltering) {
+    if (enableFiltering || ignoreGlobs.length > 0) {
       // Reading rush-project.json may be problematic if, e.g. rush install has not yet occurred and rigs are in use
       await Async.forEachAsync(
         changesByProject,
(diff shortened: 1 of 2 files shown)


dmichon-msft:
I'd be more a fan of removing the glob handling from `ProjectChangeAnalyzer` altogether and instead take as input a `filterByProject?: Map<RushConfigurationProject, (relativePath: string) => boolean>` parameter (if not provided, default filter is `(relativePath: string) => true`). Said map could get constructed by reading the `RushProjectConfiguration files`, but reading config files shouldn't really be in the scope of `ProjectChangeAnalyzer`.
It also comes up because for the existing filtering, the ignore/additional globbing will need to be scoped to a specific phase, not an entire project.

elliot-nelson:
@dmichon-msft Makes sense to me!

Maybe even `filterByProject?: (project: RushConfigurationProject, relativePath: string) => boolean`?

I realize either approach would allow Rush to lazy-lookup the project configuration, but I think a single function that gets passed both args will be a lot easier for casual tool developers to reason about.

dmichon-msft:
@elliot-nelson The `filterByProject?: (project: RushConfigurationProject, relativePath: string) => boolean` syntax would work fine for this particular API. Since the per-operation deps scenario is going to need to support querying the same project data multiple times with different filters, it's probably not worth worrying about optimizing for the ability to query project, then filter, and skip the filter calls if the project didn't return a filter. For `getChangedProjectsAsync` the filter will only be passed changed files and only needs to be called on files from a particular project until it finds at least one match.

I can go either way between both arguments in one function or having a function/map that takes the project and returns a filter for later querying.

libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -178,17 +196,29 @@ export class ProjectChangeAnalyzer {
...
   ): Promise<Map<string, T>> {
-    const ignoreMatcher: Ignore | undefined = await this._getIgnoreMatcherForProjectAsync(project, terminal);
-    if (!ignoreMatcher) {
+    let ignoreMatcher: Ignore | undefined = await this._getIgnoreMatcherForProjectAsync(project, terminal);
+    if (!ignoreMatcher && !filterByProject) {
       return unfilteredProjectData;
     }
+    if (!ignoreMatcher) {
+      ignoreMatcher = ignore();
+    }
```
dmichon-msft, line 217: I seem to recall this call being somewhat expensive, which is why `_getIgnoreMatcherForProjectAsync` returns `undefined` in the first place.

libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -178,17 +196,29 @@ export class ProjectChangeAnalyzer {
...
+   * "ignore globs" settings, as well as an optional filtering function, returning a new list of
+   * changed files that have passed through the filter.
+   */
   public async _filterProjectDataAsync<T>(
     project: RushConfigurationProject,
     unfilteredProjectData: Map<string, T>,
     rootDir: string,
-    terminal: ITerminal
+    terminal: ITerminal,
+    filterByProject?: ProjectFilterFactory
```
dmichon-msft, line 209: This should just be the filter itself (which is synchronous), which the caller is expected to provide in its entirety (i.e. including any behavior from `RushProjectConfiguration`). The motivation is to make the filtering call be completely synchronous and not perform any file system calls.

libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -292,20 +323,16 @@ export class ProjectChangeAnalyzer {
...
-          }
-          projectChanges.set(file, diffStatus);
-        } else {
-          changedProjects.add(project);
+        let projectChanges: Map<string, IFileDiffStatus> | undefined = changesByProject.get(project);
+        if (!projectChanges) {
+          projectChanges = new Map();
+          changesByProject.set(project, projectChanges);
         }
+        projectChanges.set(file, diffStatus);
```
dmichon-msft, line 331: The motivation to take the filters as a parameter is that the logic here can become:
```suggestion
      const filterForProject: ((relativePath: string) => boolean) | undefined = await options.filterByProject(project);
      if (filterByProject) {
        const projectRelativePath: string = file.slice(match.index + 1);
        if (filterByProject(projectRelativePath)) {
          changedProjects.add(project);
        }
      } else {
        changedProjects.add(project);
      }
```

dmichon-msft: You'll need to use `lookup.findLongestPrefixMatch` instead of `lookup.findChildPath`. It returns `{ value: RushConfigurationProject, index: number } | undefined`.

libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -42,8 +42,26 @@ export interface IGetChangedProjectsOptions {
...
+  /**
+   * If provided, this filtering function will restrict the set of changed files, and only projects
+   * containing files that pass the filter will be included.
+   *
+   * The filtering function accepts a project as an argument, and should return a promise for a function that
+   * accepts a path as a string and returns a boolean. This "factory pattern" allows the caller to
+   * perform additional asynchronous logic (loading configuration files, etc.) when encountering a
+   * project for the first time.
+   */
+  filterByProject?: ProjectFilterFactory;
```
dmichon-msft, line 55: `filterByProject` should be completely replacing `enableFiltering`. If you want you can provide a helper method that provides an implementation of `filterByProject` backed by `RushProjectConfiguration`, though. The intent is to make it so that ProjectChangeAnalyzer no longer contains any reference to `RushProjectConfiguration` internally.

PR #4046, opened 2023-04-02
[rush] Rush version can generate version manifest

### LINKED ISSUE #3957: [rush] Versioning + Publishing Data Reports
## Summary

The commands `rush version` and `rush publish` should accept an optional report parameter, which will produce a structured report of the actions taken by the command, for consumption by following commands in a CI/CD pipeline.

## Details

As a monorepo maintainer, I may have various jobs that produce many published packages, but checking the published packages of a recent run involves scrolling through lots of job logs.

For example, after running this command:

```console
rush version --bump -b main --report common/temp/publish/version.json
```

This should produce a `version.json` file that looks like this:

```json
{
  "bumped": [
    {
      "name": "@acme/runner",
      "priorVersion": "0.3.0",
      "version": "0.4.0",
      "tag": "@acme/runner_v0.4.0",
      "date": "Tue, 09 Jan 2018 17:05:51 GMT",
      "comments": {
        "patch": [
          {
            "author": "Somebody",
            "commit": "d00b6549d13610fbb6f84be3478b532be9da0747",
            "comment": "Runner is 50% more awesome"
          }
        ]
      },
      "rendered": "## 0.4.0\nTue, 09 Jan 2018 17:05:51 GMT\n\n### Patches\n\n - Runner is 50% more awesome"
    },
    {
      "name": "@acme/dynamite",
      "priorVersion": "0.5.0",
      "version": "1.0.0",
      "tag": "@acme/dynamite_v.1.0.0",
      "date": "Tue, 09 Jan 2018 17:05:51 GMT",
      "comments": {
        "major": [
          {
            "author": "Somebody",
            "commit": "d00b6549d13610fbb6f84be3478b532be9da0747",
            "comment": "Dynamite no longer explodes randomly"
          }
        ]
      },
      "rendered": "## 1.0.0\nTue, 09 Jan 2018 15:05:51 GMT\n\n### Major\n\n - Dynamite no longer explodes randomly"
    }
  ]
}
```

The CI/CD pipeline can then produce a nice markdown table with a few extra lines of formatting (in a GHA workflow, for example, this would be appended to the `$GITHUB_STEP_SUMMARY`):

---

### Published Packages

Name | Version | Changelog
--- | --- | ---
@acme/runner | 0.4.0 | <h3>0.4.0</h3><h4>Patches</h4><ul><li>Runner is 50% more awesome</li></ul>
@acme/dynamite | 1.0.0 | <h3>1.0.0</h3><h4>Major</h4><ul><li>Dynamite no longer explodes randomly</li></ul>

---

The JSON structure above is nearly identical to the format generated for CHANGELOG.json, with the addition of a couple fields, so it should be relatively easy to output.

The file produced by `rush publish` should be similar, but without the `changes` entries and rendered markdown -- it's simply a list of published packages and their versions.

## Standard questions

| Question | Answer |
| -------- | -------- |
| Operating system? |  |
| Would you consider contributing a PR? |  Yes |

elliot-nelson:
## Summary

Allow the `rush version` command to optionally produce a new "version manifest" file, which describes the results of a _versioning event_ (in this case, version bumping of any projects during rush version).

This PR fixes #3957.

## Details

Create a schema for the _version manifest_, as well as a simple file save / file load class.

I've wired up just enough to show off the `rush version` command producing the report, but not all the details are there -- I can fill in the additional wiring if people feel the overall schema and naming make sense.

> Notes: there's a lot of possible names for this thing: it's a version _report_, it's a publishing _plan_, etc. I think "version manifest" might be the best option because it is, essentially, just a list of the projects we've versioned, without putting any spin on what you might do with it next (for example: convert it into a nice markdown report; use it as an input for rush publishing; use it as a github actions matrix for an infrastructure deployment workflow; etc.).

## How it was tested

 - In local monorepo, run: `(rush) version --bump --manifest-file=common/temp/manifest.json` produces the expected manifest.

## Impacted documentation

 - Add "version manifest" to the list of JSON schemas

### DIFF
--- a/libraries/rush-lib/src/logic/VersionManager.ts
+++ b/libraries/rush-lib/src/logic/VersionManager.ts
@@ -54,7 +54,8 @@ export class VersionManager {
   }
 
   /**
-   * Bumps versions following version policies.
+   * Bumps versions following version policies. Returns a map of updated packages and the corresponding
+   * updated Package JSON data. If there is not at least one updated package, returns undefined instead.
    *
    * @param lockStepVersionPolicyName - a specified lock step version policy name. Without this value,
    * versions for all lock step policies and all individual policies will be bumped.
@@ -68,7 +69,7 @@ export class VersionManager {
     bumpType?: BumpType,
     identifier?: string,
     shouldCommit?: boolean
-  ): Promise<void> {
+  ): Promise<Map<string, IPackageJson> | undefined> {
     // Bump all the lock step version policies.
     this._versionPolicyConfiguration.bump(lockStepVersionPolicyName, bumpType, identifier, shouldCommit);
 
@@ -88,9 +89,11 @@ export class VersionManager {
     );
 
     changeManager.load(this._rushConfiguration.changesFolder);
-    if (changeManager.hasChanges()) {
+    const updatedPackages: Map<string, IPackageJson> | undefined = changeManager.apply(!!shouldCommit);
+
+    if (updatedPackages) {
       changeManager.validateChanges(this._versionPolicyConfiguration);
-      changeManager.apply(!!shouldCommit)!.forEach((packageJson) => {
+      updatedPackages.forEach((packageJson) => {
         this.updatedProjects.set(packageJson.name, packageJson);
       });
       changeManager.updateChangelog(!!shouldCommit);
@@ -101,6 +104,8 @@ export class VersionManager {
     this._rushConfiguration = RushConfiguration.loadFromConfigurationFile(
       this._rushConfiguration.rushJsonFile
     );
+
+    return updatedPackages;
   }
 
   private _ensure(versionPolicyName?: string, shouldCommit?: boolean, force?: boolean): void {
--- a/libraries/rush-lib/src/logic/VersionManifest.ts
+++ b/libraries/rush-lib/src/logic/VersionManifest.ts
@@ -1,0 +1,54 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import { FileSystem, JsonFile, JsonSchema, IPackageJson } from '@rushstack/node-core-library';
+
+import schemaJson from '../schemas/version-manifest.schema.json';
+
+interface IVersionManifestJson {
+  projects: IVersionManifestProjectJson[];
+}
+
+interface IVersionManifestProjectJson {
+  name: string;
+  relativeFolderPath: string;
+  version: string;
+  previousVersion?: string;
+}
+
+/**
+ * A "version manifest" is a small JSON file that describes the results of a versioning
+ * event (for example, running "rush version --bump").
+ */
+export class VersionManifest {
+  private static _jsonSchema: JsonSchema = JsonSchema.fromLoadedObject(schemaJson);
+
+  private _data: IVersionManifestJson;
+
+  private constructor(data: IVersionManifestJson) {
+    this._data = data;
+  }
+
+  public async save(filePath: string): Promise<void> {
+    await FileSystem.writeFileAsync(filePath, JSON.stringify(this._data, undefined, 2));
+  }
+
+  public static fromUpdatedProjects(updatedProjects: Map<string, IPackageJson>): VersionManifest {
+    const data: IVersionManifestJson = { projects: [] };
+
+    for (const [name, packageJson] of updatedProjects.entries()) {
+      data.projects.push({
+        name: name,
+        relativeFolderPath: '.',
+        version: packageJson.version
+      });
+    }
+
+    return new VersionManifest(data);
+  }
+
+  public static async load(filePath: string): Promise<VersionManifest> {
+    const data: IVersionManifestJson = JsonFile.loadAndValidate(filePath, VersionManifest._jsonSchema);
+    return new VersionManifest(data);
+  }
+}
--- a/libraries/rush-lib/src/schemas/version-manifest.schema.json
+++ b/libraries/rush-lib/src/schemas/version-manifest.schema.json
@@ -1,0 +1,41 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "title": "Version manifest output file format",
+  "description": "A version manifest is an output file produced by the 'rush version' command in Rush, describing the set of projects that have new versions and should be published and/or deployed.",
+  "type": "object",
+  "required": ["projects"],
+  "additionalProperties": false,
+  "properties": {
+    "$schema": {
+      "description": "Part of the JSON Schema standard, this optional keyword declares the URL of the schema that the file conforms to. Editors may download the schema and use it to perform syntax highlighting.",
+      "type": "string"
+    },
+    "projects": {
+      "description": "A list of versioned projects.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["name", "relativeFolderPath", "version"],
+        "additionalProperties": false,
+        "properties": {
+          "name": {
+            "description": "The name of the versioned project.",
+            "type": "string"
+          },
+          "relativeFolderPath": {
+            "description": "The relative path to the folder containing this project.",
+            "type": "string"
+          },
+          "version": {
+            "description": "The new version of the project to be published and/or deployed.",
+            "type": "string"
+          },
+          "previousVersion": {
+            "description": "(Optional) The previous version of this project.",
+            "type": "string"
+          }
+        }
+      }
+    }
+  }
+}
--- a/libraries/rush-lib/src/cli/actions/VersionAction.ts
+++ b/libraries/rush-lib/src/cli/actions/VersionAction.ts
@@ -14,6 +14,7 @@ import { PolicyValidator } from '../../logic/policy/PolicyValidator';
 import { BaseRushAction } from './BaseRushAction';
 import { PublishGit } from '../../logic/PublishGit';
 import { Git } from '../../logic/Git';
+import { VersionManifest } from '../../logic/VersionManifest';
 
 import type * as VersionManagerType from '../../logic/VersionManager';
 
@@ -30,6 +31,7 @@ export class VersionAction extends BaseRushAction {
   private readonly _overwriteBump: CommandLineStringParameter;
   private readonly _prereleaseIdentifier: CommandLineStringParameter;
   private readonly _ignoreGitHooksParameter: CommandLineFlagParameter;
+  private readonly _manifestFileParameter: CommandLineStringParameter;
 
   public constructor(parser: RushCommandLineParser) {
     super({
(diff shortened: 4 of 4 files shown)


libraries/rush-lib/src/logic/VersionManager.ts
```diff
@@ -88,9 +89,11 @@ export class VersionManager {
...
     changeManager.load(this._rushConfiguration.changesFolder);
-    if (changeManager.hasChanges()) {
+    const updatedPackages: Map<string, IPackageJson> | undefined = changeManager.apply(!!shouldCommit);
+
+    if (updatedPackages) {
       changeManager.validateChanges(this._versionPolicyConfiguration);
-      changeManager.apply(!!shouldCommit)!.forEach((packageJson) => {
+      updatedPackages.forEach((packageJson) => {
         this.updatedProjects.set(packageJson.name, packageJson);
       });
```
dmichon-msft, line 98:
```suggestion
      for (const [packageName, packageJson] of updatedPackages) {
        this.updatedProjects.set(packageName, packageJson);
      }
```

libraries/rush-lib/src/logic/VersionManifest.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/**
+ * A "version manifest" is a small JSON file that describes the results of a versioning
+ * event (for example, running "rush version --bump").
+ */
+export class VersionManifest {
+  private static _jsonSchema: JsonSchema = JsonSchema.fromLoadedObject(schemaJson);
+
+  private _data: IVersionManifestJson;
```
dmichon-msft, line 26:
```suggestion
  private readonly _data: IVersionManifestJson;
```

libraries/rush-lib/src/logic/VersionManifest.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  private constructor(data: IVersionManifestJson) {
+    this._data = data;
+  }
+
+  public async save(filePath: string): Promise<void> {
+    await FileSystem.writeFileAsync(filePath, JSON.stringify(this._data, undefined, 2));
```
dmichon-msft, line 33: JsonFile.saveAsync?

libraries/rush-lib/src/logic/VersionManifest.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  public static fromUpdatedProjects(updatedProjects: Map<string, IPackageJson>): VersionManifest {
+    const data: IVersionManifestJson = { projects: [] };
+
+    for (const [name, packageJson] of updatedProjects.entries()) {
+      data.projects.push({
+        name: name,
+        relativeFolderPath: '.',
```
dmichon-msft, line 42: This seems incorrect. Is this supposed to be the `projectRelativeFolder` from `RushConfigurationProject`?

libraries/rush-lib/src/logic/VersionManifest.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  public async save(filePath: string): Promise<void> {
+    await FileSystem.writeFileAsync(filePath, JSON.stringify(this._data, undefined, 2));
+  }
+
+  public static fromUpdatedProjects(updatedProjects: Map<string, IPackageJson>): VersionManifest {
+    const data: IVersionManifestJson = { projects: [] };
+
+    for (const [name, packageJson] of updatedProjects.entries()) {
```
dmichon-msft, line 39:
```suggestion
    for (const [name, packageJson] of updatedProjects) {
```

`[Symbol.iterator]()` returns the same iterator as `.entries()`

libraries/rush-lib/src/schemas/version-manifest.schema.json
```diff
@@ -0,0 +1,41 @@
...
+    },
+    "projects": {
+      "description": "A list of versioned projects.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["name", "relativeFolderPath", "version"],
+        "additionalProperties": false,
+        "properties": {
+          "name": {
```
dmichon-msft, line 21: Unless duplicates are allowed, it makes more semantic sense to have this be a dictionary of package name to change data. I know @octogonz prefers named fields for auto-complete, but this isn't a manually authored file.

libraries/rush-lib/src/schemas/version-manifest.schema.json
```diff
@@ -0,0 +1,41 @@
...
+            "description": "The name of the versioned project.",
+            "type": "string"
+          },
+          "relativeFolderPath": {
+            "description": "The relative path to the folder containing this project.",
```
dmichon-msft, line 26: Relative to what?

libraries/rush-lib/src/schemas/version-manifest.schema.json
elliot-nelson: 🤔 I'm fine with either. I think I was trying to optimize for ease-of-use across other tooling, but they might be too similar to pick a favorite.

### Use case 1: some kind of GHA workflow matrix

```console
# Array
echo "FOLDERS=$(cat manifest.json | jq '[.[].relativeFolderPath]' -c)" >> $GITHUB_OUTPUT

# Dictionary
echo "FOLDERS=$(cat manifest.json | jq '[values[].relativeFolderPath]' -c)" >> $GITHUB_OUTPUT
```

### Use case 2: some kind of inline report

In this case I think array is a bit easier but they are comparable:

```console
# Array
node -pe '"### Publishing\n\nName | Version\n--- | ---\n" + require("./manifest.json").map(x => x.name + " | " + x.version).join("\n")'

# Dictionary
node -pe 'a=require("./manifest.json"),"### Publishing\n\nName | Version\n--- | ---\n" + Object.keys(a).map(x => x + " | " + a[x].version).join("\n")'
```

libraries/rush-lib/src/schemas/version-manifest.schema.json
elliot-nelson: Ah this is intended to be from root of the monorepo (relative as opposed to absolute).

libraries/rush-lib/src/logic/VersionManifest.ts
```diff
@@ -0,0 +1,73 @@
...
+export class VersionManifest {
+  private static _jsonSchema: JsonSchema = JsonSchema.fromLoadedObject(schemaJson);
+
+  private readonly _data: IVersionManifestJson;
+
+  private constructor(data: IVersionManifestJson) {
+    this._data = data;
+  }
+
+  public async save(filePath: string): Promise<void> {
```
iclanton, line 35:
```suggestion
  public async saveAsync(filePath: string): Promise<void> {
```

libraries/rush-lib/src/logic/VersionManifest.ts
```diff
@@ -0,0 +1,73 @@
...
+          version: packageJson.version,
+          previousVersion: rushProject.packageJson.version
+        });
+      }
+    }
+
+    return new VersionManifest(data);
+  }
+
+  public static async load(filePath: string): Promise<VersionManifest> {
```
iclanton, line 69:
```suggestion
  public static async loadAsync(filePath: string): Promise<VersionManifest> {
```

libraries/rush-lib/src/logic/VersionManifest.ts
```diff
@@ -0,0 +1,73 @@
...
+    const data: IVersionManifestJson = JsonFile.loadAndValidate(filePath, VersionManifest._jsonSchema);
```
iclanton, line 70:
```suggestion
    const data: IVersionManifestJson = await JsonFile.loadAndValidateAsync(filePath, VersionManifest._jsonSchema);
```

iclanton:
So this will just generate a manifest of what was changed by that particular version event? It won't be able to replace [this functionality](https://github.com/microsoft/rushstack/blob/main/common/config/azure-pipelines/templates/record-published-versions.yaml)?

elliot-nelson:
> So this will just generate a manifest of what was changed by that particular version event? It won't be able to replace [this functionality](https://github.com/microsoft/rushstack/blob/main/common/config/azure-pipelines/templates/record-published-versions.yaml)?

@iclanton 🤔 This is a good question!

The way I envision it, a "version manifest" is a file that describes only the _actions taken in a particular versioning event_, and would not include what this file you linked to includes (which is more like generating a summary of all versions in the repo).

A nuance there is that the toolbox is using the `shouldPublish` boolean flag to query whether something is a published package. I hope to eventually eliminate this flag altogether -- e.g. if your monorepo contained 100 packages, with 14 of them published to NPM, 11 of them "published" to public GitHub Actions repos in GitHub, and 9 of them deployed as Azure Functions, you might be _versioning_ all of them using the version tooling -- but in this script you are interested in only those 14 published to NPM.

The way I envision Versioning 2.0, eventually, we'd scrap `shouldPublish` and run a command like this:

```console
rush version --scenario publish-npm --manifest manifest.json
rush publish --manifest manifest.json
```

What packages does this publish? Well, that scenario in `version-scenarios.json` might look like:

```json
{
  "name": "publish-npm",
  "projectScope": { "tag": "publish-npm" }
}
```

So, in this world, how would you get the report file that you wanted? My first suggestion would be to see if you can modify the tooling that consumes the file to use the JSON output of rush list instead.

```console
rush list --only tag:publish-npm --json > report.json
```

If you can't, then I'd suggest two changes to `rush list`: let's add a formal "output file" option, and a formal "format" option.

```console
rush list --only tag:publish-npm -o report.json --format version-report
```

Then you could scrap the custom toolbox action.

libraries/rush-lib/src/schemas/version-manifest.schema.json
dmichon-msft: For the node example, `Object.entries` will convert it back to an array of `[key, value]` tuples, so:
```
node -pe '`### Publishing\n\nName | Version\n--- | ---\n${Object.entries(require(`./manifest.json`)).map(([name, { version }]) => `${name} | ${version}`).join(`\n`)}`'
```

Pretty negligible complexity difference, but makes the file guaranteed to not contain duplicates.

libraries/rush-lib/src/schemas/version-manifest.schema.json
```diff
@@ -0,0 +1,41 @@
...
+        "properties": {
+          "name": {
+            "description": "The name of the versioned project.",
+            "type": "string"
+          },
+          "relativeFolderPath": {
+            "description": "The path to the folder containing this project, relative to the root of the monorepo.",
+            "type": "string"
+          },
+          "version": {
```
chengcyber, line 29: I am thinking about whether "version" and "previousVersion" is useful here.

Assuming the we have a `my-lib@1.0.0`, and going to do a patch version.

Running `rush version` generate the publishing plan file, including

```
[
  {
    "name": "my-lib",
    "version": "1.0.1",
    "changes": [...],   // <-- I can see "comments" property in design, where is it? 🤔
  }
]
```

But, in the workflow, the version is not the "final" version during publishing. By running `rush publish --plan <plan_file_above> --suffix alpha.0`. (Correct me if I am wrong).

The `version`s here in the manifest becomes less useful, since it no longer shows the "final" version.

libraries/rush-lib/src/schemas/version-manifest.schema.json
```diff
@@ -0,0 +1,41 @@
...
+    "$schema": {
+      "description": "Part of the JSON Schema standard, this optional keyword declares the URL of the schema that the file conforms to. Editors may download the schema and use it to perform syntax highlighting.",
+      "type": "string"
+    },
+    "projects": {
+      "description": "A list of versioned projects.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["name", "relativeFolderPath", "version"],
```
chengcyber, line 18: I can see `comments` property in design. Is it removed?

PR #4059, opened 2023-04-17
[heft-next] Merge main into release/heft-0.50.0-rc

dmichon-msft:
## Summary
Pulls in changes from main (namely the jest 29.5 bump) to the heft RC branch.

## Details
Also updates cyclic dependency on Heft.

## How it was tested
Full `rush test -v`

## Impacted documentation
None.


apps/heft/src/pluginFramework/InternalHeftSession.ts
```diff
@@ -32,8 +32,6 @@ function* getAllTasks(phases: Iterable<HeftPhase>): Iterable<HeftTask> {
   }
 }
 
-const FORBIDDEN_SOURCE_FILE_GLOBS: string[] = ['package.json', '.gitingore', 'config/**/*', '.rush/**/*'];
```
iclanton, line 35: Are we no longer forbidding these?

iclanton (approved):

apps/heft/src/pluginFramework/InternalHeftSession.ts
dmichon-msft: There isn't currently any special watcher logic in Heft to detect changes to them. It's something we could add back, but would require having a code path that expressly listens for changes to said files.

PR #4073, opened 2023-04-24
[module-minifier-plugin] Emit position data

dmichon-msft:
## Summary
Extends `ModuleMinifierPlugin` (both `webpack@4` and `webpack@5` versions) with data about the rendered locations of each minified module in the final assets, so that they may be sliced out for diffing purposes.

## Details
Emits an additional metadata table mapping asset names to the map from module ID to character offset and length.

## How it was tested
Added unit tests to validate the emitted data.

## Impacted documentation
Metadata fields currently not documented.


dmichon-msft:
This may need some attention in `LocalizationPlugin` since that performs further post-processing of asset content.

TheLarkInn (approved):

webpack/webpack4-module-minifier-plugin/src/ModuleMinifierPlugin.ts
```diff
@@ -38,7 +38,9 @@ import {
   IPostProcessFragmentContext,
   IDehydratedAssets,
   _IWebpackCompilationData,
-  _IAcornComment
+  _IAcornComment,
```
iclanton, line 41: Change this import statement to an `import type`?

iclanton (approved):

PR #4074, opened 2023-04-25
Fix look by path

chengcyber:
## Summary

Fixes the logic of prefix match in `LookupByPath` when the last segment of project folder is a single character.

## Details

I am testing build cache in a demo Rush.js monorepo with several projects, the rush.json is like

```json
"projects": [
  {
    "packageName": "a",
    "projectFolder": "projects/a"
  },
  {
    "packageName": "b",
    "projectFolder": "projects/b"
  },
  // ...
]
```

However, when i run `rush build` with build cache enabled. I found there is a issue to calculate the correct cache id. After dive into the code, it seems there is an edge case when prefix matching with each project folders.

Actual behaviour:

When creating the lookup tree with `projects/a`, it now ignores the trailing character a. The path tree would be like

```
{
  value: undefined,
  children: { // It's a map, use object here to show idea.
    "projects": { 
      value: RushConfigurationProject // <-- which is project a.
    }
  }
}
```

The incorrect path tree leads to a issue to match the files under "projects" folder, which means Rush.js now treats `projects/b/package.json`, `projects/c/package.json`, etc.. belongs to project A, and finally the "ProjectBuildCache#_getCacheId" doesn't get the right scope of the project B, C....

Expected:

Create the correct path tree

```
{
  value: undefined,
  children: { // It's a map, use object here to show idea.
    "projects": {
      value: undefined
      children: {
        a: {
          value: RushConfigurationProject // <-- a
        },
        b: {
          value: RushConfigurationProject // <-- b
        },
        ...
      }
    }
  }
}
```

`projects/a/package.json` belongs to `projects/a`
`projects/b/package.json` belongs to `projects/b`

## How it was tested

Add a new test case to cover this edge case.

## Impacted documentation

Nope

### DIFF
--- a/common/changes/@microsoft/rush/fix-look-by-path_2023-04-25-09-41.json
+++ b/common/changes/@microsoft/rush/fix-look-by-path_2023-04-25-09-41.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Correctly match the prefix with the last single character segment",
+      "comment": "Fix an issue where the last character in a project's path is ignored when determining which files contribute to the project's cache ID.",
       "type": "none"
     }
   ],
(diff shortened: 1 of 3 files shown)


dmichon-msft (approved):
Good catch.

PR #4087, opened 2023-04-28
[heft] Fix PrefixProxyTerminalProvider Newline Handling

D4N14L:
## Summary

Previously, using `terminal.write(...)` with the `PrefixProxyTerminalProvider` would cause multiple calls to appear like the following:

```typescript
terminal.write('test');
terminal.write('ing')
// [prefix] test[prefix] ing
```

Additionally, newlines would be ignored with respect to appending the prefix:
```typescript
terminal.write('test\ning');
// [prefix] test
// ing
```

This change addresses both of the above scenarios by ensuring that newlines are handled by:
- ensuring that multiple consecutive calls to `terminal.write` remember the newline state
- ensuring that inline newlines are handled and written separately

Ex.
```typescript
terminal.write('test');
terminal.write('ing');
// [prefix] testing
```

```typescript
terminal.write('test\ning');
// [prefix] test
// [prefix] ing
```

Additionally, this PR moves the `PrefixProxyTerminalProvider` class to `@rushstack/node-core-library` alongside the other existing `TerminalProvider` classes.

## How it was tested

Added Jest unit tests, validated using Heft that existing, correct logging continues to appear as-is.

## Impacted documentation

Added a class to `@rushstack/node-core-library`


libraries/node-core-library/src/Terminal/PrefixProxyTerminalProvider.ts
```diff
@@ -0,0 +1,85 @@
...
+  public write(data: string, severity: TerminalProviderSeverity): void {
+    // We need to track newlines to ensure that the prefix is added to each line
+    let currentIndex: number = 0;
+    // eslint-disable-next-line @rushstack/no-new-null
+    let newlineMatch: RegExpExecArray | null;
+    while ((newlineMatch = this._newlineRegex.exec(data))) {
+      // Extract the line, add the prefix, and write it out with the newline
+      const newlineIndex: number = newlineMatch.index;
+      const newIndex: number = newlineIndex + newlineMatch[0].length;
+      const dataToWrite: string = `${this._isNewline ? this._prefix : ''}${data.substring(
```
dmichon-msft, line 68: You might as well just have `this._isNewline` be a string that alternates between the values `this._prefix` and `''`. Would make the code a bit more concise.

dmichon-msft (approved):
Minor nit.

PR #4089, opened 2023-04-30
[rush] Replace some more `Import.lazy` calls with dynamic imports.

iclanton:
## Summary

Replace some more `Import.lazy` calls with dynamic imports.

## How it was tested

Need to do some more testing.

### DIFF
--- a/libraries/rush-lib/src/cli/RushPnpmCommandLineParser.ts
+++ b/libraries/rush-lib/src/cli/RushPnpmCommandLineParser.ts
@@ -12,7 +12,6 @@ import {
   Executable,
   FileConstants,
   FileSystem,
-  Import,
   ITerminal,
   ITerminalProvider,
   JsonFile,
@@ -29,9 +28,6 @@ import type { SpawnSyncReturns } from 'child_process';
 import type { BaseInstallManager } from '../logic/base/BaseInstallManager';
 import type { IInstallManagerOptions } from '../logic/base/BaseInstallManagerTypes';
 
-const lodash: typeof import('lodash') = Import.lazy('lodash', require);
-const semver: typeof import('semver') = Import.lazy('semver', require);
-
 const RUSH_SKIP_CHECKS_PARAMETER: string = '--rush-skip-checks';
 
 /**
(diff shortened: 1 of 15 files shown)


libraries/rush-lib/src/logic/ChangeFiles.ts
```diff
@@ -122,15 +122,16 @@ export class ChangeFiles {
...
       // Skip changes files if the package's change log is not updated.
       const packagesToInclude: Set<string> = new Set<string>();
       updatedChangelogs.forEach((changelog) => {
         packagesToInclude.add(changelog.name);
       });
 
-      const filesToDelete: string[] = this.getFiles().filter((filePath) => {
+      const files: string[] = await this.getFilesAsync();
+      const filesToDelete: string[] = files.filter((filePath) => {
         const changeRequest: IChangeInfo = JsonFile.load(filePath);
```
D4N14L, line 135: load
```suggestion
        const changeRequest: IChangeInfo = await JsonFile.loadAsync(filePath);
```

libraries/rush-lib/src/logic/ChangeFiles.ts
```diff
@@ -143,7 +144,8 @@ export class ChangeFiles {
       return this._deleteFiles(filesToDelete, shouldDelete);
     } else {
       // Delete all change files.
-      return this._deleteFiles(this.getFiles(), shouldDelete);
+      const files: string[] = await this.getFilesAsync();
+      return this._deleteFiles(files, shouldDelete);
```
D4N14L, line 148: Can this be made async?

D4N14L (approved):

libraries/rush-lib/src/logic/ChangeFiles.ts
iclanton: Should be.

libraries/rush-lib/src/cli/RushPnpmCommandLineParser.ts
```diff
@@ -43,6 +39,25 @@ export interface IRushPnpmCommandLineParserOptions {
...
+    // If catchSyncErrors() called this, then show a call stack similar to what Node.js
+    // would show for an uncaught error
+    terminal.writeErrorLine('\n' + error.stack);
+  }
+
+  if (process.exitCode !== undefined) {
+    process.exit(process.exitCode);
+  } else {
+    process.exit(1);
+  }
```
dmichon-msft, line 58:
```suggestion
  process.exit(process.exitCode ?? 1);
```

libraries/rush-lib/src/cli/RushPnpmCommandLineParser.ts
```diff
@@ -51,77 +66,97 @@ export class RushPnpmCommandLineParser {
...
+    this._debugEnabled = debugEnabled;
+    this._verboseEnabled = verboseEnabled;
+    this._terminal = terminal;
+
+    // Are we in a Rush repo?
+    let rushConfiguration: RushConfiguration | undefined = undefined;
+    if (RushConfiguration.tryFindRushJsonLocation()) {
+      // showVerbose is false because the logging message may break JSON output
+      rushConfiguration = RushConfiguration.loadFromDefaultLocation({ showVerbose: false });
+    }
```
dmichon-msft, line 84: Surprised we don't have a single call for this.

dmichon-msft (approved):

libraries/rush-lib/src/cli/RushPnpmCommandLineParser.ts
iclanton: Guess it isn't something that's really used all that often.

PR #4091, opened 2023-04-30
[package-exractor] Bundle the create-links script to ensure any imports that are added in the future don't cause issues.

iclanton:
## Summary

This is a follow-up to https://github.com/microsoft/rushstack/pull/4064.

## How it was tested

This needs some more testing.

### DIFF
--- a/libraries/package-extractor/webpack.config.js
+++ b/libraries/package-extractor/webpack.config.js
@@ -12,11 +12,11 @@ module.exports = () => {
     entry: {
       [PathConstants.createLinksScriptFilename]: {
         import: `${__dirname}/lib-esnext/scripts/create-links.js`,
-        filename: `${PathConstants.scriptsFolderName}/[name]`
+        filename: `[name]`
       }
     },
     output: {
-      path: `${__dirname}/lib`,
+      path: PathConstants.scriptsFolderPath,
       filename: '[name].js',
       chunkFilename: 'chunks/[name].js', // TODO: Don't allow any chunks to be created
       library: {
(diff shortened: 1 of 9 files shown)


libraries/rush-lib/src/cli/actions/DeployAction.ts
```diff
@@ -174,7 +171,7 @@ export class DeployAction extends BaseRushAction {
     // Call the deploy manager
     const { PackageExtractor } = await import(
       /* webpackChunkName: 'DeployManager' */
```
D4N14L, line 173:
```suggestion
      /* webpackChunkName: 'PackageExtractor' */
```

libraries/package-extractor/webpack.config.js
```diff
@@ -0,0 +1,34 @@
...
+    mode: 'development', // So the output isn't minified
+    devtool: 'source-map',
+    entry: {
+      [PathConstants.createLinksScriptFilename]: {
+        import: `${__dirname}/lib-esnext/scripts/create-links.js`,
+        filename: `${PathConstants.scriptsFolderName}/[name]`
+      }
+    },
+    output: {
+      path: `${__dirname}/lib`,
```
D4N14L, line 19: Maybe we should put the output into a `dist/scripts/*` path? That way the code that references it can point to the same location when using the `lib-esnext` or `lib` versions? Cause as it stands right now, the `lib-esnext` version would break because it paths to it using `${__dirname}/scripts/create-links.js`  (ln. 314 in PackageExtractor.ts)

iclanton: Yeah probably

PR #4096, opened 2023-05-02
[rush] Some minor cleanup in Rush.

iclanton:


libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -935,6 +935,14 @@ export class RushConfiguration {
     return new RushConfiguration(rushConfigurationJson, resolvedRushJsonFilename);
   }
 
+  public static tryLoadFromDefaultLocation(
+    options?: ITryFindRushJsonLocationOptions
+  ): RushConfiguration | undefined {
+    if (RushConfiguration.tryFindRushJsonLocation()) {
```
dmichon-msft, line 941: This returns the folder containing `rush.json`, shouldn't we promptly use that location instead of redoing the search?

dmichon-msft (approved):

PR #4105, opened 2023-05-08
[debug-cert] Add env, fix IP address

dmichon-msft:
## Summary
Adds an environment variable `DISABLE_DEV_CERT_GENERATION` that can be set to `1` on the machine to cause all calls to generate a new dev certificate to be an error. This is most relevant to GitHub codespaces, where the certificate needs to be kept in sync with a certificate on the client machine to successfully connect.

## Details
Removes `rushstack.localhost` from this list of required `subjectAltNames` again to avoid forcing certificate regeneration.

Switches `127.0.0.1` to be properly flagged in the X.509 certificate as an IP Address, rather than a DNS name.

## How it was tested
Local run.
Validated that `https://127.0.0.1:8080` now has a secure connection to `npm start` in `heft-webpack5-everything-test`.

## Impacted documentation
None, currently.

### DIFF
--- a/libraries/debug-certificate-manager/src/CertificateManager.ts
+++ b/libraries/debug-certificate-manager/src/CertificateManager.ts
@@ -29,6 +29,9 @@ export const DEFAULT_CERTIFICATE_SUBJECT_NAMES: ReadonlyArray<string> = ['localh
  */
 export const DEFAULT_CERTIFICATE_SUBJECT_IP_ADDRESSES: ReadonlyArray<string> = ['127.0.0.1'];
 
+const DISABLE_CERT_GENERATION_VARIABLE_NAME: 'RUSHSTACK_DISABLE_DEV_CERT_GENERATION' =
+  'RUSHSTACK_DISABLE_DEV_CERT_GENERATION';
+
 /**
  * The interface for a debug certificate instance
  *
@@ -137,10 +140,10 @@ export class CertificateManager {
 
     const { certificateData: existingCert, keyData: existingKey } = this._certificateStore;
 
-    if (process.env.RUSHSTACK_DISABLE_DEV_CERT_GENERATION === '1') {
+    if (process.env[DISABLE_CERT_GENERATION_VARIABLE_NAME] === '1') {
       // Allow the environment (e.g. GitHub codespaces) to forcibly disable dev cert generation
       terminal.writeLine(
-        `Found environment variable RUSHSTACK_DISABLE_DEV_CERT_GENERATION=1, disabling certificate generation.`
+        `Found environment variable ${DISABLE_CERT_GENERATION_VARIABLE_NAME}=1, disabling certificate generation.`
       );
       canGenerateNewCertificate = false;
     }
--- a/libraries/debug-certificate-manager/src/CertificateManager.ts
+++ b/libraries/debug-certificate-manager/src/CertificateManager.ts
@@ -84,6 +84,7 @@ interface IDnsAltName {
 }
 /**
  * Fields for a Subject Alternative Name of type IP Address
+ * `node-forge` requires the field name to be "ip" instead of "value", likely due to subtle encoding differences.
  */
 interface IIPAddressAltName {
   type: 7;
@@ -103,7 +104,7 @@ export interface ICertificateGenerationOptions {
   /**
    * The IP Address Subject names to issue the certificate for. Defaults to ['127.0.0.1'].
    */
-  subjectIpAddresses?: ReadonlyArray<string>;
+  subjectIPAddresses?: ReadonlyArray<string>;
   /**
    * How many days the certificate should be valid for.
    */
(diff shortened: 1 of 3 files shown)


libraries/debug-certificate-manager/src/CertificateManager.ts
```diff
@@ -70,10 +66,15 @@ interface ISubjectAltNameExtension {
   altNames: readonly IAltName[];
 }
 
-interface IAltName {
+interface IDNSName {
```
iclanton, line 69:
```suggestion
interface IDnsName {
```

libraries/debug-certificate-manager/src/CertificateManager.ts
```diff
@@ -120,6 +121,14 @@ export class CertificateManager {
 
     const { certificateData: existingCert, keyData: existingKey } = this._certificateStore;
 
+    if (process.env.DISABLE_DEV_CERT_GENERATION === '1') {
```
iclanton, line 124:
```suggestion
    if (process.env.RUSHSTACK_DISABLE_DEV_CERT_GENERATION === '1') {
```

And put the env var name in a constant?

libraries/debug-certificate-manager/src/CertificateManager.ts
```diff
@@ -120,6 +121,14 @@ export class CertificateManager {
...
+      // Allow the environment (e.g. GitHub codespaces) to forcibly disable dev cert generation
+      terminal.writeLine(
+        `Found environment variable DISABLE_DEV_CERT_GENERATION=1, force disabling certificate generation.`
```
iclanton, line 127:
```suggestion
        `Found environment variable DISABLE_DEV_CERT_GENERATION=1, disabling certificate generation.`
```

libraries/debug-certificate-manager/src/CertificateManager.ts
```diff
@@ -414,10 +425,14 @@ export class CertificateManager {
     certificate.setSubject(subjectAttrs);
     certificate.setIssuer(issuerAttrs);
 
-    const subjectAltNames: readonly IAltName[] = subjectNames.map((subjectName) => ({
+    const subjectAltNames: IAltName[] = subjectNames.map((subjectName) => ({
       type: 2, // DNS
       value: subjectName
     }));
+    subjectAltNames.push({
```
iclanton, line 432: So we're always adding `127.0.0.1`? The idea being that this is only ever going to be used for local development?

iclanton (approved):

iclanton:
https://www.youtube.com/watch?v=07Y0cy-nvAg - except for `Brooklyn`, `Ship it`.

libraries/debug-certificate-manager/src/CertificateManager.ts
dmichon-msft: I'm open to adding another parameter for IP Addresses to authenticate, though since it's a self-signed cert it should probably only be used for local development.

libraries/debug-certificate-manager/src/CertificateManager.ts
```diff
@@ -21,11 +21,16 @@ const ONE_DAY_IN_MILLISECONDS: number = 24 * 60 * 60 * 1000;
  * The set of names the certificate should be generated for, by default.
  * @public
  */
-export const DEFAULT_CERTIFICATE_SUBJECT_NAMES: ReadonlyArray<string> = [
-  'localhost',
-  'rushstack.localhost',
-  '127.0.0.1'
-];
+export const DEFAULT_CERTIFICATE_SUBJECT_NAMES: ReadonlyArray<string> = ['localhost'];
```
D4N14L, line 24: Removing `rushstack.localhost`?

libraries/debug-certificate-manager/src/CertificateManager.ts
```diff
@@ -70,20 +75,35 @@ interface ISubjectAltNameExtension {
...
 export interface ICertificateGenerationOptions {
   /**
-   * The DNS Subject names to issue the certificate for.
+   * The DNS Subject names to issue the certificate for. Defaults to ['localhost'].
    */
   subjectAltNames?: ReadonlyArray<string>;
+  /**
+   * The IP Address Subject names to issue the certificate for. Defaults to ['127.0.0.1'].
+   */
+  subjectIpAddresses?: ReadonlyArray<string>;
```
D4N14L, line 106:
```suggestion
  subjectIPAddresses?: ReadonlyArray<string>;
```

Though obviously that's just personal preference

libraries/debug-certificate-manager/src/CertificateManager.ts
```diff
@@ -70,20 +75,35 @@ interface ISubjectAltNameExtension {
...
+interface IDnsAltName {
   type: 2;
   value: string;
 }
+/**
+ * Fields for a Subject Alternative Name of type IP Address
+ */
+interface IIPAddressAltName {
+  type: 7;
+  ip: string;
```
D4N14L, line 90: Why not make this `value` as well? Could avoid all the mapping below.

libraries/debug-certificate-manager/src/CertificateManager.ts
dmichon-msft: The issue here mostly has to do with that adding it, due to the rules, forces cert regeneration, which is a headache for developers consuming it.

libraries/debug-certificate-manager/src/CertificateManager.ts
dmichon-msft: Updated comments here. `node-forge` requires this to be named `ip`, otherwise it emits a malformed certificate.

D4N14L (approved):

PR #4111, opened 2023-05-11
[node-core-library] Add an option to the PrefixProxyTerminalProvider to create a dynamic prefix.

iclanton:
## Summary

This PR adds an option to `PrefixProxyTerminalProvider` called `getPrefix` which takes a function that returns a string to dynamically generate the prefix. This can be used for prefixing logging lines with timestamps.

## How it was tested

Included a unit test.

### DIFF
--- a/libraries/node-core-library/src/Terminal/PrefixProxyTerminalProvider.ts
+++ b/libraries/node-core-library/src/Terminal/PrefixProxyTerminalProvider.ts
@@ -93,34 +93,24 @@ export class PrefixProxyTerminalProvider implements ITerminalProvider {
     let currentIndex: number = 0;
     // eslint-disable-next-line @rushstack/no-new-null
     let newlineMatch: RegExpExecArray | null;
-    let isFirstLoop: boolean = true;
-    const startedOnNewLine: boolean = this._isOnNewline;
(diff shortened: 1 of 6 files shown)


libraries/node-core-library/src/Terminal/PrefixProxyTerminalProvider.ts
```diff
@@ -58,22 +93,35 @@ export class PrefixProxyTerminalProvider implements ITerminalProvider {
     let currentIndex: number = 0;
     // eslint-disable-next-line @rushstack/no-new-null
     let newlineMatch: RegExpExecArray | null;
+    let isFirstLoop: boolean = true;
+    const startedOnNewLine: boolean = this._isOnNewline;
+    let currentPrefix: string = this._isOnNewline ? this._getPrefix() : '';
+
     while ((newlineMatch = this._newlineRegex.exec(data))) {
+      if (!isFirstLoop && !startedOnNewLine) {
+        currentPrefix = this._getPrefix();
```
D4N14L, line 102: I wonder if maintaining the same prefix across multiple lines of a string containing newlines is what's expected? `getPrefix` seems to imply that it'd be called each time a prefix is required, which is not what's happening here.

D4N14L (changes requested):

D4N14L (approved):

PR #4123, opened 2023-05-19
[heft-jest-plugin] Add --test-path-ignore-patterns CLI option.

DavidRieman:
## Summary
Users sometimes need to run "all tests except..."
For example, one might have both unit tests and integration tests in the same project (even mixed in the same folders) and could need an exclusion pattern to run only the unit tests, or run only the integration tests.
Jest accepts the "testPathIgnorePatterns" CLI argument to support this but Heft-jest-plugin was not piping such an argument through.  This PR adds that missing support.

## Details
I simply followed the pattern of the existing similar "testPathPattern", to add the new "--test-path-ignore-patterns" argument which maps to the "testPathIgnorePatterns" Jest argument.

## How it was tested
After building, in my rushstack mono-repo I went into libraries/rush-lib as it has a lot of tests.  I executed commands like these, which all seem to operate as expected, reducing the total number of tests executed from the 325 tests that normally execute:
* `rushx test --test-path-ignore-patterns Shrink`
* `rushx test --test-path-ignore-patterns Shrink\|Com\|Oper\|Look\|Telem\|Last`

## Impacted documentation
From what I'm aware of, the CLI output for the new option would need to be updated at, for example, https://rushstack.io/pages/heft/cli/#heft-test - not sure if that is automated or whatnot?

### DIFF
--- a/apps/api-documenter/package.json
+++ b/apps/api-documenter/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft/api-documenter",
-  "version": "7.22.13",
+  "version": "7.22.14",
   "description": "Read JSON files from api-extractor, generate documentation pages",
   "repository": {
     "type": "git",
--- a/apps/heft/package.json
+++ b/apps/heft/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rushstack/heft",
-  "version": "0.52.2",
+  "version": "0.53.0",
   "description": "Build all your JavaScript projects the same way: A way that works.",
   "keywords": [
     "toolchain",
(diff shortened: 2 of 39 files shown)


DavidRieman:
Strange, commit and build passed green on my own fork.  Local `rush change` says `Nothing to do.` - not sure how to rectify this now.  Do I manually seek out and edit the changes file?

iclanton:
We're currently working on releasing a rework of Heft. (@D4N14L can comment further), so we should probably wait on this until that gets merged into main.

iclanton:
@DavidRieman - The refactor of Heft is in and has been released.

DavidRieman:
Closed in favor of https://github.com/microsoft/rushstack/pull/4184 as it was easier to just start over due to the code churn.

PR #4127, opened 2023-05-22
Add a test case for --allowArbitraryExtensions.

iclanton:
This is a follow-up to https://github.com/microsoft/rushstack/pull/4103, testing https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#supporting-multiple-configuration-files-in-extends.

This feature is currently broken, and need to be fixed in API Extractor.

### DIFF
--- a/build-tests/api-extractor-test-04/tsconfig.json
+++ b/build-tests/api-extractor-test-04/tsconfig.json
@@ -11,7 +11,8 @@
     "esModuleInterop": true,
     "types": [],
     "lib": ["es5", "scripthost", "es2015.collection", "es2015.promise", "es2015.iterable", "dom"],
-    "outDir": "lib"
+    "outDir": "lib",
+    "allowArbitraryExtensions": true
   },
-  "include": ["src/**/*.ts", "typings/tsd.d.ts"]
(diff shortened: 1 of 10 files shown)


build-tests/api-extractor-test-04/tsconfig.json
```diff
@@ -11,7 +11,8 @@
     "esModuleInterop": true,
     "types": [],
     "lib": ["es5", "scripthost", "es2015.collection", "es2015.promise", "es2015.iterable", "dom"],
-    "outDir": "lib"
+    "outDir": "lib",
+    "allowArbitraryExtensions": true
   },
-  "include": ["src/**/*.ts", "typings/tsd.d.ts"]
+  "include": ["src/**/*.ts", "typings/tsd.d.ts", "src/AlphaClass.tsz"]
```
dmichon-msft, line 17: Is the reference to the nonexistent file here deliberate?

iclanton: Oops, thought I removed that.
```suggestion
  "include": ["src/**/*.ts", "typings/tsd.d.ts"]
```

PR #4130, opened 2023-05-23
Feat log for build cache

chengcyber:
## Summary

Fixes https://github.com/microsoft/rushstack/issues/4102 https://github.com/microsoft/rushstack/issues/4117

## Details

Use a separate terminal for logging cache subsystem. Log for build cache only displayed in terminal instead of appending to all.log

terminal: 
CleanShot 2023-05-23 at 19 25 22@2x

log file:
CleanShot 2023-05-23 at 19 27 07@2x

## How it was tested

Tested in a local monorepo.

### DIFF
--- a/common/changes/@microsoft/rush/feat-log-for-build-cache_2023-05-23-11-22.json
+++ b/common/changes/@microsoft/rush/feat-log-for-build-cache_2023-05-23-11-22.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Use a separate temrinal for logging cache subsystem",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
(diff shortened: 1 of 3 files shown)


dmichon-msft (approved):

ankhzet:
Getting "EPERM: operation not permitted, open '<project>\torque\shared\utils/utils.build.log'" on 5.100.0, win 11
`rush purge`/`rush update`/deleting build cache etc. does not help. `--debug` adds 0 context except where top-level error is thrown: 
```Error: An error occurred.
    at PhasedScriptAction._executeOperations (<user>\.rush\node-v18.15.0\rush-5.100.0\node_modules\@microsoft\rush-lib\dist\commons.js:10290:19)```

Is there a config change/cleanup/rebuild process associated with this change?

dmichon-msft:
There's a bug in the ProjectLogWritable that we are accidentally attempting to write the cache log and the build log to the same file if phased commands are not enabled. @iclanton is putting together a fix.

PR #4131, opened 2023-05-23
[rush-lib] Expose beforeLog hook to augment/modify telemetry logs

IsaacSNK:
## Summary
Exposes a _beforeLog_ hook that can be used to modify the telemetry object before it gets written in the temporary telemetry folder. This allows plugin writers to add custom fields in the telemetry.

## How it was tested
Manual tests where a custom Rush plugin was created and attached to the _beforeLog_ hook to add more data to the telemetry payload.

### DIFF
--- a/common/changes/@microsoft/rush/feature-custom-hooks-log-telemetry_2023-05-23-19-53.json
+++ b/common/changes/@microsoft/rush/feature-custom-hooks-log-telemetry_2023-05-23-19-53.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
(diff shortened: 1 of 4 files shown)


dmichon-msft:
Is the reason you are adding this in `PhasedCommandHooks` so that the closure has access to context from the `PhasedScriptAction`? Otherwise it would make more sense to put this in `RushLifecycleHooks` where the existing `flushTelemetry` hook is.

IsaacSNK:
Yes, that is the reason to add it there. I want to be able to write a plugin for PhasedCommands that can connect to this hook.

dmichon-msft (approved):

PR #4137, opened 2023-05-30
[heft-typescript@next] More transpile tuning

dmichon-msft:
## Summary
Optimize the cleanup of the worker process.
Remove the `TypeScriptCachedFileSystem`, since TypeScript does caching internally.

## Details
Removes up to half a second of dead time from expressly waiting for the worker to exit.
Instead include error/exit handlers in the individual transpile callbacks. If the worker exits with non-zero after all pending transpiles have already finished, we don't really care.

Removes unneeded code in the form of the caching file system. TypeScript's internal fs wrapper already does caching.

## How it was tested
Local runs.

## Impacted documentation
None.


dmichon-msft:
Wrong target branch...

PR #4138, opened 2023-05-30
[heft-typescript@next] More transpile tuning

dmichon-msft:
## Summary
Optimize the cleanup of the worker process.
Remove the `TypeScriptCachedFileSystem`, since TypeScript does caching internally.

## Details
Removes up to half a second of dead time from expressly waiting for the worker to exit.
Instead include error/exit handlers in the individual transpile callbacks. If the worker exits with non-zero after all pending transpiles have already finished, we don't really care.

Removes unneeded code in the form of the caching file system. TypeScript's internal fs wrapper already does caching.

## How it was tested
Local runs.

## Impacted documentation
None.

### DIFF
--- a/heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
+++ b/heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
@@ -7,20 +7,12 @@ import { Worker } from 'worker_threads';
 
 import * as semver from 'semver';
 import type * as TTypescript from 'typescript';
-import {
-  type ITerminal,
-  JsonFile,
-  type IPackageJson,
-  Path,
-  Async,
-  FileError
-} from '@rushstack/node-core-library';
+import { type ITerminal, JsonFile, type IPackageJson, Path, FileError } from '@rushstack/node-core-library';
 import type { IScopedLogger } from '@rushstack/heft';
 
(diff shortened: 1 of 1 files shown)


heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
```diff
@@ -534,15 +499,17 @@ export class TypeScriptBuilder {
...
-    const filesToWrite: IFileToWrite[] = [];
-
-    const writeFileCallback: TTypescript.WriteFileCallback = (filePath: string, data: string) => {
-      filesToWrite.push({ filePath, data });
-    };
-
     const { changedFiles } = configureProgramForMultiEmit(innerProgram, ts, this._moduleKindsToEmit, mode);
 
-    const emitResult: TTypescript.EmitResult = genericProgram.emit(undefined, writeFileCallback);
+    const emitResult: TTypescript.EmitResult = genericProgram.emit(
```
D4N14L, line 504: Maybe add a comment for what's going on here? All the `undefined`s seem suspicious

D4N14L (approved):

heft-plugins/heft-typescript-plugin/src/TypeScriptBuilder.ts
dmichon-msft: They're just there to match the arity of the underlying function.

PR #4169, opened 2023-06-06
[rush] Silence cache logs in quiet mode and write cache logs to a `*.cache.log` file.

iclanton:
## Summary

This fixes a regression where cache logs were always written to stdout. It also creates `*cache.log` files for that data.

## How it was tested

Ran `rush build` in a repo with caching enabled.

### DIFF
--- a/common/changes/@microsoft/rush/silence-cache-logging-in-quiet-mode_2023-06-06-22-19.json
+++ b/common/changes/@microsoft/rush/silence-cache-logging-in-quiet-mode_2023-06-06-22-19.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Write cache logs to their own file(s).",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/common/changes/@microsoft/rush/silence-cache-logging-in-quiet-mode_2023-06-06-22-19.json
+++ b/common/changes/@microsoft/rush/silence-cache-logging-in-quiet-mode_2023-06-06-22-19.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Write cache logs to their own file(s).",
+      "comment": "Fix an issue where cache logging data was always written to stdout.",
       "type": "none"
     }
   ],
--- a/libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
+++ b/libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
@@ -188,7 +188,9 @@ export class ShellOperationRunner implements IOperationRunner {
       const terminal: Terminal = new Terminal(terminalProvider);
 
       // Controls the log for the cache subsystem
-      const buildCacheCollatedTerminal: CollatedTerminal = new CollatedTerminal(context.collatedWriter);
+      const buildCacheCollatedTerminal: CollatedTerminal = new CollatedTerminal(
+        context.quietMode ? discardTransform : context.collatedWriter
+      );
       const buildCacheTerminalProvider: CollatedTerminalProvider = new CollatedTerminalProvider(
         buildCacheCollatedTerminal,
         {
--- a/libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
+++ b/libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
@@ -23,6 +23,8 @@ import {
   PrintUtilities
 } from '@rushstack/terminal';
 import { CollatedTerminal } from '@rushstack/stream-collator';
+import type { TerminalWritable } from '@rushstack/terminal';
+
 import { Utilities, UNINITIALIZED } from '../../utilities/Utilities';
 import { OperationStatus } from './OperationStatus';
 import { OperationError } from './OperationError';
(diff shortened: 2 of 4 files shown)


common/changes/@microsoft/rush/silence-cache-logging-in-quiet-mode_2023-06-06-22-19.json
```diff
@@ -0,0 +1,10 @@
+{
```
dmichon-msft, line 1: Accidental double change file.

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
```diff
@@ -150,6 +150,11 @@ export class ShellOperationRunner implements IOperationRunner {
       context.collatedWriter.terminal,
       this._logFilenameIdentifier
     );
+    const cacheProjectLogWritable: ProjectLogWritable = new ProjectLogWritable(
```
dmichon-msft, line 153: The act of creating the object opens the log file. This (1) needs to always get cleaned up in a `finally` branch, (2) needs to not prevent the main projectLogWritable from getting cleaned up if it fails to open, and (3) doesn't need to be opened if the build cache feature is completely disabled.

dmichon-msft: Also, the cache log should technically be the outermost log, and also responsible for the skip detection and such.

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
```diff
@@ -420,6 +436,13 @@ export class ShellOperationRunner implements IOperationRunner {
 
         const [, cacheWriteSuccess] = await Promise.all([writeProjectStatePromise, setCacheEntryPromise]);
 
+        cacheSplitterTransform.close();
```
dmichon-msft, line 439: This also needs to get cleaned up in the `finally` branch.

common/changes/@microsoft/rush/silence-cache-logging-in-quiet-mode_2023-06-06-22-19.json
iclanton: Weird, these were supposed to be different.

common/changes/@microsoft/rush/silence-cache-logging-in-quiet-mode_2023-06-06-22-19.json
```diff
@@ -0,0 +1,10 @@
...
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Write cache logs to their own file(s).",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue where cache logging data was always written to stdout.",
```

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
```diff
@@ -420,7 +450,7 @@ export class ShellOperationRunner implements IOperationRunner {
 
         const [, cacheWriteSuccess] = await Promise.all([writeProjectStatePromise, setCacheEntryPromise]);
 
-        if (buildCacheTerminalProvider.hasErrors) {
+        if (terminalProvider.hasErrors) {
```
iclanton, line 453: Pretty sure this was incorrectly changed here: https://github.com/microsoft/rushstack/pull/4130/files#diff-f19d02b6f5da5124f2fe15eba7801b17015e12f2d44fa9b78b8fd250abce2a17R423

dmichon-msft, line 453: I'm pretty sure this was right. This check is in the cache write code path; if the build failed or wrote to stderr that was already handled outside of the parent `if` check on line 425.

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
```diff
@@ -138,20 +140,57 @@ export class ShellOperationRunner implements IOperationRunner {
...
+
     const projectLogWritable: ProjectLogWritable = new ProjectLogWritable(
       this._rushProject,
       context.collatedWriter.terminal,
       this._logFilenameIdentifier
     );
 
     try {
+      //#region CACHE LOGGING
+      const cacheDiscardTransform: DiscardStdoutTransform = new DiscardStdoutTransform({
```
dmichon-msft, line 160: Can we inline this so we don't create the DiscardStdoutTransform only to immediately discard it in the `--verbose` case?

libraries/rush-lib/src/logic/operations/ShellOperationRunner.ts
```diff
@@ -138,20 +140,57 @@ export class ShellOperationRunner implements IOperationRunner {
...
+        cacheCollatedTerminal = new CollatedTerminal(cacheConsoleWritable);
+      }
+
+      const buildCacheTerminalProvider: CollatedTerminalProvider = new CollatedTerminalProvider(
+        cacheCollatedTerminal,
+        {
+          debugEnabled: context.debugMode
+        }
+      );
+      const buildCacheTerminal: Terminal | undefined = new Terminal(buildCacheTerminalProvider);
```
dmichon-msft, line 183: This can't be undefined when it is `const` and assigned the result of a constructor, unless said constructor contains a `return undefined` inside of itself (which would be quite unusual).

iclanton: Oops, didn't fix this one.

dmichon-msft (approved):

PR #4171, opened 2023-06-07
forcing newer version of cosmiconfig to fix CVE for yaml

Marija-Pet:
## Summary
Package cosmiconfig was updated from 7.1.0 to 8.1.3 
 
## Details
The yaml package version 1.10.2 has a vulnerability. The cosmiconfig 7.1.0 has a dependencie on the yaml 1.10.2. Newer versions of cosmioconfig depend on newer versions of yaml which dont have the vunerability.
Using the globalOverrides, forcing cosmiconfig version to 8.13 to remove vulnerability

## How it was tested
Manual rush install and build

### DIFF
--- a/common/config/rush/pnpm-config.json
+++ b/common/config/rush/pnpm-config.json
@@ -93,7 +93,8 @@
     // "example2": "npm:@company/example2@^1.0.0"
     // TODO: Remove once https://github.com/dylang/npm-check/issues/499
     // has been closed and a new version of `npm-check` is published.
-    "package-json": "^7"
+    "package-json": "^7",
+    "cosmiconfig": "^8.1.3"
   },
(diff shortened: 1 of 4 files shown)


common/config/rush/pnpm-config.json
```diff
@@ -93,7 +93,8 @@
     // "example2": "npm:@company/example2@^1.0.0"
     // TODO: Remove once https://github.com/dylang/npm-check/issues/499
     // has been closed and a new version of `npm-check` is published.
-    "package-json": "^7"
+    "package-json": "^7",
+    "cosmiconfig": "^8.1.3"
```
iclanton, line 97: This won't upgrade this package for consumers.

PR #4176, opened 2023-06-07
[heft] Remove the concept of the cache folder

dmichon-msft:
## Summary
Removes the concept of the "cache" folder from Heft. This concept mostly just led to bugs and unexpected behavior when running multiple builds consecutively.

## Details
Due to cyclic dependencies, the `cacheFolderPath` properties are left in place, but return the value of `tempFolderPath` instead. The properties have been marked deprecated.

## How it was tested
rush retest -v

## Impacted documentation
Any docs regarding the heft cache folder.

### DIFF
--- a/apps/heft/src/pluginFramework/HeftLifecycleSession.ts
+++ b/apps/heft/src/pluginFramework/HeftLifecycleSession.ts
@@ -37,11 +37,9 @@ export interface IHeftLifecycleSession {
   readonly parameters: IHeftParameters;
 
   /**
-   * The cache folder for the lifecycle plugin. This folder is unique for each lifecycle plugin,
-   * and will not be cleaned when Heft is run with `--clean`. However, it will be cleaned when
-   * Heft is run with `--clean` and `--clean-cache`.
+   * The cache folder for the lifecycle plugin. This is an alias for tempFolderPath.
    *
-   * @public
+   * @deprecated Use `tempFolderPath` instead.
    */
   readonly cacheFolderPath: string;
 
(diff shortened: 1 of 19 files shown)


apps/heft/src/pluginFramework/HeftLifecycleSession.ts
```diff
@@ -37,11 +37,9 @@ export interface IHeftLifecycleSession {
...
   /**
-   * The cache folder for the lifecycle plugin. This folder is unique for each lifecycle plugin,
-   * and will not be cleaned when Heft is run with `--clean`. However, it will be cleaned when
-   * Heft is run with `--clean` and `--clean-cache`.
+   * The cache folder for the lifecycle plugin. This is an alias for tempFolderPath.
    *
-   * @public
+   * @deprecated Use `tempFolderPath` instead.
    */
   readonly cacheFolderPath: string;
```
D4N14L, line 44: I would just straight up remove it since it's so new.

apps/heft/src/pluginFramework/HeftTaskSession.ts
```diff
@@ -97,11 +97,9 @@ export interface IHeftTaskSession {
...
   /**
-   * The cache folder for the task. This folder is unique for each task, and will not be
-   * cleaned when Heft is run with `--clean`. However, it will be cleaned when Heft is run
-   * with `--clean` and `--clean-cache`.
+   * The cache folder for the task. This is an alias for the temp folder.
    *
-   * @public
+   * @deprecated Use `tempFolderPath` instead.
    */
   readonly cacheFolderPath: string;
```
D4N14L, line 104: Same here

D4N14L (approved):

D4N14L:
Could also remove the `.cache` folder from the gitignore

apps/heft/src/pluginFramework/HeftLifecycleSession.ts
dmichon-msft: I can't until we do another publish and another cyclic dependency bump.

apps/heft/src/pluginFramework/HeftTaskSession.ts
dmichon-msft: I can't until we do another publish and another cyclic dependency bump.

PR #4180, opened 2023-06-08
[heft] Remove `taskEvent` configuration and replace it with directly-referenced plugins

D4N14L:
## Summary

This PR removes the `taskEvent` option from tasks in heft.json, and replaces it with referencing the `@rushstack/heft` package plugins directly. It also allows for self-referencing of plugin packages.

## Details

The `taskEvent` API was duplicative and unnecessary when plugins provided by Heft could be provided by simply referencing the `@rushstack/heft` package directly as with any other plugin package. Additionally, this change also allows for the `pluginPackage` field to reference the owning package. This will allow for loading plugins specified within the same package that defines the configuration.

Plugin name mappings for previously-existing task events are:
- `copyFiles` -> `copy-files-plugin`
- `deleteFiles` -> `delete-files-plugin`
- `runScript` -> `run-script-plugin`
- `nodeService` -> `node-service-plugin`

Old format:
```json
{
  "phasesByName": {
    "build": {
      "tasksbyName": {
        "perform-copy": {
          "taskEvent": {
            "eventKind": "copyFiles",
            "options": {
              ...
            }
          }
        }
      }
    }
  }
}
```
New format:
```json
{
  "phasesByName": {
    "build": {
      "tasksbyName": {
        "perform-copy": {
          "taskPlugin": {
            "pluginPackage": "@rushstack/heft",
            "pluginName": "copy-files-plugin",
            "options": {
              ...
            }
          }
        }
      }
    }
  }
}
```

## How it was tested
Locally, building the repo.

## Impacted documentation
All documentation referencing Heft `taskEvent` usage will need to be updated. I have updated `UPGRADING.MD` in this PR, and the updated schema will also need to be uploaded to the CDN.

### DIFF
--- a/apps/heft/UPGRADING.md
+++ b/apps/heft/UPGRADING.md
@@ -1,5 +1,48 @@
 # Upgrade notes for @rushstack/heft
 
+### Heft 0.53.0
+The `taskEvent` configuration option in heft.json has been removed, and use of any `taskEvent`-based functionality is now accomplished by referencing the plugins directly within the `@rushstack/heft` package.
+
+Old format:
+```json
+{
+  "phasesByName": {
+    "build": {
+      "tasksbyName": {
+        "perform-copy": {
+          "taskEvent": {
+            "eventKind": "copyFiles",
+            "options": {
+              ...
+            }
+          }
+        }
+      }
+    }
+  }
+}
+```
+New format:
+```json
+{
+  "phasesByName": {
+    "build": {
+      "tasksbyName": {
+        "perform-copy": {
+          "taskPlugin": {
+            "pluginPackage": "@rushstack/heft",
+            "pluginName": "copy-files-plugin",
+            "options": {
+              ...
+            }
+          }
+        }
+      }
+    }
+  }
+}
+```
+
 ### Heft 0.52.0
 
 The `nodeService` built-in plugin now supports the `--serve` parameter, to be consistent with the `@rushstack/heft-webpack5-plugin` dev server.
@@ -141,8 +184,9 @@ The following is an example "heft.json" file defining both a "build" and a "test
           }
         },
         "copy-assets": {
-          "taskEvent": {
-            "eventKind": "copyFiles",
+          "taskPlugin": {
+            "packageName": "@rushstack/heft",
+            "pluginName": "copy-files-plugin",
             "options": {
               "copyOperations": [
                 {
--- a/apps/heft/UPGRADING.md
+++ b/apps/heft/UPGRADING.md
@@ -3,6 +3,12 @@
 ### Heft 0.53.0
 The `taskEvent` configuration option in heft.json has been removed, and use of any `taskEvent`-based functionality is now accomplished by referencing the plugins directly within the `@rushstack/heft` package.
 
+Plugin name mappings for previously-existing task events are:
+- `copyFiles` -> `copy-files-plugin`
+- `deleteFiles` -> `delete-files-plugin`
+- `runScript` -> `run-script-plugin`
+- `nodeService` -> `node-service-plugin`
+
 Old format:
 ```json
 {
--- a/apps/heft/UPGRADING.md
+++ b/apps/heft/UPGRADING.md
@@ -118,13 +118,13 @@ Lifecycle plugins are specified in the top-level `heftPlugins` array. Plugins ca
 
   "heftPlugins": [
     {
-      "packageName": "@rushstack/heft-metrics-reporter",
+      "pluginPackage": "@rushstack/heft-metrics-reporter",
       "options": {
         "disableMetrics": true
       }
     },
     {
-      "packageName": "@rushstack/heft-initialization-plugin",
+      "pluginPackage": "@rushstack/heft-initialization-plugin",
       "pluginName": "my-lifecycle-plugin"
     }
   ]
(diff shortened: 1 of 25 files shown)


apps/heft/src/plugins/RunScriptPlugin.ts
```diff
@@ -53,10 +54,10 @@ export default class RunScriptPlugin implements IHeftTaskPlugin<IRunScriptPlugin
     pluginOptions: IRunScriptPluginOptions,
     runOptions: IHeftTaskRunHookOptions
   ): Promise<void> {
-    // The scriptPath property should be fully resolved since it is included in the resolution logic used by
-    // HeftConfiguration
-    const resolvedModulePath: string = pluginOptions.scriptPath;
-
+    const resolvedModulePath: string = path.resolve(
```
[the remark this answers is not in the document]
D4N14L: The old version of the script didn't support this, so I'm considering this out of scope. However, in the past I had discussed with @iclanton that we should introduce special path tokens like `<configDir>` or `<rootDir>` to be parsed in the path. But that's for another PR.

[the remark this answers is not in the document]
D4N14L: To answer your question though, they would path through the `node_modules` directory

apps/heft/UPGRADING.md
```diff
@@ -141,8 +190,9 @@ The following is an example "heft.json" file defining both a "build" and a "test
           }
         },
         "copy-assets": {
-          "taskEvent": {
-            "eventKind": "copyFiles",
+          "taskPlugin": {
+            "packageName": "@rushstack/heft",
+            "pluginName": "copy-files-plugin",
```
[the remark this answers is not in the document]
D4N14L: I think it makes sense to allow this for now since it's still brand new and this is intended to be an overview of the new version of Heft. However, I do agree in general that this shouldn't be changed retroactively. Ideally this should be the last breaking schema change.

apps/heft/UPGRADING.md
```diff
@@ -141,8 +173,9 @@ The following is an example "heft.json" file defining both a "build" and a "test
           }
         },
         "copy-assets": {
-          "taskEvent": {
-            "eventKind": "copyFiles",
+          "taskPlugin": {
+            "packageName": "@rushstack/heft",
```
iclanton, line 177: We were talking about omitting this for built-in plugins.

apps/heft/src/pluginFramework/HeftTask.ts
```diff
@@ -193,57 +121,22 @@ export class HeftTask {
...
-      return pluginDefinition as HeftTaskPluginDefinition;
-    } else {
-      // This is validated in the schema so this shouldn't happen, but throw just in case
-      throw new InternalError(
-        `Task ${JSON.stringify(this._taskName)} has no specified task event or task plugin.`
+        `Plugin ${JSON.stringify(pluginSpecifier.pluginName)} specified by task ` +
+          `${JSON.stringify(this._taskName)} is not a task plugin.`
       );
     }
+    return pluginDefinition as HeftTaskPluginDefinition;
```
iclanton, line 139: This cast seems odd.

iclanton:
What would happen if a project and a rig have dependencies on different versions of Heft, and the rig defines tasks using some of the built-in plugins. Will we potentially load two copies of Heft? Will that cause issues?

Similarly, what if a project is using `npm` as their package manager and there are two instances of the `@rushstack/heft` package on disk (even if they're the same version). Will both be loaded? Will that cause issues?

octogonz:
> What would happen if a project and a rig have dependencies on different versions of Heft, and the rig defines tasks using some of the built-in plugins. Will we potentially load two copies of Heft? Will that cause issues?
> 
> Similarly, what if a project is using `npm` as their package manager and there are two instances of the `@rushstack/heft` package on disk (even if they're the same version). Will both be loaded? Will that cause issues?

@iclanton when chatting with @D4N14L, we had said that Heft would contain special logic to guarantee that `@rushstack/heft` always resolves to the actively running Heft instance.  If you wanted to explicitly load a different version of heft, you should use a relative file path or an alias (`"other-heft": "@rushstack/heft@1.2.3"`).

apps/heft/UPGRADING.md
D4N14L: @iclanton I left this out of this PR intentionally. Right now, `pluginName` is optional, and defaults to the first plugin iff there is only one plugin contained in the plugin package. I think this makes sense, because the plugin name isn't terribly important other than to direct Heft to which plugin to load. And I also think there will be a lot of plugin packages with a single plugin, so this could slim down the config significantly. If we _also_ defaulted non-provided plugin packages to `@rushstack/heft`, then we would be in a situation where we would have to then enforce that the plugin name is always provided, but _only_ in the case where `@rushstack/heft` is the package. This seemed overly complex, and leaving it with the original rules (`pluginPackage` must be specified, `pluginName` must be specified if there is more than one plugin) feels much more straightforward

dmichon-msft: There ended up being some potential ambiguity about defaults between "current package" vs "@rushstack/heft"; both sound like a reasonable expectation for default behavior, so might be clearer to not have a default.

octogonz (approved):

apps/heft/heft-plugin.json
```diff
@@ -0,0 +1,35 @@
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/heft/heft-plugin.schema.json",
```
octogonz, line 2: This is the wrong schema URL

D4N14L: Resolving in your schema PR.

PR #4194, opened 2023-06-13
Bump cyclic dependencies.

iclanton:


D4N14L (approved):
Now that we're using @rushstack/heft@0.54.0, could you remove the `cacheFolderPath` property in Heft? This was kept around to avoid issues with mixed-versioning with plugins that target an older version of Heft.

PR #4201, opened 2023-06-14
[rush-lib] Git hook arg passthrough

### LINKED ISSUE #4200: [rush] Delegate git hook files do not pass arguments through
## Summary

We use custom git hooks to check contents of commit messages in a `pre-commit` hook.

Since upgrading to rush 5.100.0 these hooks have stopped working.

The issue seems to be that the new delegate git hook scripts (that were added in 5.100.0) invoke the hooks defined in the `common/git-hooks/` folder, but do not pass any of the git args through.

This is a problem because for the hooks we use, we need access to the temporary commit message file which is passed in as a command line arg to the hook scripts (which we access using $1)

## Repro steps

Define a `pre-commit` hook which reads the commit message file as passed in via `$1`

`$1` does not evaluate to anything as it's not passed through.

Example:

`common/git-hooks/pre-commit`
```
#!/bin/bash

echo "Commit footer text" >> $1
```

This should append the `Commit footer text` to the end of the commit message.

Actual result is an error `$1: ambiguous redirect`

## Details

I believe the fix is just to pass through all args in the delegate scripts.

This can be achieved using `$@` so ` "$SCRIPT_IMPLEMENTATION_PATH"` becomes ` "$SCRIPT_IMPLEMENTATION_PATH" $@`

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? |  5.100.0 |
| `rushVersion` from rush.json? | 5.100.0 |
| Operating system? | Linux (via WSL) |
| Node.js version (`node -v`)? | 18.15.0 |

egmacke:
## Summary

New delegate git hook scripts do not pass through the args that git normally passes to commit hooks.

This ensures that all args are passed through.

Fixes #4200

## How it was tested

Manually changed the commit hook delegate scripts in a test repo to mirror this change. Confirmed that this did then pass through the arguments from git.

### DIFF
--- a/common/changes/@microsoft/rush/git-hook-passthrough_2023-06-14-16-10.json
+++ b/common/changes/@microsoft/rush/git-hook-passthrough_2023-06-14-16-10.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix git hook passthrough",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/common/changes/@microsoft/rush/git-hook-passthrough_2023-06-14-16-10.json
+++ b/common/changes/@microsoft/rush/git-hook-passthrough_2023-06-14-16-10.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Fix git hook passthrough",
+      "comment": "Fix an issue where arguments weren't passed to git hook scripts.",
       "type": "none"
     }
   ],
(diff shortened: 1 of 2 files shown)


common/changes/@microsoft/rush/git-hook-passthrough_2023-06-14-16-10.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix git hook passthrough",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue where arguments weren't passed to git hook scripts.",
```

iclanton (approved):

iclanton:
This has been released in Rush `5.100.1`.

PR #4210, opened 2023-06-16
[package-extractor] Allow for `patternsToInclude` and `patternsToExclude` to filter project files

D4N14L:
## Summary

When searching for all source files to be included in an extraction of a package, allow for providing `patternsToInclude` and `patternsToExclude` to limit which files get included from the local project.

## How it was tested

Locally using various provided patterns.

### DIFF
--- a/libraries/package-extractor/src/PackageExtractor.ts
+++ b/libraries/package-extractor/src/PackageExtractor.ts
@@ -3,6 +3,7 @@
 
 import * as path from 'path';
 import * as fs from 'fs';
+import { IMinimatch, Minimatch } from 'minimatch';
 import npmPacklist from 'npm-packlist';
 import pnpmLinkBins from '@pnpm/link-bins';
 import ignore, { Ignore } from 'ignore';
@@ -94,12 +95,12 @@ export interface IExtractorProjectConfiguration {
    */
   projectFolder: string;
   /**
-   * A list of file patterns to include when extracting this project. If undefined,
+   * A list of glob patterns to include when extracting this project. If undefined,
    * all files will be included.
    */
   patternsToInclude?: string[];
   /**
-   * A list of file patterns to exclude when extracting this project. If undefined,
+   * A list of glob patterns to exclude when extracting this project. If undefined,
    * no files will be excluded.
    */
   patternsToExclude?: string[];
@@ -608,8 +609,8 @@ export class PackageExtractor {
     const { includeNpmIgnoreFiles, targetRootFolder } = options;
     const { projectConfigurationsByPath, archiver } = state;
     let useNpmIgnoreFilter: boolean = false;
-    let includeFilter: Ignore | undefined;
-    let excludeFilter: Ignore | undefined;
+    let includeFilters: IMinimatch[] | undefined;
+    let excludeFilters: IMinimatch[] | undefined;
 
     const sourceFolderRealPath: string = await FileSystem.getRealPathAsync(sourceFolderPath);
     const sourceProjectConfiguration: IExtractorProjectConfiguration | undefined =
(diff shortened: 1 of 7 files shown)


libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -93,6 +93,16 @@ export interface IExtractorProjectConfiguration {
...
+  /**
+   * A list of file patterns to include when extracting this project. If undefined,
+   * all files will be included.
+   */
+  patternsToInclude?: string[];
+  /**
+   * A list of file patterns to exclude when extracting this project. If undefined,
+   * no files will be excluded.
+   */
+  patternsToExclude?: string[];
```
iclanton, line 105: Mention in the doc comment what kind of pattern it is. Glob? `.gitignore`-style?

iclanton (approved):

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -93,6 +94,16 @@ export interface IExtractorProjectConfiguration {
...
+  /**
+   * A list of glob patterns to include when extracting this project. If undefined,
+   * all files will be included.
+   */
+  patternsToInclude?: string[];
+  /**
+   * A list of glob patterns to exclude when extracting this project. If undefined,
+   * no files will be excluded.
+   */
+  patternsToExclude?: string[];
```
dmichon-msft, line 106: Document which wins if a file covered by both `patternsToInclude` and `patternsToExclude`.

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -598,14 +609,42 @@ export class PackageExtractor {
...
+    }
+
+    function isFileExcluded(filePath: string): boolean {
+      // If we're not in a project folder, or if there are no filters, then we can't exclude anything.
+      if (!includeFilters && !excludeFilters) {
+        return false;
+      }
+
+      // Since the includeFilter is used to match against included files, if `includeFilter.ignores()`
+      // returns true, we know that the file is included.
```
dmichon-msft, line 642: Update comment to match the new code

D4N14L:
@EscapeB where is this folder existing? It looks like the path you're providing isn't in a node_modules folder. You could try using the `patternsToExclude` field to exclude `node_gyp_bins` if this file lives outside of the node_modules folder. Package extractor will not traverse a dependency if it's included in the `dependenciesToExclude` field so not sure how that didn't work for you by specifying `dependenciesToExclude: ["@my-team/foo"]`

EscapeB:
@D4N14L This folder is inside node_modules. To be more clear, this folder path is `monorepo-root-path/packages/my-apps/node_modules/@my-team/foo/node_gyp_bins/python3 ` and it is linked to global path `/usr/bin/python3`.

We want to include `@my-team/foo` this dependencies(as it is needed for this app package), but want to exclude some of its subfolders, like `@my-team/foo/node_gyp_bins/python3` .

EscapeB:
I can help with the implementation to provide field that allow users to exclude subfolders of dependencies inside node_modules. 

But I'm not sure if this field can be merged with `patternsToInclude`.

D4N14L:
@EscapeB Ahh okay I think I understand. In that case, I believe you need to set the `patternsToExclude` to include `node_gyp_bins` for the project `@my-team/foo` in your deploy configuration. This might be a feature that needs to be surfaced on the configuration still, but it should work (the code exists to support this I believe). If you'd like to make a PR to do this, I'd be happy to review it

D4N14L:
@EscapeB  I've created a PR over here surfacing the properties on the deploy.json for Rush: https://github.com/microsoft/rushstack/pull/4259

Something to note is that I explicitly did _not_ make changes to exclude files from dependencies. This can get problematic, ex. for scenarios where we want to exclude files within a dependency of a dependency. The PR itself should address the issues you mention having in this thread by surfacing already-existing functionality on the deploy.json configuration file.

jltxwesley:
Thanks for adding this, we found two packages in our project have this issue, 

- `common/temp/node_modules/.pnpm/fsevents@1.2.13/node_modules/fsevents/build/node_gyp_bins/python3`
-  `common/temp/node_modules/.pnpm/@newrelic+native-metrics@7.1.2/node_modules/@newrelic/native-metrics/build/node_gyp_bins/python3`

PR #4212, opened 2023-06-19
[rig-package] Expose an `IRigConfig` interface and use it in consuming projects' APIs.

iclanton:
## Summary

Similar to `Terminal` vs `ITerminal`, this PR is creating an `IRigConfig` interface to be used in APIs to avoid issues with mismatched versions of `@rushstack/rig-package`.

## How it was tested

Built.

## Impacted documentation

The `IRigConfig` interface needs to be added to the API docs.

### DIFF
--- a/apps/api-extractor/src/api/ExtractorConfig.ts
+++ b/apps/api-extractor/src/api/ExtractorConfig.ts
@@ -16,7 +16,7 @@ import {
   Path,
   NewlineKind
 } from '@rushstack/node-core-library';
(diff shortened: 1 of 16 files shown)


dmichon-msft (approved):
Only thing I might change is to differentiate the docs on the `RigConfig` class vs. the `IRigConfig` interface, since the latter is only for instances and the former has the loader API.

iclanton:
Ehh I'm not really concerned about that in the docs. It should be pretty obvious, and people can keep using the `RigConfig` type if they want.

PR #4229, opened 2023-07-10
[DEMO ONLY] Perform CI installation via phased command orchestration

dmichon-msft:
## Summary
This PR is a tech demonstration of a streamlined dependency installation flow for use on CI, to be used as reference for propagating the performance enhancements upstream into package managers.

## Details
The most interesting component of this PR is the tarball hash/decompress/parse/extract worker.
The next most interesting is the networking layer that directly uses the `node:https` module for reliability and performance.

## How it was tested
This is code currently used in use, with various secrets redacted and left as placeholders.

### DIFF
--- a/rush-plugins/rush-phased-install-plugin/src/worker/tarballWorker.ts
+++ b/rush-plugins/rush-phased-install-plugin/src/worker/tarballWorker.ts
@@ -1,0 +1,229 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import * as crypto from 'node:crypto';
+import * as fs from 'node:fs';
+import { parentPort } from 'node:worker_threads';
+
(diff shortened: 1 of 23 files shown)


rush-plugins/rush-phased-install-plugin/src/worker/tarballWorker.ts
```diff
@@ -0,0 +1,229 @@
...
+  size: number;
+}
+
+const ZERO: number = '0'.charCodeAt(0);
+const SPACE: number = ' '.charCodeAt(0);
+const SLASH: number = '/'.charCodeAt(0);
+const BACKSLASH: number = '\\'.charCodeAt(0);
+
+// See TAR specification here: https://www.gnu.org/software/tar/manual/html_node/Standard.html
+function parseTarball(buffer: Buffer): IParseResult {
```
dmichon-msft, line 55: This needs some updates to parse PAX extended headers.

PR #4232, opened 2023-07-11
[heft-typescript] Fix unexpected warning to error promotion

dmichon-msft:
## Summary
Ensures that the "Typescript encountered X warnings and is configured to build project references" error message only is displayed when `buildProjectReferences` is `true`. This error message is only supposed to occur when using the `createSolutionBuilder` APIs in the TypeScript compiler, because when using those APIs, the compiler implicitly enables the `noEmitOnError` behavior.

## Details
Added a missing check to the condition for emitting the error.

## How it was tested
- [x] Added an unused variable in `heft-typescript-composite-test` (which uses `buildProjectReferences: true`) and verified the error is emitted, yielding `FAILURE` state.
- [x] Added an unused variable in `heft-webpack4-everything-test` (which has `buildProjectReferences: false`) and verified the error is not emitted, yielding `SUCCESS WITH WARNINGS` state.

## Impacted documentation
None

### DIFF
--- a/common/changes/@rushstack/heft-typescript-plugin/fix-typescript-error_2023-07-11-21-40.json
+++ b/common/changes/@rushstack/heft-typescript-plugin/fix-typescript-error_2023-07-11-21-40.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-typescript-plugin",
+      "comment": "Only make warnings terminal if \"buildProjectReferences\" is true.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/heft-typescript-plugin"
+}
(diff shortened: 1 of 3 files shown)


D4N14L (approved):

common/changes/@rushstack/heft-typescript-plugin/fix-typescript-error_2023-07-11-21-40.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-typescript-plugin",
+      "comment": "Only make warnings terminal if \"buildProjectReferences\" is true.",
```
iclanton, line 5: What does "make warnings terminal" mean?

iclanton: Do you mean "make warnings a hard error?" Using the word "terminal" makes it sound like you're conditionally writing to a console.

dmichon-msft: "terminal" as in "stops the build"

PR #4246, opened 2023-07-19
[heft] Update Heft clean and `delete-files-plugin` to delete files by default instead of folders

D4N14L:
## Summary

This PR updates the `cleanFiles` property in heft.json as well as plugin options for the `delete-files-plugin` to delete the contents of folders by default (instead of the current behavior of deleting the folder if no `includeGlobs` property is specified). The `sourcePath` will continue to resolve to the root of the package targeted by Heft. Additionally, `includeGlobs` will default to `**/*` for both copy and delete operations.

This is a breaking change, and deletion options once specified as the following:
```json
{
  "cleanFiles": [{ "sourcePath": "folder" }]
}
```

will now delete all files and folders under the `folder` path, but not `folder` itself. To restore the previous behavior, use the following:
```json
{
  "cleanFiles": [{ "includeGlobs": ["folder"] }]
}
```

## How it was tested

Tested locally.

## Impacted documentation

The schema will need to be updated on the service to ensure that at least one of the targeting properties is provided to the file glob specification for cleans, copies, or deletes.


dmichon-msft:
The intention was that if not specified, the value was implied to be `.`. Did that not end up being reflected in the parsing?

iclanton (approved):

octogonz:
@D4N14L the website docs also need to be updated to mark this field as `/** (REQUIRED)`

octogonz:
@D4N14L the website docs also need to be updated to mark this field as `/** (REQUIRED)`

D4N14L:
> @D4N14L the website docs also need to be updated to mark this field as `/** (REQUIRED)`
> 
> https://heft.rushstack.io/pages/plugins/copy-files/

@octogonz so I've since updated the PR, but the intention of the original code was actually to allow the `sourcePath` to be left blank. If it is, it will default to the root of the package.

I've updated the description on the PR as well, and marked the changelog as "BREAKING" since I've modified the default behaviour of the delete operations (for both the delete-files-plugin and when cleaning with Heft)

apps/heft/src/plugins/FileGlobSpecifier.ts
```diff
@@ -129,39 +136,55 @@ export async function watchGlobAsync(
...
+  if (fileSystemAdapter && isWatchFileSystemAdapter(fileSystemAdapter)) {
+    results = new Map();
     await Async.forEachAsync(
-      rawFiles,
-      async (file: string) => {
-        const state: IWatchedFileState = await fs.getStateAndTrackAsync(path.normalize(file));
+      rawEntries,
+      async (entry: Entry) => {
+        const { path: filePath, dirent } = entry;
+        if (entry.dirent.isDirectory()) {
```
D4N14L, line 160: @dmichon-msft  does it make sense to skip directories here? Only include the watched files?

apps/heft/src/plugins/FileGlobSpecifier.ts
```diff
@@ -129,39 +136,55 @@ export async function watchGlobAsync(
...
+  const { fileGlobSpecifier, includeFolders, fileSystemAdapter } = options;
+  const rawEntries: Entry[] = await glob(fileGlobSpecifier.includeGlobs!, {
+    fs: fileSystemAdapter,
     cwd: fileGlobSpecifier.sourcePath,
     ignore: fileGlobSpecifier.excludeGlobs,
     dot: true,
-    absolute: true
+    absolute: true,
+    onlyFiles: !includeFolders,
+    stats: true
```
dmichon-msft, line 150: I just know I'm going to have to write my own glob implementation sooner or later so that I can stop wasting CPU cycles making `stat` calls when `fs.readdir` returns everything any decent globber needs to know.

apps/heft/src/plugins/FileGlobSpecifier.ts
dmichon-msft: Yes. The `readdir` calls will have already been tracked.

apps/heft/src/plugins/FileGlobSpecifier.ts
dmichon-msft: Aha! Use `objectMode: true` instead: https://github.com/mrmlnc/fast-glob#objectmode

apps/heft/src/plugins/DeleteFilesPlugin.ts
```diff
@@ -78,16 +94,26 @@ async function _deleteFilesInnerAsync(pathsToDelete: Set<string>, terminal: ITer
...
+          throw error;
+        }
+      }
+    },
+    { concurrency: Constants.maxParallelism }
+  );
+
+  // Clear out any folders that were encountered during the file deletion process. These
+  // folders should already be empty.
+  await Async.forEachAsync(
```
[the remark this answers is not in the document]
D4N14L: I'm confused what your point here is? These folders should already be empty by virtue of clearing out the children in the code above (for example, in a `**/*` scenario). If a directory is matched by a glob pattern my assumption is that all children are either a) already deleted, or b) will need to be deleted. We also get the advantage of not needing to try-catch the deletion of the target path since we already know what is a directory and what is a file.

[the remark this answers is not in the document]
D4N14L: Yes, though it does accomplish the task it's intended to (it clears out empty folders from patterns like `**/*`, and deletes folders that match the pattern directly). So not sure this is an issue?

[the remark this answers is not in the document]
dmichon-msft: Figured it might be worth including a code comment since this isn't the traversal order normally associated with deleting a folder tee. Normally when doing a recursive delete folders would be unlinked in post-order tree-walk order to address only being able to delete empty folders.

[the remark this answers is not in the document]
D4N14L: Fair. Will add that.

apps/heft/src/plugins/CopyFilesPlugin.ts
```diff
@@ -218,7 +211,9 @@ export default class CopyFilesPlugin implements IHeftTaskPlugin<ICopyFilesPlugin
     taskSession.hooks.registerFileOperations.tap(
       PLUGIN_NAME,
       (operations: IHeftTaskFileOperations): IHeftTaskFileOperations => {
-        for (const operation of _resolveCopyOperationPaths(heftConfiguration, pluginOptions.copyOperations)) {
+        for (const operation of pluginOptions.copyOperations) {
+          // Do this here so that we only have to do it once
+          normalizeCopyOperation(heftConfiguration.buildFolderPath, operation);
```
dmichon-msft, line 216: The result of this hook evaluation is passed to the `normalizeCopyOperation` loop in `TaskOperationRunner`, so converting here is redundant. You've merged the `_resolveCopyOperationPaths` with `normalizeFileGlobSpecifier` in `normalizeCopyOperation`, so this call and the call in `TaskOperationRunner` now each contain both halves of the normalization.

apps/heft/src/plugins/FileGlobSpecifier.ts
```diff
@@ -129,39 +136,55 @@ export async function watchGlobAsync(
...
 }
 
-export function normalizeFileSelectionSpecifier(fileGlobSpecifier: IFileSelectionSpecifier): void {
+export function normalizeFileSelectionSpecifier(
+  rootPath: string,
+  fileGlobSpecifier: IFileSelectionSpecifier
+): void {
+  fileGlobSpecifier.sourcePath = fileGlobSpecifier.sourcePath
+    ? path.resolve(rootPath, fileGlobSpecifier.sourcePath)
+    : rootPath;
```
dmichon-msft, line 187:
```suggestion
  const { sourcePath } = fileGlobSpecifier;
  fileGlobSpecifier.sourcePath = sourcePath
    ? path.resolve(rootPath, sourcePath)
    : rootPath;
```
I find it makes it slightly more readable and clearer that the value is the same in both the condition and the resolve call.

dmichon-msft (approved):

PR #4252, opened 2023-07-25
[rush] Improve accuracy of detecting pnpm shrinkwrap is modified

chengcyber:
## Summary

Improve accuracy of detecting pnpm shrinkwrap modified

### Case 1: pnpm.overrides

With `pnpm.overrides`, the version in package.json is different from the specifier in pnpm lockfile. This MR changes the logic to compare the overrided version with specifier in lockfile when necessary.

### Case 2: duplicated inconsistent version in dep and devDep.

There is a chance that one package declared both in `dependencies` and `devDependencies` with inconsistent versions (absurd). This MR ignores the version in devDependencies when detecting for pnpm lockfile v6

NOTE: when detecting pnpm lockfile v5, the implementation iterates `specifiers` in lockfile where dev dependencies' versions never show up.

## How it was tested

Test cases are added

## Impacted documentation

No

### DIFF
--- a/common/changes/@microsoft/rush/improve-pnpm-shrinkwrap-modified_2023-07-25-12-11.json
+++ b/common/changes/@microsoft/rush/improve-pnpm-shrinkwrap-modified_2023-07-25-12-11.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "improve accuracy of detecting pnpm shrinkwrap is modified",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/common/changes/@microsoft/rush/improve-pnpm-shrinkwrap-modified_2023-07-25-12-11.json
+++ b/common/changes/@microsoft/rush/improve-pnpm-shrinkwrap-modified_2023-07-25-12-11.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "improve accuracy of detecting pnpm shrinkwrap is modified",
+      "comment": "Reduce false positive detections of the pnpm shrinkwrap file being out of date in the presence of the `globalOverrides` setting in `pnpm-config.json`, or when a dependency is listed in both `dependencies` and `devDependencies` in the same package.",
       "type": "none"
     }
   ],
--- a/libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
+++ b/libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
@@ -778,6 +778,7 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
       for (const { dependencyType, name, version } of allDependencies) {
         let isOptional: boolean = false;
         let specifierFromLockfile: IPnpmVersionSpecifier | undefined;
+        let isDevDepFallThrough: boolean = false;
         switch (dependencyType) {
           case DependencyType.Optional: {
             specifierFromLockfile = importer.optionalDependencies?.[name];
@@ -798,6 +799,8 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
               importerDevDependencies.delete(name);
               break;
             }
+            // If fall through, there is a chance the package declares an inconsistent version, ignore it.
+            isDevDepFallThrough = true;
           }
 
           // eslint-disable-next-line no-fallthrough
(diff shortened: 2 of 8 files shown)


libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -755,7 +759,11 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
       for (const [importerPackageName, importerVersionSpecifier] of Object.entries(specifiers)) {
         const foundDependency: PackageJsonDependency | undefined =
           dependencyVersions.get(importerPackageName);
-        if (!foundDependency || foundDependency.version !== importerVersionSpecifier) {
+        if (!foundDependency) {
+          return true;
+        }
+        const resolvedVersion: string = this.overrides.get(importerPackageName) ?? foundDependency.version;
```
dmichon-msft, line 765: The `overrides` field also accepts keys like `@rushstack/heft@<=0.50.6`, or `@rushstack/heft>@rushstack/node-core-library`, not just package names, but I guess this will catch the basic case.

dmichon-msft (commented):
The scenario of using `pnpm.overrides` to override the installed versions of dependencies declared in your own workspace package.json files is something that I would honestly consider an error case. You shouldn't be specifying one value in package.json and then using pnpm.overrides to install an incompatible version.

The case of having two different versions between `dependencies` and `devDependencies` is something I've considered doing for the purpose of having `dependencies` specify a range and `devDependencies` specify an exact version, but haven't seen particularly consistent behavior when that is attempted. Some tools error, some pick one or the other, etc.

As currently written I'm not sure exactly what this is solving, other than at least making the behavior in the aforementioned edge case of overriding the dependencies declared in the very same repository be consistent with the pnpm installation.

chengcyber:
HI @dmichon-msft . Thank you for reviewing this!

For case 1, when the `pnpm.overrides` is specified, the projects which has dependency over the overrided packages will be always marked as "out of date". which blocks `rush install` forever.

I created a repro repo at https://github.com/chengcyber/rush-pnpm-overrides-repro

For case 2, even though the case is specifying a range in dependencies and a fixed version in devDependencies, the project will be marked as "out of date". which blocks `rush install` forever.

I also created a repro repo at https://github.com/chengcyber/rush-pnpm-dep-devdep-repro

common/changes/@microsoft/rush/improve-pnpm-shrinkwrap-modified_2023-07-25-12-11.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "improve accuracy of detecting pnpm shrinkwrap is modified",
```
dmichon-msft, line 5:
```suggestion
      "comment": "Reduce false positive detections of the pnpm shrinkwrap file being out of date in the presence of the `globalOverrides` setting in `pnpm-config.json`, or when a dependency is listed in both `dependencies` and `devDependencies` in the same package.",
```

dmichon-msft (approved):

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -810,7 +821,8 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
                 `"${specifierFromLockfile}" instead of an object.`
             );
           } else {
-            if (specifierFromLockfile.specifier !== version && !isOptional) {
+            const resolvedVersion: string = this.overrides.get(name) ?? version;
```
iclanton, line 824:
```suggestion
            // TODO: Add an error message when someone tries to override a version of something in one of their
            // local repo packages.
            const resolvedVersion: string = this.overrides.get(name) ?? version;
```

iclanton (approved):

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -810,7 +821,10 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
                 `"${specifierFromLockfile}" instead of an object.`
             );
           } else {
-            if (specifierFromLockfile.specifier !== version && !isOptional) {
+            // TODO: Add an error message when someone tries to override a version of something in one of their
```
iclanton, line 824:
```suggestion
            // TODO: Emit an error message when someone tries to override a version of something in one of their
```

PR #4253, opened 2023-07-27
[heft] Fix sequencing of toolFinish lifecycle hook

dmichon-msft:
## Summary
Ensures that the `toolFinish` lifecycle hook is invoked strictly _after_ the `recordMetrics` lifecycle hook. Also ensures that the `toolFinish` lifecycle hook is invoked during graceful shutdown of Heft (e.g. via Ctrl+C).

## Details
Previously `toolFinish` was erroneously invoked before the `recordMetrics` hook. This inadvertently made telemetry plugins try to log events after they had already shut down.

Also documents the `recordMetrics` hook.

## How it was tested
Manual testing.

## Impacted documentation
None.


apps/heft/src/cli/HeftActionRunner.ts
```diff
@@ -282,10 +282,15 @@ export class HeftActionRunner {
...
-      await this._executeOnceAsync(executionManager, cliCancellationToken);
+    try {
+      if (this._action.watch) {
+        await this._executeWatchAsync(executionManager, cliCancellationToken);
+      } else {
+        await this._executeOnceAsync(executionManager, cliCancellationToken);
+      }
+    } finally {
+      // Invoke this here both to ensure it always runs and that it does so after recordMetrics
+      await this._internalHeftSession.lifecycle.hooks.toolFinish.promise();
```
D4N14L, line 293: Should we only run this if we confirmed that `toolStart` has executed successfully? Seems almost like we should consider pulling lifecycle clean + toolStart + toolStop directly into the action runner, and leave the operation runner to handle phases + tasks?

apps/heft/src/pluginFramework/HeftLifecycleSession.ts
```diff
@@ -100,7 +100,12 @@ export interface IHeftLifecycleHooks {
    */
   toolFinish: AsyncParallelHook<IHeftLifecycleToolFinishHookOptions>;
 
-  // TODO: Wire up and document this hook.
+  /**
+   * The `recordMetrics` hook is called at the end of every Heft execution pass. It is called after all
```
D4N14L, line 104: Mention whether it runs before or after the toolstart/toolfinish hooks as well

apps/heft/src/cli/HeftActionRunner.ts
```diff
@@ -280,12 +282,24 @@ export class HeftActionRunner {
...
+      await _startLifecycleAsync(this._internalHeftSession);
+
+      if (this._action.watch) {
+        await this._executeWatchAsync(executionManager, cliCancellationToken);
+      } else {
+        await this._executeOnceAsync(executionManager, cliCancellationToken);
+      }
+    } finally {
+      // Invoke this here both to ensure it always runs and that it does so after recordMetrics
+      const lifecycleToolFinishHookOptions: IHeftLifecycleToolFinishHookOptions = {};
```
D4N14L, line 301: Punt this out into it's own helper function just like _startLifecycleAsync

D4N14L (approved):

PR #4255, opened 2023-07-28
[package-extractor]Add dependencies configurations field in package extractor

EscapeB:
## Summary
Add "dependenciesConfigurations" field in options. 
Related discussion in https://github.com/microsoft/rushstack/pull/4210

## Details
1. Provide "dependenciesConfigurations" field in options to configure behavior when extracting third party dependencies. 
2. Add similar fields named "patternToInclude" and "patternToExclude" to filter files inside third party dependencies.

## How it was tested
Added some unit test cases to test the new field.


EscapeB:
Hi @D4N14L, could you please help to review this PR?

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -74,8 +74,12 @@ export interface IExtractorMetadataJson {
 
 interface IExtractorState {
   foldersToCopy: Set<string>;
+  packageJsonByPath: Map<string, IPackageJson>;
   projectConfigurationsByPath: Map<string, IExtractorProjectConfiguration>;
   projectConfigurationsByName: Map<string, IExtractorProjectConfiguration>;
+  /* As there might be different versions of same dependency, we need take version field into consideration */
+  /* The key format `${name}@${version}` */
```
iclanton, line 81:
```suggestion
  /**
   * As there might be different versions of same dependency, we need take version field into consideration
   * The key format `${name}@${version}`
   */
```

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -120,6 +124,34 @@ export interface IExtractorProjectConfiguration {
...
+  /**
+   * The version of dependency
+   */
+  dependencyVersion: string;
+  /**
+   * A list of glob patterns to exclude when extracting this dependency. If a path is
+   * matched by both "patternsToInclude" and "patternsToExclude", the path will be
+   * excluded. If undefined, no paths will be excluded.
+   */
+  patternsToExclude?: string[];
```
iclanton, line 146: Are these paths? Maybe update the name to something like `filePathPatternsToExclude` and `filePathPatternsToInclude` to be completely clear.

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -646,6 +693,32 @@ export class PackageExtractor {
       // and there is no exclude filter, then we know that the file is not excluded. If it is included
       // and there is an exclude filter, then we need to check for a match.
       return !isIncluded || !!excludeFilters?.some((m) => m.match(filePath));
+    };
+    // Function to filter files inside local project
+    const isFileExcludedForProject = (filePath: string): boolean =>
+      isFileExcluded(filePath, {
+        ...sourceProjectConfiguration
+      });
```
iclanton, line 701:
```suggestion
      isFileExcluded(filePath, sourceProjectConfiguration);
```

The second parameter doesn't appear to get modified, so it doesn't need to be cloned.

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -608,34 +651,38 @@ export class PackageExtractor {
...
+
+    const packagesJson: IPackageJson | undefined = packageJsonByPath.get(sourceFolderPath);
+    // As this function will be used to copy folder for both project inside monorepo and third party dependencies insides node_modules
+    // Third part dependencies won't have project configurations
+    const isLocalProject: boolean = !!sourceProjectConfiguration;
+
+    // Base filter function
+    const isFileExcluded = (
+      filePath: string,
+      options: { patternsToInclude?: string[]; patternsToExclude?: string[] }
```
iclanton, line 675:
```suggestion
      { patternsToInclude, patternsToExclude}: Pick<IExtractorDependencyConfiguration, 'patternsToInclude' | 'patternsToExclude'>
```

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -608,34 +651,38 @@ export class PackageExtractor {
...
+    ): boolean => {
+      let includeFilters: IMinimatch[] | undefined;
+      let excludeFilters: IMinimatch[] | undefined;
+      if (options.patternsToInclude?.length) {
+        includeFilters = options.patternsToInclude?.map((p) => new Minimatch(p, { dot: true }));
```
iclanton, line 680:
```suggestion
      if (patternsToInclude?.length) {
        includeFilters = patternsToInclude?.map((p) => new Minimatch(p, { dot: true }));
```

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -608,34 +651,38 @@ export class PackageExtractor {
...
       }
-      if (sourceProjectConfiguration.patternsToExclude?.length) {
-        excludeFilters = sourceProjectConfiguration.patternsToExclude.map(
-          (p) => new Minimatch(p, { dot: true })
-        );
+      if (options.patternsToExclude?.length) {
+        excludeFilters = options.patternsToExclude?.map((p) => new Minimatch(p, { dot: true }));
```
iclanton, line 683:
```suggestion
      if (patternsToExclude?.length) {
        excludeFilters = patternsToExclude?.map((p) => new Minimatch(p, { dot: true }));
```

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -646,6 +693,32 @@ export class PackageExtractor {
...
+      }
+      const dependenciesConfiguration: IExtractorDependencyConfiguration | undefined =
+        dependencyConfigurationsByNameAndVersion.get(`${packagesJson.name}@${packagesJson.version}`);
+      if (!dependenciesConfiguration) {
+        return false;
+      }
+
+      return isFileExcluded(filePath, {
+        ...dependenciesConfiguration
+      });
```
iclanton, line 715:
```suggestion
      return isFileExcluded(filePath, dependenciesConfiguration);
```

libraries/package-extractor/src/tests/PackageExtractor.test.ts
```diff
@@ -0,0 +1,155 @@
+import path from 'path';
```
iclanton, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
// See LICENSE in the project root for license information.

import path from 'path';
```

libraries/package-extractor/src/tests/PackageExtractor.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    projectFolder: p.projectFolder,
+    projectName: p.packageName
+  }));
+  beforeAll(() => {
+    FileSystem.ensureEmptyFolder(testRepoBaseTargetFolder);
+  });
+  afterAll(() => {
+    // clean up target folder after each test run
+    // FileSystem.ensureEmptyFolder(testRepoBaseTargetFolder);
+  });
```
iclanton, line 29: Intentionally commented out?

libraries/package-extractor/src/tests/PackageExtractor.test.ts
```diff
@@ -0,0 +1,155 @@
...
+import {
+  PackageExtractor,
+  type IExtractorProjectConfiguration,
+  type IExtractorDependencyConfiguration
+} from '../PackageExtractor';
+
+describe(PackageExtractor.name, () => {
+  // because this files will be compiled to js, and the path will change if using __dirname
+  // using path inside src folder
+  const testRepoRoot = path.join(path.resolve(__dirname, '../../'), 'src/tests/package-extractor-test-repo');
```
iclanton, line 14:
```suggestion
  const testRepoRoot = path.resolve(__dirname, '../../src/tests/package-extractor-test-repo');
```

libraries/package-extractor/src/tests/Utils.ts
```diff
@@ -0,0 +1,22 @@
+import path from 'path';
```
iclanton, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
// See LICENSE in the project root for license information.

import path from 'path';
```

libraries/package-extractor/src/tests/PackageExtractor.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  const testRepoBaseTargetFolder = path.join(testRepoRoot, 'output');
+  const projects = loadRushProjectFromConfigurationFile(path.resolve(testRepoRoot, 'rush.json'));
```
iclanton, line 16: This should be passed in via an argument instead of parsing it again here.

libraries/package-extractor/src/PackageExtractor.ts
EscapeB: They are glob patterns, same as option filed in projectConfigurations.

libraries/package-extractor/src/tests/PackageExtractor.test.ts
EscapeB: No, this line should be kept to clean test files generated in unit tests, will fix it

libraries/package-extractor/src/tests/PackageExtractor.test.ts
EscapeB: I'm not sure how can it be passed in...As it's inside jest test function.
Could you please provide more details?

libraries/package-extractor/src/tests/package-extractor-test-repo/packages/foo/node_modules/.pnpm/baz/package.json
```diff
@@ -0,0 +1,5 @@
+{
```
D4N14L, line 1: The node_modules directory should not be checked in and should be gitignored. If this is required, an install should be performed as a part of the build process.

D4N14L: Ahh hmm, this is intentional. Seems a bit janky of a way to test this... it's not actually an install, so it doesn't have symlinks etc. that would normally exist... Yeah, I would suggest doing an actual install to get the state on-disk to match what you plan to test. Really, you could make two test projects within the repo, and one of them would run the deploy and validate output content. You can add the deploy scenario to the config in the Rushstack repo, so that everything should be prepared for you.

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -608,34 +653,41 @@ export class PackageExtractor {
...
-      if (!includeNpmIgnoreFiles) {
-        // Only use the npmignore filter if the project configuration explicitly asks for it
-        useNpmIgnoreFilter = true;
-      }
-      if (sourceProjectConfiguration.patternsToInclude?.length) {
-        includeFilters = sourceProjectConfiguration.patternsToInclude.map(
-          (p) => new Minimatch(p, { dot: true })
-        );
+
+    const packagesJson: IPackageJson | undefined = packageJsonByPath.get(sourceFolderPath);
```
D4N14L, line 669: It feels like there's a lot more going on here than is needed... I have to run, but I'll try to take a closer look tonight to see if there's a way to handle this a bit more simply

libraries/package-extractor/src/tests/package-extractor-test-repo/packages/foo/node_modules/.pnpm/baz/package.json
EscapeB: Yes, just created a fake node_modules structure for testing. It had symlinks like pnpm created but without hardlinks to pnpm global store.

1. packages/foo/node_modules/bar - > packages/bar                                                 (local project dependencies)
2. packages/foo/node_modules/baz -> packages/foo/node_modules/.pnpm/baz    (third party dependencies)

EscapeB: Two test projects inside the repo only mocks the local project dependency scenario...

D4N14L:
I just merged my PR which should address your concerns: https://github.com/microsoft/rushstack/pull/4259

I'm a bit iffy on adding general 3rd party pattern include/exclude support given that it's a bit ambiguous if it applies only to the version of the package consumed directly by a project, or another dependency, or otherwise. The original issue should be addressed by my PR though.

EscapeB:
> I just merged my PR which should address your concerns: #4259
> 
> I'm a bit iffy on adding general 3rd party pattern include/exclude support given that it's a bit ambiguous if it applies only to the version of the package consumed directly by a project, or another dependency, or otherwise. The original issue should be addressed by my PR though.

I will rebase my PR branch~

But seems this PR can't solve my original issue, the symlink(which linked to global path) inside third party dependencies still will cause errors :(

EscapeB:
Hi @D4N14L , what I need to do if I want to get this PR merged?

D4N14L:
@EscapeB I'm so sorry that I've let this slide. I've been pretty busy of late and just haven't had a chance to take a closer look. Plan on doing so tomorrow when I can find a moment. Again, apologies for holding this up.

libraries/package-extractor/src/PackageExtractor.ts
D4N14L, line 669: Use the `sourceFolderRealPath` instead. Looks like that is what's used when it's originally added to the map

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -681,12 +753,22 @@ export class PackageExtractor {
...
           // We can detect the duplicates by comparing the path.resolve() result.
           const copySourcePath: string = path.resolve(sourceFolderPath, npmPackFile);
+
+          // Filter out files that are excluded by the project configuration or dependency configuration.
+          if (
+            (isLocalProject && isFileExcludedForProject(npmPackFile)) ||
+            (!isLocalProject && isFileExcludedForDependency(npmPackFile))
+          ) {
+            return;
+          }
```
D4N14L, line 763: Move this before the `path.resolve`. It's more expensive than it has any right to be, so we should avoid doing it if we can.

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -734,13 +816,20 @@ export class PackageExtractor {
             return;
           }
 
+          // Filter out files that are excluded by the project configuration or dependency configuration.
+          if (
+            relativeSourcePath !== '' &&
+            ((isLocalProject && isFileExcludedForProject(relativeSourcePath)) ||
+              (!isLocalProject && isFileExcludedForDependency(relativeSourcePath)))
+          ) {
+            terminal.writeLine(`Extra exclude file ${relativeSourcePath} in ${sourceFolderPath}.`);
```
D4N14L, line 825: I think we should drop this, it'll get pretty verbose.

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -734,13 +816,20 @@ export class PackageExtractor {
             return;
           }
 
+          // Filter out files that are excluded by the project configuration or dependency configuration.
+          if (
+            relativeSourcePath !== '' &&
```
D4N14L, line 821: Both of these depend on the relative source path, Just put them under the same if block

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -120,6 +126,34 @@ export interface IExtractorProjectConfiguration {
...
+ */
+export interface IExtractorDependencyConfiguration {
+  /**
+   * The name of dependency
+   */
+  dependencyName: string;
+  /**
+   * The version of dependency
+   */
+  dependencyVersion: string;
```
D4N14L, line 142: We should make this a semver version range

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -274,8 +314,12 @@ export class PackageExtractor {
     // Create a new state for each run
     const state: IExtractorState = {
       foldersToCopy: new Set(),
+      packageJsonByPath: new Map(),
       projectConfigurationsByName: new Map(projectConfigurations.map((p) => [p.projectName, p])),
       projectConfigurationsByPath: new Map(projectConfigurations.map((p) => [p.projectFolder, p])),
+      dependencyConfigurationsByNameAndVersion: new Map(
+        dependenciesConfigurations.map((p) => [`${p.dependencyName}@${p.dependencyVersion}`, p])
```
D4N14L, line 321: If we turn the version into a version range, this will likely have to turn into a `Map<string, IExtractorDependencyConfiguration[]>` which is name -> config, and then cycle through the included configs to check the ranges with semver.

common/reviews/api/package-extractor.api.md
```diff
@@ -18,6 +26,7 @@ export interface IExtractorMetadataJson {
 export interface IExtractorOptions {
     createArchiveFilePath?: string;
     createArchiveOnly?: boolean;
+    dependenciesConfigurations?: IExtractorDependencyConfiguration[];
```
D4N14L, line 29:
```suggestion
    dependencyConfigurations?: IExtractorDependencyConfiguration[];
```

D4N14L (changes requested):

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -646,6 +698,31 @@ export class PackageExtractor {
...
+      const dependenciesConfiguration: IExtractorDependencyConfiguration | undefined =
+        dependencyConfigurationsByNameAndVersion.get(`${packagesJson.name}@${packagesJson.version}`);
+      if (!dependenciesConfiguration) {
+        return false;
+      }
+
+      return isFileExcluded(filePath, dependenciesConfiguration);
+    };
+    if (sourceProjectConfiguration) {
+      if (!includeNpmIgnoreFiles) {
```
D4N14L, line 722: Combine this with the outer if statement

D4N14L (changes requested):

libraries/package-extractor/src/PackageExtractor.ts
D4N14L: Better yet, just set the property to `sourceProjectConfiguration && !includeNpmIgnoreFiles`

libraries/package-extractor/src/PackageExtractor.ts
EscapeB: Nice idea, have changed to use semver version range.

libraries/package-extractor/src/PackageExtractor.ts
EscapeB: LGTM.
But as the Map structure has changed, it can't be easily convert from Array. So I add several lines to covert origin Array to Map.

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -666,6 +719,31 @@ export class PackageExtractor {
...
+      const dependenciesConfigurations: IExtractorDependencyConfiguration[] | undefined =
+        dependencyConfigurationsByName.get(packagesJson.name);
+      if (!dependenciesConfigurations) {
+        return false;
+      }
+      const matchedDependenciesConfigurations: IExtractorDependencyConfiguration[] =
+        dependenciesConfigurations.filter((d) => semver.satisfies(packagesJson.version, d.dependencyVersion));
+
+      return matchedDependenciesConfigurations.some((d) => isFileExcluded(filePath, d));
+    };
```
D4N14L, line 743: You can combine these into the `isFileExcluded` check by checking `isLocalProject` there instead of twice below.

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -629,33 +679,36 @@ export class PackageExtractor {
...
+    // Third part dependencies won't have project configurations
+    const isLocalProject: boolean = !!sourceProjectConfiguration;
+
+    // Base filter function
+    const isFileExcluded = (
+      filePath: string,
+      {
+        patternsToInclude,
+        patternsToExclude
+      }: Pick<IExtractorDependencyConfiguration, 'patternsToInclude' | 'patternsToExclude'>
```
D4N14L, line 701: Since we already know `isLocalProject` and have the package.json, we can just get the correct patterns within this function without needing to pass them in.

libraries/package-extractor/src/PackageExtractor.ts
D4N14L: Let's also rename it to `dependencyVersionRange` as well to make it clear that it accepts a semver range.

build-tests/package-extractor-test-03/package.json
```diff
@@ -5,5 +5,8 @@
   "private": true,
   "scripts": {
     "_phase:build": ""
+  },
+  "devDependencies": {
+    "@types/node": "ts3.9"
```
D4N14L, line 10: Curious... I've never seen this version specifier format. How does this work? Why is this needed?

D4N14L (changes requested):
Looking good! Just a few more changes then I think it's good to merge.

build-tests/package-extractor-test-03/package.json
EscapeB: I saw them in `common/config/rush/common-versions.json`, and the comment says these are versions are used by testing projects.

libraries/package-extractor/src/PackageExtractor.ts
EscapeB: Done

EscapeB:
> Looking good! Just a few more changes then I think it's good to merge.

Hi @D4N14L, I think all the changes mentioned in the comments have now been made, please help merge this PR when you have time.

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -74,8 +75,14 @@ export interface IExtractorMetadataJson {
 
 interface IExtractorState {
   foldersToCopy: Set<string>;
+  packageJsonByPath: Map<string, IPackageJson>;
   projectConfigurationsByPath: Map<string, IExtractorProjectConfiguration>;
   projectConfigurationsByName: Map<string, IExtractorProjectConfiguration>;
+  /**
+   * As there might be different versions of same dependency, we need take version field into consideration
+   * The key format is `${name}@${version}`
+   */
```
D4N14L, line 84:
```suggestion
```

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -629,43 +684,78 @@ export class PackageExtractor {
...
+    // Function to filter files inside local project or third party dependencies.
+    const isFileExcluded = (filePath: string): boolean => {
+      // Encapsulate exclude logic into a function, so it can be reused.
+      const excludeFileByPatterns = (
+        filePath: string,
+        option: Pick<IExtractorDependencyConfiguration, 'patternsToInclude' | 'patternsToExclude'>
+      ): boolean => {
+        let includeFilters: IMinimatch[] | undefined;
+        let excludeFilters: IMinimatch[] | undefined;
+        const { patternsToExclude, patternsToInclude } = option;
```
D4N14L, line 709:
```suggestion
        filePath: string,
        patternsToInclude: string[] | undefined,
        patternsToExclude: string[] | undefined
      ): boolean => {
        let includeFilters: IMinimatch[] | undefined;
        let excludeFilters: IMinimatch[] | undefined;
```

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -629,43 +684,78 @@ export class PackageExtractor {
...
+        // and there is no exclude filter, then we know that the file is not excluded. If it is included
+        // and there is an exclude filter, then we need to check for a match.
+        return !isIncluded || !!excludeFilters?.some((m) => m.match(filePath));
+      };
+
+      if (isLocalProject) {
+        return excludeFileByPatterns(filePath, {
+          patternsToExclude: sourceProjectConfiguration?.patternsToExclude,
+          patternsToInclude: sourceProjectConfiguration?.patternsToInclude
+        });
```
D4N14L, line 733:
```suggestion
        return excludeFileByPatterns(
          filePath,
          sourceProjectConfiguration?.patternsToInclude
          sourceProjectConfiguration?.patternsToExclude
        );
```

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -629,43 +684,78 @@ export class PackageExtractor {
...
+        const matchedDependenciesConfigurations: IExtractorDependencyConfiguration[] =
+          dependenciesConfigurations.filter((d) =>
+            semver.satisfies(packagesJson.version, d.dependencyVersionRange)
+          );
+        return matchedDependenciesConfigurations.some((d) =>
+          excludeFileByPatterns(filePath, {
+            patternsToExclude: d.patternsToExclude,
+            patternsToInclude: d.patternsToInclude
+          })
+        );
```
D4N14L, line 752:
```suggestion
        return matchedDependenciesConfigurations.some(
          (d) => excludeFileByPatterns(filePath, d.patternsToInclude, d.patternsToExclude)
        );
```

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -739,18 +832,21 @@ export class PackageExtractor {
...
+            if (ignoreFilter.ignores(relativeSourcePath)) {
+              callback();
+              return;
+            }
+            // Filter out files that are excluded by the project configuration or dependency configuration.
+            if (isFileExcluded(relativeSourcePath)) {
               callback();
               return;
             }
+          }
```
D4N14L, line 845:
```suggestion
          if (
            relativeSourcePath !== '' &&
            (ignoreFilter.ignores(relativeSourcePath) || isFileExcluded(relativeSourcePath))
          ) {
            callback();
            return;
          }
```

D4N14L (approved):

libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -629,43 +680,76 @@ export class PackageExtractor {
...
+        // If the file is not included, then we don't need to check the excludeFilter. If it is included
+        // and there is no exclude filter, then we know that the file is not excluded. If it is included
+        // and there is an exclude filter, then we need to check for a match.
+        return !isIncluded || !!excludeFilters?.some((m) => m.match(filePath));
+      };
+
+      if (isLocalProject) {
+        return excludeFileByPatterns(
+          filePath,
+          sourceProjectConfiguration?.patternsToInclude
```
D4N14L, line 728:
```suggestion
          sourceProjectConfiguration?.patternsToInclude,
```

PR #4257, opened 2023-07-30
[heft-jest-plugin] Make `jest-environment-jsdom` and `jest-environment-node` optional peerDependencies.

iclanton:
`heft-jest-plugin` currently declares `testEnvironment` properties (`jest-environment-jsdom` for Web and `jest-environment-node` for Node), but isn't able to resolve those packages because they're devDependencies, meaning that a consumer extending that built-in Jest configs needs to declare its own dependency on one of those and re-declare a `testEnvironment` property.

This PR adds an optional peerDependency on those two packages, allowing a consumer (either a project or a rig) to just declare a dependency on the `jest-environment-` package.

### DIFF
--- a/common/changes/@rushstack/heft-jest-plugin/fix-jest-environment-dep_2023-07-30-03-41.json
+++ b/common/changes/@rushstack/heft-jest-plugin/fix-jest-environment-dep_2023-07-30-03-41.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-jest-plugin",
+      "comment": "Make `jest-environment-jsdom` and `jest-environment-node` peerDependencies.",
+      "type": "minor"
+    }
+  ],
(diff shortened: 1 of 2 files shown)


D4N14L (approved):

common/changes/@rushstack/heft-jest-plugin/fix-jest-environment-dep_2023-07-30-03-41.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-jest-plugin",
+      "comment": "Make `jest-environment-jsdom` and `jest-environment-node` peerDependencies.",
```
iclanton, line 5:
```suggestion
      "comment": "Make `jest-environment-jsdom` and `jest-environment-node` optional peerDependencies.",
```

PR #4262, opened 2023-08-01
[heft] Expose a _metricsCollector property on the sessions to allow plugins to instrument performance.

iclanton:
## Summary

This property was already there, but it wasn't exposed on the session interfaces.

## How it was tested

N/A

## Impacted documentation

API docs will probably need to be updated.

### DIFF
--- a/apps/heft/src/metrics/MetricsCollector.ts
+++ b/apps/heft/src/metrics/MetricsCollector.ts
@@ -80,21 +80,17 @@ export interface IPerformanceData {
 }
 
 /**
- * @internal
  * A simple performance metrics collector. A plugin is required to pipe data anywhere.
+ *
+ * @internal
  */
-export class MetricsCollector {
-  public readonly recordMetricsHook: AsyncParallelHook<IHeftRecordMetricsHookOptions> =
-    new AsyncParallelHook<IHeftRecordMetricsHookOptions>(['recordMetricsHookOptions']);
-
-  private _startTimeMs: number | undefined;
+export interface IMetricsCollector {
+  readonly recordMetricsHook: AsyncParallelHook<IHeftRecordMetricsHookOptions>;
 
   /**
    * Start metrics log timer.
    */
-  public setStartTime(): void {
-    this._startTimeMs = performance.now();
-  }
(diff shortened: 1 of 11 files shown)


apps/heft/src/metrics/MetricsCollector.ts
```diff
@@ -80,28 +80,50 @@ export interface IPerformanceData {
...
+   * Start metrics log timer.
+   */
+  setStartTime(): void;
+
+  /**
+   * Record metrics to the installed plugin(s).
+   *
+   * @param command - Describe the user command, e.g. `start` or `build`
+   * @param parameterMap - Optional map of parameters to their values
+   * @param performanceData - Optional performance data
```
dmichon-msft, line 100: These don't match the arguments below.

apps/heft/src/pluginFramework/HeftLifecycleSession.ts
```diff
@@ -162,12 +173,12 @@ export class HeftLifecycleSession implements IHeftLifecycleSession {
   /**
    * @internal
    */
-  public readonly metricsCollector: MetricsCollector;
+  public readonly _metricsCollector: MetricsCollector;
```
dmichon-msft, line 176: The mixing of the meaning of the `_` prefix is really confusing, since the common case is it indicates "private property", yet sometimes it mean "public, but only intended for use by some consumers", which are very different things.

dmichon-msft (commented):
What's the use case for this? I'm having trouble coming up with a scenario that wouldn't be better served by allowing lifecycle plugins to listen to task start/finish and phase start/finish events.

apps/heft/src/pluginFramework/HeftLifecycleSession.ts
iclanton: We've historically used these interchangeably.

iclanton:
@dmichon-msft - This is intended for feature parity with Heft legacy. That does sound like a better experience, though. We'll just need to make sure plugins can augment the metrics with additional data.

dmichon-msft (approved):

PR #4264, opened 2023-08-03
[various] Switch from glob to fast-glob

dmichon-msft:
## Summary
Switches from `glob` to `fast-glob` for globbing. The latter ships built-in typings, has a promise-based API, and leverages the `withFileTypes: true` option of `readdir` for better performance.

## Details
Affects 3 packages.

## How it was tested
Full build and test.

## Impacted documentation
None.


libraries/typings-generator/src/TypingsGenerator.ts
```diff
@@ -134,11 +133,10 @@ export class TypingsGenerator {
     let checkFilePaths: boolean = true;
     if (!relativeFilePaths?.length) {
       checkFilePaths = false; // Don't check file paths if we generate them
-      relativeFilePaths = await LegacyAdapters.convertCallbackToPromise(glob, this.inputFileGlob, {
+      relativeFilePaths = await glob(this.inputFileGlob, {
         cwd: this.sourceFolderPath,
-        ignore: this.ignoredFileGlobs,
-        nosort: true,
-        nodir: true
+        ignore: this.ignoredFileGlobs as string[],
```
D4N14L, line 138: Hmm, should we allocate a new array for this? It's unlikely it'll get changed so this is not really a concern of mine, but more asking for my own knowledge of what the best design here would be.

webpack/hashed-folder-copy-plugin/src/HashedFolderCopyPlugin.ts
```diff
@@ -267,21 +267,19 @@ export class HashedFolderCopyPlugin implements webpack.Plugin {
...
+          compilation.errors.push(new Error(errorMessage));
+          return this._renderError(errorMessage);
+        }
 
-          if (assetsToAdd.has(globResult)) {
-            const errorMessage: string = `Two files resolve to the same output path "${globResult}"`;
-            compilation.errors.push(new Error(errorMessage));
-            return this._renderError(errorMessage);
-          }
+        const globResultFullPath: string = path.resolve(resolvedGlobsBase, globResult);
```
D4N14L, line 278: Should we have the globber return absolute paths so we don't have to do this all the time?

libraries/rush-lib/src/logic/buildCache/getHashesForGlobsAsync.ts
```diff
@@ -13,16 +13,16 @@ async function expandGlobPatternsAsync(
 ): Promise<string[]> {
   const allMatches: Set<string> = new Set<string>();
 
-  const { default: glob } = await import('glob');
-  const globAsync = (pattern: string, options: IOptions = {}): Promise<string[]> => {
-    return LegacyAdapters.convertCallbackToPromise(glob, pattern, options);
+  const { default: glob } = await import('fast-glob');
+  const globAsync = (pattern: string, options: Options = {}): Promise<string[]> => {
```
D4N14L, line 17: Don't think this is needed anymore IMO

D4N14L (approved):

libraries/typings-generator/src/TypingsGenerator.ts
dmichon-msft: This is really just a case of poor typings. I expect `ReadonlyArray<string>` is probably the correct input type.

webpack/hashed-folder-copy-plugin/src/HashedFolderCopyPlugin.ts
dmichon-msft: The relative path is used separately in this function, so in this scenario it is best to defer the resolving.

libraries/rush-lib/src/logic/buildCache/getHashesForGlobsAsync.ts
dmichon-msft: I want to kill this function but haven't yet. That'll happen when all cache hashes get computed in advance of execution.

libraries/rush-lib/src/logic/buildCache/getHashesForGlobsAsync.ts
```diff
@@ -1,55 +1,40 @@
...
+  });
+
+  if (matches.length === 0) {
     throw new Error(
       `Couldn't find any files matching provided glob patterns: ["${Array.from(globPatterns).join('", "')}"].`
     );
   }
 
-  return Array.from(allMatches);
+  return Array.from(matches);
```
D4N14L, line 37: Shouldn't need to allocate a new array for this

D4N14L (approved):

libraries/typings-generator/src/TypingsGenerator.ts
iclanton: Should we then change the signature on our side?

D4N14L (approved):

PR #4266, opened 2023-08-03
[typings-generator] Support custom readFile

dmichon-msft:
## Summary
Adds the ability for consumers of `TypingsGenerator` to customize how files are read. Possible use cases include:
 - Skipping reading files altogether, such as when generating .d.ts files that don't depend on the file content
 - Reading from somewhere other than the file system
 - Reading files as `Buffer` rather than `string`, or with alternate encoding

## Details
Figuring out the options interfaces for backwards compatibility was the toughest part of making this work. The `readFile` property needs to be optional if its return type is `string`, and required otherwise.

## How it was tested
Local build.
Lab build.
Manual validation that the `readFile` property is required if the type of `fileContents` is not `string`.

## Impacted documentation
Docs for TypingsGenerator.

### DIFF
--- a/libraries/typings-generator/src/TypingsGenerator.ts
+++ b/libraries/typings-generator/src/TypingsGenerator.ts
@@ -51,9 +51,10 @@ export interface ITypingsGeneratorOptionsWithoutReadFile<
 /**
  * @public
  */
-export interface IReadFile<TFileContents = string> {
-  (filePath: string, relativePath: string): Promise<TFileContents> | TFileContents;
-}
+export type ReadFile<TFileContents = string> = (
+  filePath: string,
+  relativePath: string
+) => Promise<TFileContents> | TFileContents;
 
 /**
  * @public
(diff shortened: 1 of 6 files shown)


libraries/typings-generator/src/TypingsGenerator.ts
```diff
@@ -46,12 +48,41 @@ export interface ITypingsGeneratorOptions<TTypingsResult = string | undefined>
   filesToIgnore?: string[];
 }
 
+/**
+ * @public
+ */
+export interface IReadFile<TFileContents = string> {
+  (filePath: string, relativePath: string): Promise<TFileContents> | TFileContents;
+}
```
iclanton, line 56:
```suggestion
export type ReadFile<TFileContents = string> = (filePath: string, relativePath: string): Promise<TFileContents> | TFileContents;
```

iclanton: I think of types that are defined as interfaces (and start with `I`) as objects.

iclanton (approved):

PR #4271, opened 2023-08-06
[rush] Selector Expressions ("json:" parser)

elliot-nelson:
## Summary

 - Create JSON-based selector expressions, allowing for complex selection of projects.
 - Implement an API-level Project Selector, which can process these expressions.
 - Add a new selector parser, `json:`, which loads a JSON file containing an expression and selects it.

## Details

This PR is sliced off of the prototype in https://github.com/microsoft/rushstack/pull/4250, based on discussion from August Rush Hour West.

Some key highlights/changes from the Rush Hour demo:

 - All "string-based" expressions and parsers are removed, so reviewers can focus on the JSON syntax.
 - The JSON syntax has been significantly reworked. Previously, it was kind of an "internal" syntax that was frustrating for humans to write and understand; now it has been modified to be human-focused first, with ease of crafting and understanding expressions prioritized.
 - "Filters" have been renamed back to "Parameters" to map better to existing CLI concepts.

### Selector Expressions

This table represents a quick, at-a-glance introduction to selector expressions, which can be saved in a JSON file or used directly in API calls from a node script:

Description | Example
--- | ---
Simple named project | `"@microsoft/rush"`
Simple scoped selector | `"tag:app"`
Detailed selector | `{ "scope": "tag", value: "app" }`
Parameter | `{ "--to": "@microsoft/rush" }`
Operator | `{ "and": ["git:origin/main", "tag:app"] }`
Complex | `{ "and": [{ "--to": "@microsoft/rush" }, { "not": "tag:sdk" }] }`

All of the above are examples of valid _expressions_, which means any of them can be inserted into any of the places above where expressions are valid.

## How it was tested

 - Some simple unit tests
 - Testing on CLI

## Impacted documentation

 - Document new selector expressions


libraries/rush-lib/src/api/RushProjectSelector.ts
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * A central interface for selecting a subset of Rush projects from a given monorepo,
+ * using standardized selector expressions. Note that the types of selectors available
+ * in a monorepo may be influenced in the future by plugins, so project selection
+ * is always done in the context of a particular Rush configuration.
+ */
+export class RushProjectSelector {
+  private _rushConfig: RushConfiguration;
+  private _scopes: Map<string, ISelectorParser<RushConfigurationProject>> = new Map();
+  private _options: IProjectSelectionOptions;
```
dmichon-msft, line 53:
```suggestion
  private readonly _rushConfig: RushConfiguration;
  private readonly _scopes: Map<string, ISelectorParser<RushConfigurationProject>> = new Map();
  private readonly _options: IProjectSelectionOptions;
```

libraries/rush-lib/src/api/RushProjectSelector.ts
```diff
@@ -0,0 +1,183 @@
...
+    this._options = options;
+
+    this._scopes.set('name', new NamedProjectSelectorParser(this._rushConfig));
+    this._scopes.set(
+      'git',
+      new GitChangedProjectSelectorParser(this._rushConfig, this._options.gitSelectorParserOptions)
+    );
+    this._scopes.set('tag', new TagProjectSelectorParser(this._rushConfig));
+    this._scopes.set('version-policy', new VersionPolicyProjectSelectorParser(this._rushConfig));
+    this._scopes.set('json', new JsonFileSelectorParser(this._rushConfig, this));
```
dmichon-msft, line 66:
```suggestion
    const scopes: Map<string, ISelectorParser<RushConfigurationProject>> = new Map();
    this._scopes = scopes;
    scopes.set('name', new NamedProjectSelectorParser(rushConfig));
    scopes.set(
      'git',
      new GitChangedProjectSelectorParser(rushConfig, options.gitSelectorParserOptions)
    );
    scopes.set('tag', new TagProjectSelectorParser(rushConfig));
    scopes.set('version-policy', new VersionPolicyProjectSelectorParser(rushConfig));
    scopes.set('json', new JsonFileSelectorParser(rushConfig, this));
```
Reduction in verbosity, removes unnecessary property accesses, helps the code compress better.

libraries/rush-lib/src/api/RushProjectSelector.ts
```diff
@@ -0,0 +1,183 @@
...
+   * an error in user input. The default string "expression" is used if no context is provided.
+   */
+  public async selectExpression(
+    expr: SelectorExpression,
+    context: string = 'expression'
+  ): Promise<RushConfigurationProject[]> {
+    if (isAnd(expr)) {
+      return this._evaluateAnd(expr, context);
+    } else if (isOr(expr)) {
+      return this._evaluateOr(expr, context);
```
dmichon-msft, line 82: Considering that this operates in set space, `Union` and `Intersect` are probably clearer terms to use.

libraries/rush-lib/src/api/RushProjectSelector.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  private async _evaluateAnd(
+    expr: IExpressionOperatorAnd,
+    context: string
+  ): Promise<RushConfigurationProject[]> {
+    const result: Array<RushConfigurationProject>[] = [];
+    for (const operand of expr.and) {
+      result.push(await this.selectExpression(operand, context));
+    }
+    return [...Selection.intersection(new Set(result[0]), ...result.slice(1).map((x) => new Set(x)))];
```
dmichon-msft, line 106: Leave this as a `Set<RushConfigurationProject>`. In general we should prefer `Set<RushConfigurationProject>` to `RushConfigurationProject[]`.

libraries/rush-lib/src/api/IRushProjectSelector.ts
```diff
@@ -0,0 +1,15 @@
...
+import { SelectorExpression } from './SelectorExpressions';
+
+/**
+ * This interface allows a previously constructed RushProjectSelector to be passed around
+ * and used by other lower-level objects. (For example, the "json:" selector reads and
+ * parses an entire new selector expression, which might in turn load another selector
+ * expression and parse it.)
+ */
+export interface IRushProjectSelector {
+  selectExpression(expr: SelectorExpression, context: string): Promise<RushConfigurationProject[]>;
```
dmichon-msft, line 14:
```suggestion
  selectExpression(expr: SelectorExpression, context: string): Promise<ReadonlySet<RushConfigurationProject>>;
```
Sets are the semantically correct type for selectors to return, since duplicates are not permitted. Using `ReadonlySet` allows for caching internally.

libraries/rush-lib/src/logic/selectors/JsonFileSelectorParser.ts
```diff
@@ -0,0 +1,59 @@
...
+      if (!expr) {
+        throw new SelectorError(`Unable to find JSON file at "${absolutePath}" in ${context}.`);
+      }
+    }
+
+    return this._projectSelector.selectExpression(expr!, `JSON file ${unscopedSelector} in ${context}`);
+  }
+
+  public getCompletions(): Iterable<string> {
+    return this._rushConfiguration.projectsByTag.keys();
```
dmichon-msft, line 49: This should probably return `[]` unless you are predefining a set of known JSON files somewhere.

libraries/rush-lib/src/api/SelectorExpressions.ts
```diff
@@ -0,0 +1,59 @@
...
+  scope: string;
+  value: string;
+
+  // Reserved for future use
+  filters?: Record<string, string>;
+}
+
+export type ExpressionParameter = {
+  [K in `--${string}`]: string;
+};
```
dmichon-msft, line 22:
```suggestion
export type ExpressionParameter = Record<`--${string}`, string>;
```

libraries/rush-lib/src/api/RushProjectSelector.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+    return [...Selection.union(new Set(result[0]), ...result.slice(1).map((x) => new Set(x)))];
+  }
+
+  private async _evaluateNot(
+    expr: IExpressionOperatorNot,
+    context: string
+  ): Promise<RushConfigurationProject[]> {
+    const result: RushConfigurationProject[] = await this.selectExpression(expr.not, context);
+    return this._rushConfig.projects.filter((p) => !result.includes(p));
```
dmichon-msft, line 125: Add a `subtract` method to the `Selection` namespace (that will return a Set)

libraries/rush-lib/src/api/RushProjectSelector.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private async _evaluateParameter(
+    expr: ExpressionParameter,
+    context: string
+  ): Promise<RushConfigurationProject[]> {
+    const key: string = Object.keys(expr)[0];
+
+    if (key === '--to') {
```
dmichon-msft, line 134: Shouldn't we still include `--impacted-by`, `--impacted-by-except`, and `--to-except`?

libraries/rush-lib/src/api/RushProjectSelector.ts
```diff
@@ -0,0 +1,183 @@
...
+  ): Promise<RushConfigurationProject[]> {
+    const result: RushConfigurationProject[] = await this.selectExpression(expr.not, context);
+    return this._rushConfig.projects.filter((p) => !result.includes(p));
+  }
+
+  private async _evaluateParameter(
+    expr: ExpressionParameter,
+    context: string
+  ): Promise<RushConfigurationProject[]> {
+    const key: string = Object.keys(expr)[0];
```
dmichon-msft, line 132:
```suggestion
    const [key, value] = Object.entries(expr)[0];
```

dmichon-msft (commented):
Mostly looks good, but it'd be better to leave all the selections as `Set<RushConfigurationProject>` (or `ReadonlySet<RushConfigurationProject>`).

elliot-nelson:
@dmichon-msft Thanks for the review! I've taken care of all feedback except the suggestion on operator names. I wanted to loop @octogonz in here as well before I dive into changing them everywhere.

My suggestion is "logical" (query) operators. In infix form, `(a AND b) OR (c AND NOT d)`.

David's recommendation is "set" operators, so, `(a INTERSECT b) UNION (c SUBTRACT d)`.

Is that everyone's preference? I guess I can't argue that it's explicit.

My original reasoning was that AND and OR are quite clear in SQL and JQL, so people should be familiar with them. I think the advantage those contexts have is that the expression looks like this: `name = A and name = B` (clearly both can't be true), whereas `A and B` is a little more ambiguous.

elliot-nelson:
I have updated the types and json schema to use set operators (UNION, INTERSECT, SUBTRACT).

A complex expression might now be expressed like this (`example.json`):

```json
{
  "intersect": [
    "--impacted-by": {
      "intersect": [
        { "--impacted-by": "git:origin/main" },
        "tag:terraform"
      ]
    },
    "tag:app"
  ]
}
```

In English, this might be "the set of application projects that could be impacted by projects with terraform components that were impacted by this pull request".

Then, this set of application projects can be used on the command line, like so:

```console
rush install

# A "safe" build up to and including the selected projects
rush build --to json:example.json

# An "unsafe" custom operation, designed to be used this way
rush deploy-previews --only json:example.json
```

libraries/rush-lib/src/schemas/selector-expression.schema.json
```diff
@@ -0,0 +1,97 @@
...
+    },
+    "simpleSelector": {
+      "type": "string"
+    },
+    "detailedSelector": {
+      "type": "object",
+      "additionalProperties": false,
+      "requiredProperties": ["scope", "value"],
+      "properties": {
+        "scope": {
```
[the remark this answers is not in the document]
elliot-nelson: In https://rushjs.io/pages/developer/selecting_subsets/, we say that in "--to X", the "X" is a _selector_, and that there are different _types_ of selectors.

We then describe a `tag:` selector, a `git:` selector, etc.

Perhaps words like "scope" and "provider" are just wishful thinking, and ways to avoid "selectorType" or simply `"selector": "git"`?

libraries/rush-lib/src/logic/selectors/ISelectorParser.ts
```diff
@@ -4,9 +4,27 @@
...
+   * can interact with other Rush subsystems (which often require a terminal instance).
+   *
+   * If your selector produces an error, instead of writing it to terminal, consider throwing
+   * a SelectorError, so your select is usable both within Rush and in user scripts.
+   */
   terminal: ITerminal;
-  parameterName: string;
+
+  /**
+   * A short, human-readable description of the context where this selector was encountered.
```
octogonz-proxy, line 23: Is `context` always a noun phrase?

libraries/rush-lib/src/api/SelectorExpressions.ts
```diff
@@ -0,0 +1,61 @@
...
+ * A "Selector Expression" is a JSON description of a complex selection of
+ * projects, using concepts familiar to users of the Rush CLI. This type represents
+ * the type-safe version of these JSON objects.
+ */
+export type SelectorExpression = ExpressionSelector | ExpressionParameter | ExpressionOperator;
+
+export type ExpressionSelector = string | IExpressionDetailedSelector;
+
+export interface IExpressionDetailedSelector {
+  scope: string;
```
octogonz-proxy, line 14: What is `scope`? Give example strings

libraries/rush-lib/src/api/SelectorExpressions.ts
```diff
@@ -0,0 +1,61 @@
...
+  value: string;
```
octogonz-proxy, line 15: What is `value`? Give example strings

libraries/rush-lib/src/api/test/RushProjectSelector.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+function createProjectSelector(): RushProjectSelector {
+  const rushJsonFile: string = `${__dirname}/repo/rush-pnpm.json`;
+  const rushConfig: RushConfiguration = RushConfiguration.loadFromConfigurationFile(rushJsonFile);
+  const projectSelector: RushProjectSelector = new RushProjectSelector(rushConfig, {
+    gitSelectorParserOptions: {
+      includeExternalDependencies: true,
+      enableFiltering: false
+    }
+  });
```
octogonz-proxy, line 16: Screenshot 2023-09-13 at 11 22 56 PM

octogonz-proxy: What is a "selector"? Looking at these docs, a "selector" is an engine that chooses a list of Rush projects, kind of like how a CSS selector chooses DOM elements.

You have created a class called `RushProjectSelector` whose state is a **rush.json** configuration plus some options controlling how the Git selector behaves.  If we use SQL terminology, this class is like a SQL table plus some global options that affect query behavior.  It has the ability to execute queries, but the class is not a query -- instead the actual query is passed into `projectSelector.selectExpression()`.

Is a "selector" that query? Well, the query's type is `SelectorExpression` (not to be confused with `ExpressionSelector`, let's ignore that for now).

What is a "selector expression"? Well apparently it can be a selector:

```ts
const projects: ReadonlySet<RushConfigurationProject> = await projectSelector.selectExpression({
  '--to': 'project1'
});
```

Is it then the case that `--to X` and `X` are both selectors? Is everything a "selector"?  Is a "selector" just any AST node for this query language?

libraries/rush-lib/src/api/SelectorExpressions.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface IExpressionOperatorUnion {
+  union: SelectorExpression[];
+}
+
+export interface IExpressionOperatorIntersect {
+  intersect: SelectorExpression[];
+}
+
+export interface IExpressionOperatorSubtract {
+  subtract: SelectorExpression[];
```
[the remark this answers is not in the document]
elliot-nelson: 🤔 I guess the only pushback I have is where to enforce such a restriction.

I could attempt to use a `FixedLengthArray` type for this everywhere, and enforce a min/max length in the JSON schema, and then throw Error in the function if it's not exactly 2 elements, but it seems like a lot of extra work to _prevent_ the user from providing multiple elements.

Whereas if you allow it, and all your examples only show 2 elements, I think most people will typically just subtract two elements.

(For what it's worth, your first result seems like the clear answer, it is the 1st operand minus all the other operands.)

libraries/rush-lib/src/schemas/selector-expression.schema.json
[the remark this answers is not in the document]
octogonz-proxy: @elliot-nelson this terminology arises from the grammar.  Your original prototype looked like this:
```
rush build --select 'to(tag:sdk and git:release/v3.0.0) and example-project'
```
...with the idea that Rush plugins might introduce new notations like `approver:team1`.

In this grammar, operators like `to()` and `and` use a simple standard tokenizer.  Whereas `release/v3.0.0` and `sdk` have arbitrary notations. When designing them, we need to be very careful that the syntax does not conflict with the grammar.  For example, `release/v3.0.0)` is a perfectly legal Git branch name, but is inexpressible in your notation, because the `)` will be misinterpreted as part of the `to()` function. (Yes yes, we can handle that by introducing escapes, or by banning unpleasant branch name choices, but awareness of this problem is my point.)

Beyond this parsing annoying, `release/v3.0.0` is also a learning curve for end users: Imagine a person doesn't know what Git is. They might wonder whether `/` is part of Rush's query language, or if it is part of the `git:` operator. In order to know, they need to learn the specialized syntax for Git branch names.

So, if you look at the docs, the word "selector" was really introduced to mean "AST leaf nodes with their own weird syntaxes you have to learn." The island grammars waiting to break our parser if we aren't careful.

`--to` selects a set of projects, but it is not a "selector" because `--to` has standard well-behaved syntax.  It is a grammatically normal friendly selection "operator" or "function", being represented as as a CLI "parameter".

libraries/rush-lib/src/api/test/RushProjectSelector.test.ts
octogonz-proxy: Constructive suggestions:

1. Per my [other comment](https://github.com/microsoft/rushstack/pull/4271/files#r1324763400), let's return to the idea that a **"selector"** is a leaf node like `git:origin/main` or `tag:release`.
2. An expression like `to(tag:release and git:origin/main)` should get some other name, like **selection expression**.  `to()` itself is really a **"selection function"** but its JSON notation is an expression involving that function. 


`--to` is the CLI parameter representation of `to()`. I feel like the JSON design went off the rails a bit by putting `--to` in a JSON key.  JSON keys are JavaScript variable names. We don't write `let --to: string;`.  Of course we do want the JSON syntax to be concise to write, but it also needs to be an easy API to use.  The API scenarios are things like:
- a program that generates a selection expression 
- a program that optimizes a selection expression by reducing it to a more efficient form

Such programs aren't very easy to write if we can't even do:

```ts
switch (node.kind) {
  case 'union': . . .
  case 'intersect': . . .
  . . .
}
```

and instead must do:

```ts
    if (isUnion(expr)) {
      return this._evaluateUnion(expr, context);
    } else if (isIntersect(expr)) {
      return this._evaluateIntersect(expr, context);
    } else if (isSubtract(expr)) {
      return this._evaluateSubtract(expr, context);
    } else if (isParameter(expr)) {
      return this._evaluateParameter(expr, context);
    } else if (isDetailedSelector(expr)) {
```
3. So I recommend going back to a more conventional AST design.
4. Rename `RushProjectSelector.selectExpression` to something like `ProjectQueryContext.select()`

libraries/rush-lib/src/api/SelectorExpressions.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export interface IExpressionDetailedSelector {
+  scope: string;
+  value: string;
+
+  // Reserved for future use
+  filters?: Record<string, string>;
+}
+
+export type ExpressionParameter = Record<`--${string}`, string>;
```
octogonz-proxy, line 21: Seems like this should be allowed:

Screenshot 2023-09-14 at 1 06 08 AM

PR #4272, opened 2023-08-07
fix(eslint-patch): fix for VS Code eslint extension with flat config enabled

jamescostian:
## Summary

Fixes a bug where VS Code's eslint extension + eslint-patch confuse each other

### Problem

If all of the following apply to you:
- You're using VS Code with the eslint extension from Microsoft
- One of your eslint packages uses eslint-patch
- You're using the new flat config eslint has
- You've followed the instructions for the VS Code eslint extension and set `"eslint.experimental.useFlatConfig": true` in your VS Code settings (required, otherwise you get an error)

Eslint still won't work inside VS Code. This MR fixes that.

### Root cause

1. VS Code's eslint extension [uses `eslint/use-at-your-own-risk`](https://github.com/microsoft/vscode-eslint/blob/release/2.4.2/server/src/eslint.ts#L872) when you turn on flat config
2. eslint's `use-at-your-own-risk` [maps to `lib/unsupported-api.js`](https://github.com/eslint/eslint/blob/v8.46.0/package.json#L13)
3. `lib/unsupported-api.js` [loads `lib/cli-engine/file-enumerator.js`](https://github.com/eslint/eslint/blob/v8.46.0/lib/unsupported-api.js#L14), not `lib/cli-engine/cli-engine.js`
4. eslint-patch prior to this commit specifically searched for `lib/cli-engine/cli-engine.js` in the import stack (see the diff) and if it wasn't found, it would assume an older version of eslint, and that would cause an error

### The fix

This commit allows for any file inside the eslint package to count when searching for an eslint package, allowing VS Code's eslint extension to work when you use eslint's flat config.

### Alternatives considered
I considered is to allow `lib/cli-engine/file-enumerator.js`, or `lib/cli-engine/*`, or `lib/unsupported-api.js` instead. However, I decided to consider what the original purpose of this code was. The comment right above the code said this:

> Make sure we actually resolved the module in our call path and not some other spurious dependency.

Based on that comment, it shouldn't matter which exact file in eslint is being used; one of them is being used, and that's all that matters. Eslint's implementation details shouldn't matter.

### Performance

Personally, I believe this performs well enough. However, a micro optimizer could make the case that it would be more efficient to check for `lib/cli-engine/file-enumerator.js` specifically. I'll make that change if requested, but personally I believe the right trade-off has been made in this MR as is.

## How it was tested

Looking at the links mentioned in the details, I think it's clear that this works. There was no algorithmic code here, and there are no unit tests for this library, so I simply patched the file locally, quit VS Code, and restarted it to see if it would work in my test repo.

## Impacted documentation

None.

### DIFF
--- a/eslint/eslint-patch/src/modern-module-resolution.ts
+++ b/eslint/eslint-patch/src/modern-module-resolution.ts
@@ -68,7 +68,7 @@ for (let currentModule = module; ; ) {
 
       // Make sure we actually resolved the module in our call path
       // and not some other spurious dependency.
-      if (path.join(eslintCandidateFolder, 'lib/cli-engine/cli-engine.js') === currentModule.filename) {
+      if (currentModule.filename.startsWith(eslintCandidateFolder + path.sep)) {
         eslintFolder = eslintCandidateFolder;
         break;
       }
--- a/common/changes/@rushstack/eslint-patch/eslint-flat-config-support_2023-08-07-03-59.json
+++ b/common/changes/@rushstack/eslint-patch/eslint-flat-config-support_2023-08-07-03-59.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/eslint-patch",
+      "comment": "Fix patching for running eslint via eslint/use-at-your-own-risk, which VS Code's eslint extension does when enabling flat config support",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/eslint-patch"
+}


jamescostian:

iclanton (approved):

eslint/eslint-patch/src/modern-module-resolution.ts
```diff
@@ -68,7 +68,7 @@ for (let currentModule = module; ; ) {
 
       // Make sure we actually resolved the module in our call path
       // and not some other spurious dependency.
-      if (path.join(eslintCandidateFolder, 'lib/cli-engine/cli-engine.js') === currentModule.filename) {
+      if (currentModule.filename.startsWith(eslintCandidateFolder)) {
```
iclanton, line 71: The only open question I have is whether `currentModule.filename` and `eslintCandidateFolder` are both alwahs fully resolved and have normalized/matching path separators, and that `eslintCandidateFolder` ends with a path separator.

Have you tested this both on POSIX and Windows? Can you confirm the terminating slash?

jamescostian: Hi, thanks for the prompt review!

1. `currentModule.filename` and `eslintCandidateFolder` are both always fully resolved and have matching path separators on Windows and POSIX - I've confirmed this by hand
2. `eslintCandidateFolder` *never* ends in a path separator, I've confirmed this by hand and by reading the docs and considering how it's set up. My bad! I've adjusted this MR to add `path.sep` at the end, and have tested that on both Windows and POSIX
3. All that said, this doesn't work in Windows. But it doesn't make any better or worse - it just had no effect at all, at least in my testing. There was already a bug in Windows. If that bug were taken care of, this code would behave properly.

<details>
<summary>Details about what went wrong with Windows</summary>

Here's how I tested in Windows:

1. Rented an OVH VPS with a fresh Windows Server 2022 installation
2. Installed Node.js 20 and VS Code on it
3. Installed the eslint plugin for VS Code
4. Added the same repo I've been testing locally, which uses the latest version of eslint + a flat config file + eslint-config-next imported via [FlatCompat](https://www.npmjs.com/package/@eslint/eslintrc#Usage). Note that this repo works fine for me on macOS with the same set up and versions of everything, as long as I put in the patch here

I noticed a crucial difference between macOS and windows. This code works by walking up the stack of module imports like `module.parent.parent.parent...` and looking for a module with a `.filename` that's within an eslint installation. The issue is, no such file can be found. Instead, by the time eslint-patch is imported, entire stack of module imports has been truncated, and nothing before import-fresh appears:

windows import stack does not include eslint

For comparison, here's the import stack on my mac:

macos import stack does include eslint

Note how it includes `node_modules/eslint/` (that bottom line actually is from the VS Code extension). I tried adding the same debugging to import-fresh, and that got me interesting results. Initially, it had the full stack. But later on, import-fresh got imported yet again, and when that happened and it logged its stack, it only showed itself and nothing before it:

Screenshot 2023-08-07 at 10 17 08 PM

 I tried replacing import-fresh with a simple function that just calls require, thinking maybe it was the culprit - that didn't work.

Anyway, I've seen that this MR doesn't hurt anything in Windows either - see my very first 2 points at the very top of this comment.
</details>

The current code, as it is in the main branch and as it is in my branch, simply would never work on the Windows set up I tried. More substantial effort is required to get it working, and I'm personally not invested enough to get sort through it.

But I believe this MR does contain a valuable fix for anyone running into the same issue as me - where the existing code in eslint-patch is almost perfect, but just needs a slight adjustment to deal with a new change in eslint.

iclanton: Sounds good! Thanks for testing.

We'll have to investigate the issue on Windows separately. Thanks for doing some initial investigation.

PR #4277, opened 2023-08-11
[rush] Fix an issue where changefiles would not be correctly discovered when running on Windows and add a Windows run.

iclanton:
## Summary

`rush-lib`'s unit tests are currently failing on Windows after https://github.com/microsoft/rushstack/pull/4264. This PR fixes the introduced issue and adds a Windows PR pipeline.

## How it was tested

Ran `heft test` in `rush-lib` on Windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,17 @@ on:
   workflow_dispatch:
 jobs:
   build:
-    name: Node.js v${{ matrix.NodeVersion }}
-    runs-on: ubuntu-latest
     strategy:
       matrix:
-        NodeVersion: [14, 16]
+        include:
+          - NodeVersion: 14
+            OS: ubuntu-latest
+          - NodeVersion: 16
+            OS: ubuntu-latest
+          - NodeVersion: 16
+            OS: windows-latest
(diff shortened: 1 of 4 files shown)


octogonz (approved):

D4N14L (approved):

.github/workflows/ci.yml
```diff
@@ -8,11 +8,17 @@ on:
...
     strategy:
       matrix:
-        NodeVersion: [14, 16]
+        include:
+          - NodeVersion: 14
+            OS: ubuntu-latest
+          - NodeVersion: 16
+            OS: ubuntu-latest
+          - NodeVersion: 16
+            OS: windows-latest
```
octogonz, line 19: If we're already doing Node 16 on Windows, do we need Node 16 on Linux?

octogonz: @D4N14L suggested to build both 14 and 16 on Windows. 

I think we also need to get a Node 18 pipeline enabled, even if we have to skip some Webpack projects.  I will address this in a separate PR

Faithfinder: You might want to do node 20 already. It becomes the "Active" version fairly soon
image

PR #4280, opened 2023-08-15
[rush] Support splitting `rush update` into "update lockfile" and "install from lockfile"

dmichon-msft:
## Summary
Adds a new experiment flag "usePnpmLockfileOnlyThenFrozenLockfileForRushUpdate" that causes two separate invocations of `pnpm install` when running `rush update`: the first with `--lockfile-only`, and the second with `--frozen-lockfile`. This is intended to support repositories that leverage the `afterAllResolved` hook in `.pnpmfile.cjs` to alter the shape of the lockfile following dependency resolution.

## Details
Addresses an issue wherein rewriting the lockfile during `afterAllResolved` would result in broken symbolic links.

## How it was tested
Running `rush update` against the local version and validating that `pnpm install` was invoked twice, the first time with `--lockfile-only --prefer-frozen-lockfile` and the second with `--frozen-lockfile`.

## Impacted documentation
Only the experiment flag. Or guidance for when using the `afterAllResolved` hook.

### DIFF
--- a/common/changes/@microsoft/rush/shrinkwrap-only_2023-08-15-01-41.json
+++ b/common/changes/@microsoft/rush/shrinkwrap-only_2023-08-15-01-41.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Add experiment \"usePnpmLockfileOnlyThenFrozenLockfileForRushUpdate\" that, when running `rush update`, performs first a `--lockfile-only` update to the lockfile, then a `--frozen-lockfile` installation.",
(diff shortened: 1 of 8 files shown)


common/changes/@microsoft/rush/shrinkwrap-only_2023-08-15-01-41.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Add experiment \"usePnpmLockfileOnlyThenFrozenLockfileForRushUpdate\" that, when running `rush update`, performs first a `--lockfile-only` update to the lockfile, then a `--frozen-lockfile` installation.",
```
iclanton, line 5:
```suggestion
      "comment": "Add experiment \"usePnpmLockfileOnlyThenFrozenLockfileForRushUpdate\" that, when running `rush update`, performs first a `--lockfile-only` update to the lockfile, then a `--frozen-lockfile` installation. This mitigates issues that may arise when using the `afterAllResolved` hook in `.pnpmfile.cjs`.",
```

iclanton (approved):

PR #4296, opened 2023-08-25
[rush] Surface dependencySettings on the deploy.json configuration

EscapeB:
## Summary
Add dependencySettings field to Rush deploy.json configurations. This will allow developers to customize how third party dependencies are processed when running `rush deploy` 
Related PR https://github.com/microsoft/rushstack/pull/4255 cc @D4N14L 

## How it was tested
Tested locally

## Impacted documentation
Schema for deploy.json has been updated and need to be updated.

### DIFF
--- a/libraries/rush-lib/src/schemas/deploy-scenario.schema.json
+++ b/libraries/rush-lib/src/schemas/deploy-scenario.schema.json
@@ -111,14 +111,14 @@
             "type": "string"
           },
           "patternsToInclude": {
-            "description": "A list of glob patterns to include when extracting specific dependency. If a path is matched by both \"patternsToInclude\" and \"patternsToExclude\", the path will be excluded. If undefined, all paths will be included.",
+            "description": "A list of glob patterns to include when extracting the dependency specified in this object. If a path is matched by both \"patternsToInclude\" and \"patternsToExclude\", the path will be excluded. If undefined, all paths will be included.",
             "type": "array",
             "items": {
               "type": "string"
             }
           },
           "patternsToExclude": {
-            "description": "A list of glob patterns to exclude when extracting specific dependency. If a path is matched by both \"patternsToInclude\" and \"patternsToExclude\", the path will be excluded. If undefined, no paths will be excluded.",
+            "description": "A list of glob patterns to include when extracting the dependency specified in this object. If a path is matched by both \"patternsToInclude\" and \"patternsToExclude\", the path will be excluded. If undefined, no paths will be excluded.",
             "type": "array",
             "items": {
               "type": "string"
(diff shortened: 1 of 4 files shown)


libraries/rush-lib/src/schemas/deploy-scenario.schema.json
```diff
@@ -94,6 +94,40 @@
...
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "dependencyName": {
+            "description": "The full package name of third party dependency",
+            "type": "string"
+          },
+          "dependencyVersionRange": {
+            "description": "The semver version range of third party dependency",
```
dmichon-msft, line 110: Nit: either "semantic version range" or "semver range".

dmichon-msft (approved):
Looks good, just a nit on the field description.

libraries/rush-lib/src/schemas/deploy-scenario.schema.json
EscapeB: Fixed~

libraries/rush-lib/src/schemas/deploy-scenario.schema.json
```diff
@@ -94,6 +94,40 @@
...
+          "dependencyName": {
+            "description": "The full package name of third party dependency",
+            "type": "string"
+          },
+          "dependencyVersionRange": {
+            "description": "The semantic version range of third party dependency",
+            "type": "string"
+          },
+          "patternsToInclude": {
+            "description": "A list of glob patterns to include when extracting specific dependency. If a path is matched by both \"patternsToInclude\" and \"patternsToExclude\", the path will be excluded. If undefined, all paths will be included.",
```
iclanton, line 114:
```suggestion
            "description": "A list of glob patterns to include when extracting the dependency specified in this object. If a path is matched by both \"patternsToInclude\" and \"patternsToExclude\", the path will be excluded. If undefined, all paths will be included.",
```

libraries/rush-lib/src/schemas/deploy-scenario.schema.json
```diff
@@ -94,6 +94,40 @@
...
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          },
+          "patternsToExclude": {
+            "description": "A list of glob patterns to exclude when extracting specific dependency. If a path is matched by both \"patternsToInclude\" and \"patternsToExclude\", the path will be excluded. If undefined, no paths will be excluded.",
```
iclanton, line 121:
```suggestion
            "description": "A list of glob patterns to include when extracting the dependency specified in this object. If a path is matched by both \"patternsToInclude\" and \"patternsToExclude\", the path will be excluded. If undefined, no paths will be excluded.",
```

PR #4310, opened 2023-09-06
[heft-webpack4-plugin] Add an option to heft-webpack4-plugin to replace md4 hashes with md5.

iclanton:
## Summary

This PR adds an option called `patchMd4WithMd5Hash` to `@rushstack/heft-webpack4-plugin` that patches Webpack to replace MD4 hashes with MD5.

## Details

This specifically performs these patches in-memory when Webpack is loaded by the plugin:
- wraps the [`createHash`](https://github.com/webpack/webpack/blob/webpack-4/lib/util/createHash.js#L126-L137) function with a function that conditionally replaces the argument's value with `"md5"` if `"md4"` was passed in
- Replaces `crypto.createHash("md4")` with `crypto.createHash("md5")` in [`SplitChunksPlugin`](https://github.com/webpack/webpack/blob/webpack-4/lib/optimize/SplitChunksPlugin.js#L25C21-L25C21).

## How it was tested

Tested in a large internal repo that uses Webpack 4 under Node 18 and by removing the `--openssl-legacy-provider=true` env var workaround in this repo.

### DIFF
--- a/heft-plugins/heft-webpack4-plugin/src/webpackPackageUtilities.ts
+++ b/heft-plugins/heft-webpack4-plugin/src/webpackPackageUtilities.ts
@@ -74,10 +74,9 @@ async function _patchWebpackCreateHashModule(logger: IWarningErrorEmitter): Prom
 
   type HashFunction = (algorithm?: string | (new () => import('crypto').Hash)) => import('crypto').Hash;
   const webpackCreateHashModule: HashFunction | { default: HashFunction } = await import(createHashFullPath);
-  const webpackCreateHashFunction: HashFunction = (webpackCreateHashModule as { default: HashFunction })
-    .default
-    ? (webpackCreateHashModule as { default: HashFunction }).default
-    : (webpackCreateHashModule as HashFunction);
+  const webpackCreateHashFunction: HashFunction =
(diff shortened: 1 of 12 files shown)


heft-plugins/heft-webpack4-plugin/src/webpackPackageUtilities.ts
```diff
@@ -0,0 +1,153 @@
...
+      )
+    );
+  }
+
+  type HashFunction = (algorithm?: string | (new () => import('crypto').Hash)) => import('crypto').Hash;
+  const webpackCreateHashModule: HashFunction | { default: HashFunction } = await import(createHashFullPath);
+  const webpackCreateHashFunction: HashFunction = (webpackCreateHashModule as { default: HashFunction })
+    .default
+    ? (webpackCreateHashModule as { default: HashFunction }).default
+    : (webpackCreateHashModule as HashFunction);
```
dmichon-msft, line 72:
```suggestion
    ?? (webpackCreateHashModule as HashFunction);
```

dmichon-msft (approved):

iclanton:
Superseded by https://github.com/webpack/webpack/pull/17628

PR #4313, opened 2023-09-06
Fix deployment when an include pattern specifies a subdirectory

D4N14L:
## Summary

Fixes an issue in the `@rushstack/package-extractor` package introduced in PR #4255 which causes subdirectory patterns to get ignored. 

## Details

Originally, package-extractor would enumerate all subdirectories within a project (excluding node_modules and a couple other specific folders) before determining whether or not a file is included in the extraction. This changed and folder paths were now also being analyzed on if they are ignored before enumerating its children, which was originally intended to increase traversal. This had the unintentional affect that inclusion patterns such as `src/subdir/**/*` would not match files in this directory since `src/subdir` itself does not match the pattern, thus ending traversal of this folder path.

This change re-introduces the legacy behaviour while maintaining the other changes introduced in PR #4255

## How it was tested

Added test, tested locally


libraries/package-extractor/src/PackageExtractor.ts
```diff
@@ -826,20 +826,25 @@ export class PackageExtractor {
...
             return;
           }
 
           const sourcePathNode: PathNode = await state.symlinkAnalyzer.analyzePathAsync(sourcePath);
-
           if (sourcePathNode.kind === 'file') {
+            // Only ignore files and not folders to ensure that we traverse the contents of all folders. This is
+            // done so that we can match against subfolder patterns, ex. "src/subfolder/**/*"
+            if (relativeSourcePath !== '' && isFileExcluded(relativeSourcePath)) {
+              callback();
```
iclanton, line 839: Can this callback be converted to a promise?

iclanton (approved):

libraries/package-extractor/src/PackageExtractor.ts
D4N14L: The callback is built into the AsyncQueue and is required to notify that the queue can be iterated. I'm open to suggestions on how the AsyncQueue class works for sure, but I'd call it out of scope for this PR.

EscapeB:
@D4N14L Hi, we got issues after this PR get merged. 

```typescript
// PackageExtractor.ts | line 841
const sourcePathNode: PathNode = await state.symlinkAnalyzer.analyzePathAsync(sourcePath);
```

In this line, we use `symlinkAnalyzer` to get sourcePathNode info, which means even thought we have configured some files should be excluded in extracting process, they still will be analyzed by symlinkAnalyzer.

If there are any errors occur inside analyzePathAsync function, the whole extracting process exit with errors. 

Usually our errors are (more context can be found at https://github.com/microsoft/rushstack/pull/4210)
> Symlink targets not under folder  xxxxx

Can we get all files inside a project/dependency folder without using symlinkAnalyzer?

D4N14L:
@EscapeB the symlink analyzer is used because otherwise trawling through all those folders would be slow since we go over folder paths multiple times, so we're using it as a cache. We should instead fix the bug that is allowing that folder to get analyzed. Can you provide a repro? Then I can narrow down the search a bit better.

EscapeB:
Sure, here is my reproduce repo. https://github.com/EscapeB/rush-playground/tree/main/projects/symlink-demo

D4N14L:
@EscapeB I tried to get that repo cloned and set up, however I got this error:
```
Running "pnpm install" in D:\playground\rush-playground\common\temp

Scope: all 2 workspace projects
../../projects/symlink-demo              |  WARN  Moving dependency-with-error that was installed by a different package manager to "node_modules/.ignored
D:\playground\rush-playground\projects\symlink-demo:
 ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/dependency-with-error: Not Found - 404

This error happened while installing a direct dependency of D:\playground\rush-playground\projects\symlink-demo

dependency-with-error is not in the npm registry, or you have no permission to fetch it.
```

Could you update this branch to include the dependency that is being referenced?/Reference a valid NPM package that reproduces?

EscapeB:
@D4N14L Sorry for late response. 

Our issues comes from several in house packages, which can not be referenced outside company.  

I'm not sure which package in official npm has such kind of issues. 

Actually, you don't need to run `rush update` to set up, just clone that repo and following steps in  README file.
image

octogonz:
@EscapeB opened https://github.com/microsoft/rushstack/issues/4421 to track this regression

PR #4314, opened 2023-09-06
[rushx-command-line] Add telemetry to rushx

nebosite:
## Summary

Reuse the telemetry code in rush to record telemetry for rush x

## Details

Changes:
* Added telemetry initialization and logging to rushx
* Refactored telemetry to take a hook instead of a rush session to prevent code creep from rush to rushx
* Refactored telemetry to allow an undefined hook (rushx does not support this hook)
* Refactored telemetry to take an executable name to distinquish telemetry output by file name. (This is a breaking change for downstream implementations that depend on the file name having a particular structure.  This seems like a safer change than added an exeName property to the data because that can be missed and cause data processing errors by including the wrong files)

## How it was tested

* Invoked `rushx build` and `rush update` to verify that in both cases, the correct filename is written with correct data.
* Invoked `rushx not_exist` and verified correct error data was logged in the telemetry output 

## Impacted documentation

* https://rushjs.io/pages/advanced/rush_files_and_folders/

### DIFF
--- a/libraries/rush-lib/src/cli/RushCommandLineParser.ts
+++ b/libraries/rush-lib/src/cli/RushCommandLineParser.ts
@@ -230,7 +230,7 @@ export class RushCommandLineParser extends CommandLineParser {
 
   private async _wrapOnExecuteAsync(): Promise<void> {
     if (this.rushConfiguration) {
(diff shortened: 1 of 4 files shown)


dmichon-msft:
This PR is still marked as a draft, so assumed to not be ready for review.

PR #4316, opened 2023-09-06
[rush-lib]: Add resolutionMode to pnpm-config.json

### LINKED ISSUE #4283: [rush] package version gets downgraded when `rush update --full` with pnpm v8
## Summary

- with pnpm v7, when we run `rush update --full` (with `"strictPeerDependencies": true` in pnpm-config.json), the latest version of dependency that satisfies the semver range is installed and saved in the lock file.
- however, with pnpm v8, `rush update --full`, in additional to switching to use v6 of the lockfile, also downgraded package versions to the exact version in the semver

## Repro steps

This is demostrated in this branch of test repo: https://github.com/iclanton/rush-test-repo/compare/main...jeremymeng:rush-test-repo:pnpm7

- add `"jssha": "^3.1.0"` to project3 package.json
- bump rush version to latest 5.102.0
- Set `"strictPeerDependencies": false` in pnpm-config.json
- `rush update --full` with pnpm v7, note that v5 lock file is generated, and jssha 3.31.0 installed and stored in pnpm-lock.yamlhttps://github.com/iclanton/rush-test-repo/commit/d5071ed3b5ab2b729622e8d8bc92821cd4a17efe
- change pnpm version to 8.6.12 in rush.json and run `rush update --full` again. Note that v6 lock file is used, but `jssha` 3.1.0 is installed and saved in lock file. https://github.com/iclanton/rush-test-repo/commit/9704d85214bf255e93ddc06cf9289d53fc2ba2f4

I'd expect jssha 3.31.0 too. Not sure if this is a pnpm behavior change or rush one, or combination of them.

Background:
we have a process that 
- Set `"strictPeerDependencies": false` in pnpm-config.json
- Set `"implicitlyPreferredVersions": true` in common-versions.json
- Run `rush update --full` and parse the output to see what version upgrades are available then create or update github workitems for those.

we also have a process that regularly runs `rush update --full` to upgrade to latest possible versions. 

The new behavior is breaking the above processes, because the lower versions are installed, and latest versions are no longer installed by `rush update --full`

are there other ways to achieve what we want? 1) upgrade to latest possible version of the semver 2) report versions available for upgrade but cannot done by 1) e.g., new major versions available.

  **Expected result:** 

  **Actual result:** 

## Standard questions

| Question | Answer |
| -------- | -------- |
| Operating system? |  |

g-chao:
## Summary

This PR is to close #4283, which we want to have some controls on resolution-mode in Rush when using PNPM as package manager.
## Details

We will add a new config option, resolutionMode, which is equivalent to the one on [PNPM](https://pnpm.io/npmrc#resolution-mode), but with some enhancements on Rush side. 

Here is how it works in Rush:

If user set resolution-mode in pnpm-config.json only, Rush use the value in pnpm-config.json
If user set resolution-mode in pnpm-config.json and .npmrc, Rush will ignore value in .npmrc, use the one in pnpm-config.json and print a warning message to user.
If user set resolution-mode in .npmrc only, do nothing, let pnpm handle it
If user does not set resolution-mode in pnpm-config.json and .npmrc, Rush will default resolution-mode to "highest"
## How it was tested

Manually tested with rush-example repo


g-chao:
@microsoft-github-policy-service agree [company="TikTok, Inc."]

g-chao:
@microsoft-github-policy-service agree company="TikTok"

common/reviews/api/rush-lib.api.md
```diff
@@ -520,6 +520,11 @@ export interface _IPnpmOptionsJson extends IPackageManagerOptionsJsonBase {
     globalPeerDependencyRules?: IPnpmPeerDependencyRules;
     pnpmStore?: PnpmStoreOptions;
     preventManualShrinkwrapChanges?: boolean;
+    // Warning: (ae-forgotten-export) The symbol "PnpmResolutionModeOptions" needs to be exported by the entry point index.d.ts
```
dmichon-msft, line 523: Fix missing export from index.ts

libraries/rush-lib/src/schemas/pnpm-config.schema.json
```diff
@@ -159,6 +159,12 @@
     "unsupportedPackageJsonSettings": {
       "description": "(USE AT YOUR OWN RISK) This is a free-form property bag that will be copied into the `common/temp/package.json` file that is generated by Rush during installation. This provides a way to experiment with new PNPM features.  These settings will override any other Rush configuration associated with a given JSON field except for `.pnpmfile.cjs`.",
       "type": "object"
+    },
+
+    "resolutionMode": {
+      "description": "This option override the resolution-mode in pnpm. Use it if you want to change the default resolution behavior when install dependencies. \n\nPNPM documentation: https://pnpm.io/npmrc#resolution-mode",
```
dmichon-msft, line 165:
```suggestion
      "description": "This option overrides the resolution-mode in PNPM. Use it if you want to change the default resolution behavior when installing dependencies. Rush defaults to "highest", regardless of the default in your current version of PNPM. \n\nPNPM documentation: https://pnpm.io/npmrc#resolution-mode",
```

libraries/rush-lib/src/utilities/npmrcUtilities.ts
```diff
@@ -12,22 +12,20 @@ export interface ILogger {
...
-  logger.info(`Transforming ${sourceNpmrcPath}`); // Verbose
-  logger.info(`  --> "${targetNpmrcPath}"`);
+
+// create a global _combinedNpmrc for cache purpose
+const _combinedNpmrcMap: Map<string, string> = new Map();
+
+function _trimNpmrcFile(sourceNpmrcPath: string): string {
+  if (_combinedNpmrcMap.has(sourceNpmrcPath) && _combinedNpmrcMap.get(sourceNpmrcPath) !== undefined) {
+    return _combinedNpmrcMap.get(sourceNpmrcPath)!;
+  }
```
dmichon-msft, line 28:
```suggestion
  const combinedNpmrcFromCache: string = _combinedNpmrcMap.get(sourceNpmrcPath);
  if (combinedNpmrcFromCache !== undefined) {
    return combinedNpmrcFromCache;
  }
```

libraries/rush-lib/src/utilities/npmrcUtilities.ts
```diff
@@ -110,3 +135,9 @@ export function syncNpmrc(
     throw new Error(`Error syncing .npmrc file: ${e}`);
   }
 }
+
+export function isVariableSetInNpmrcFile(sourceNpmrcFolder: string, variableKey: string): boolean {
+  const sourceNpmrcPath: string = path.join(sourceNpmrcFolder, '.npmrc');
+  const trimNpmrcFile: string = _trimNpmrcFile(sourceNpmrcPath);
+  return trimNpmrcFile.includes(`${variableKey}=`);
```
dmichon-msft, line 142: This needs to verify that a line starts with this expression, otherwise this will match if `variableKey` is a suffix of some other variable name, or is in a comment, etc.

g-chao: Thanks! Changed to use RegExp to find the matches.

common/changes/@microsoft/rush/chao-update-pnpm-config_2023-08-30-00-23.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "feat: add resolutionMode in pnpm-config",
```
octogonz, line 5:
```suggestion
      "comment": "(IMPORTANT) Add a new setting `resolutionMode` in pnpm-config.json; be aware that Rush now overrides the default behavior if you are using PNPM 8.0.0 through 8.6.12 (GitHub #4283)",
```

octogonz: FYI there's some tips for writing `rush change` messages in this doc: [Authoring change logs](https://rushjs.io/pages/best_practices/change_logs/)

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -16,6 +16,12 @@ import schemaJson from '../../schemas/pnpm-config.schema.json';
  */
 export type PnpmStoreOptions = 'local' | 'global';
 
+/**
+ * This represents the available resolution-mode options
```
octogonz, line 20:
```suggestion
 * Possible values for the `resolutionMode` setting in Rush's pnpm-config.json file.
 * @remarks
 * These modes correspond to PNPM's `resolution-mode` values, which are documented here:
 *
 * {@link https://pnpm.io/npmrc#resolution-mode }
```

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -116,6 +126,17 @@ export class PnpmOptionsConfiguration extends PackageManagerOptionsConfiguration
...
   public readonly pnpmStore: PnpmStoreOptions;
 
+  /**
+   * The method used to set resolution mode used by PNPM.
+   *
+   * @remarks
+   * The resolution mode used by PNPM.
+   * Available options: highest, time-based, lowest-direct
+   *
+   * PNPM documentation: https://pnpm.io/npmrc#resolution-mode
```
octogonz, line 136: * This setting determines PNPM's `resolution-mode`  option. The default value is `highest`.
 * 
 * @remarks
 * Be aware that the PNPM 8 initially defaulted to `lowest` instead of  `highest`, but this 
 * was reverted in 8.6.12 because it caused confusion for users.  Rush 5.105.0 and newer
 * avoids this problem by always defaulting to `highest` when `resolutionMode` is not 
 * explicitly set in pnpm-config.json or .npmrc.
 * 
 * PNPM documentation: https://pnpm.io/npmrc#resolution-mode

octogonz, line 136:
```suggestion
   * This setting determines PNPM's `resolution-mode`  option. The default value is `highest`.
   * 
   * @remarks
   * Be aware that the PNPM 8 initially defaulted to `lowest` instead of  `highest`, but PNPM
   * reverted this decision in 8.6.12 because it caused confusion for users.  Rush 5.105.0 and newer
   * avoids this confusion by always defaulting to `highest` when `resolutionMode` is not 
   * explicitly set in pnpm-config.json or .npmrc.
   * 
   * PNPM documentation: https://pnpm.io/npmrc#resolution-mode
```

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -597,6 +598,30 @@ ${gitLfsHookHandling}
...
+      */
+      const isResolutionModeInNpmrc: boolean = isVariableSetInNpmrcFile(
+        this.rushConfiguration.commonRushConfigFolder,
+        'resolution-mode'
+      );
+      if (this.rushConfiguration.pnpmOptions.resolutionMode) {
+        if (isResolutionModeInNpmrc) {
+          console.log(
+            colors.yellow(
+              `Warning: you have set resolution-mode in both .npmrc and pnpm-config.json! We will use the value in pnpm-config.json.`
```
octogonz, line 615:
```suggestion
              `Warning: PNPM's resolution-mode is specified in both .npmrc and pnpm-config.json. The value in pnpm-config.json will take precedence.`
```

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -16,6 +16,16 @@ import schemaJson from '../../schemas/pnpm-config.schema.json';
  */
 export type PnpmStoreOptions = 'local' | 'global';
 
+/**
+ * Possible values for the `resolutionMode` setting in Rush's pnpm-config.json file.
+ * @remarks
+ * These modes correspond to PNPM's `resolution-mode` values, which are documented here:
+ *
+ * {@link https://pnpm.io/npmrc#resolution-mode }
+ * @public
```
iclanton, line 25:
```suggestion
 * These modes correspond to PNPM's `resolution-mode` values, which are documented here:
 * {@link https://pnpm.io/npmrc#resolution-mode}
 *
 * @public
```

libraries/rush-lib/src/utilities/npmrcUtilities.ts
```diff
@@ -70,6 +69,33 @@ function _copyAndTrimNpmrcFile(logger: ILogger, sourceNpmrcPath: string, targetN
...
+ *
+ * Why are we trimming the .npmrc lines?  NPM allows environment variables to be specified in
+ * the .npmrc file to provide different authentication tokens for different registry.
+ * However, if the environment variable is undefined, it expands to an empty string, which
+ * produces a valid-looking mapping with an invalid URL that causes an error.  Instead,
+ * we'd prefer to skip that line and continue looking in other places such as the user's
+ * home directory.
+ *
+ * @returns
+ * The text of the the .npmrc.
```
iclanton, line 91:
```suggestion
 * The text of the the .npmrc with lines containing undefined variables commented out.
```

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -597,6 +598,30 @@ ${gitLfsHookHandling}
...
+        this.rushConfiguration.commonRushConfigFolder,
+        'resolution-mode'
+      );
+      if (this.rushConfiguration.pnpmOptions.resolutionMode) {
+        if (isResolutionModeInNpmrc) {
+          console.log(
+            colors.yellow(
+              `Warning: PNPM's resolution-mode is specified in both .npmrc and pnpm-config.json. The value in pnpm-config.json will take precedence.`
+            )
+          );
```
iclanton, line 617:
```suggestion
          this._terminal.writeWarningLine(
            `Warning: PNPM's resolution-mode is specified in both .npmrc and pnpm-config.json. ` +
              `The value in pnpm-config.json will take precedence.`
          );
```

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -16,6 +16,16 @@ import schemaJson from '../../schemas/pnpm-config.schema.json';
...
+ */
+export type PnpmResolutionModeOptions = 'highest' | 'time-based' | 'lowest-direct';
```
iclanton, line 27:
```suggestion
export type PnpmResolutionMode = 'highest' | 'time-based' | 'lowest-direct';
```

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -597,6 +598,30 @@ ${gitLfsHookHandling}
...
+        }
+        args.push(`--config.resolutionMode=${this.rushConfiguration.pnpmOptions.resolutionMode}`);
+      } else if (!isResolutionModeInNpmrc) {
+        // if you don't specify the resolution-mode in both .npmrc and pnpm-config.json, then rush will default it to "highest"
+        args.push('--config.resolutionMode=highest');
+      }
```
iclanton, line 623:
```suggestion
      let resolutionMode: PnpmResolutionModeOptions | undefined =
        this.rushConfiguration.pnpmOptions.resolutionMode;
      if (resolutionMode) {
        if (isResolutionModeInNpmrc) {
          this._terminal.writeWarningLine(
            `Warning: PNPM's resolution-mode is specified in both .npmrc and pnpm-config.json. ` +
              `The value in pnpm-config.json will take precedence.`
          );
        }
      } else if (!isResolutionModeInNpmrc) {
        // if resolution-mode isn't specified in either .npmrc or pnpm-config.json,
        // then rush will default it to "highest"
        resolutionMode = 'highest';
      }

      if (resolutionMode) {
        args.push(`--config.resolutionMode=${resolutionMode}`);
      }
```

libraries/rush-lib/src/utilities/npmrcUtilities.ts
```diff
@@ -110,3 +136,17 @@ export function syncNpmrc(
...
+
+  //if .npmrc file does not exist, return false directly
+  if (!fs.existsSync(sourceNpmrcPath)) {
+    return false;
+  }
+
+  const trimNpmrcFile: string = _trimNpmrcFile(sourceNpmrcPath);
+
+  const variableKeyRegExp: RegExp = new RegExp(`^${variableKey}=`, 'm');
+  return trimNpmrcFile.match(variableKeyRegExp) !== null;
```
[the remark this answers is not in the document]
octogonz: @iclanton @g-chao  It is not safe to generate a RegExp like this:
```
const variableKeyRegExp: RegExp = new RegExp(`^${variableKey}=`, 'm');
```

We either need:

- a `//` comment containing a convincing proof that `variableKey` cannot possibly ever contain any RegExp operators. For example, what if the `variableKey` is `this.that`? Then the `.` would be matched as a RegExp wildcard.  


Or else:

-  we need to call `Text.escapeRegExp(variableKey)` to properly escape any such characters.

However, generating RegExp's on the fly is not very efficient.  Each time `variableKey` changes, the RegExp has to be recompiled.

For such a very simple test, could we instead do a simple string comparison, something like `trimmedNpmrcFile.startsWith(variableKey + '=')`?  This avoids the complexity of dynamically generated RegExp matchers.

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -116,6 +130,19 @@ export class PnpmOptionsConfiguration extends PackageManagerOptionsConfiguration
    */
   public readonly pnpmStore: PnpmStoreOptions;
 
+  /**
+   * This setting determines PNPM's `resolution-mode`  option. The default value is `highest`.
+   *
+   * @remarks
+   * Be aware that the PNPM 8 initially defaulted to `lowest` instead of  `highest`, but PNPM
+   * reverted this decision in 8.6.12 because it caused confusion for users.  Rush 5.105.0 and newer
```
iclanton, line 138:
```suggestion
   * reverted this decision in 8.6.12 because it caused confusion for users.  Rush 5.106.0 and newer
```

libraries/rush-lib/src/utilities/npmrcUtilities.ts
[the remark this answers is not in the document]
g-chao: But, `trimmedNpmrcFile.startsWith(variableKey + '=')` can not match multiple lines? Or you mean, we use a for loop to to check each line?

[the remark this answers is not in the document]
octogonz: Okay looking more closely at the code, I think the RegExp is actually not considering the possibility of whitespace.  For example, wouldn't NPM consider the following two lines to be equivalent?

**.npmrc**
```bash
my-variable=example value
```

```bash
     my-variable     =     example value
```

And this I believe will get wrongly matched by the current implementation:
```bash
# my-variable=example value
```


So, maybe this file format is complex enough that RegExp is the right way.  But we don't need to hardwire the name of the variable into the expression.  We could instead calculate it as follows:

```js
// an "=" preceded by a key that doesn't contain "=" or  comment "#"
// Use lazy matching to trim whitespace
const pattern: RegExp = /^\s*([^#=]+?)\s*=/gm
```

Then you can use `pattern.exec(input)` to obtain each key and compare against `variableKey`.

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
iclanton: We're trying to move everything to `Terminal`. `Terminal.writeWarningLine` by default prints in Yellow.

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
g-chao: Maybe we can stick with this naming `PnpmResolutionModeOptions`? The existing `PnpmStoreOptions` using the same naming convention

libraries/rush-lib/src/utilities/npmrcUtilities.ts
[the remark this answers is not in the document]
g-chao: It seems ` /^\s*([^#=]+?)\s*=/gm`  can't remove spaces in between.
so 
// "b c    =  1" or become "b c    =" not "bc="

And, I found the existing function `_trimNpmrcFile()`, we already trimmed all spaces before and after each line, why we take a step further, to remove spaces before and after key and value as well?
So 
// "  b c  =  1 2 " will be to "b c=1 2"

I made some updates to this `_trimNpmrcFile()`, you can take a look. 

After this function, all key value pairs in `.npmrc` will be in perfect format

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
iclanton: `PnpmStoreOptions` is also a non-ideal name for that type. In almost all cases, when we call a type `___Options`, it's an object that contains named parameters for a function.

Actually, here's a PR to rename that type: https://github.com/microsoft/rushstack/pull/4320

g-chao: Thanks for the explanation. I will rename it.

iclanton (approved):

PR #4320, opened 2023-09-11
[rush] Rename the PnpmStoreOptions type to PnpmStoreLocation.

iclanton:
Renaming the `PnpmStoreOptions` type to `PnpmStoreLocation` because it isn't an options object.


octogonz-proxy (approved):
LGTM -- I don't think we need to consider this a "breaking" change, since this only affects a compile time type, and the type itself is probably rarely used.

PR #4336, opened 2023-09-18
[rush] Cleanup to the handling of PNPM custom tips.

iclanton:
## Summary

This changes a few things:

- Removes fields from the `custom-tips.json` schema that had been removed in https://github.com/microsoft/rushstack/pull/4292
- Adds functionality to the `PrintUtilities.wrapWords` function to accept a line prefix, and use that in the new custom tips formatting.
- Fixes an issue where pnpm's installation progress line wasn't reprinted on a TTY console when there aren't any custom tips specified. Note that this is still an issue when there are custom tips, which needs more investigation.

## How it was tested

Added more unit tests for `PrintUtilities` and tested custom tips in a small test repo.

### DIFF
--- a/libraries/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
+++ b/libraries/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
@@ -391,13 +391,11 @@ export class WorkspaceInstallManager extends BaseInstallManager {
         // The try-finally is to avoid the tips NOT being printed if the install fails.
         // NOT catching the error because we want to keep the other behaviors (i.e., the error will be caught and handle in upper layers).
 
-        const tipIDsToBePrintedArray: CustomTipId[] = Array.from(tipIDsToBePrinted);
-        for (let i: number = 0; i < tipIDsToBePrintedArray.length; i++) {
-          if (i !== 0) {
-            this._terminal.writeLine();
+        if (tipIDsToBePrinted.size > 0) {
+          this._terminal.writeLine();
+          for (const tipID of tipIDsToBePrinted) {
+            this.rushConfiguration.customTipsConfiguration._showTip(this._terminal, tipID);
           }
-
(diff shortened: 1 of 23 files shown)


libraries/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
```diff
@@ -388,15 +391,14 @@ export class WorkspaceInstallManager extends BaseInstallManager {
...
+        for (let i: number = 0; i < tipIDsToBePrintedArray.length; i++) {
+          if (i !== 0) {
             this._terminal.writeLine();
           }
-          this.rushConfiguration.customTipsConfiguration._showTip(this._terminal, tipID);
-          idx++;
-        });
+
+          this.rushConfiguration.customTipsConfiguration._showTip(this._terminal, tipIDsToBePrintedArray[i]);
+        }
```
dmichon-msft, line 401:
```suggestion
        let isFirstLine: boolean = true;
        for (const tipID of tipIDsToBePrinted) {
          if (isFirstLine) {
            this._terminal.writeLine();
            isFirstLine = false;
          }
        
          this.rushConfiguration.customTipsConfiguration._showTip(this._terminal, tipID);
        }
```

libraries/terminal/src/PrintUtilities.ts
```diff
@@ -22,40 +20,127 @@ export class PrintUtilities {
...
+  public static wrapWords(text: string, maxLineLength?: number, linePrefix?: string): string;
+  public static wrapWords(
+    text: string,
+    maxLineLength?: number,
+    indentOrLinePrefix?: number | string
+  ): string {
+    const wrappedLines: string[] = PrintUtilities.wrapWordsToLines(
+      text,
+      maxLineLength,
+      indentOrLinePrefix as string | undefined // TS is confused by the overloads
```
dmichon-msft, line 50: This is because there isn't an overload that expressly accepts `string | number`. The last one is treated as the implementation and not one of the overloads.

dmichon-msft (approved):
Few minor notes.

PR #4338, opened 2023-09-19
[rush] Fix pnpm's install status printing when pnpm custom tips are defined.

iclanton:
## Summary

This change adds a `--reporter default` parameter to the args passed to pnpm if Rush is running in an environment with a TTY stdout. This fixes an issue where, when the pnpm output is being monitored to match custom tips, pnpm will detect a non-TTY stdout stream and use the `append-only` reporter. This reporter doesn't incrementally update the installation progress line, and instead repeatedly reprints that line down the console.
 
Docs on that parameter: https://pnpm.io/cli/install#--reportername

## How it was tested

Tested in a repo with a nonexistent dependency and a `TIP_PNPM_NO_MATCHING_VERSION` custom tip defined.

### DIFF
--- a/common/changes/@microsoft/rush/fix-tty_2023-09-19-10-01.json
+++ b/common/changes/@microsoft/rush/fix-tty_2023-09-19-10-01.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix pnpm's install status printing when pnpm custom tips are defined.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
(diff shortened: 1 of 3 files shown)


common/changes/@microsoft/rush/fix-tty_2023-09-19-10-01.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fix pnpm's install status printing when pnpm custom tips are defined.",
```
octogonz-proxy, line 5: @iclanton let's get this merged+released asap. The regression is conspicuous and visually annoying.

theJiawen:
thanks for the fix!

dmichon-msft (approved):

PR #4342, opened 2023-09-20
[rush] Add "resolutionMode" to "rush init" template

octogonz-proxy:
## Summary

 
This is a bit of work we missed in #4316

## How it was tested

Confirmed the `rush init` output is correct JSON.

@g-chao


dmichon-msft (approved):

octogonz-proxy:

iclanton (approved):

PR #4348, opened 2023-09-22
Create and use a stricter local ESLint configuration.

iclanton:


eslint/eslint-config-local/mixins/react.js
```diff
@@ -0,0 +1,57 @@
...
+      files: ['*.ts', '*.tsx'],
+
+      // New rules and changes to existing rules
+      rules: {
+        // =====================================================================
+        // eslint-plugin-react-hooks
+        // =====================================================================
+        'react-hooks/rules-of-hooks': 'error',
+        'react-hooks/exhaustive-deps': 'warn',
+        'deprecation/deprecation': 'off'
```
D4N14L, line 29: Why include the plugin but disable?

eslint/eslint-config-local/profile/_common.js
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
```
D4N14L, line 1: Why re-declare all this stuff? I know internally we have a change that should allow for extending from configs in a package with any name in order to extend from `@rushstack/eslint-config`. Maybe we should bring that in with this PR to reduce duplication?

eslint/eslint-config-local/mixins/react.js
iclanton, line 29:
```suggestion
        'react-hooks/exhaustive-deps': 'warn'
```

PR #4350, opened 2023-09-24
[rush] Update the functionality that runs external lifecycle processes to be async.

iclanton:
## Summary

This PR refactors the `executeCommand` and related functions to be async. This is in preparation for some follow-up work to merge `executeCommandAndProcessOutputWithRetryAsync` into the `executeCommandAsync` function.

## How it was tested

This change needs some more testing.

### DIFF
--- a/libraries/rush-lib/src/cli/actions/ChangeAction.ts
+++ b/libraries/rush-lib/src/cli/actions/ChangeAction.ts
@@ -164,8 +164,9 @@ export class ChangeAction extends BaseRushAction {
   }
 
   public async runAsync(): Promise<void> {
+    const targetBranch: string = await this._getTargetBranchAsync();
     // eslint-disable-next-line no-console
-    console.log(`The target branch is ${this._targetBranch}`);
+    console.log(`The target branch is ${targetBranch}`);
 
     if (this._verifyParameter.value) {
       const errors: string[] = [
@@ -197,11 +198,11 @@ export class ChangeAction extends BaseRushAction {
     const sortedProjectList: string[] = (await this._getChangedProjectNamesAsync()).sort();
     if (sortedProjectList.length === 0) {
       this._logNoChangeFileRequired();
-      this._warnUnstagedChanges();
+      await this._warnUnstagedChangesAsync();
       return;
     }
 
-    this._warnUnstagedChanges();
+    await this._warnUnstagedChangesAsync();
 
     const inquirer: typeof InquirerType = await import('inquirer');
     const promptModule: InquirerType.PromptModule = inquirer.createPromptModule();
@@ -277,7 +278,7 @@ export class ChangeAction extends BaseRushAction {
       interactiveMode = true;
 
       const existingChangeComments: Map<string, string[]> = ChangeFiles.getChangeComments(
-        this._getChangeFiles()
+        await this._getChangeFilesAsync()
       );
       changeFileData = await this._promptForChangeFileDataAsync(
         promptModule,
@@ -310,7 +311,7 @@ export class ChangeAction extends BaseRushAction {
     }
     if (this._commitChangesFlagParameter.value || this._commitChangesMessageStringParameter.value) {
       if (changefiles && changefiles.length !== 0) {
-        this._stageAndCommitGitChanges(
+        await this._stageAndCommitGitChangesAsync(
           changefiles,
           this._commitChangesMessageStringParameter.value ||
             this.rushConfiguration.gitChangefilesCommitMessage ||
@@ -340,15 +341,16 @@ export class ChangeAction extends BaseRushAction {
   private async _verifyAsync(): Promise<void> {
     const changedPackages: string[] = await this._getChangedProjectNamesAsync();
     if (changedPackages.length > 0) {
-      this._validateChangeFile(changedPackages);
+      await this._validateChangeFileAsync(changedPackages);
     } else {
       this._logNoChangeFileRequired();
     }
   }
 
-  private get _targetBranch(): string {
+  private async _getTargetBranchAsync(): Promise<string> {
     if (!this._targetBranchName) {
-      this._targetBranchName = this._targetBranchParameter.value || this._git.getRemoteDefaultBranch();
+      this._targetBranchName =
+        this._targetBranchParameter.value || (await this._git.getRemoteDefaultBranchAsync());
     }
 
     return this._targetBranchName;
@@ -358,7 +360,7 @@ export class ChangeAction extends BaseRushAction {
     const projectChangeAnalyzer: ProjectChangeAnalyzer = new ProjectChangeAnalyzer(this.rushConfiguration);
     const changedProjects: Set<RushConfigurationProject> =
       await projectChangeAnalyzer.getChangedProjectsAsync({
-        targetBranchName: this._targetBranch,
+        targetBranchName: await this._getTargetBranchAsync(),
         terminal: this._terminal,
         shouldFetch: !this._noFetchParameter.value,
         // Lockfile evaluation will expand the set of projects that request change files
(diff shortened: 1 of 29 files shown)


libraries/rush-lib/src/utilities/Utilities.ts
```diff
@@ -773,39 +785,52 @@ export class Utilities {
...
-    if (result.error && (result.error as any).errno === 'ENOENT') {
-      // This is a workaround for GitHub issue #25330
-      // https://github.com/nodejs/node-v0.x-archive/issues/25330
-      //
-      // TODO: The fully worked out solution for this problem is now provided by the "Executable" API
-      // from @rushstack/node-core-library
-      result = child_process.spawnSync(command + '.cmd', args, options);
+    const stdoutBuffers: string[] = [];
+    process.stdout?.on('data', (chunk) => {
+      stdoutBuffers.push(chunk.toString());
```
D4N14L, line 792: Should we care about encoding here?

libraries/rush-lib/src/utilities/Utilities.ts
```diff
@@ -773,39 +785,52 @@ export class Utilities {
...
+    });
+
+    const stderrBuffers: string[] = [];
+    process.stderr?.on('data', (chunk) => {
+      stderrBuffers.push(chunk.toString());
```
D4N14L, line 797: Or here?

libraries/rush-lib/src/utilities/Utilities.ts
```diff
@@ -773,39 +785,52 @@ export class Utilities {
...
+    });
+
+    let error: Error | undefined;
+    try {
+      await once(process, 'close');
```
D4N14L, line 802: There was some outcome that we came to with which event we should wait on before exiting. Is this the one we settled on?

libraries/rush-lib/src/utilities/Utilities.ts
```diff
@@ -773,39 +785,52 @@ export class Utilities {
...
 
-    if (result.status) {
-      throw new Error(
-        'The command failed with exit code ' +
-          result.status +
-          '\n' +
-          (result.stderr ? result.stderr.toString() : '')
-      );
+    if (status) {
+      throw new Error(`The command failed with exit code ${status}\n${stderr}`);
```
D4N14L, line 833: Did non-zero exit codes always throw `spawnSync`? I know some tools intentionally return non-zero exit codes. Whatever we do here should be compatible with what `spawnSync` does.

libraries/rush-lib/src/cli/actions/ChangeAction.ts
```diff
@@ -340,15 +341,16 @@ export class ChangeAction extends BaseRushAction {
...
       this._logNoChangeFileRequired();
     }
   }
 
-  private get _targetBranch(): string {
+  private async _getTargetBranchAsync(): Promise<string> {
     if (!this._targetBranchName) {
-      this._targetBranchName = this._targetBranchParameter.value || this._git.getRemoteDefaultBranch();
+      this._targetBranchName =
+        this._targetBranchParameter.value || (await this._git.getRemoteDefaultBranch());
```
D4N14L, line 353: Label this as an async method

libraries/rush-lib/src/cli/actions/PublishAction.ts
```diff
@@ -337,50 +337,53 @@ export class PublishAction extends BaseRushAction {
...
+    git: PublishGit,
+    allPackages: Map<string, RushConfigurationProject>
+  ): Promise<void> {
     // eslint-disable-next-line no-console
     console.log(`Rush publish starts with includeAll and version policy ${this._versionPolicy.value}`);
 
     let updated: boolean = false;
 
-    allPackages.forEach((packageConfig, packageName) => {
+    for (const [packageName, packageConfig] of allPackages) {
```
D4N14L, line 365: Async.forEachAsync? Though not sure what issues multiple async git commands may cause

libraries/rush-lib/src/cli/test/Cli.test.ts
```diff
@@ -10,8 +10,8 @@ describe('CLI', () => {
     const workingDir: string = '/';
     const startPath: string = path.resolve(__dirname, '../../../lib-commonjs/start.js');
 
-    expect(() => {
-      Utilities.executeCommand({
+    expect(async () => {
```
D4N14L, line 13: Do you have to await these expects?

libraries/rush-lib/src/utilities/Utilities.ts
```diff
@@ -854,39 +866,52 @@ export class Utilities {
...
 
-    if (result.status) {
-      throw new Error(
-        'The command failed with exit code ' +
-          result.status +
-          '\n' +
-          (result.stderr ? result.stderr.toString() : '')
-      );
+    const stderr: string = stderrBuffers.join('');
+    const stdout: string = stdoutBuffers.join('');
```
D4N14L, line 889: I feel like a lot of this should exist under the Executable class... but it may not? (On mobile)

libraries/rush-lib/src/cli/actions/PublishAction.ts
iclanton: Yeah that feels a little unsafe. Let's leave this sequential.

libraries/rush-lib/src/utilities/Utilities.ts
iclanton: Updated to use utf8

libraries/rush-lib/src/utilities/Utilities.ts
iclanton: Updated to use `Executable.waitForExitAsync`

libraries/rush-lib/src/utilities/Utilities.ts
iclanton: No, `spawnSync` returns an object that includes the exit code.

libraries/rush-lib/src/utilities/Utilities.ts
```diff
@@ -302,119 +312,57 @@ export class Utilities {
...
+    suppressOutput,
+    onStdoutStreamChunk,
+    environment,
+    keepEnvironment
+  }: IExecuteCommandOptions): Promise<void> {
+    await Utilities._executeCommandInternalAsync({
+      command,
+      args,
+      workingDirectory,
+      stdio: onStdoutStreamChunk ? ['inherit', 'pipe', 'inherit'] : suppressOutput ? undefined : [0, 1, 2],
```
D4N14L, line 328: Does undefined prevent capturing stderr? It looks like it's used below in error scenarios. Seems like it would be valuable to capture regardless of the suppressOutput argument. Also, what does [0,1,2] do? Maybe add a comment?

libraries/rush-lib/src/utilities/Utilities.ts
```diff
@@ -870,39 +766,58 @@ export class Utilities {
...
+    error,
+    stderr,
+    status
+  }: {
+    error: Error | undefined;
+    stderr: string;
+    status: number | null;
+  }): void {
+    if (error) {
+      error.message += `\n${stderr}`;
```
D4N14L, line 815: Should the exit code be optionally included? Also maybe only print the newline if stderr is truthy

D4N14L (approved):

libraries/rush-lib/src/logic/Git.ts
```diff
@@ -318,40 +306,43 @@ export class Git {
...
-
-          if (!remoteUrl) {
-            return false;
+      const matchingRemotes: string[] = [];
+      await Async.forEachAsync(
+        output.split('\n'),
+        async (remoteName) => {
+          if (remoteName) {
+            const remoteUrl: string = (
+              await this._executeGitCommandAndCaptureOutputAsync(gitPath, [
```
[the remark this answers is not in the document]
iclanton: In local testing, this works fine.

D4N14L (approved):

PR #4358, opened 2023-09-26
[heft] Fix an issue where 'heft clean' crashes.

iclanton:
## Summary

Fixes https://github.com/microsoft/rushstack/issues/4351

## Details

`heft clean` was directly constructing an `AbortSignal` object via `new AbortSignal()`, which isn't allowed. This change refactors the code used in `HeftActionRunner` to construct the `AbortSignal` to also be used by `heft clean`.

## How it was tested

Tested in this repo and in the repro provided in https://github.com/microsoft/rushstack/issues/4351.

### DIFF
--- a/apps/heft/src/cli/HeftActionRunner.ts
+++ b/apps/heft/src/cli/HeftActionRunner.ts
@@ -76,6 +76,35 @@ export function initializeHeft(
   terminal.writeVerboseLine('');
 }
 
+let _cliAbortSignal: AbortSignal | undefined;
+export function ensureCliAbortSignal(terminal: ITerminal): AbortSignal {
+  if (!_cliAbortSignal) {
+    // Set up the ability to terminate the build via Ctrl+C and have it exit gracefully if pressed once,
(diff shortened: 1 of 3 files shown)


dmichon-msft (approved):

D4N14L (approved):

apps/heft/src/cli/HeftActionRunner.ts
```diff
@@ -76,6 +76,35 @@ export function initializeHeft(
   terminal.writeVerboseLine('');
 }
 
+let _cliAbortSignal: AbortSignal | undefined;
+export function ensureCliAbortSignal(terminal: ITerminal): AbortSignal {
```
dmichon-msft, line 80: Might consider migrating this over to operation-runner.

iclanton: @dmichon-msft - you want to do that in a follow-up?

dmichon-msft: Sure.

Issue #4364, opened 2023-09-28
[rush] watch mode fails as "process.stdin.setRawMode is not a function" since 5.107.0


antoine-coulon:
## Summary

Hello folks, Rush **5.107.0** release and specifically [this commit](https://github.com/microsoft/rushstack/commit/122f164e7eb50768df5be794e3ccc637d4cb276b#diff-06f3bbc9b62600e841944eabfec9d427c69b26ca7d6af979072cfbbac0b41816) introduced a great feature, which is enabling pausing/resuming the watch mode. However, when running Rush commands on watch mode via child_processes, it produces the following error `process.stdin.setRawMode is not a function` as this requires the Node process to be initialized with process.stdin as an instance of `tty.ReadStream`, otherwise these methods are not available. This currently broke the custom toolchain we wrote on top of Rush.js, starting from `5.107.0`.

This can be fixed by providing `stdio: "inherit"` to the child_process so that the child process shares the TTY instance however I feel like we should be able to run watched commands in the background whether it's simply by piping (`stdio: "pipe"`) or ignoring the stdio (`stdio: "ignore"`). In other words, **we should be able to benefit from interactivity is the process is spawned in a way that is meant to be interactive, but it should not crash the process otherwise**.

## Repro steps

Just run a Rush command under watch mode from a child_process with no TTY enabled:

```js

import { spawn } from "node:child_process";

spawn("<whatever_rush_command_under_watch_mode>", { stdio: "pipe" });
```

Will provoke the following: `process.stdin.setRawMode is not a function since 5.107.0`

  **Expected result:** Run the watch mode, with no interactivity whatsoever

  **Actual result:** `process.stdin.setRawMode is not a function`

## Details

I'll land a PR that first checks if the process.stdin is an instance of TTY before registering keyboard listeners (coming from `_registerWatchModeInterface`)

## Standard questions

Please answer these questions to help us investigate your issue more quickly:

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.107.0 |
| `rushVersion` from rush.json? | 5.107.0 |
| `useWorkspaces` from rush.json? | yes |
| Operating system? | Mac |
| Would you consider contributing a PR? | Yes |
| Node.js version (`node -v`)? | 18.18.0 |

antoine-coulon:
This was merged in #4365 by @dmichon-msft, should I close it or wait for the release?

D4N14L:
[...]

PR #4365, opened 2023-09-28
[rush] Fix project watcher crash when the started process is not in TTY mode

antoine-coulon:
## Summary

Hello, this follows up the issue #4364 I opened where the project watcher would fail when the process is not started in interactive mode (i.e: no TTY). This is linked to a nice feature that was introduced in `5.107.0`.

## Details

In our own toolchain, we usually spawn Rush commands as child processes and most of the time we only want to pipe the output between processes, meaning that processes are spawned with no interactivity purposes. In that specific case, we are not able to do so and forced to inherit stdin to prevent the error from happening. 

About the fix, this is pretty straightforward, this just ensures that the TTY API + keyboard event listeners are only registered when the process is in TTY mode itself.

## How it was tested

On our own Rush monorepo where the bug was discovered.

## Impacted documentation

None, this is the desired behavior that was probably described when the feature was introduced

### DIFF
--- a/common/changes/@microsoft/rush/use-process-interactivity-when-being-in-tty_2023-09-28-12-51.json
+++ b/common/changes/@microsoft/rush/use-process-interactivity-when-being-in-tty_2023-09-28-12-51.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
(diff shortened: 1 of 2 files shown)


antoine-coulon:

dmichon-msft (approved):
Nice and simple.

PR #4398, opened 2023-10-18
[rush-lib] Use configured shrinkwrapFilename for logging shrinkwrapFilePhrase

davemyersworld:
## Summary

When using a rush-managed project with pnpm, occasionally, there are problems with the shrinkwrap file. The log message goes like `The shrinkwrap file is out of date. You need to run "rush update".` When working with yarn, there is a hard-coded hint as to _what_ the shrinkwrap file is: `The shrinkwrap file (yarn.lock) is out of date. You need to run "rush update".`

Since we have access to the name of the shrinkwrap file, I'm suggesting that we change to implementation to use that so that all package managers include a hint as to what shrinkwrap file it's talking about.

## Details

(covered above)

## How it was tested
I tested this by running tests and manually reviewing in a pnpm and a yarn codebase.

### DIFF
--- a/common/changes/@microsoft/rush/dmyers-shrinkwrap-filename_2023-10-18-19-31.json
+++ b/common/changes/@microsoft/rush/dmyers-shrinkwrap-filename_2023-10-18-19-31.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "Use configured shrinkwrapFilename when printing out the Shrinkwrap phrase",
+      "comment": "Include the filename of the shrinkwrap file in logging messages for all package managers, not just Yarn.",
       "type": "none"
     }
   ],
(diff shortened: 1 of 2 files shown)


common/changes/@microsoft/rush/dmyers-shrinkwrap-filename_2023-10-18-19-31.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Use configured shrinkwrapFilename when printing out the Shrinkwrap phrase",
```
dmichon-msft, line 5:
```suggestion
      "comment": "Include the filename of the shrinkwrap file in logging messages for all package managers, not just Yarn.",
```

dmichon-msft (approved):

iclanton (approved):

PR #4401, opened 2023-10-20
[rush-lib] Enhancing Performance with Concurrent Execution using `Promise.all` ✈

sanjaiyan-dev:
Hi, this PR aims to bring performance improvements by running asynchronous code concurrently using `Promise.all`. This approach not only enhances the overall speed and efficiency of the code but also allows for better resource utilization, reduced latency, and improved scalability.

### DIFF
--- a/libraries/rush-lib/scripts/copyEmptyModules.js
+++ b/libraries/rush-lib/scripts/copyEmptyModules.js
@@ -73,7 +73,7 @@ module.exports = {
             terminal.writeVerboseLine(`Copying ${inDtsPath} to ${outDtsPath}`);
             // We know this is a file, don't need the redundant checks in FileSystem.copyFileAsync
             const buffer = await FileSystem.readFileToBufferAsync(inDtsPath);
-            return FileSystem.writeFileAsync(outDtsPath, buffer, { ensureFolderExists: true });
+            await FileSystem.writeFileAsync(outDtsPath, buffer, { ensureFolderExists: true });
           }
         }
       });
--- a/libraries/rush-lib/scripts/copyEmptyModules.js
+++ b/libraries/rush-lib/scripts/copyEmptyModules.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { FileSystem } = require('@rushstack/node-core-library');
+const { FileSystem, Async } = require('@rushstack/node-core-library');
 
 const JS_FILE_EXTENSION = '.js';
 const DTS_FILE_EXTENSION = '.d.ts';
@@ -50,7 +50,7 @@ module.exports = {
       const folderItems = await FileSystem.readFolderItemsAsync(
         relativeFolderPath ? `${jsInFolderPath}/${relativeFolderPath}` : jsInFolderPath
       );
-      await Promise.all(folderItems.map(async (folderItem) => {
+      await Async.forEachAsync(folderItems, async (folderItem) => {
         const itemName = folderItem.name;
         const relativeItemPath = relativeFolderPath ? `${relativeFolderPath}/${itemName}` : itemName;
 
@@ -76,7 +76,7 @@ module.exports = {
             return FileSystem.writeFileAsync(outDtsPath, buffer, { ensureFolderExists: true });
           }
         }
-      }));
+      });
     }
 
     await searchAsync(undefined);
--- a/libraries/rush-lib/scripts/copyEmptyModules.js
+++ b/libraries/rush-lib/scripts/copyEmptyModules.js
@@ -55,7 +55,7 @@ module.exports = {
         const relativeItemPath = relativeFolderPath ? `${relativeFolderPath}/${itemName}` : itemName;
 
         if (folderItem.isDirectory()) {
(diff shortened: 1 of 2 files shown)


libraries/rush-lib/scripts/copyEmptyModules.js
```diff
@@ -50,12 +50,12 @@ module.exports = {
       const folderItems = await FileSystem.readFolderItemsAsync(
         relativeFolderPath ? `${jsInFolderPath}/${relativeFolderPath}` : jsInFolderPath
       );
-      for (const folderItem of folderItems) {
+      await Promise.all(folderItems.map(async (folderItem) => {
```
D4N14L, line 53: We should probably use `Async.forEachAsync(...)` for this instead of Promise.all. Then we can limit the concurrency of the search (normally I believe we limit file ops to 10)

D4N14L (changes requested):

libraries/rush-lib/scripts/copyEmptyModules.js
sanjaiyan-dev: > We should probably use `Async.forEachAsync(...)` for this instead of Promise.all. Then we can limit the concurrency of the search (normally I believe we limit file ops to 10)

Could you please clarify if I should simply replace `Promise.all` with `Async.forEachAsync(...)` in the code? It would help me understand your suggestion better. Thanks @D4N14L!

D4N14L: @sanjaiyan-dev For the portion of code you are touching, it performs file operations concurrently. Using `Async.forEachAsync` _instead of_ `Promise.all` will have a similar impact on performance, while allowing us to limit the amount of concurrency when performing the file operations.

sanjaiyan-dev: > @sanjaiyan-dev For the portion of code you are touching, it performs file operations concurrently. Using `Async.forEachAsync` _instead of_ `Promise.all` will have a similar impact on performance, while allowing us to limit the amount of concurrency when performing the file operations.

I understand the point you're making. Are the current code suggestions correct ?

```suggestion
      await Async.forEachAsync(folderItems.map(async (folderItem) => {
```

D4N14L: Not quite. Make the changes locally and you'll see the signature for Async.forEachAsync is slightly different (first param is an iterator of the objects you want to iterate, the second parameter is an async method used to perform the work)

sanjaiyan-dev: > Not quite. Make the changes locally and you'll see the signature for Async.forEachAsync is slightly different (first param is an iterator of the objects you want to iterate, the second parameter is an async method used to perform the work)

Sure, I will update you regarding this ASAP :)

sanjaiyan-dev: @D4N14L Is this correct now ? I hope so.

libraries/rush-lib/scripts/copyEmptyModules.js
```diff
@@ -73,10 +73,10 @@ module.exports = {
             terminal.writeVerboseLine(`Copying ${inDtsPath} to ${outDtsPath}`);
             // We know this is a file, don't need the redundant checks in FileSystem.copyFileAsync
             const buffer = await FileSystem.readFileToBufferAsync(inDtsPath);
-            await FileSystem.writeFileAsync(outDtsPath, buffer, { ensureFolderExists: true });
+            return FileSystem.writeFileAsync(outDtsPath, buffer, { ensureFolderExists: true });
```
D4N14L, line 76:
```suggestion
            await FileSystem.writeFileAsync(outDtsPath, buffer, { ensureFolderExists: true });
```

libraries/rush-lib/scripts/copyEmptyModules.js
```diff
@@ -50,12 +50,12 @@ module.exports = {
...
         relativeFolderPath ? `${jsInFolderPath}/${relativeFolderPath}` : jsInFolderPath
       );
-      for (const folderItem of folderItems) {
+      await Async.forEachAsync(folderItems, async (folderItem) => {
         const itemName = folderItem.name;
         const relativeItemPath = relativeFolderPath ? `${relativeFolderPath}/${itemName}` : itemName;
 
         if (folderItem.isDirectory()) {
-          await searchAsync(relativeItemPath);
+          return searchAsync(relativeItemPath);
```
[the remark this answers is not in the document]
D4N14L: Our repo generally likes to await the promises inline rather than return them from async functions because it provides a more informative callstack if it does fail.

D4N14L (changes requested):

libraries/rush-lib/scripts/copyEmptyModules.js
D4N14L: @sanjaiyan-dev almost! I added a couple comments and reasoning behind them. Once those are resolved this can be merged.

D4N14L (approved):

PR #4411, opened 2023-10-27
[rush-lib] Fix concurrency of rush-lib build script

D4N14L:
## Summary

Recent change #4401 added concurrent path traversal to the buildscript used by rush-lib. This however would allow for ballooning of the concurrency since it doesn't use the `AsyncQueue` class. This PR addresses this issue.

## How it was tested

Tested locally by building rush-lib and confirming output.

### DIFF
--- a/libraries/rush-lib/scripts/copyEmptyModules.js
+++ b/libraries/rush-lib/scripts/copyEmptyModules.js
@@ -46,6 +46,7 @@ module.exports = {
     const jsInFolderPath = `${buildFolderPath}/lib-esnext`;
     const dtsInFolderPath = `${buildFolderPath}/lib-commonjs`;
     const outFolderPath = `${buildFolderPath}/lib`;
+    const emptyModuleBuffer = Buffer.from('module.exports = {};', 'utf8');
     const folderPathQueue = new AsyncQueue([undefined]);
 
     await Async.forEachAsync(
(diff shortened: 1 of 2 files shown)


libraries/rush-lib/scripts/copyEmptyModules.js
```diff
@@ -46,39 +46,44 @@ module.exports = {
...
+          if (folderItem.isDirectory()) {
+            folderPathQueue.push(relativeItemPath);
+          } else if (folderItem.isFile() && itemName.endsWith(JS_FILE_EXTENSION)) {
+            const jsInPath = `${jsInFolderPath}/${relativeItemPath}`;
+            const jsFileText = await FileSystem.readFileAsync(jsInPath);
+            const strippedJsFileText = stripCommentsFromJsFile(jsFileText);
+            if (strippedJsFileText === 'export {};') {
+              const outJsPath = `${outFolderPath}/${relativeItemPath}`;
+              terminal.writeVerboseLine(`Writing stub to ${outJsPath}`);
+              await FileSystem.writeFileAsync(outJsPath, 'module.exports = {};', {
```
dmichon-msft, line 69: Minor nit, for performance we should store the result of `Buffer.from('module.exports = {};', 'utf8')` and reuse teh same buffer for all of these calls.

dmichon-msft (approved):

dmichon-msft (approved):

PR #4442, opened 2023-12-04
[rush] Subspace Configuration Files

william2958:
## Summary
This MR introduces the configuration files for the upcoming subspaces feature, who's RFC is located here: https://github.com/microsoft/rushstack/blob/octogonz/rfc-4320-rush-subspaces/common/docs/rfcs/rfc-4230-rush-subspaces.md

This MR sets up:
- The reading of the root level `subspace.json` configuration file for individual subspaces (As well as accompanying schema).
- Modifications to project configurations in `rush.json` to support subspace specification.
- Sets up the basics of reading subspace information in the rush-lib core library.

## How it was tested
Currently being manually tested

## Impacted documentation

- **subspaces.json** needs to be added to [rush_files_and_folders.md](https://rushjs.io/pages/advanced/rush_files_and_folders/) on the website


libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * This represents the JSON data structure for the "subspace.json" configuration file.
+ * See subspace.schema.json for documentation.
+ */
+export interface ISubspaceConfigurationJson {
+  $schema: string;
+  enabled: boolean;
+  depreciatedTTSupport?: boolean;
```
octogonz, line 20: Rather than referencing the company, `pr3481Compatibility` or `splitWorkspaceCompatibility` and prefix your `/** */` comment and `"description"` summary with `(DEPRECATED)`

libraries/rush-lib/src/schemas/subspace.schema.json
```diff
@@ -0,0 +1,31 @@
+{
```
octogonz, line 1: This file was called "subspaces.json" in the RFC - are we changing the name?

libraries/rush-lib/src/schemas/subspace.schema.json
```diff
@@ -0,0 +1,31 @@
...
+  "title": "Rush subspace config file.",
+  "description": "The configuration file for enabling the subspaces feature in rush.",
+  "type": "object",
+
+  "properties": {
+    "$schema": {
+      "description": "Part of the JSON Schema standard, this optional keyword declares the URL of the schema that the file conforms to. Editors may download the schema and use it to perform syntax highlighting.",
+      "type": "string"
+    },
+    "enabled": {
```
octogonz, line 12: For a casual observer (who upgraded Rush but has never heard of subspaces), we need to clearly indicate that:
- This feature is still "EXPERIMENTAL" 
- The implementation is incomplete and not yet usable
- Which setting is used to opt-in to the experiment

During code reviews, it will need to be easy for us to prove that the experimental algorithms will not be invoked unless the user has opted-in to the experiment.  (Because for any code that is reachable in the opted-out state, we would need to do careful testing to ensure no regressions are introduced.)

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,84 @@
...
+    };
+    this.subspaceJsonFile = subspaceJsonFilename;
+    this.availableSubspaceSet = new Set();
+
+    if (FileSystem.exists(this.subspaceJsonFile)) {
+      this.configuration = JsonFile.loadAndValidate(this.subspaceJsonFile, SubspaceConfiguration._jsonSchema);
+    }
+
+    for (const { subspaceName } of Object.values(this.configuration.availableSubspaces)) {
+      this.availableSubspaceSet.add(subspaceName);
```
octogonz, line 61: Report an error if the subspaceName doesn't conform to the usual `/^[a-z][a-z0-9]*([-][a-z0-9]+)*$/` pattern.

libraries/rush-lib/src/api/packageManager/PnpmPackageManager.ts
```diff
@@ -22,6 +22,14 @@ export class PnpmPackageManager extends PackageManager {
    */
   public readonly pnpmfileFilename: string;
 
+  /**
+   * The filename of the shrinkwrap file of a subspace that is used by the package manager
+   *
+   * @remarks
+   * Example: `.pnpmfile-<subspace_name>.cjs
+   */
+  public subspacePnpmfileFilename: (subspaceName: string) => string;
```
octogonz, line 31: Why is this a dynamic function?

The RFC said:
image

libraries/rush-lib/src/api/packageManager/PnpmPackageManager.ts
```diff
@@ -35,6 +43,10 @@ export class PnpmPackageManager extends PackageManager {
       this.pnpmfileFilename = RushConstants.pnpmfileV1Filename;
     }
 
+    this.subspacePnpmfileFilename = (subspaceName: string) => {
+      return `.pnpmfile-${subspaceName}.cjs`;
```
octogonz, line 47: The RFC said:
image

libraries/rush-lib/src/schemas/subspaces.schema.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    "enabled": {
+      "description": "If true, rush will use the subspaces configuration",
+      "type": "boolean"
+    },
+    "splitWorkspaceCompatibility": {
+      "description": "(DEPRECATED) A depreciated option that allows individual subspaces to be configured at the package level if that package is the only project in the subspace. Used to help migrate from a split workspace state.",
+      "type": "boolean"
+    },
+    "availableSubspaces": {
```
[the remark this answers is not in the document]
octogonz: I could imagine that we might want to configure additional properties for a subspace in the future, although each subspace already gets its own folder under `common/config/subspaces/<subspace-name>/` so we could easily add config files there.

The main purpose for listing them in **subspaces.json** was to avoid having to enumerate `common/config/subspaces/*`, which has a few benefits:

- The `CODEOWNERS` for **subspaces.json** can be used to intercept people trying to add new subspaces
- A subspace can be temporarily removed without having to delete all its files
- If a previous Git checkout leaves behind some old files in **common/config/subspaces**, Rush won't try to process them

libraries/rush-lib/assets/rush-init/subspaces.json
```diff
@@ -0,0 +1,30 @@
+/**
+ * This is the main configuration file for Rush.
+ * For full documentation, please see https://rushjs.io
+ */
+ {
+  /**
+   * Temporarily use the default schema until the subspaces schema is uploaded
+   */
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
octogonz, line 9: I think you can just put the real path name here. Rush doesn't process it; it is only used by VS Code IntelliSense underlining

william2958: do you mean a local link to the schema?

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: fixed

libraries/rush-lib/src/schemas/subspace.schema.json
william2958: changed

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: added

libraries/rush-lib/src/api/packageManager/PnpmPackageManager.ts
william2958: Removed the dynamic function as it's just going to use the same name as the regular pnpmfile

libraries/rush-lib/src/api/packageManager/PnpmPackageManager.ts
william2958: fixed

libraries/rush-lib/src/schemas/subspaces.schema.json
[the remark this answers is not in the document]
william2958: Changed to just use string array for the subspace names, I think it does make sense to keep any future subspace configuration files in the individual config folder

libraries/rush-lib/assets/rush-init/subspaces.json
octogonz: No, use this string:
```
  "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/subspaces.schema.json",
```

After this PR is merged, we will publish your **subspaces.schema.json** file to that location.

FYI the GitHub source is here:

https://github.com/microsoft/json-schemas/tree/main/rush/v5

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -880,6 +959,14 @@ export class RushConfiguration {
         );
       }
       this._projectsByName.set(project.packageName, project);
+      if (projectJson.subspace) {
+        const subspaceName: string = projectJson.subspace;
+        if (this._subspaceProjectsCache.has(subspaceName)) {
+          (this._subspaceProjectsCache.get(subspaceName) as RushConfigurationProject[]).push(project);
+        } else {
+          this._subspaceProjectsCache.set(subspaceName, [project]);
+        }
```
dmichon-msft, line 968:
```suggestion
        const projectsForSubspace: RushConfigurationProject[] | undefined = this._subspaceProjectsCache.get(subspaceName);
        if (projectsForSubspace) {
          projectsForSubspace.push(project);
        } else {
          this._subspaceProjectsCache.set(subspaceName, [project]);
        }
```
Avoid the unnecessary double lookup in the success path.

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -1188,6 +1276,16 @@ export class RushConfiguration {
...
 
+  /**
+   * A list of all the available subspaces in this workspace.
+   */
+  public get subspaceNames(): string[] {
+    if (!this._projects) {
+      this._initializeAndValidateLocalProjects();
+    }
+    return Array.from(this._subspaceProjectsCache.keys());
+  }
```
dmichon-msft, line 1287:
```suggestion
  public get subspaceNames(): Iterable<string> {
    if (!this._projects) {
      this._initializeAndValidateLocalProjects();
    }
    return this._subspaceProjectsCache.keys();
  }
```

Unless there's a strong reason to force the type into an array and copy the object upon using this method?

libraries/rush-lib/assets/rush-init/subspaces.json
william2958: Changed

common/reviews/api/rush-lib.api.md
```diff
@@ -1300,6 +1308,21 @@ export class RushUserConfiguration {
...
+    // Warning: (ae-forgotten-export) The symbol "ISubspaceConfigurationJson" needs to be exported by the entry point index.d.ts
+    //
+    // @internal
+    readonly configuration: Readonly<ISubspaceConfigurationJson>;
+    // (undocumented)
+    static loadFromConfigurationFile(subspaceJsonFilename: string): SubspaceConfiguration;
+    // (undocumented)
+    static loadFromDefaultLocation(): SubspaceConfiguration | undefined;
+    readonly subspaceJsonFile: string;
+    static _subspaceRegex: RegExp;
```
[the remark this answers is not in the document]
william2958: Moved to constant

common/reviews/api/rush-lib.api.md
```diff
@@ -1300,6 +1308,21 @@ export class RushUserConfiguration {
     static initializeAsync(): Promise<RushUserConfiguration>;
 }
 
+// @beta
+export class SubspaceConfiguration {
+    readonly availableSubspaceSet: Set<string>;
+    // Warning: (ae-forgotten-export) The symbol "ISubspaceConfigurationJson" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
william2958: Fixed!

common/reviews/api/rush-lib.api.md
```diff
@@ -734,6 +734,18 @@ export interface IStopwatchResult {
     toString(): string;
 }
 
+// @beta
+export interface ISubspaceConfigurationJson {
+    // Warning: (ae-forgotten-export) The symbol "ISubspaceConfig" needs to be exported by the entry point index.d.ts
```
iclanton, line 739: Fix this

common/reviews/api/rush-lib.api.md
```diff
@@ -734,6 +734,18 @@ export interface IStopwatchResult {
...
+    //
+    // (undocumented)
+    availableSubspaces: ISubspaceConfig[];
```
iclanton, line 742: Available?

common/reviews/api/rush-lib.api.md
```diff
@@ -1300,6 +1320,18 @@ export class RushUserConfiguration {
     static initializeAsync(): Promise<RushUserConfiguration>;
 }
 
+// @beta
+export class SubspaceConfiguration {
+    readonly availableSubspaceSet: Set<string>;
+    // @internal
+    readonly configuration: Readonly<ISubspaceConfigurationJson>;
```
iclanton, line 1327: Is exposing the raw JSON file useful?

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -222,6 +233,11 @@ export class RushConfiguration {
   // Lazily loaded when the projectsByTag() getter is called.
   private _projectsByTag: ReadonlyMap<string, ReadonlySet<RushConfigurationProject>> | undefined;
 
+  // Cache subspace projects
+  private _subspaceProjectsCache: Map<string, RushConfigurationProject[]>;
```
iclanton, line 237:
```suggestion
  private _rushProjectsBySubspaceNameCache: Map<string, RushConfigurationProject[]>;
```

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -348,6 +372,31 @@ export class RushConfiguration {
...
+  /**
+   * The full path of the temporary shrinkwrap file for a specific subspace.
+   * This function takes the subspace name, and returns the full path for the subspace's shrinkwrap file.
+   * This function also consults the depreciated option to allow for shrinkwraps to be stored under a package folder.
+   * This shrinkwrap file is used during "rush install", and may be rewritten by the package manager during installation
+   * @remarks
+   * This property merely reports the filename, the file itself may not actually exist.
+   * example: `C:\MyRepo\common\<subspace_name>\pnpm-lock.yaml`
+   */
+  public readonly tempSubspaceShrinkwrapFileName: (subspaceName: string) => string;
```
iclanton, line 398:
```suggestion
  public readonly getTempSubspaceShrinkwrapFileName: (subspaceName: string) => string;
```

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -622,6 +673,15 @@ export class RushConfiguration {
 
     this.ensureConsistentVersions = !!rushConfigurationJson.ensureConsistentVersions;
 
+    // Check if we have a subspace configuration file
+    const subspaceConfigLocation: string = path.join(this.rushJsonFolder, 'subspaces.json');
+    if (FileSystem.exists(subspaceConfigLocation)) {
+      // Try getting a subspace configuration
+      this.subspaceConfiguration = SubspaceConfiguration.loadFromConfigurationFile(subspaceConfigLocation);
+    }
```
iclanton, line 681: Can you rename `loadFromConfigurationFile` to `tryLoadFromConfigurationFile` and just return undefined if the file doesn't exist?

iclanton: And ideally this load should be async.

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -703,6 +763,15 @@ export class RushConfiguration {
 
     this.shrinkwrapFilename = this.packageManagerWrapper.shrinkwrapFilename;
 
+    // From "pnpm-lock.yaml" --> "subspace-pnpm-lock.yaml"
+    this.subspaceShrinkwrapFilenames = (subspaceName: string): string => {
+      const shrinkwrapFilenameParsedPath: path.ParsedPath = path.parse(this.shrinkwrapFilename);
+      return path.join(
+        shrinkwrapFilenameParsedPath.dir,
+        `${subspaceName}-` + shrinkwrapFilenameParsedPath.name + shrinkwrapFilenameParsedPath.ext
+      );
```
iclanton, line 772:
```suggestion
      const lastSlashIndex: number = Math.max(this.shrinkwrapFilename.lastIndexOf('/'), this.shrinkwrapFilename.lastIndexOf('\\'));
      return `${this.shrinkwrapFilename.substring(0, lastSlashIndex)}/${subspaceName}-${this.shrinkwrapFilename.substring(lastSlashIndex + 1)}`;
```

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -714,6 +783,16 @@ export class RushConfiguration {
       )
     );
 
+    this.tempSubspaceShrinkwrapFileName = (subspaceName: string): string => {
+      // TODO: do subspace name validation here
+      const fullSubspacePath: string = path.join(
+        this.commonTempSubspaceFolderRoot,
+        subspaceName,
+        this.shrinkwrapFilename
+      );
```
iclanton, line 792: `path.join` is actually pretty expensive. It's safe to just concatenate with slashes.

```suggestion
      const fullSubspacePath: string = `${this.commonTempSubspaceFolderRoot}/${subspaceName}/${this.shrinkwrapFilename}`;
```

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -714,6 +783,16 @@ export class RushConfiguration {
       )
     );
 
+    this.tempSubspaceShrinkwrapFileName = (subspaceName: string): string => {
```
iclanton, line 786: Why is this function defined here instead of being declared as a normal member function?

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -1262,6 +1361,13 @@ export class RushConfiguration {
     return commonVersionsFilename;
   }
 
+  /**
+   * Does this project have subspaces
+   */
+  public get hasSubspaces(): boolean {
+    return this._subspaceProjectsCache.size > 0;
```
iclanton, line 1368: Will this always be hydrated? The "Cache" part of the name make this sound like it may or may not be fully initialized.

libraries/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -184,6 +185,11 @@ export class RushConfigurationProject {
    */
   public readonly tags: ReadonlySet<string>;
 
+  /**
+   * If this project is in a subspace, and which one
+   */
+  public readonly subspace?: string;
```
iclanton, line 191:
```suggestion
  public readonly subspaceName?: string;
```

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+
+import { FileSystem, JsonFile, JsonSchema } from '@rushstack/node-core-library';
+import schemaJson from '../schemas/subspaces.schema.json';
+import path from 'path';
+import { trueCasePathSync } from 'true-case-path';
+import { RushConfiguration } from './RushConfiguration';
```
iclanton, line 8:
```suggestion
import { FileSystem, JsonFile, JsonSchema } from '@rushstack/node-core-library';
import path from 'path';
import { trueCasePathSync } from 'true-case-path';

import { RushConfiguration } from './RushConfiguration';
import schemaJson from '../schemas/subspaces.schema.json';
```

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+export interface ISubspaceConfigurationJson {
+  enabled: boolean;
+  splitWorkspaceCompatibility?: boolean;
+  availableSubspaces: ISubspaceConfig[];
+}
+
+/**
+ * The allowed naming convention for subspace names
+ */
+export const SubspaceRegex: RegExp = new RegExp('/^[a-z][a-z0-9]*([-][a-z0-9]+)*$/');
```
iclanton, line 28:
```suggestion
export const SUBSPACE_NAME_REGEXP: RegExp = new RegExp('/^[a-z][a-z0-9]*([-][a-z0-9]+)*$/');
```

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+ * @beta
+ */
+export interface ISubspaceConfigurationJson {
+  enabled: boolean;
+  splitWorkspaceCompatibility?: boolean;
+  availableSubspaces: ISubspaceConfig[];
+}
+
+/**
+ * The allowed naming convention for subspace names
```
iclanton, line 26: Document what this matches.

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+   * Gets the JSON data structure for the "subspaces.json" configuration file.
+   *
+   * @internal
+   */
+  public readonly configuration: Readonly<ISubspaceConfigurationJson>;
+
+  /**
+   * A set of the available subspaces
+   */
+  public readonly availableSubspaceSet: Set<string>;
```
iclanton, line 53:
```suggestion
  public readonly subspaceNames: Set<string>;
```

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.configuration = {
+      enabled: false,
+      availableSubspaces: []
+    };
+    this.subspaceJsonFile = subspaceJsonFilename;
+    this.availableSubspaceSet = new Set();
+
+    if (FileSystem.exists(this.subspaceJsonFile)) {
+      this.configuration = JsonFile.loadAndValidate(this.subspaceJsonFile, SubspaceConfiguration._jsonSchema);
+    }
```
iclanton, line 65:
```suggestion
    try {
      this.configuration = JsonFile.loadAndValidate(this.subspaceJsonFile, SubspaceConfiguration._jsonSchema);
    } catch (e) {
      if (!FileSystem.isNotExistError(e)) {
        throw e;
     }
   }
```

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+      } else {
+        throw new Error(
+          `Invalid subspace name: ${subspaceName}. Subspace names must only consist of lowercase letters, numbers, and hyphens (-).`
+        );
+      }
+    }
+  }
+
+  public static loadFromConfigurationFile(subspaceJsonFilename: string): SubspaceConfiguration {
+    let resolvedSubspaceJsonFilename: string = path.resolve(subspaceJsonFilename);
```
iclanton, line 79: When would this path not be resolved?

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    try {
+      resolvedSubspaceJsonFilename = trueCasePathSync(resolvedSubspaceJsonFilename);
```
iclanton, line 82: Doesn't this get initialized from the already-case-corrected repo root path?

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+    } catch (error) {
+      /* ignore errors from true-case-path */
+    }
+
+    return new SubspaceConfiguration(resolvedSubspaceJsonFilename);
+  }
+
+  public static loadFromDefaultLocation(): SubspaceConfiguration | undefined {
+    const rushJsonLocation: string | undefined = RushConfiguration.tryFindRushJsonLocation();
```
iclanton, line 91: Instead of doing this, have this function take a `RushConfiguration` as an argument.

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (rushJsonLocation) {
+      const subspaceJsonLocation: string = path.join(path.dirname(rushJsonLocation), 'subspaces.json');
```
iclanton, line 93: The `subspaces.json` sits in the repo root? Not under `common/config/rush`? That doesn't feel right.

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,97 @@
...
+        this.availableSubspaceSet.add(subspaceName);
+      } else {
+        throw new Error(
+          `Invalid subspace name: ${subspaceName}. Subspace names must only consist of lowercase letters, numbers, and hyphens (-).`
+        );
+      }
+    }
+  }
+
+  public static loadFromConfigurationFile(subspaceJsonFilename: string): SubspaceConfiguration {
```
iclanton, line 78: Ideally all of the things that touch the filesystem should be async.

libraries/rush-lib/src/api/SubspaceConfiguration.ts
iclanton, line 93: Filenames should be defined in https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/src/logic/RushConstants.ts

libraries/rush-lib/src/cli/actions/InitAction.ts
```diff
@@ -178,6 +185,12 @@ export class InitAction extends BaseConfiglessRushAction {
       'rush.json'
     ];
 
+    const experimentalTemplateFilePaths: string[] = ['subspaces.json'];
```
iclanton, line 188: Put this filename in the constants file and grab it from there.

libraries/rush-lib/src/schemas/rush.schema.json
```diff
@@ -307,6 +307,10 @@
               "type": "string",
               "pattern": "^[a-z0-9.@]+([-/][a-z0-9.@]+)*$"
             }
+          },
+          "subspace": {
+            "description": "An optional entry for specifying which subspace this project belongs to if the subspaces feature is enabled.",
```
iclanton, line 312:
```suggestion
            "description": "(EXPERIMENTAL) An optional entry for specifying which subspace this project belongs to if the subspaces feature is enabled.",
```

libraries/rush-lib/src/schemas/subspaces.schema.json
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/subspaces.schema.json",
```
iclanton, line 2:
```suggestion
  "$schema": "http://json-schema.org/draft-04/schema#",
```

libraries/rush-lib/src/schemas/subspaces.schema.json
```diff
@@ -0,0 +1,28 @@
...
+    "$schema": {
+      "description": "Part of the JSON Schema standard, this optional keyword declares the URL of the schema that the file conforms to. Editors may download the schema and use it to perform syntax highlighting.",
+      "type": "string"
+    },
+    "enabled": {
+      "description": "If true, rush will use the subspaces configuration.",
+      "type": "boolean"
+    },
+    "splitWorkspaceCompatibility": {
+      "description": "(DEPRECATED) A depreciated option that allows individual subspaces to be configured at the package level if that package is the only project in the subspace. Used to help migrate from a split workspace state.",
```
iclanton, line 17: We're introducing this with a deprecated feature? That doesn't seem right.

iclanton (changes requested):

libraries/rush-lib/assets/rush-init/subspaces.json
```diff
@@ -0,0 +1,30 @@
...
+
+  /**
+   * This specifies whether or not to use the "subspaces" feature in rush. This allows grouping of
+   * projects into discrete workspaces known as "subspaces".
+   */
+  "enabled": false,
```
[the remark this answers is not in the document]
octogonz: Yes, if they are optional in the JSON schema.

But I think `/*[LINE "HYPOTHETICAL"]*/` maybe should NOT be applied to the feature-gating setting (`"useSubspaces"`? `"enabled"`?) even if it is optional, since we want a casual observer to see that it is `false` without wondering what is the default value.

common/reviews/api/rush-lib.api.md
octogonz: In the RFC we called it `"subspaceNames"`:
```js
{
  "useSubspaces": true,

  // Names must be lowercase and separated by dashes.
  // To avoid mistakes, common/config/subspaces/ subfolder
  // cannot be used unless its name appears in this array.
  "subspaceNames": [ "default", "react19", "install-test" ]
}
```

I'm not sure "available" is even correct -- adding a string to this list will probably cause Rush to read and validate its config files whether or not any projects are actually using that subspace.

common/reviews/api/rush-lib.api.md
octogonz: +1 Let's avoid that unless TikTok specifically needs it for something

libraries/rush-lib/src/api/RushConfiguration.ts
octogonz: @iclanton our naming convention is `<name of map value>By<name of map key>`.  So if we want to give more adjectives about the value, it should go on the front, like `_cachedRushProjectsBySubspaceName`.

@D4N14L @william2958

libraries/rush-lib/src/api/RushConfigurationProject.ts
octogonz: Also  `public readonly subspaceName: string | undefined;` seems like a more natural notation.  

Practically they are equivalent, but conceptually this example is "possibly entirely missing":

```ts
function makeReport(options: { verbose?: string});
```

...whereas this example is "always present but sometimes assigned to `undefined`":
```ts
class ReportConfig {
  public verbose: string | undefined;
  
  public constructor() { this.verbose = undefined; }
}
```

libraries/rush-lib/src/api/SubspaceConfiguration.ts
octogonz: According to the RFC it is **common/config/rush/subspaces.json**

common/reviews/api/rush-lib.api.md
octogonz: @william2958 Could you update your PR description to list any design changes you made that are different from the RFC doc?

common/changes/@microsoft/rush/will-intro-subspace_2023-12-04-16-28.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Add subspace configuration file structure",
```
octogonz, line 5:
```suggestion
      "comment": "(EXPERIMENTAL) Implemented config file loader for the not-yet-released \"subspaces\" feature (GitHub #4230)",
```

common/reviews/api/rush-lib.api.md
william2958: fixed

common/reviews/api/rush-lib.api.md
william2958: Changed to match the RFC spec

common/reviews/api/rush-lib.api.md
william2958: Changed configuration to a private property

libraries/rush-lib/src/api/RushConfiguration.ts
william2958: changed to _cachedRushProjectsBySubspaceName

libraries/rush-lib/src/api/RushConfiguration.ts
william2958: Changed to tryLoadFromConfigurationFile

libraries/rush-lib/src/api/RushConfiguration.ts
william2958: Changed to a normal member function

libraries/rush-lib/src/api/RushConfigurationProject.ts
william2958: Changed to `public readonly subspaceName: string | undefined;`

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: changed

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: changed

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: that makes sense, removed the conversion with trueCasePathSync

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: changed

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: I guess this is not necessary to use path here, so I removed it

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: This function only calls FileSystem.exists, which is a synchronous function - does it still need to be labelled async?

libraries/rush-lib/src/cli/actions/InitAction.ts
william2958: Changed

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: Added

libraries/rush-lib/src/cli/actions/InitAction.ts
```diff
@@ -185,7 +186,7 @@ export class InitAction extends BaseConfiglessRushAction {
       'rush.json'
     ];
 
-    const experimentalTemplateFilePaths: string[] = ['subspaces.json'];
+    const experimentalTemplateFilePaths: string[] = [`common/config/rush/${RushConstants.subspacesFilename}`];
```
iclanton, line 189: Feels like the `common/config/rush` part should also be in a const (or consts?) somewhere.

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -1365,6 +1363,9 @@ export class RushConfiguration {
    * Does this project have subspaces
    */
   public get hasSubspaces(): boolean {
+    if (!this._projects) {
```
iclanton, line 1366: The `_cachedRushProjectsBySubspaceName` name is still confusing. A "cache" isn't generally guaranteed to be a complete dataset. Caches are used for improving efficiency by deduplicating work. If this is guaranteed to be complete after initialization has completed, it probably shouldn't be called a "cache."

libraries/rush-lib/src/api/RushConfiguration.ts
iclanton: Can this be made async?

octogonz: (If the other similar Rush functions are not already `async`, then we shouldn't insist that Will fixes it in this PR.)

libraries/rush-lib/src/cli/actions/InitAction.ts
octogonz: That string appears throughout this file. Seems like it's probably out of scope for this PR.

libraries/rush-lib/src/api/SubspaceConfiguration.ts
octogonz: @iclanton `RushConfiguration.tryLoadFromDefaultLocation()` itself is synchronous.  Synchronous functions cannot call `async` functions.  Therefore I think we may be asking @william2958 to do nontrivial work that is out of scope for this PR.

libraries/rush-lib/src/api/RushConfiguration.ts
william2958: Renamed to _rushProjectsBySubspaceName

libraries/rush-lib/src/api/RushConfiguration.ts
iclanton: There is a `FileSystem.existsAsync`, and there is a `JsonFile.loadAndValidateAsync`. Anything that touches the filesystem (including JSON file loading) should ideally be async to allow those calls to be parallelized.

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -287,6 +301,14 @@ export class RushConfiguration {
    */
   public readonly commonTempFolder: string;
 
+  /**
+   * The folder where temporary files will be stored for subspaces. The specific folder will
+   * append the subspace name to this path.
+   *
+   * Example: `C:\MyRepo\common\temp\<subspace_name>`
```
iclanton, line 308: When this is initialized on line 647, it looks like it's set to just `<repo root>/common/temp`.

Is this property useful to have around, since it appears to actually just be the same as `commonTempFolder`?

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -703,6 +752,18 @@ export class RushConfiguration {
...
+    this.subspaceShrinkwrapFilenames = (subspaceName: string): string => {
+      const lastSlashIndex: number = Math.max(
+        this.shrinkwrapFilename.lastIndexOf('/'),
+        this.shrinkwrapFilename.lastIndexOf('\\')
+      );
+      return `${this.shrinkwrapFilename.substring(
+        0,
+        lastSlashIndex
+      )}/${subspaceName}-${this.shrinkwrapFilename.substring(lastSlashIndex + 1)}`;
+    };
```
iclanton, line 765: Define this as a normal member function?

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,105 @@
...
+  ): SubspaceConfiguration | undefined {
+    const rushJsonLocation: string = rushConfiguration.rushJsonFolder;
+    if (rushJsonLocation) {
+      const subspaceJsonLocation: string = path.join(
+        path.dirname(rushJsonLocation),
+        'common/config/rush',
+        RushConstants.subspacesFilename
+      );
+      return SubspaceConfiguration.tryLoadFromConfigurationFile(subspaceJsonLocation);
+    }
```
iclanton, line 103:
```suggestion
    const commonRushConfigFolder: string = rushConfiguration.commonRushConfigFolder;
    if (commonRushConfigFolder) {
      const subspaceJsonLocation: string = `${commonRushConfigFolder}/${RushConstants.subspacesFilename}`
      return SubspaceConfiguration.tryLoadFromConfigurationFile(subspaceJsonLocation);
    }
```

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,105 @@
...
+  public static tryLoadFromConfigurationFile(
+    subspaceJsonFilename: string
+  ): SubspaceConfiguration | undefined {
+    if (FileSystem.exists(subspaceJsonFilename)) {
+      return undefined;
+    }
+    return new SubspaceConfiguration(subspaceJsonFilename);
+  }
+
+  public static loadFromDefaultLocation(
```
iclanton, line 92:
```suggestion
  public static tryLoadFromDefaultLocation(
```

libraries/rush-lib/src/api/RushConfiguration.ts
```diff
@@ -622,6 +660,17 @@ export class RushConfiguration {
 
     this.ensureConsistentVersions = !!rushConfigurationJson.ensureConsistentVersions;
 
+    // Check if we have a subspace configuration file
+    const subspaceConfigLocation: string = path.join(
```
iclanton, line 664: This path is already generated inside `SubspaceConfiguration.loadFromDefaultLocation`. Just call that with `this` as an argument instead of generating this path again.

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  public static tryLoadFromConfigurationFile(
+    subspaceJsonFilename: string
+  ): SubspaceConfiguration | undefined {
+    if (FileSystem.exists(subspaceJsonFilename)) {
+      return undefined;
+    }
+    return new SubspaceConfiguration(subspaceJsonFilename);
+  }
```
iclanton, line 90: This will avoid one filesystem access:

```suggestion
  private constructor(configuration: Readonly<ISubspaceConfigurationJson>, subspaceJsonFilePath: string) {
    this._configuration = configuration;
    this.subspaceJsonFile = subspaceJsonFilePath;
    this.subspaceNames = new Set();

    for (const subspaceName of configuration.subspaceNames) {
      if (SUBSPACE_NAME_REGEXP.test(subspaceName)) {
        this.subspaceNames.add(subspaceName);
      } else {
        throw new Error(
          `Invalid subspace name: ${subspaceName}. Subspace names must only consist of lowercase letters, numbers, and hyphens (-).`
        );
      }
    }
  }

  public static tryLoadFromConfigurationFile(
    subspaceJsonFilePath: string
  ): SubspaceConfiguration | undefined {
    let configuration: Readonly<ISubspaceConfigurationJson> | undefined;
    try {
      configuration = JsonFile.loadAndValidate(
        subspaceJsonFilePath,
        SubspaceConfiguration._jsonSchema
      );
    } catch (e) {
      if (!FileSystem.isNotExistError(e)) {
        throw e;
      }
    }

    if (configuration) {
      return new SubspaceConfiguration(configuration, subspaceJsonFilePath);
    }
  }
```

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,105 @@
...
+ * configuration file.
+ * @beta
+ */
+export class SubspaceConfiguration {
+  private static _jsonSchema: JsonSchema = JsonSchema.fromLoadedObject(schemaJson);
+
+  /**
+   * The absolute path to the "subspaces.json" configuration file that was loaded to construct this object.
+   */
+  public readonly subspaceJsonFile: string;
```
iclanton, line 39:
```suggestion
  public readonly subspaceJsonFilePath: string;
```

libraries/rush-lib/src/index.ts
```diff
@@ -10,6 +10,8 @@ export { ApprovedPackagesPolicy } from './api/ApprovedPackagesPolicy';
 
 export { RushConfiguration, ITryFindRushJsonLocationOptions } from './api/RushConfiguration';
 
+export { SubspaceConfiguration, ISubspaceConfigurationJson } from './api/SubspaceConfiguration';
```
iclanton, line 13:
```suggestion
export { SubspaceConfiguration } from './api/SubspaceConfiguration';
```

libraries/rush-lib/src/api/SubspaceConfiguration.ts
```diff
@@ -0,0 +1,105 @@
...
+import path from 'path';
+
+import type { RushConfiguration } from './RushConfiguration';
+import schemaJson from '../schemas/subspaces.schema.json';
+import { RushConstants } from '../logic/RushConstants';
+
+/**
+ * This represents the JSON data structure for the "subspaces.json" configuration file.
+ * See subspace.schema.json for documentation.
+ * @beta
```
iclanton, line 14:
```suggestion
```

libraries/rush-lib/src/logic/RushConstants.ts
```diff
@@ -178,6 +178,11 @@ export class RushConstants {
    */
   public static readonly artifactoryFilename: string = 'artifactory.json';
 
+  /**
+   * The subspaces.json configuration file name
+   */
+  public static readonly subspacesFilename: string = 'subspaces.json';
```
iclanton, line 184:
```suggestion
  public static readonly subspacesConfigFilename: string = 'subspaces.json';
```

libraries/rush-lib/src/api/RushConfiguration.ts
william2958: I guess this is not necessary right now, so removed

common/reviews/api/rush-lib.api.md
```diff
@@ -734,6 +734,16 @@ export interface IStopwatchResult {
     toString(): string;
 }
 
+// @beta
+export interface ISubspaceConfigurationJson {
```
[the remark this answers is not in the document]
william2958: Removed

libraries/rush-lib/src/api/RushConfiguration.ts
william2958: changed

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: changed

libraries/rush-lib/src/api/SubspaceConfiguration.ts
william2958: changed

octogonz (approved):

PR #4443, opened 2023-12-05
[node-core-library] Add ability to list processes on Windows and Unix-based platforms

D4N14L:
## Summary

Add the ability to list out processes and their child process tree for Windows and Unix-based systems.

## Details

While there exists a few different tools that can accomplish similar tasks (ex. `ps-tree`, `ps-list`), these tools are missing some functionality. For example, `ps-tree` is able to return the child tree of a specific process ID, but is unable to list out all processes. `ps-list` can list out all processes, but does not build a child tree. Additionally, `ps-list` is not bundlable due to it's dependency on the `fastlist` binary on Windows.

This PR brings all this functionality under the same method.

## How it was tested

- Tested on Windows and Ubuntu
- Added tests


libraries/node-core-library/src/Executable.ts
```diff
@@ -113,11 +114,221 @@ interface IExecutableContext {
...
+      if (line.length) {
+        yield line;
+      }
+      remaining = remaining.substring(index + match[0].length);
+    }
+  }
+  if (remaining.length) {
+    yield remaining;
+  }
+}
```
dmichon-msft, line 204: These are generic. Put them in a different file or files.

libraries/node-core-library/src/Executable.ts
```diff
@@ -113,11 +114,221 @@ interface IExecutableContext {
...
+// of them.
+const NEWLINE_REGEX: RegExp = /\r*\n/;
+
+async function* readLinesFromStreamAsync(stream: NodeJS.ReadableStream): AsyncGenerator<string> {
+  let remaining: string = '';
+  for await (const chunk of stream) {
+    remaining += chunk;
+    // eslint-disable-next-line @rushstack/no-new-null
+    let match: RegExpMatchArray | null;
+    while ((match = remaining.match(NEWLINE_REGEX)) && match.index !== undefined && match.index >= 0) {
```
dmichon-msft, line 167: I think this will be simpler if you just use indexOf for `\n` and then trim off all `/[\r\n]/` from returned lines.

libraries/node-core-library/src/Executable.ts
```diff
@@ -113,11 +114,221 @@ interface IExecutableContext {
...
+      remaining = remaining.substring(index + match[0].length);
+    }
+  }
+  if (remaining.length) {
+    yield remaining;
+  }
+}
+
+// eslint-disable-next-line @rushstack/no-new-null
+function* readLinesFromStringArray(text: (string | null)[]): Generator<string> {
```
dmichon-msft, line 182:
```suggestion
function* readLinesFromStringArray(text: Iterable<string | null>): Generator<string> {
```

Since you're only using the iterator, best to maximize applicability of the API.

libraries/node-core-library/src/Executable.ts
```diff
@@ -113,11 +114,221 @@ interface IExecutableContext {
...
+  let remaining: string = '';
+  let index: number;
+  for (const chunk of text) {
+    if (chunk === null) {
+      continue;
+    }
```
dmichon-msft, line 188:
```suggestion
    if (!chunk) {
      continue;
    }
```
An empty string doesn't alter the state either.

libraries/node-core-library/src/Executable.ts
```diff
@@ -113,11 +114,221 @@ interface IExecutableContext {
...
+   */
+  childProcessInfos: IProcessInfo[];
+}
+
+// Match the newline character(s) at the end of a line of text in order to split on it. On Windows,
+// it is possible for multiple \r characters may precede the \n character, so we need to match all
+// of them.
+const NEWLINE_REGEX: RegExp = /\r*\n/;
+
+async function* readLinesFromStreamAsync(stream: NodeJS.ReadableStream): AsyncGenerator<string> {
```
dmichon-msft, line 161:
```suggestion
async function* readLinesFromStreamAsync(stream: AsyncIterable<string>): AsyncGenerator<string> {
```
Since you're only using the `for await` code path.

libraries/node-core-library/src/Executable.ts
```diff
@@ -113,11 +114,221 @@ interface IExecutableContext {
...
+      remaining = remaining.substring(index + match[0].length);
+    }
+  }
+  if (remaining.length) {
+    yield remaining;
+  }
+}
+
+export async function parseProcessListOutputAsync(
+  stream: NodeJS.ReadableStream
```
dmichon-msft, line 207:
```suggestion
  stream: AsyncIterable<string>
```

libraries/node-core-library/src/Executable.ts
```diff
@@ -113,11 +114,221 @@ interface IExecutableContext {
...
+      seenHeaders = true;
+    } else {
+      parseProcessInfoEntry(line, processInfoById);
+    }
+  }
+  return processInfoById;
+}
+
+// eslint-disable-next-line @rushstack/no-new-null
+export function parseProcessListOutput(output: (string | null)[]): Map<number, IProcessInfo> {
```
dmichon-msft, line 222:
```suggestion
export function parseProcessListOutput(output: Iterable<string | null>): Map<number, IProcessInfo> {
```

libraries/node-core-library/src/test/Executable.test.ts
```diff
@@ -5,217 +5,301 @@ import * as os from 'os';
...
-    // %^&|<>  newline
-    '~!@#$*()_+`={}[]:";\'?,./',
-    '~!@#$*()_+`={}[]:";\'?,./'
+  const UNIX_PROCESS_LIST_OUTPUT: string[] = [
+    'COMMAND                PPID   PID\n',
+    // Test that the parser can handle referencing a parent that doesn't exist
+    'init                      0     1\n',
+    // Test that the parser can handle a line that is truncated in the middle of a field
+    // Test that the parser can handle an entry referencing a parent that hasn't been seen yet
+    'process2           ' + '       2     4\n',
```
dmichon-msft, line 256:
```suggestion
    'process2           ', '       2     4\n',
```

dmichon-msft: Also throw some empty strings and/or nulls in there.

libraries/node-core-library/src/Executable.ts
D4N14L: Moved to `Text` class and it makes more sense to use the included regexes there. However, I'm now using `matches` to avoid multiple calls to the regex.

libraries/node-core-library/src/Text.ts
```diff
@@ -178,4 +200,72 @@ export class Text {
...
+      yield remaining;
+    }
+  }
+
+  private static *_readLinesFromChunk(
+    chunk: string | Buffer | null,
+    encoding: Encoding,
+    skipEmptyLines: boolean,
+    state: IReadLinesFromIterableState
+  ): Generator<string> {
```
iclanton, line 254: Make this a non-member function? That'll improve webpack optimization a bit.

libraries/node-core-library/src/Text.ts
```diff
@@ -178,4 +200,72 @@ export class Text {
...
+    if (!chunk) {
+      return;
+    }
+    const remaining: string =
+      state.remaining + (typeof chunk === 'string' ? chunk : chunk.toString(encoding));
```
iclanton, line 259: Will this occasionally return a buffer?

D4N14L: The iterable ReadableStream is typed such that it _could_ return either a buffer or a string (it's `AsyncIterable<string | Buffer>`, not `AsyncIterable<string> | AsyncIterable<buffer>`, so I don't think we can depend on one or the other being returned. Given the main use case for the method is for handling streams, it's a frustrating check to have to make, but IMO it may be needed. At least it's rather low overhead.

libraries/node-core-library/src/Executable.ts
```diff
@@ -276,6 +437,89 @@ export class Executable {
...
+            // We've already rejected the promise
+            return;
+          }
+          if (code !== 0) {
+            reject(new InternalError(`Unable to list processes: ${command} exited with code ${code}`));
+          } else {
+            resolve();
+          }
+        });
+      }
```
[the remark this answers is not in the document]
dmichon-msft: `ChildProcess.spawn` will emit the `error` event if the process fails to spawn.

libraries/node-core-library/src/Text.ts
dmichon-msft: In practice, no, but TypeScript doesn't know that because `setEncoding` does not alter the declared type of the object.

libraries/node-core-library/src/Executable.ts
```diff
@@ -438,56 +498,125 @@ export class Executable {
...
+  ): Promise<IWaitForExitResult<never>>;
+
+  public static async waitForExitAsync<T extends Buffer | string | never = never>(
+    childProcess: child_process.ChildProcess,
+    options: IWaitForExitOptions = {}
+  ): Promise<IWaitForExitResult<T>> {
+    const { throwOnNonZeroExitCode = false, encoding } = options;
+    if (encoding && (!childProcess.stdout || !childProcess.stderr)) {
+      throw new Error(
+        'An encoding was specified, but stdout and/or stderr are not piped. Did you set stdio?'
```
iclanton, line 532: You may be able to check for `childProcess.stdio`.

libraries/node-core-library/src/Executable.ts
```diff
@@ -438,56 +498,125 @@ export class Executable {
...
+      );
+    }
+
+    const collectedStdout: T[] = [];
+    const collectedStderr: T[] = [];
+    const useBufferEncoding: boolean = encoding === 'buffer';
```
iclanton, line 538: You can't say this type is `encoding is 'buffer'`?

iclanton (approved):

Issue #4454, opened 2023-12-13
[eslint-patch] bulk suppressions patch fails to find eslintrc in git root


SpenserJ:
## Summary

When trying to use `@rushstack/eslint-patch/eslint-bulk-suppressions` in a package that is not a monorepo, the `.eslintrc.js` may be at the root of the git repository. Enabling the patch for bulk suppression will throw an error of `Cannot locate eslintrc`

## Repro steps

Reproduction: https://github.com/SpenserJ/rushstack/tree/repro-bulk-suppression-in-polyrepo

1. Create a new repository with a new package (not a workspace/monorepo)
2. Install ESLint and `@rushstack/eslint-patch`, and configure it with the bulk suppression patch and any ESLint config or preset you'd like
3. Create a file that will result in a linting error with your config
4. Run ESLint against that file

  **Expected result:** The linter should run and report any linting errors

  **Actual result:** `Error: Cannot locate eslintrc`

## Details

https://github.com/microsoft/rushstack/blob/d9e0c73c31a125df8b39871f73c4f6c5a706f638/eslint/eslint-patch/src/eslint-bulk-suppressions/bulk-suppressions-patch.ts#L76-L81

In a polyrepo, the eslintrc file will usually be in the root directory of the git repository, however `findEslintrcDirectory` will stop as soon as it has reached the `GitRootPath`. Changing this to check if we're outside/above the root path resolves this issue:
```diff
- for (let currentDir = fileAbsolutePath; currentDir !== GitRootPath; currentDir = path.dirname(currentDir)) 
+ for (let currentDir = fileAbsolutePath; currentDir.startsWith(GitRootPath); currentDir = path.dirname(currentDir)) 
```

## Standard questions

Please answer these questions to help us investigate your issue more quickly:

| Question | Answer |
| -------- | -------- |
| `@rushstack/eslint-config` version? | N/A |
| Operating system? | Mac |
| Would you consider contributing a PR? | [Yes](https://github.com/microsoft/rushstack/pull/4455) |
| TypeScript compiler version? | N/A |
| Node.js version (`node -v`)? | v20.10.0 |

PR #4455, opened 2023-12-13
fix(eslint-patch): fix bulk suppression eslintrc detection in polyrepo

SpenserJ:
## Summary

In a polyrepo, the eslintrc file will usually be in the root directory of the git repository, however `findEslintrcDirectory` will stop as soon as it has reached the `GitRootPath`. Changing this to check if we're outside/above the root path resolves this issue.

## Details

Please see #4454 for more details and a reproduction

## How it was tested

Tested against the reproduction case in #4454


SpenserJ:

kevin-y-ang:
I reviewed the bug, and it's just as you say, bulk suppressions exits out after searching for eslintrc in the root directory, so it doesn't account for eslintrc files in the root directory.

I ran the reproduced bug, and also ran the fixed code on the build-test files and it works as expected.

Thank you for the fix, SpenserJ!

octogonz (approved):

PR #4462, opened 2023-12-22
Add fallback option for formatting [un]localized assetpaths. 

TheLarkInn:
Currently the localization plugin doesn't have a way to provide a fallback folder path/options for non-localized assets. This PR introduces a fallback function which the user can provide to override unlocalized assets being placed in the `'/none'` folder.


webpack/webpack5-localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -231,9 +232,11 @@ export class LocalizationPlugin implements WebpackPluginInstance {
 
                   const noLocaleExpression: string = JSON.stringify(this._noStringsLocaleName);
 
-                  return `" + (${JSON.stringify(chunkMapping)}[chunkId]?${
-                    isLocalizedSmaller ? localeExpression : noLocaleExpression
-                  }:${isLocalizedSmaller ? noLocaleExpression : localeExpression}) + "`;
+                  return this._formatLocaleForFilename(
+                    `" + (${JSON.stringify(chunkMapping)}[chunkId]?${
```
TheLarkInn, line 236: @dmichon-msft think this is incorrect usage here. If I understand correctly is this the path getting output into the chunk referenced code? Not sure what this logic branch does.

webpack/webpack5-localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -207,9 +208,9 @@ export class LocalizationPlugin implements WebpackPluginInstance {
                   Constants.JSONP_PLACEHOLDER;
 
                 if (chunkIdsWithStrings.size === 0) {
-                  return this._noStringsLocaleName;
+                  return this._formatLocaleForFilename(this._noStringsLocaleName);
                 } else if (chunkIdsWithoutStrings.size === 0) {
-                  return `" + ${localeExpression} + "`;
+                  return this._formatLocaleForFilename(`" + ${localeExpression} + "`);
```
dmichon-msft, line 213: Don't call the function on this value. The input text is raw JavaScript that will be executed by the webpack runtime.

You will instead need to inject it inside of this call:
https://github.com/microsoft/rushstack/blob/main/webpack/webpack5-localization-plugin/src/AssetProcessor.ts#L290

webpack/webpack5-localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -231,9 +232,11 @@ export class LocalizationPlugin implements WebpackPluginInstance {
 
                   const noLocaleExpression: string = JSON.stringify(this._noStringsLocaleName);
```
dmichon-msft, line 233:
```suggestion
                  const noLocaleExpression: string = JSON.stringify(this._formatLocaleForFilename(this._noStringsLocaleName));
```

webpack/webpack5-localization-plugin/src/LocalizationPlugin.ts
```diff
@@ -700,6 +705,14 @@ export class LocalizationPlugin implements WebpackPluginInstance {
     }
     // END options.noStringsLocaleName
 
+    // START options.formatLocaleForFilename
+    const { formatLocaleForFilename } = this._options;
+    if (!formatLocaleForFilename) {
+      this._formatLocaleForFilename = (localeName: string) => localeName;
+    } else {
+      this._formatLocaleForFilename = formatLocaleForFilename;
+    }
```
dmichon-msft, line 714:
```suggestion
    const { formatLocaleForFilename = (localeName: string) => localeName } = this._options;
    this._formatLocaleForFilename = formatLocaleForFilename;
```

webpack/webpack5-localization-plugin/src/interfaces.ts
```diff
@@ -134,6 +134,13 @@ export interface ILocalizationPluginOptions {
    * Options for how localization stats data should be produced.
    */
   localizationStats?: ILocalizationStatsOptions;
+
+  /**
+   * Custom function for controlling how locale names are formatted based on the locale specified.
+   * This is useful if you want to emit non-localized files to the root output directory instead
+   * of a '/none' subdirectory.
```
dmichon-msft, line 141: If combining with `runtimeLocaleExpression`, ensure that the runtime output of `runtimeLocaleExpression` produces the same output as `formatLocaleForFilename`.

webpack/webpack5-localization-plugin/src/test/LocalizedAsyncDynamicFormatWithNoLocaleFallback.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    localizationStats: {
+      callback: statsCallback
+    },
+    formatLocaleForFilename: (locale: string) => {
+      if (locale === 'none') {
+        return '';
+      } else {
+        return locale;
+      }
+    }
```
dmichon-msft, line 61: The initial expectation of how to use this was to have `output.filename` be `[locale][name].js` and have this function return `''` for `'none'` and `${locale}/` for all other values. This would then entail `runtimeLocaleExpression` being `self.__locale + "/"`.

webpack/webpack5-localization-plugin/src/test/__snapshots__/LocalizedAsyncDynamicFormatWithNoLocaleFallback.test.ts.snap
```diff
@@ -0,0 +1,338 @@
...
...Scripts&&(e=o.g.location+\\"\\");var r=o.g.document;if(!e&&r&&(r.currentScript&&(e=r.currentScript.src),!e)){var t=r.getElementsByTagName(\\"script\\");if(t.length)for(var n=t.length-1;n>-1&&!e;)e=t[n--].src}if(!e)throw new Error(\\"Automatic publicPath is not supported in this browser\\");e=e.replace(/#.*$/,\\"\\").replace(/\\\\?.*$/,\\"\\").replace(/\\\\/[^\\\\/]+$/,\\"/\\"),o.p=e+\\"../\\"})(),(()=>{var e={179:0};o.f.j=(r,t)=>{var n=o.o(e,r)?e[r]:void 0;if(0!==n)if(n)t.push(n[2]);else{var i=new Promise(((t,o)=>n=e[r]=[t,o]));t.push(n[2]=i);var a=o.p+o.u(r),l=new Error;o.l(a,(t=>{if(o.o(e,r)&&(0!==(n=e[r])&&(e[r]=void 0),n)){var i=t&&(\\"load\\"===t.type?\\"missing\\":t.type),a=t&&t.target&&t.target.src;l.message=\\"Loading chunk \\"+r+\\" failed.\\\\n(\\"+i+\\": \\"+a+\\")\\",l.name=\\"ChunkLoadError\\",l.type=i,l.request=a,n[1](l)}}),\\"chunk-\\"+r,r)}};var r=(r,t)=>{var n,i,[a,l,c]=t,s=0;if(a.some((r=>0!==e[r]))){for(n in l)o.o(l,n)&&(o.m[n]=l[n]);c&&c(o)}for(r&&r(t);s<a.length;s++)i=a[s],o.o(e,i)&&e[i]&&e[i][0](),e[i]=0},t=self.webpackChunk=self.webpackChunk||[];t.forEach(r.bind(null,0)),t.push=r.bind(null,t.push.bind(t))})(),o.e(931).then(o.bind(o,\\"./a/async.js\\"))})();",
```
dmichon-msft, line 7: The virtual file system is emitting the files at the desired paths, but are the keys in webpack's compilation.assets correct? Or do they contain `//`? The latter could cause interoperability issues.

iclanton:
Can you fill in the description?

I thought the problem that you were trying to solve was putting non-localized assets in a different folder, or something like that.

webpack/webpack5-localization-plugin/src/test/LocalizedAsyncDynamicFormatWithNoLocaleFallback.test.ts
TheLarkInn: So does this mean that `runtimeLocaleExpression` must always be set when `formatLocaleForFilename` is also set?

webpack/webpack5-localization-plugin/src/test/__snapshots__/LocalizedAsyncDynamicFormatWithNoLocaleFallback.test.ts.snap
TheLarkInn: Included `expect(stats.compilation.assets).toMatchSnapshot("Assets")` in latest commit. It appears paths are correctly handled.

TheLarkInn:
> Can you fill in the description?
> 
> I thought the problem that you were trying to solve was putting non-localized assets in a different folder, or something like that.

Filled in, and yes that is technically what we're introducing here.

common/changes/@rushstack/webpack5-localization-plugin/thelarkinn-createfallbacklocformatter_2024-01-03-23-27.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/webpack5-localization-plugin",
+      "comment": "Introduce new formatLocaleForFilename feature for customizable output path for non localized files.",
```
dmichon-msft, line 5:
```suggestion
      "comment": "Introduce a `formatLocaleForFilename` option to customize how a locale (or the lack thereof) is rendered in file paths.",
```

dmichon-msft (approved):

PR #4476, opened 2024-01-09
[rush] Split ProjectChangeAnalyzer, fix build cache hashes

### LINKED ISSUE #4400: [rush] Build cache keys do not correctly account for configuration changes in dependencies
## Summary
The following options that may impact the cache key of a given operation incorrectly do not currently affect operations that depend on them:
- `dependsOnEnvVars`
- `dependsOnAdditionalFiles`
- Any CLI parameter that affects some phases but does not (directly) affect others

## Repro steps
Create a repository with two projects, `A`, and `B` as root folders in the repo.

Define two phases:
`_phase:first`. no dependencies.
`_phase:second`, dependency on `upstream: ['_phase:first']`.

Define `build` as a phased command with `phases: ['_phase:first', '_phase:second']`.

1)  Define a CLI flag parameter `--production` that is applicable to the command `build` and the phase `_phase:first`.

2) Configure `dependsOnEnvVars` in `A/config/rush-project.json` for `_phase:first` to include `TEST_VAR`.

3) Configure `dependsOnAdditionalFiles` in `A/config/rush-project.json` for `_phase:first` to include `../common/tmp.log`.

Observe that when running `rush build`, the cache key for `B (_phase:second)` is not affected by:
1) The presence or absence of the `--production` flag.
2) Any change to the environment variable `TEST_VAR`.
3) Any change to the file `tmp.log`

## Details
The root issue here is that the cache key for an operation is computed based on the project dependencies, not the operation dependencies. Cache key for an operation should be a function of:
1) The "configuration" of the operation, i.e. the npm script text, the CLI parameters, any configured environment variables, etc.
2) The "local state" of the operation, i.e. the content hashes of all Git tracked files and any files identified by the `dependsOnAdditionalFiles` property, sorted deterministically.
3) The final computed cache keys of all operations upon which the operation depends, sorted deterministically.

dmichon-msft:
## Summary
Fixes #4400

Makes #3994 more feasible.

## Details
Splits the functionality of ProjectChangeAnalyzer into two components:
1) `ProjectChangeAnalyzer#_tryGetSnapshotProviderAsync` - Performs all the async I/O operations to determine the current state of the repository, including taking a snapshot of `process.env` so that per-operation hashes can be lazily computed in a stable manner.
2) `InputsSnapshot` - Synchronous data structure representing the state of the repository, which can be queried for operation-level state hashes, or tracked file lists. Performs the evaluation of input/output collisions as part of evaluating said queries. Includes evaluation of `incrementalBuildIgnoredGlobs` and `dependsOnEnvVars`, but does not include the hashes from dependencies, since the Operation graph is not provided to this API.

Moves build cache hash computation to the `beforeExecuteOperations` tap in `CacheableOperationPlugin` and updates the computation to factor in the runtime operation graph, rather than only the npm dependencies.

For a specific example of how this alters the build cache, consider a repository with one project and two phases in `rush build`:
Project `A`
Phases `_phase:first` and `_phase:second`, with `_phase:second` depends on `self: ['_phase:first']`

Define a command line parameter `--some-flag-for-first` that is forwarded only to `_phase:first`.

Before this change, running `rush build` and `rush build --some-flag-for-first` would produce different cache IDs for `A (_phase:first)`, but the same cache id for `A (_phase:second)`.

With this change the cache IDs for `A (_phase:second)` are also impacted by the presence of the `--some-flag-for-first`, since it can change the output of `A (_phase:first)`.

Added a new flag `enabled` to `Operation` to control whether or not a given operation in the graph should be executed. Updated graph construction to use this flag for operations that are dependencies of the user's selection, but not directly included (e.g. when using `--only` or `--impacted-by`), or during watch-mode incremental builds.

Added a new property `cacheHashSalt` to `build-cache.json` to allow repository maintainers to globally force cache misses if they have altered the build process in a way that Rush can't currently detect, for example if Rush plugins emit files that would be picked up by the cache.

## How it was tested
For the splitting, added unit tests for `InputSnapshot` and `ProjectChangeAnalyzer#_tryGetSnapshotProviderAsync`.

For the build cache IDs, temporarily removed applicability of the `--production` CLI parameter from `_phase:test` and validated that running `node ./apps/rush/lib/start-dev.js test` and `node ./apps/rush/lib/start-dev.js test --production` still produced different cache keys for `_phase:test`.

For comparison, validated that the command-line.json alternation and existing `rush test` and `rush test --production` had a collision on the cache keys for `_phase:test`.

Validated cache reads/writes using the `heft-node-everything-test` project. Confirmed that cache is not written during a `rush rebuild -o`, but can be read during `rush build -o`. Confirmed that cache reads can occur during `rush start -o`, even on subsequent iterations that leverage the new `:incremental` suffixed scripts.

Validate the `cacheHashSalt` property by setting the salt to different values and validating cache/miss hit depending on whether or not a prior build had happened with the same salt.

## Impacted documentation
Any documentation about how build cache keys are computed.


libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -402,58 +315,178 @@ export class ProjectChangeAnalyzer {
...
-            if (!(await FileSystem.existsAsync(filePath))) {
-              throw new Error(
-                `A project dependency file (${filePath}) is missing. You may need to run ` +
-                  '"rush install" or "rush update".'
-              );
+    // Do not use getGitInfo().root; it is the root of the *primary* worktree, not the *current* one.
+    const rootDirectory: string = getRepoRoot(this._rushConfiguration.rushJsonFolder, gitPath);
+
+    // Load the rush-project.json files for the whole repository
+    const additionalGlobs: IAdditionalGlob[] = [];
```
D4N14L, line 341: Missing something here :)

libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -402,58 +315,178 @@ export class ProjectChangeAnalyzer {
...
+      // That being said, we want to keep `absolute` options here as false:
+      absolute: false
+    });
+
+    for (const match of matches) {
+      if (path.isAbsolute(match)) {
+        additionalFilesFromRushProjectConfiguration.add(match);
+        additionalFilesForOperation.add(match);
+      } else {
+        const filePathForGit: string = path.posix.join(project.projectRelativeFolder, match);
```
D4N14L, line 472: This will be from the root of the rush.json folder. Should verify if that is the correct path root for the git hash for files method.

libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -402,58 +315,178 @@ export class ProjectChangeAnalyzer {
...
+
+        const projectMatch: RushConfigurationProject | undefined = lookupByPath.findChildPath(filePathForGit);
+        if (projectMatch && projectMatch !== project) {
+          terminal.writeErrorLine(
+            `In project "${project.packageName}" ("${project.projectRelativeFolder}"), ` +
+              `config for operation "${operationName}" specifies a glob "${pattern}" that selects a file in a different workspace project ` +
+              `"${projectMatch.packageName}" ("${projectMatch.projectRelativeFolder}"). ` +
+              `This is forbidden. The "dependsOnAdditionalFiles" property may only be used to refer non-workspace files, non-project files, or untracked files in the current project. ` +
```
D4N14L, line 480: May want to specify which file these fields exist in.

libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -402,58 +315,178 @@ export class ProjectChangeAnalyzer {
...
+        additionalFilesFromRushProjectConfiguration.add(match);
+        additionalFilesForOperation.add(match);
+      } else {
+        const filePathForGit: string = path.posix.join(project.projectRelativeFolder, match);
+
+        const projectMatch: RushConfigurationProject | undefined = lookupByPath.findChildPath(filePathForGit);
+        if (projectMatch && projectMatch !== project) {
+          terminal.writeErrorLine(
+            `In project "${project.packageName}" ("${project.projectRelativeFolder}"), ` +
+              `config for operation "${operationName}" specifies a glob "${pattern}" that selects a file in a different workspace project ` +
```
D4N14L, line 478:
```suggestion
              `config for operation "${operationName}" specifies a glob "${pattern}" that selects a file <filename> in a different workspace project ` +
```

libraries/rush-lib/src/logic/operations/CacheableOperationPlugin.ts
```diff
@@ -90,81 +84,133 @@ export class CacheableOperationPlugin implements IPhasedCommandPlugin {
...
+      ): void => {
+        const { isIncrementalBuildAllowed, inputSnapshot, projectConfigurations, isInitial } = context;
+
+        if (!inputSnapshot) {
+          throw new Error(
+            `Build cache is only supported if running in a Git repository. Either disable the build cache or run Rush in a Git repository.`
+          );
+        }
+
+        const definitelyDefinedInputSnapshot: IInputSnapshot = inputSnapshot;
```
D4N14L, line 103: This is silly but I get why

libraries/rush-lib/src/logic/snapshots/InputSnapshot.ts
```diff
@@ -0,0 +1,437 @@
...
+
+    // Route hashes to individual projects
+    for (const [file, hash] of hashes) {
+      const project: IRushConfigurationProjectForSnapshot | undefined = lookupByPath.findChildPath(file);
+      if (!project) {
+        continue;
+      }
+
+      let record: IInternalRushSnapshotProjectMetadata | undefined = projectMetadataMap.get(project);
+      if (!record) {
```
D4N14L, line 222: DRY this up

libraries/rush-lib/src/logic/snapshots/InputSnapshot.ts
```diff
@@ -0,0 +1,437 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public getTrackedFileHashesForOperation(
+    project: IRushConfigurationProjectForSnapshot,
+    operationName?: string
+  ): ReadonlyMap<string, string> {
+    const record: IInternalRushSnapshotProjectMetadata | undefined = this._projectMetadataMap.get(project);
+    if (!record) {
+      throw new Error(`No information available for project at ${project.projectFolder}`);
```
D4N14L, line 281: InternalError?

libraries/rush-lib/src/logic/snapshots/InputSnapshot.ts
```diff
@@ -0,0 +1,437 @@
...
+      }
+
+      // Hash the base project files
+      for (const [filePath, hash] of record.hashes) {
+        if (filter(filePath)) {
+          hashes.set(filePath, hash);
+          const outputMatch: string | undefined = outputValidator?.findChildPath(filePath);
+          if (outputMatch) {
+            throw new Error(
+              `Configured output folder "${outputMatch}" for operation "${operationName}" in project "${project.projectRelativeFolder}" contains tracked input file "${filePath}".`
```
D4N14L, line 331: Maybe suggest what the user would need to modify/fix to address this?

libraries/rush-lib/src/logic/snapshots/InputSnapshot.ts
```diff
@@ -0,0 +1,437 @@
...
+
+      hash = hasher.digest('hex');
+
+      hashByOperationName.set(operationName, hash);
+    }
+
+    return hash;
+  }
+
+  private *_resolveHashes(filePaths: Iterable<string>): IterableIterator<[string, string]> {
```
D4N14L, line 395: Generator type?

D4N14L (approved):
We should do a pre-publish to verify that everything that is changing here doesn't break our flow, since we're very likely the largest customer on this tool.

libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -509,13 +512,18 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
...
 
     const phaseOriginal: Set<IPhase> = new Set(this._watchPhases);
     const phaseSelection: Set<IPhase> = new Set(this._watchPhases);
 
-    const { projectChangeAnalyzer: initialState, projectSelection: projectsToWatch } =
-      initialCreateOperationsContext;
+    const { inputSnapshot: initialState, projectSelection: projectsToWatch } = initialCreateOperationsContext;
+
+    if (!initialState) {
+      terminal.writeErrorLine(`Cannot watch for changes if not in a Git repository, exiting.`);
```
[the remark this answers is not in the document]
iclanton: Are there other reasons this wouldn't be defined?

libraries/rush-lib/src/logic/buildCache/test/ProjectBuildCache.test.ts
```diff
@@ -19,13 +18,8 @@ interface ITestOptions {
...
   async function prepareSubject(options: Partial<ITestOptions>): Promise<ProjectBuildCache | undefined> {
     const terminal: Terminal = new Terminal(new StringBufferTerminalProvider());
-    const projectChangeAnalyzer = {
-      [ProjectChangeAnalyzer.prototype._tryGetProjectStateHashAsync.name]: async () => {
-        return 'state_hash';
-      }
-    } as unknown as ProjectChangeAnalyzer;
 
-    const subject: ProjectBuildCache | undefined = await ProjectBuildCache.tryGetProjectBuildCache({
+    const subject: ProjectBuildCache | undefined = await ProjectBuildCache.getProjectBuildCache({
```
iclanton, line 22: If this can return undefined, it should probably still be called `tryGetProjectBuildCache`

libraries/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
```diff
@@ -24,11 +22,12 @@ import { EnvironmentVariableNames } from '../../api/EnvironmentConfiguration';
...
   project: RushConfigurationProject;
+
   projectOutputFolderNames: ReadonlyArray<string>;
   additionalProjectOutputFilePaths?: ReadonlyArray<string>;
-  additionalContext?: Record<string, string>;
-  configHash: string;
-  projectChangeAnalyzer: ProjectChangeAnalyzer;
+
+  operationStateHash: string;
+
```
iclanton, line 30: Slightly weird line breaks here.

libraries/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
```diff
@@ -79,10 +78,10 @@ export class ProjectBuildCache {
     return this._cacheId;
   }
 
-  public static async tryGetProjectBuildCache(
+  public static async getProjectBuildCache(
```
iclanton, line 81:
```suggestion
  public static async getProjectBuildCacheAsync(
```

iclanton: Actually, this doesn't need to be async anymore.

libraries/rush-lib/src/logic/operations/CacheableOperationPlugin.ts
```diff
@@ -447,8 +489,13 @@ export class CacheableOperationPlugin implements IPhasedCommandPlugin {
           }
         }
 
-        const { cobuildLock, projectBuildCache, isCacheWriteAllowed, buildCacheTerminal, cacheRestored } =
-          buildCacheContext;
+        const {
+          cobuildLock,
+          operationBuildCache: projectBuildCache,
```
iclanton, line 494: Rename the local variable to `operationBuildCache`?

libraries/rush-lib/src/logic/snapshots/InputSnapshot.ts
```diff
@@ -0,0 +1,437 @@
...
+   */
+  projectFilePathFilter?: (filePath: string) => boolean;
+  /**
+   * The cached Git hashes for all files in the project folder.
+   */
+  hashes: Map<string, string>;
+  /**
+   * The list of additional file paths that are applicable to each operation name, if any.
+   */
+  additionalFilesByOperationName?: ReadonlyMap<string, ReadonlySet<string>>;
```
iclanton, line 55:
```suggestion
interface IInternalRushSnapshotProjectMetadata extends IRushSnapshotProjectMetadata {
  /**
   * Cached filter of files that are not ignored by the project's `incrementalBuildIgnoredGlobs`.
   * @param filePath - The path to the file to check
   * @returns true if the file path is an input to all operations in the project, false otherwise
   */
  projectFilePathFilter?: (filePath: string) => boolean;
  /**
   * The cached Git hashes for all files in the project folder.
   */
  hashes: Map<string, string>;
```

libraries/rush-lib/src/logic/snapshots/InputSnapshot.ts
```diff
@@ -0,0 +1,437 @@
...
+      globalAdditionalFiles,
+      hashes,
+      lookupByPath,
+      rootDir
+    } = params;
+    const projectMetadataMap: Map<
+      IRushConfigurationProjectForSnapshot,
+      IInternalRushSnapshotProjectMetadata
+    > = new Map();
+    for (const [project, record] of params.projectMap) {
```
iclanton, line 200: Is this just the projects that the command invocation will actually run operations on?

libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
iclanton: Any reason to not just concatenate the path parts together with a slash?

common/reviews/api/rush-lib.api.md
```diff
@@ -491,6 +491,12 @@ export interface IGetChangedProjectsOptions {
 export interface IGlobalCommand extends IRushCommand {
 }
 
+// @beta
+export interface IInputSnapshot {
+    getLocalStateHashForOperation(project: IRushConfigurationProjectForSnapshot, operationName?: string): string;
```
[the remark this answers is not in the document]
dmichon-msft: It is a hash of the projects's own files and the npm dependency shrinkwrap. `getOwnStateHashForOperation`?

libraries/rush-lib/src/logic/buildCache/ProjectBuildCache.ts
dmichon-msft: That was why the rename, but I forgot to de-async it.

libraries/rush-lib/src/logic/snapshots/InputSnapshot.ts
```diff
@@ -0,0 +1,455 @@
...
+
+const { hashDelimiter } = RushConstants;
+
+/**
+ * Represents a synchronously-queryable in-memory snapshot of the state of the inputs to a Rush repository.
+ *
+ * The methods on this interface are idempotent and will return the same result regardless of when they are executed.
+ * @beta
+ */
+export interface IInputSnapshot {
```
iclanton, line 127:
```suggestion
export interface IInputsSnapshot {
```

libraries/rush-lib/src/logic/snapshots/InputSnapshot.ts
```diff
@@ -0,0 +1,455 @@
...
+/**
+ * Represents a synchronously-queryable in-memory snapshot of the state of the inputs to a Rush repository.
+ * Any asynchronous work needs to be performed by the caller and the results passed to the constructor.
+ *
+ * @remarks
+ * All operations on this class will return the same result regardless of when they are executed.
+ *
+ * @internal
+ */
+export class InputSnapshot implements IInputSnapshot {
```
iclanton, line 170:
```suggestion
export class InputsSnapshot implements IInputsSnapshot {
```

libraries/rush-lib/src/logic/buildCache/test/ProjectBuildCache.test.ts
```diff
@@ -42,21 +36,18 @@ describe(ProjectBuildCache.name, () => {
...
-  describe(ProjectBuildCache.tryGetProjectBuildCacheAsync.name, () => {
-    it('returns a ProjectBuildCache with a calculated cacheId value', async () => {
-      const subject: ProjectBuildCache = (await prepareSubject({}))!;
-      expect(subject['_cacheId']).toMatchInlineSnapshot(
-        `"acme-wizard/1926f30e8ed24cb47be89aea39e7efd70fcda075"`
-      );
+  describe(ProjectBuildCache.getProjectBuildCache.name, () => {
+    it('returns a ProjectBuildCache with a calculated cacheId value', () => {
+      const subject: ProjectBuildCache = prepareSubject({});
+      expect(subject['_cacheId']).toMatchInlineSnapshot(`"acme-wizard/build"`);
```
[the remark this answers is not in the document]
iclanton: Part of the value of this test was to ensure that the cache IDs don't get unexpectedly changed.

dmichon-msft (commented):
Addressed feedback.

libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
```diff
@@ -337,66 +200,187 @@ export class ProjectChangeAnalyzer {
...
+      // Load the rush-project.json files for the whole repository
+      const additionalGlobs: IAdditionalGlob[] = [];
 
-      const shrinkwrapHash: string | undefined = repoDeps.get(shrinkwrapFile);
+      const projectMap: Map<RushConfigurationProject, IRushSnapshotProjectMetadata> = new Map();
 
-      for (const projectDeps of projectHashDeps.values()) {
-        if (shrinkwrapHash) {
-          projectDeps.set(shrinkwrapFile, shrinkwrapHash);
+      for (const project of rushConfiguration.projects) {
```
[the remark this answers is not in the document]
dmichon-msft: Hm, this would cause problems with partial installs.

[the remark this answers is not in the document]
dmichon-msft: After further examination, this isn't a behavior change from what Rush does today.

libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -536,24 +538,34 @@ export class PhasedScriptAction extends BaseScriptAction<IPhasedCommandConfig> {
...
+    const snapshotProvider: (() => Promise<IInputSnapshot | undefined>) | undefined =
+      await analyzer._tryGetSnapshotProviderAsync(projectConfigurations, terminal);
+    const initialSnapshot: IInputSnapshot | undefined = await snapshotProvider?.();
+
     repoStateStopwatch.stop();
     terminal.writeLine(`DONE (${repoStateStopwatch.toString()})`);
+    if (!initialSnapshot) {
+      terminal.writeLine(
+        `The Rush monorepo is not in a Git repository. Rush will proceed without incremental build support.`
+      );
```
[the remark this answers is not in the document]
dmichon-msft: Moved the logging into the snapshot calculation.

Issue #3517, opened 2022-07-04
[rush] repo directory symlinks passed to git hash-object cannot be hashed 


hawkett:
## Summary

If a repo contains symlinked folders that are not in `.gitignore`, the diff/hashing process fails, because `git hash-object` cannot hash directories or symlinks to directories. The impact is that monorepos with git managed symlinks cannot take advantage of incremental builds, because rush attempts to hash every git managed file (including symlinks).

```
Error calculating the state of the repo. (inner error: Error: git hash-object exited with status 128: fatal: Unable to hash apps/infra/containers/containerA/<symlink-to-repo-project>
). Continuing without diffing files.
```

In this case the structure above is used to support building Docker containers using projects in the monorepo.

~Same issue in Turborepo: https://github.com/vercel/turborepo/issues/531

[Zulip discussion](https://rushstack.zulipchat.com/#narrow/stream/262513-general/topic/symlink.20causes.20hash.20failure.20during.20rush.20build/near/288264852)

## Repro steps

- Create a folder symlink anywhere (don't add it to .gitignore).
- Run `rush build`.

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.75.0 |
| `rushVersion` from rush.json? | 5.75.0 |
| `useWorkspaces` from rush.json? | true |
| Operating system? | Linux |
| Would you consider contributing a PR? | No |
| Node.js version (`node -v`)? | 16.15.1 |

elliot-nelson:
Bringing over from zulipchat discussion...

The `git ls-tree` does show `120000` for symlinks (on MacOS -- need to confirm behavior on Windows), so one approach might be to filter out these files from the list before calling `git hash-object`.

Then we could do one of the following:

 - (Option 1) Ignore all symlinks for the purposes of hashing. Simplest, but is a potential regression for existing users that were symlinking files cross-folders explicitly to get hashing.
 - (Option 2) Call `git hash-object` individually on each symlink, discarding errors but honoring files that return a hash.
 - (Option 3) Stat each symlink, discarding folders and missing files, and then make a single repeat git hash-object call with all the legit file symlinks (which might be to files inside or outside the repo).

[...]

PR #4478, opened 2024-01-11
[rush](package-deps-hash) Avoid defect to be produced when spawned "git hash-object" process stdin fails while computing repo-state

### LINKED ISSUE #4475: [rush](package-deps-hash): uncaught errors happening when using Rush commands spawning git commands to calculate the repo state
## Summary

Hello,

While trying to bootstrap a Rush toolchain with Docker, I encountered an issue when running commands associated with repo state (`rush build` or any other bulk command) and [more precisely interacting with git under the hood](https://github.com/microsoft/rushstack/blob/d351d37f5407a5f173f054d018e4dc607a6c25ca/libraries/package-deps-hash/src/getRepoState.ts#L259) (I might report other issues soon related to that, but one problem at a time), I'll put the full error below.

When interacting with git in a Docker environment or any other environment, git might be there or might not, or maybe wrongly configured, whatever. I think Rush should work anyway and just skip repo state calculation when it can't rely on git in restrained environments and not crash.

## Repro steps

1. Clone the following repository https://github.com/antoine-coulon/rushelp. FYI this is a repository I created to simplify my way of reproducing Rush.js issues and working on them. The Dockerfile might be wrongly configuring `git` for Rush to work properly, but in that case its helpful (maybe we could fix it afterwards).  

2. Build the image and run the container. Then `cd /build` and from there you can run `rush my-bulk-command` or `rush build` and you'll get the error described just below.

  **Expected result:** 

I expect all the Rush commands relying on repo state to be working even if the repo state could not be computed for some reason.
Also, [I would expect the warning from `_getRepoDepsAsync`](https://github.com/microsoft/rushstack/blob/d351d37f5407a5f173f054d018e4dc607a6c25ca/libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts#L452) to be emitted such that the one I got by investigating (I will open the PR):

`Error calculating the state of the repo. (inner error: Error: git --no-optional-locks exited with code 128: fatal: Unable to hash bin
). Continuing without diffing files.`

  **Actual result:** 

The actual result is an uncaught error that makes the process exit hence does not allow Rush to complete its command:

```
node:events:495
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:160:15)
    at writeGeneric (node:internal/stream_base_commons:151:3)
    at Socket._writeGeneric (node:net:962:11)
    at Socket._write (node:net:974:8)
    at writeOrBuffer (node:internal/streams/writable:392:12)
    at _write (node:internal/streams/writable:333:10)
    at Writable.write (node:internal/streams/writable:337:10)
    at Readable.ondata (node:internal/streams/readable:809:22)
    at Readable.emit (node:events:517:28)
    at addChunk (node:internal/streams/readable:368:12)
Emitted 'error' event on Socket instance at:
    at Socket.onerror (node:internal/streams/readable:828:14)
    at Socket.emit (node:events:517:28)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -32,
  code: 'EPIPE',
  syscall: 'write'
}
```
## Details

**What do you think is the cause of this problem?**

The cause of this problem is basically no `error` handlers [when spawning git commands there](https://github.com/microsoft/rushstack/blob/d351d37f5407a5f173f054d018e4dc607a6c25ca/libraries/package-deps-hash/src/getRepoState.ts#L259) on various streams on the spawned child process and given its async nature it just bubbles up in an uncaught way. More precisely, it seems to be related to `process.stdin` in my example but it could be due to any other streams.

Note also that it might leak memory it these specific cases because we don't close streams in case of errors.

**How do you think we should fix this?**

I think we should fix it by attaching error handlers when spawning git commands hence allowing `rush-lib` to process the remaining work and fallbacking to no cache. I'll open a PR with a suggestion, feel free to discuss it there.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.112.0 |
| `rushVersion` from rush.json? | 5.112.2 |
| Operating system? | Mac / Linux |
| Would you consider contributing a PR? | Yes, I'm already working on the PR |
| Node.js version (`node -v`)? | 18.17.0 |

antoine-coulon:
## Summary

Fixes #4475 

As explained in #3517, symlinked directories can't be hashed by `git hash-object`. Whatever the underlying reason is for that, as I already stated in the issue, executed commands should not crash due to this problem and just emit a warning.

Note that we might consider this to be a workaround while we fix the root cause which is discarding symlinked directories from the analysis. However I feel like having error handlers there will be worth it anyway as other errors could occur in the future and we definitely don't want to crash the process at this step which is not mandatory for Rush to properly work (file diffing).

## Details

I solved the problem by simply attaching an "error" handler on the underlying child process' stdin stream to avoid EPIPE errors (fatal) to occur. This error happens because the git child process closes because of the mentioned errors, but the `stdin` Readable keeps pushing file paths while the Writable is closed.

## How it was tested

Mainly through manual testing using https://github.com/antoine-coulon/rushelp and the Docker container created from it. The issue can easily be re-created, just run the container, move to `/build` and run `rush my-bulk-command`, EPIPE will occur. This is because `git status` includes `/bin` and `/lib` symlinked directories so as I said git child process will emit errors.

## Impacted documentation

None

### DIFF
--- a/libraries/package-deps-hash/src/getRepoState.ts
+++ b/libraries/package-deps-hash/src/getRepoState.ts
@@ -3,7 +3,7 @@
 
 import type * as child_process from 'child_process';
 import { once } from 'events';
-import { Readable } from 'stream';
+import { Readable, pipeline } from 'stream';
 
 import { Executable, FileSystem, type IExecutableSpawnOptions } from '@rushstack/node-core-library';
 
@@ -282,7 +282,13 @@ async function spawnGitAsync(
   });
 
   if (stdin) {
-    stdin.pipe(proc.stdin!);
+    /**
+     * We are just interested in catching "error" events there from child process'
+     * proc.stdin to avoid stdin based git commands to throw unhandled exceptions
+     * hence to let a chance for the error to be thrown in a context we can recover
+     * from.
+     */
+    pipeline(stdin, proc.stdin!, (err) => {});
   }
 
   const [status] = await once(proc, 'exit');
--- a/common/changes/@rushstack/package-deps-hash/avoid-getrepostate-to-crash_2024-01-11-09-03.json
+++ b/common/changes/@rushstack/package-deps-hash/avoid-getrepostate-to-crash_2024-01-11-09-03.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/package-deps-hash",
+      "comment": "Avoid defect to be produced bit spawnGitAsync when the underlying child process stdin stream fails while computing repo-state",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/package-deps-hash"
+}
--- a/libraries/package-deps-hash/package.json
+++ b/libraries/package-deps-hash/package.json
@@ -17,7 +17,6 @@
   },
   "devDependencies": {
     "@rushstack/heft": "workspace:*",
-    "@rushstack/node-core-library": "workspace:*",
     "local-node-rig": "workspace:*"
   },
   "dependencies": {


libraries/package-deps-hash/src/getRepoState.ts
```diff
@@ -282,7 +282,13 @@ async function spawnGitAsync(
   });
 
   if (stdin) {
-    stdin.pipe(proc.stdin!);
+    /**
+     * We are just interested in catching "error" events there from child process'
+     * proc.stdin to avoid stdin based git commands to throw unhandled exceptions
+     * hence to let a chance for the error to be thrown in a context we can recover
+     * from.
```
dmichon-msft, line 289:
```suggestion
     * For `git hash-object` data is piped in asynchronously. In the event that one of the
     * passed filenames cannot be hashed, `proc.stdin` will be closed, causing writes to fail.
     * Silence this error since it will be handled by the non-zero exit code of the process.
```

dmichon-msft (approved):
Updating the wording of the comments for clarity about the exact issue being fixed and how the change mitigates it.

iclanton (approved):

PR #4497, opened 2024-01-29
[node-core-library] Gracefully handle irregular LockFile.tryAcquire fail on macOS/Linux

### LINKED ISSUE #4491: [node-core-library] LockFile.acquire fails irregularly when multiple processes try to acquire lock
## Summary

Hi! A little context: we have a rush-managed monorepo with multiple projects. For most of the projects, build step involves calling a certain external API. Multiple parallel calls to this API lead to API rate limit errors, that’s why we’re trying to wrap those calls with LockFile from `@rushstack/node-core-library` to prevent making multiple calls in parallel.

The problem is that sometimes our CI fails with the following error:

<details>
<summary>Stack trace</summary>

```
[2024-01-24T11:09:59.217Z] node:internal/fs/sync:58
[2024-01-24T11:09:59.217Z]   const stats = binding.statSync(
[2024-01-24T11:09:59.217Z]                         ^
[2024-01-24T11:09:59.217Z] 
[2024-01-24T11:09:59.217Z] Error: File does not exist: /home/ubuntu/workspace/web-monorepo/packages/joom-crowdin/src/lib/lock/joom-crowdin#2650.lock
[2024-01-24T11:09:59.217Z] ENOENT: no such file or directory, stat '/home/ubuntu/workspace/web-monorepo/packages/joom-crowdin/src/lib/lock/joom-crowdin#2650.lock'
[2024-01-24T11:09:59.217Z]     at Object.stat (node:internal/fs/sync:58:25)
[2024-01-24T11:09:59.217Z]     at Object.statSync (node:fs:1669:17)
[2024-01-24T11:09:59.217Z]     at Object.<anonymous> (/home/ubuntu/workspace/web-monorepo/common/temp/node_modules/.pnpm/graceful-fs@4.2.10/node_modules/graceful-fs/polyfills.js:319:16)
[2024-01-24T11:09:59.217Z]     at /home/ubuntu/workspace/web-monorepo/common/temp/node_modules/.pnpm/@rushstack+node-core-library@3.61.0_@types+node@20.9.0/node_modules/@rushstack/node-core-library/lib/FileSystem.js:152:24
[2024-01-24T11:09:59.217Z]     at FileSystem._wrapException (/home/ubuntu/workspace/web-monorepo/common/temp/node_modules/.pnpm/@rushstack+node-core-library@3.61.0_@types+node@20.9.0/node_modules/@rushstack/node-core-library/lib/FileSystem.js:1050:20)
[2024-01-24T11:09:59.217Z]     at FileSystem.getStatistics (/home/ubuntu/workspace/web-monorepo/common/temp/node_modules/.pnpm/@rushstack+node-core-library@3.61.0_@types+node@20.9.0/node_modules/@rushstack/node-core-library/lib/FileSystem.js:151:27)
[2024-01-24T11:09:59.217Z]     at LockFile._tryAcquireMacOrLinux (/home/ubuntu/workspace/web-monorepo/common/temp/node_modules/.pnpm/@rushstack+node-core-library@3.61.0_@types+node@20.9.0/node_modules/@rushstack/node-core-library/lib/LockFile.js:256:64)
[2024-01-24T11:09:59.217Z]     at LockFile.tryAcquire (/home/ubuntu/workspace/web-monorepo/common/temp/node_modules/.pnpm/@rushstack+node-core-library@3.61.0_@types+node@20.9.0/node_modules/@rushstack/node-core-library/lib/LockFile.js:202:29)
[2024-01-24T11:09:59.217Z]     at retryLoop (/home/ubuntu/workspace/web-monorepo/common/temp/node_modules/.pnpm/@rushstack+node-core-library@3.61.0_@types+node@20.9.0/node_modules/@rushstack/node-core-library/lib/LockFile.js:224:35)
[2024-01-24T11:09:59.217Z]     at LockFile.acquire (/home/ubuntu/workspace/web-monorepo/common/temp/node_modules/.pnpm/@rushstack+node-core-library@3.61.0_@types+node@20.9.0/node_modules/@rushstack/node-core-library/lib/LockFile.js:234:16) {
[2024-01-24T11:09:59.217Z]   errno: -2,
[2024-01-24T11:09:59.217Z]   code: 'ENOENT',
[2024-01-24T11:09:59.217Z]   syscall: 'stat',
[2024-01-24T11:09:59.217Z]   path: '/home/ubuntu/workspace/web-monorepo/packages/joom-crowdin/src/lib/lock/joom-crowdin#2650.lock'
[2024-01-24T11:09:59.217Z] }
[2024-01-24T11:09:59.217Z] 
[2024-01-24T11:09:59.217Z] Node.js v20.9.0
[2024-01-24T11:09:59.217Z] Returned error code: 1
```

</details>

## Repro steps

I was not able to reproduce this error reliably, it happens randomly. I was able to catch it a few times on macOS besides those errors in CI (which runs on Linux).

Here is the code that I used to reproduce the error: https://gist.github.com/andrew--r/6e45f2b1661cabec9ea7d0d0bb972e50

**Update**: see [reproduction steps](https://github.com/microsoft/rushstack/issues/4491#issuecomment-1925900671)

## Details

The stack trace leads to this line of code:

https://github.com/microsoft/rushstack/blob/a1c83337d44c8409f74c9ae13d9d4f8249b49ebf/libraries/node-core-library/src/LockFile.ts#L267

As I understood, there is some logic that deletes other processes lockfiles under certain conditions:

https://github.com/microsoft/rushstack/blob/a1c83337d44c8409f74c9ae13d9d4f8249b49ebf/libraries/node-core-library/src/LockFile.ts#L329-L334

My hypothesis is that the lockfile is deleted by other process between the calls to `lockFileHandle.write(startTime)` and `FileSystem.getStatistics(pidLockFilePath)`, which leads to error, but I’m not entirely sure about it. Either way, I think wrapping [these lines](https://github.com/microsoft/rushstack/blob/a1c83337d44c8409f74c9ae13d9d4f8249b49ebf/libraries/node-core-library/src/LockFile.ts#L265-L267C42) with try/catch (like [here](https://github.com/microsoft/rushstack/blob/a1c83337d44c8409f74c9ae13d9d4f8249b49ebf/libraries/node-core-library/src/LockFile.ts#L296-L302)) and returning from `catch` should solve the problem.

**Update**: see [detailed description](https://github.com/microsoft/rushstack/issues/4491#issuecomment-1925900671) of the race condition causing error

## Standard questions

| Question | Answer |
| -------- | -------- |
| Package name: | @rushstack/node-core-library |
| Package version? | 3.61.0 |
| Operating system? | Mac / Linux |
| Node.js version (`node -v`)? | 20.9.0 |

andrew--r:
Fixes #4491

## Summary and details

See https://github.com/microsoft/rushstack/pull/4497#issuecomment-1918189321

There are two fixes in this PR:

1. Fixed the root cause of the unexpected ENOENT error
2. Added `getStatistics` method to `FileWriter` [as requested](https://github.com/microsoft/rushstack/pull/4497#discussion_r1470058232)

## How it was tested

I’ve added a corresponding test case to LockFile tests suite

### DIFF
--- a/libraries/node-core-library/src/FileWriter.ts
+++ b/libraries/node-core-library/src/FileWriter.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-import type * as fs from 'fs';
+import type { FileSystemStats } from './FileSystem';
 import { Import } from './Import';
 
 const fsx: typeof import('fs-extra') = Import.lazy('fs-extra', require);
@@ -33,15 +33,6 @@ export interface IFileWriterFlags {
    */
   exclusive?: boolean;
 }
-
-/**
- * An alias for the Node.js `fs.Stats` object.
- *
- * @remarks
- * This avoids the need to import the `fs` package when using the {@link FileWriter} API.
- * @public
- */
-export type FileWriterStats = fs.Stats;
 
 /**
  * API for interacting with file handles.
@@ -117,7 +108,7 @@ export class FileWriter {
    * Gets the statistics for the given file handle. Throws if the file handle has been closed.
    * Behind the scenes it uses `fs.statSync()`.
    */
-  public getStatistics(): FileWriterStats {
+  public getStatistics(): FileSystemStats {
     if (!this._fileDescriptor) {
       throw new Error(`Cannot get file statistics, file descriptor has already been released.`);
     }
--- a/libraries/node-core-library/src/FileWriter.ts
+++ b/libraries/node-core-library/src/FileWriter.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
+import type * as fs from 'fs';
 import { Import } from './Import';
 
 const fsx: typeof import('fs-extra') = Import.lazy('fs-extra', require);
@@ -32,6 +33,15 @@ export interface IFileWriterFlags {
    */
   exclusive?: boolean;
 }
+
+/**
+ * An alias for the Node.js `fs.Stats` object.
+ *
+ * @remarks
+ * This avoids the need to import the `fs` package when using the {@link FileWriter} API.
+ * @public
+ */
+export type FileWriterStats = fs.Stats;
 
 /**
  * API for interacting with file handles.
@@ -101,5 +111,17 @@ export class FileWriter {
       this._fileDescriptor = undefined;
       fsx.closeSync(fd);
     }
+  }
+
+  /**
+   * Gets the statistics for the given file handle. Throws if the file handle has been closed.
+   * Behind the scenes it uses `fs.statSync()`.
+   */
+  public getStatistics(): FileWriterStats {
+    if (!this._fileDescriptor) {
+      throw new Error(`Cannot get file statistics, file descriptor has already been released.`);
+    }
+
+    return fsx.fstatSync(this._fileDescriptor);
   }
 }
--- a/libraries/node-core-library/src/LockFile.ts
+++ b/libraries/node-core-library/src/LockFile.ts
@@ -257,14 +257,23 @@ export class LockFile {
 
     let lockFile: LockFile;
 
-    try {
-      // open in write mode since if this file exists, it cannot be from the current process
-      // TODO: This will malfunction if the same process tries to acquire two locks on the same file.
-      // We should ideally maintain a dictionary of normalized acquired filenames
-      lockFileHandle = FileWriter.open(pidLockFilePath);
-      lockFileHandle.write(startTime);
+    let currentBirthTimeMs: number;
 
-      const currentBirthTimeMs: number = FileSystem.getStatistics(pidLockFilePath).birthtime.getTime();
+    try {
+      try {
+        // open in write mode since if this file exists, it cannot be from the current process
+        // TODO: This will malfunction if the same process tries to acquire two locks on the same file.
+        // We should ideally maintain a dictionary of normalized acquired filenames
+        lockFileHandle = FileWriter.open(pidLockFilePath);
+        lockFileHandle.write(startTime);
+        currentBirthTimeMs = FileSystem.getStatistics(pidLockFilePath).birthtime.getTime();
+      } catch (error) {
+        // sometimes FileSystem.getStatistics leads to failure when multiple processes try to acquire
+        // lock for the same resource, so if we get an error, gracefully exit.
+        // see https://github.com/microsoft/rushstack/issues/4491
+        lockFileHandle?.close();
+        return undefined;
+      }
 
       let smallestBirthTimeMs: number = currentBirthTimeMs;
       let smallestBirthTimePid: string = pid.toString();
--- a/libraries/node-core-library/src/LockFile.ts
+++ b/libraries/node-core-library/src/LockFile.ts
@@ -266,7 +266,7 @@ export class LockFile {
         // We should ideally maintain a dictionary of normalized acquired filenames
         lockFileHandle = FileWriter.open(pidLockFilePath);
         lockFileHandle.write(startTime);
-        currentBirthTimeMs = FileSystem.getStatistics(pidLockFilePath).birthtime.getTime();
+        currentBirthTimeMs = lockFileHandle.getStatistics().birthtime.getTime();
       } catch (error) {
         // sometimes FileSystem.getStatistics leads to failure when multiple processes try to acquire
         // lock for the same resource, so if we get an error, gracefully exit.
--- a/libraries/node-core-library/src/LockFile.ts
+++ b/libraries/node-core-library/src/LockFile.ts
@@ -260,20 +260,12 @@ export class LockFile {
     let currentBirthTimeMs: number;
 
     try {
-      try {
-        // open in write mode since if this file exists, it cannot be from the current process
-        // TODO: This will malfunction if the same process tries to acquire two locks on the same file.
-        // We should ideally maintain a dictionary of normalized acquired filenames
-        lockFileHandle = FileWriter.open(pidLockFilePath);
-        lockFileHandle.write(startTime);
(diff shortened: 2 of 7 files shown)


andrew--r:
@microsoft-github-policy-service agree company="Joom Unipessoal LDA"

libraries/node-core-library/src/LockFile.ts
```diff
@@ -257,14 +257,23 @@ export class LockFile {
...
 
-      const currentBirthTimeMs: number = FileSystem.getStatistics(pidLockFilePath).birthtime.getTime();
+    try {
+      try {
+        // open in write mode since if this file exists, it cannot be from the current process
+        // TODO: This will malfunction if the same process tries to acquire two locks on the same file.
+        // We should ideally maintain a dictionary of normalized acquired filenames
+        lockFileHandle = FileWriter.open(pidLockFilePath);
+        lockFileHandle.write(startTime);
+        currentBirthTimeMs = FileSystem.getStatistics(pidLockFilePath).birthtime.getTime();
```
dmichon-msft, line 269: I would expect part of the problem to be that this is opening a second file handle to the same file in the same process.
We should consider adding a `getStatistics()` method to the `FileWriter` instance that returns `fsx.fstatSync(this._fileDescriptor)`, then call `lockFileHandle.getStatistics()` instead of `FileSystem.getStatistics(pidLockFilePath)`

andrew--r: Thanks for the idea! I’m not sure it’s the root cause of the problem, but it definitely makes sense in terms of overall efficiency and simplicity of file system operations. I’ve added the `getStatistics` method to `FileWriter` class and reused `lockFileHandle` to get stats.

libraries/node-core-library/src/FileWriter.ts
```diff
@@ -33,6 +34,15 @@ export interface IFileWriterFlags {
...
 }
 
+/**
+ * An alias for the Node.js `fs.Stats` object.
+ *
+ * @remarks
+ * This avoids the need to import the `fs` package when using the {@link FileWriter} API.
+ * @public
+ */
+export type FileWriterStats = fs.Stats;
```
iclanton, line 44: Just grab `FileSystemStats` from `FileSystem.ts`

libraries/node-core-library/src/LockFile.ts
```diff
@@ -257,14 +257,23 @@ export class LockFile {
...
-      const currentBirthTimeMs: number = FileSystem.getStatistics(pidLockFilePath).birthtime.getTime();
+    try {
+      try {
+        // open in write mode since if this file exists, it cannot be from the current process
+        // TODO: This will malfunction if the same process tries to acquire two locks on the same file.
+        // We should ideally maintain a dictionary of normalized acquired filenames
+        lockFileHandle = FileWriter.open(pidLockFilePath);
+        lockFileHandle.write(startTime);
+        currentBirthTimeMs = lockFileHandle.getStatistics().birthtime.getTime();
+      } catch (error) {
```
iclanton, line 270: Is there an error code that we can filter for? It feels a little uncomfortable swallowing all errors, including for `FileWriter.open`.

libraries/node-core-library/src/FileWriter.ts
andrew--r: You mean removing type alias completely or defining it like `type FileWriterStats = FileSystemStats`?

andrew--r:
I’ve dug a bit deeper into the problem, and I think I’ve managed to find the root cause.

The problem occurs when at least two processes constantly try to acquire the same lock, e.g. when processing a task queue inside each process.

Reproduction steps:
1. The first process acquires lock and starts executing something (e.g. current task).
2. The second process simultaneously tries to acquire the same lock, writes it’s lockfile, …, and [scans directory in search for other lockfiles](https://github.com/microsoft/rushstack/blob/e0a8426024c4fbfdfa40a7647c105c29d94c2d29/libraries/node-core-library/src/LockFile.ts#L272-L273).
3. Immediately after it, the first process lock is released, which leads to the [first lockfile being deleted](https://github.com/microsoft/rushstack/blob/e0a8426024c4fbfdfa40a7647c105c29d94c2d29/libraries/node-core-library/src/LockFile.ts#L435-L437)
4. The second lock [processes](https://github.com/microsoft/rushstack/blob/e0a8426024c4fbfdfa40a7647c105c29d94c2d29/libraries/node-core-library/src/LockFile.ts#L280) each file found on the step 2. The first lockfile doesn’t exist anymore, so it [fails to retrieve info about the first process lock](https://github.com/microsoft/rushstack/blob/e0a8426024c4fbfdfa40a7647c105c29d94c2d29/libraries/node-core-library/src/LockFile.ts#L296-L302), and the error is silently swallowed
5. The first process takes the next task from the queue and starts acquiring new lock, writing a new lockfile
6. The second process [uses obsolete data about the first process lockfile and deletes it](https://github.com/microsoft/rushstack/blob/e0a8426024c4fbfdfa40a7647c105c29d94c2d29/libraries/node-core-library/src/LockFile.ts#L330-L334) with an assumption that the first process is no longer executing, but the lockfile still exists for some reason
7. The first process tries to [retrieve stats for it’s new lockfile](https://github.com/microsoft/rushstack/blob/e0a8426024c4fbfdfa40a7647c105c29d94c2d29/libraries/node-core-library/src/LockFile.ts#L267), which was deleted on the previous step by the second process, which finally leads to `ENOENT` error

To be sure, in the next few days I’ll write some test cases covering this behaviour, and if it confirms, I’ll update the PR with a fix for the root cause.

libraries/node-core-library/src/FileWriter.ts
andrew--r: I’ve removed the FileWriterStats type alias in favour of FileSystemStats

libraries/node-core-library/src/LockFile.ts
andrew--r: I’ve reworked the PR to fix the root cause of the problem, there is no need in this try/catch block now, so I’ve removed it

dmichon-msft (approved):

andrew--r:
@iclanton could you please take a look at the updates when you’ll have time?

iclanton (approved):

PR #4505, opened 2024-01-31
[rush] Clean up the dependenciesMeta verification code.

iclanton:
## Summary

Refactors the dependenciesMeta validation code.

## Details

This eliminates a second loop and an unnecessary check.

## How it was tested

Tested by adding and removing dependenciesMeta in a local project and running `rush update`.

## Impacted documentation

None.

### DIFF
--- a/libraries/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
+++ b/libraries/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
@@ -292,11 +292,16 @@ export class WorkspaceInstallManager extends BaseInstallManager {
     }
 
     // Now, we compare these two objects to see if they are equal or not
-    shrinkwrapIsUpToDate =
-      objectsAreDeepEqual(
-        expectedDependenciesMetaByProjectRelativePath,
-        lockfileDependenciesMetaByProjectRelativePath
-      ) && shrinkwrapIsUpToDate;
+    const dependenciesMetaAreEqual: boolean = objectsAreDeepEqual(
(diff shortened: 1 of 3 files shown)


libraries/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
```diff
@@ -253,61 +263,44 @@ export class WorkspaceInstallManager extends BaseInstallManager {
...
-          const dependenciesMeta: IDependenciesMetaTable = {};
-          for (const dependencyMeta of dependencyMetaList) {
-            dependenciesMeta[dependencyMeta.name] = {
-              injected: dependencyMeta.injected
-            };
-          }
-          packagePathToDependenciesMetaInPackageJson[relativePathFromTempFolderToPackageFolder] =
-            dependenciesMeta;
-        }
+        const relativePathFromTempFolderToPackageFolder: string = `${relativeFromTempFolderToRootFolder}/${rushProject.projectRelativeFolder}`;
```
dmichon-msft, line 277: Did we ever fully enforce that `rushProject.projectRelativeFolder` neither starts nor ends with a slash, and does not contain `\\`?

dmichon-msft (approved):

libraries/rush-lib/src/logic/installManager/WorkspaceInstallManager.ts
iclanton: We actually don't. We probably should, but that should be a in a follow-up.

PR #4510, opened 2024-02-05
[webpack5-localization-plugin] Include a feature for generating true filename hashes.

iclanton:
## Summary

The localization plugin currently has an issue where filename hashes are 1) the same on a chunk between all locales and 2) not updated if strings are changed. This PR introduces an option to the localization plugin called `useTrueHashes` that, when set to `true`, updates hashes to match individual filenames' contents.

## Details

Updating hashes is achieved by postprocessing chunks after all localized variants are rendered. Chunks are processed in reverse-tree-order, updating the hashes of async chunks in the runtime chunk loading code in the runtime chunk's bodies after async chunks are renamed, and then hashing runtime chunks to reflect the updated async chunk filename hashes.

## How it was tested

Unit tests in the localization plugin project have been updated to exercise this new functionality. This has also been tested in a large monorepo that uses the localization plugin.

## Impacted documentation

This option should be documented.


webpack/webpack5-localization-plugin/src/TrueHashPlugin.ts
```diff
@@ -0,0 +1,401 @@
...
+
+                  if (hasAnyReplacements) {
+                    const sourceString: string = assetSource.source().toString();
+                    const replaceSource: sources.ReplaceSource = new thisWebpack.sources.ReplaceSource(
+                      assetSource,
+                      assetName
+                    );
+
+                    if (allReplacementsAreTheSameLengthAsOriginals) {
+                      // If all of the replacements are the same length as the originals, we can walk the string
```
dmichon-msft, line 216: `ReplaceSource` takes care of ordering the replacements so that the indices are correct. The call to `.replace` just queues the replacement based on indices in the original source.

webpack/webpack5-localization-plugin/src/TrueHashPlugin.ts
```diff
@@ -0,0 +1,401 @@
...
+                        const replacement: string = relevantHashReplacements.get(
+                          Text.reverse(reverseOriginalHash)
+                        )!;
+
+                        // Figure out the location in the original string
+                        const reversedMatchStart: number = sourceStringLength - matchEnd - 1;
+                        const reversedMatchEnd: number = sourceStringLength - matchStart - 1;
+                        replaceSource.replace(reversedMatchStart, reversedMatchEnd, replacement);
+                      }
+                    }
```
dmichon-msft, line 257: We don't need this.

webpack/webpack5-localization-plugin/src/TrueHashPlugin.ts
```diff
@@ -0,0 +1,401 @@
...
+      if (hasLocalizationPluginTrueHashOption) {
+        compilation.warnings.push(
+          new thisWebpack.WebpackError(
+            `The ${TrueHashPlugin.name} is not compatible with the LocalizationPlugin's "useTrueHashes" option. ` +
+              `Because the LocalizationPlugin is already handling true hashes, the ${TrueHashPlugin.name} plugin ` +
+              'will have no effect.'
+          )
+        );
+      } else {
+        const { stageOverride: processAssetsStage = thisWebpack.Compilation.PROCESS_ASSETS_STAGE_SUMMARIZE } =
```
dmichon-msft, line 383: I think this should default to some value < STAGE_OPTIMIZE_TRANSFER

webpack/webpack5-localization-plugin/src/interfaces.ts
```diff
@@ -144,6 +144,11 @@ export interface ILocalizationPluginOptions {
    * runtimeLocaleExpression produces the same output as formatLocaleForFilename.
    */
   formatLocaleForFilename?: (locale: string) => string;
+
+  /**
+   * If set to true, update usages of [contenthash] to use the true hash of the file contents
+   */
+  useTrueHashes?: boolean;
```
dmichon-msft, line 151: Should we call this `realContentHash` to match the webpack config option of the same name that does the same thing?

webpack/webpack5-localization-plugin/src/test/MixedAsync.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  expect(errors).toHaveLength(0);
+  expect(warnings).toHaveLength(0);
+}
+
+describe(LocalizationPlugin.name, () => {
+  it('Handles async localized and non-localized chunks with a runtime locale expression (unminified)', async () => {
+    await testMixedAsyncInner(false);
+  });
+
+  it('Handles async localized and non-localized chunks with a runtime locale expression (minified)', async () => {
```
dmichon-msft, line 114: Fix test names. These aren't using the runtime locale expression.

webpack/webpack5-localization-plugin/src/test/NonHashedNonLocalizedAssets.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  const trueHashPlugin: TrueHashPlugin = new TrueHashPlugin();
+
+  const compiler: webpack.Compiler = webpack({
+    entry: {
+      mainSingleChunk: '/entrySingleChunk.js',
+      mainTwoChunks: '/entryTwoChunks.js'
+    },
+    output: {
+      path: '/release',
+      filename: '[name]-[locale].js'
```
dmichon-msft, line 35:
```suggestion
      filename: '[name].js'
```
To make it obvious that there is no localization plugin present.

webpack/webpack5-localization-plugin/src/test/NonHashedNonLocalizedAssets.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      '/package.json': '{}',
+      '/entrySingleChunk.js': `console.log("Do stuff");import(/* webpackChunkName: 'async1' */ './async1.js').then(mod => mod.foo());`,
+      '/entryTwoChunks.js': `console.log("Do stuff");import(/* webpackChunkName: 'async1' */ './async1.js').then(mod => mod.foo());import(/* webpackChunkName: 'async2' */ './async2.js').then(mod => mod.foo());`,
+      '/async1.js': `export function foo() { console.log('foo1'); }`,
+      '/async2.js': `export function foo() { console.log('foo2'); }`
+    },
+    '/src'
+  );
+
+  const trueHashPlugin: TrueHashPlugin = new TrueHashPlugin();
```
dmichon-msft, line 26: If there isn't a localization plugin, callers should just use `optimization.realContentHash: true`

common/reviews/api/webpack5-localization-plugin.api.md
```diff
@@ -4,19 +4,33 @@
...
+
 // @public (undocumented)
 export interface IDefaultLocaleOptions {
     fillMissingTranslationStrings?: boolean;
     localeName: string;
 }
 
+// @public (undocumented)
+export interface IHashAlgorithmOptions extends ITrueHashPluginOptionsBase {
+    // Warning: (ae-forgotten-export) The symbol "WebpackHash" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
iclanton: I'll just go ahead and export this.

webpack/webpack5-localization-plugin/src/TrueHashPlugin.ts
iclanton: Changing this to `STAGE_OPTIMIZE_TRANSFER - 1` doesn't work. It can't find the assets for chunks at that point.

iclanton: This isn't being published anyway, so it doesn't matter.

dmichon-msft (approved):

PR #4512, opened 2024-02-05
[eslint-plugin]: Allow using `as const` in `typedef-var`

### LINKED ISSUE #4446: [eslint-config] Proposal: Allow inferred types for `const`s written `as const`
## Summary

Allow `as const` to be without type, as type inference will be more correct as falling back to `Record<string, string>` or even `string` is undesirable.

## Details

The current implementation of the `typedef-var` rule requires developers to explicitly specify types for variables, even when utilizing the `as const` assertion. This approach can lead to less accurate type annotations, especially when dealing with large objects, as developers may resort to generic types such as `Record<string, string>`.

Consider the following examples:

1. **Single Value:**
    ``` 
    const myConstant = "the constant";
    // Typed as string vvvvv, losing the real value
    const myConstant: string = "the constant";
    ```

    Instead of the more accurate:
    ``` 
    const myConstant: "the constant" = "the constant";
    ```

2. **Object:**
    ``` 
    const CONSTANTS = {
      a: "first",
      b: "another"
    };
    // With larger objects likely typed as Record<string, string>
    ```

    Instead of the more accurate:
    ``` 
    const CONSTANTS: { a: "first"; b: "another" } = {
      a: "first",
      b: "another"
    };
    ```

### Proposed Solution:

I propose allowing the omission of explicit types when using the `as const` assertion. This would enable developers to leverage TypeScript's type inference capabilities and still ensure accurate type representation.

Example of valid ✅ code:

```typescript
const CONSTANTS = {
  a: "first",
  b: "another"
} as const;
```

Example of invalid ❌ code:

```typescript
const CONSTANTS = {
  a: "first",
  b: "another"
};
```

### Benefits

1. **Accurate Type Inference:** Developers can rely on TypeScript's type inference to deduce precise types, reducing the likelihood of inaccuracies in type annotations.

2. **Improved Readability:** The code becomes more concise and readable, especially for larger objects, without sacrificing type accuracy.

3. **Consistency with Best Practices:** Aligns with best practices of leveraging TypeScript's type inference whenever possible to reduce redundancy and enhance code maintainability.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@rushstack/eslint-config` version? |  |
| Operating system? |  |
| TypeScript compiler version? |  |

StyleShit:
## Summary

Closes #4446 

## Details

Added support for `const x = 1 as const` instead of forcing the developer to add a type annotation.
In addition, added a missing test for the "happy flow" of writing `const x: number = 1;`

### DIFF
--- a/common/changes/@rushstack/eslint-plugin/feat-typedef-var-as-const_2024-02-05-18-53.json
+++ b/common/changes/@rushstack/eslint-plugin/feat-typedef-var-as-const_2024-02-05-18-53.json
@@ -3,7 +3,7 @@
     {
       "packageName": "@rushstack/eslint-plugin",
       "comment": "Allow using `as const` in `typedef-var`",
-      "type": "patch"
+      "type": "minor"
     }
   ],
   "packageName": "@rushstack/eslint-plugin"
--- a/common/changes/@rushstack/eslint-plugin/feat-typedef-var-as-const_2024-02-05-18-53.json
+++ b/common/changes/@rushstack/eslint-plugin/feat-typedef-var-as-const_2024-02-05-18-53.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/eslint-plugin",
+      "comment": "Allow using `as const` in `typedef-var`",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@rushstack/eslint-plugin"
+}
--- a/eslint/eslint-plugin/src/test/typedef-var.test.ts
+++ b/eslint/eslint-plugin/src/test/typedef-var.test.ts
@@ -47,6 +47,9 @@ ruleTester.run('typedef-var', typedefVar, {
       code: 'for (const x of []) { }'
     },
     {
+      code: 'const x = 1 as const;'
+    },
+    {
       // prettier-ignore
       code: [
         'let { a , b } = {',
--- a/eslint/eslint-plugin/src/test/typedef-var.test.ts
+++ b/eslint/eslint-plugin/src/test/typedef-var.test.ts
@@ -50,6 +50,9 @@ ruleTester.run('typedef-var', typedefVar, {
       code: 'const x = 1 as const;'
     },
     {
+      code: 'const x: 1 = 1;'
+    },
+    {
       code: 'const x: number = 1;'
     },
     {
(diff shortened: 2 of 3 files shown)


StyleShit:

eslint/eslint-plugin/src/test/typedef-var.test.ts
```diff
@@ -46,6 +46,12 @@ ruleTester.run('typedef-var', typedefVar, {
     {
       code: 'for (const x of []) { }'
     },
+    {
+      code: 'const x = 1 as const;'
+    },
+    {
+      code: 'const x: number = 1;'
```
dmichon-msft, line 53:
```suggestion
      code: 'const x: 1 = 1;'
```
Is what TypeScript would normally infer, so this should really be the reference case.

dmichon-msft (approved):

common/changes/@rushstack/eslint-plugin/feat-typedef-var-as-const_2024-02-05-18-53.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/eslint-plugin",
+      "comment": "Allow using `as const` in `typedef-var`",
+      "type": "patch"
```
iclanton, line 6:
```suggestion
      "type": "minor"
```

iclanton (approved):

iclanton (approved):

eslint/eslint-plugin/src/test/typedef-var.test.ts
iclanton: If you say `const x = 1;`, the TypeScript compiler will infer the type of `x` to be `1`. That is the most specific type that `x` can be (because it's a const, so it can never be any other number or other value).

Take a look in the [TypeScript Playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAHjAvDAjAbgFBA). If you mouse-over the `x`, IntelliSense will tell you what the inferred type is:
image

Although I kinda agree that we should have a second example case of `const x: number = 1;`.

StyleShit: yeah, I understood the part about type inference in the comment, I'm just not sure what are the actual requested changes in this case and why
anyway, added a second test case

iclanton: @dmichon-msft prefers the `const x: 1 = 1;` syntax to `const x: number = 1;` because it's more strict, and I've come to agree, so I think that's why he requested the change. It's a change we'd recommend someone make in "real" code in a PR.

PR #4517, opened 2024-02-11
fix(heft-storybook-plugin): not cancelling webpack build during storybook static build

### LINKED ISSUE #4289: [heft] storybook plugin provides no way to skip webpack build when doing a static build
## Summary

The heft storybook plugin automatically cancels the webpack plugins from running when it's run with `--serve`, but it does not do this if you omit `--serve` because you want to do a static build. Not sure if there's some reason I'm not aware of that needs it to operate this way, but it doesn't work for our setup. 

I've patched the package locally to move the webpack plugin cancellation outside of the serveMode check to resolve this for us in the meantime

## Repro steps

1. Have heft phases defined where the webpack plugin and storybook plugin are in the same phase
2. run `heft <phase> --storybook --serve` and see that the webpack plugin compilation gets cancelled
3. run `heft <phase> --storybook` and see that both the webpack and storybook builds happen in parallel

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@rushstack/heft` version? | 0.58.0 |
| `@rushstack/heft-storybook-plugin` version? | 0.3.23 |
| Operating system? | Linux |
| Node.js version (`node -v`)? | v18.17.1 |

UberMouse:
## Summary

Due to the guard around the isServeMode check in the webpack configuration taps the webpack build is only disabled when running in serve mode

This means that when doing a static storybook build it does both the webpack and storybook webpack builds when the `--storybook` cli flag is used

This changes the discarding of the webpack configuration to happen always when `--storybook` is present so static storybook builds don't double up the webpack builds, Fixes #4289 

While I'm here I added a configuration option for passing the `--webpack-stats-json` flag to Storybook to emit the webpack stats file which is required for usage with Chromatic's Turbosnap feature

## How it was tested

Applied the changes as a patch to our monorepo and ran a full chromatic run, doing static builds for all storybooks (with captureWebpackstats set to true). Launched storybook in serve mode to validate that, and did a static build to verify that the webpack build is skipped

### DIFF
--- a/common/changes/@rushstack/heft-storybook-plugin/fix-storybook-plugin-static-builds_2024-02-11-20-39.json
+++ b/common/changes/@rushstack/heft-storybook-plugin/fix-storybook-plugin-static-builds_2024-02-11-20-39.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-storybook-plugin",
+      "comment": "fix double webpack builds during static storybook builds, expose --webpack-stats-json storybook CLI flag",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@rushstack/heft-storybook-plugin"
+}
--- a/common/changes/@rushstack/heft-storybook-plugin/fix-storybook-plugin-static-builds_2024-02-11-20-40.json
+++ b/common/changes/@rushstack/heft-storybook-plugin/fix-storybook-plugin-static-builds_2024-02-11-20-40.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@rushstack/heft-storybook-plugin",
-      "comment": "Fix an issue where Webpack would run twice during static storybook builds.",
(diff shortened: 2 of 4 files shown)


UberMouse:
@microsoft-github-policy-service agree company="Koordinates Ltd"

common/changes/@rushstack/heft-storybook-plugin/fix-storybook-plugin-static-builds_2024-02-11-20-39.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@rushstack/heft-storybook-plugin",
+      "comment": "fix double webpack builds during static storybook builds, expose --webpack-stats-json storybook CLI flag",
```
iclanton, line 5:
```suggestion
      "comment": "Fix an issue where Webpack would run twice during static storybook builds.",
```

iclanton: And the other thing should go in a separate changelog file.

iclanton (approved):

PR #4536, opened 2024-02-22
[rush] Validate "projectFolder" and "publishFolder"

### LINKED ISSUE #4506: [rush] Rush should enforce that the project folder in rush.json doesn't start with a slash or contain backslashes.
Brought up here: https://github.com/microsoft/rushstack/pull/4505#discussion_r1473548547

dmichon-msft:
## Summary
Fixes #4506

## Details
Ensures that the values in `projectFolder` and `publishFolder` for each project entry in `rush.json` do not contain backslashes, are normalized POSIX relative paths, and do not end in a trailing slash.

## How it was tested
Added unit tests for the validator.
Manually tested by temporarily making rush.json invalid.

## Impacted documentation
Docs on the values of `projectFolder` and `publishFolder`.

### DIFF
--- a/libraries/rush-lib/src/api/RushConfigurationProject.ts
+++ b/libraries/rush-lib/src/api/RushConfigurationProject.ts
@@ -209,29 +209,33 @@ export class RushConfigurationProject {
   /** @internal */
   public constructor(options: IRushConfigurationProjectOptions) {
     const { projectJson, rushConfiguration, tempProjectName, allowedProjectTags } = options;
+    const { packageName, projectFolder: projectRelativeFolder } = projectJson;
     this.rushConfiguration = rushConfiguration;
-    this.packageName = projectJson.packageName;
-    this.projectRelativeFolder = projectJson.projectFolder;
+    this.packageName = packageName;
+    this.projectRelativeFolder = projectRelativeFolder;
+
+    validateRelativePathField(projectRelativeFolder, 'projectFolder');
 
     // For example, the depth of "a/b/c" would be 3.  The depth of "a" is 1.
-    const projectFolderDepth: number = projectJson.projectFolder.split('/').length;
+    const projectFolderDepth: number = projectRelativeFolder.split('/').length;
     if (projectFolderDepth < rushConfiguration.projectFolderMinDepth) {
       throw new Error(
         `To keep things organized, this repository has a projectFolderMinDepth policy` +
           ` requiring project folders to be at least ${rushConfiguration.projectFolderMinDepth} levels deep.` +
-          `  Problem folder: "${projectJson.projectFolder}"`
+          `  Problem folder: "${projectRelativeFolder}"`
       );
     }
     if (projectFolderDepth > rushConfiguration.projectFolderMaxDepth) {
       throw new Error(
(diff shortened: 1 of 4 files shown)


libraries/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -503,3 +508,29 @@ export class RushConfigurationProject {
     return isMain;
   }
 }
+
+export function validateRelativePathField(relativePath: string, name: string): void {
+  if (path.posix.isAbsolute(relativePath) || path.win32.isAbsolute(relativePath)) {
```
D4N14L, line 513: Why not just `path.isAbsolute`?

libraries/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -503,3 +508,29 @@ export class RushConfigurationProject {
     return isMain;
   }
 }
+
+export function validateRelativePathField(relativePath: string, name: string): void {
```
D4N14L, line 512: Should we also validate that the relative path does not lead above a specific target directory?

D4N14L (approved):

libraries/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: `path.isAbsolute` calls `path.posix.isAbsolute` on POSIX and `path.win32.isAbsolute` on Win32. `path.posix.isAbsolute('A:/b')` returns `false. `path.win32.isAbsolute('A:/b')` returns `true`.

dmichon-msft: I'll add comments in the code for why.

libraries/rush-lib/src/api/RushConfigurationProject.ts
dmichon-msft: Don't care. We allow rush.json to live in a child of the repo.

libraries/rush-lib/src/api/RushConfigurationProject.ts
D4N14L: I guess because that would be system dependent?

libraries/rush-lib/src/api/RushConfigurationProject.ts
```diff
@@ -503,3 +508,29 @@ export class RushConfigurationProject {
...
+  if (path.posix.isAbsolute(relativePath) || path.win32.isAbsolute(relativePath)) {
+    throw new Error(`The value "${relativePath}" in the "${name}" field must be a relative path.`);
+  }
+
+  if (relativePath.includes('\\')) {
+    throw new Error(
+      `The value "${relativePath}" in the "${name}" field may not contain backslashes ('\\'), since they are interpreted differently` +
```
iclanton, line 519: These errors should probably print which project they're mentioned in, and that it's in rush.json.

dmichon-msft: The value listed can be searched as an exact string and will find the offending value, but I agree with mentioning rush.json.

D4N14L (approved):

iclanton (approved):

PR #4540, opened 2024-02-23
[Terminal] Replace 'colors' with 'supports-color'.

iclanton:
## Summary

This PR finishes out the colors/Colorize/Terminal work. This resolves https://github.com/microsoft/rushstack/issues/3147.

## How it was tested

Ran on a console that supports color with and without the `FORCE_COLOR=0` env var, and in CI.

## Impacted documentation

None.

### DIFF
--- a/libraries/terminal/src/test/AnsiEscape.test.ts
+++ b/libraries/terminal/src/test/AnsiEscape.test.ts
@@ -1,26 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
 // See LICENSE in the project root for license information.
 
-// Explicitly use the colors package here instead of Colorize
-import colors from 'colors/safe';
 import { AnsiEscape } from '../AnsiEscape';
+import { Colorize } from '../Colorize';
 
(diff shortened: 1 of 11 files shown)


libraries/terminal/src/test/AnsiEscape.test.ts
```diff
@@ -1,26 +1,12 @@
...
-
-  afterAll(() => {
-    if (!initialColorsEnabled) {
-      colors.disable();
-    }
-  });
-
   it('calls removeCodes() successfully', () => {
-    const coloredInput: string = colors.rainbow('Hello, world!');
+    const coloredInput: string = Colorize.green('Hello, world!');
```
D4N14L, line 9: Sidenote: We should have a rainbow option as well. Is it overkill? Yes. Is it neat? Also yes.

D4N14L (approved):

libraries/terminal/src/test/AnsiEscape.test.ts
iclanton: Done.

PR #4541, opened 2024-02-24
[api-extractor] Remove "const" keyword before enum to use with typescript isolatedModules

IIIMADDINIII:
## Summary
When a typescript project is using [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules), referencing exported const enums is not allowed.

## Details
Setting isolatedModules to true is recommended for projects transpiled with babel or [esbuild](https://esbuild.github.io/content-types/#isolated-modules). Unlike tsc, these tools evaluate files independently. Setting the isolatedModules flag e.g. forces developers to indicate whether they are exporting a type or value, since transpilers evaluating modules independently cannot infer this information.

However, the flag also alerts when a const enum is used. It seems that tsc evaluates const enums across modules and inlines the constants at compile. No enum object is created, which is problematic of independent transpilation.

I found [this article helpful](https://ncjamieson.com/dont-export-const-enums/) for understanding the effect of specifying const.

By removing the const keyword, the enums can safely be used in projects using isolatedModules. This change is similar to https://github.com/microsoft/rushstack/pull/3060 and https://github.com/microsoft/rushstack/pull/3074.

## How it was tested
Build passes.

### DIFF
--- a/apps/api-extractor/src/api/ConsoleMessageId.ts
+++ b/apps/api-extractor/src/api/ConsoleMessageId.ts
@@ -11,7 +11,7 @@
  *
  * @public
  */
-export const enum ConsoleMessageId {
+export enum ConsoleMessageId {
   /**
    * "Analysis will use the bundled TypeScript version ___"
    */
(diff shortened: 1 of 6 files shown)


IIIMADDINIII:

iclanton (approved):
This change should be fine in this package.

PR #4548, opened 2024-02-27
[rush] Minor cleanup to the build cache code.

iclanton:
## Summary

Minor cleanup to the build cache code.

## How it was tested

It will be tested by the PR build.

## Impacted documentation

None.

### DIFF
--- a/libraries/rush-lib/src/logic/operations/CacheableOperationPlugin.ts
+++ b/libraries/rush-lib/src/logic/operations/CacheableOperationPlugin.ts
@@ -31,10 +31,9 @@ import type {
 import type { IPhase } from '../../api/CommandLineConfiguration';
 import type { IRawRepoState, ProjectChangeAnalyzer } from '../ProjectChangeAnalyzer';
 import type { OperationMetadataManager } from './OperationMetadataManager';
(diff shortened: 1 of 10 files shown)


libraries/rush-lib/src/logic/operations/CacheableOperationPlugin.ts
```diff
@@ -585,13 +590,13 @@ export class CacheableOperationPlugin implements IPhasedCommandPlugin {
...
     if (!buildCacheContext.projectBuildCache) {
-      const { cacheDisabledReason } = buildCacheContext;
+      const { cacheDisabledReason, operationSettings, projectChangeAnalyzer, getCacheEntryId } =
+        buildCacheContext;
       if (cacheDisabledReason) {
         terminal.writeVerboseLine(cacheDisabledReason);
         return;
       }
 
-      const { operationSettings, projectChangeAnalyzer } = buildCacheContext;
```
dmichon-msft, line 594: These were separated because of the early returns, so that the (admittedly negligible) cost of the property accesses doesn't happen in those code paths, but also to keep the meaning of the variables closer to when they are used.

dmichon-msft (approved):
Seems fine, but an explanation of the motivation of the feature would be appreciated.

PR #4561, opened 2024-03-01
[ts-command-line] Add support for required env var-backed parameters.

iclanton:
## Summary

Built on https://github.com/microsoft/rushstack/pull/4560.

This PR allows parameters that may be filled by an environment variable to be marked as required.

## Details

This PR sets the `required` property to `false` only during parsing for required parameters with env vars, and then explicitly reports errors for missing values after parsing.

## How it was tested

Added unit tests.

## Impacted documentation

API docs should be updated.

### DIFF
--- a/libraries/ts-command-line/src/parameters/BaseClasses.ts
+++ b/libraries/ts-command-line/src/parameters/BaseClasses.ts
@@ -96,7 +96,7 @@ export abstract class CommandLineParameterBase {
   /**
    * @internal
    */
-  public _postParseValidation: ((value: unknown | undefined) => void) | undefined;
+  public _validateValue: ((value: unknown | undefined) => void) | undefined;
 
   /** {@inheritDoc IBaseCommandLineDefinition.parameterLongName} */
   public readonly longName: string;
--- a/libraries/ts-command-line/src/parameters/BaseClasses.ts
+++ b/libraries/ts-command-line/src/parameters/BaseClasses.ts
@@ -83,6 +83,11 @@ export abstract class CommandLineParameterBase {
    */
   public _parserKey: string | undefined;
 
+  /**
+   * @internal
+   */
+  public _postParseValidation: ((value: unknown | undefined) => void) | undefined;
+
   /** {@inheritDoc IBaseCommandLineDefinition.parameterLongName} */
   public readonly longName: string;
 
@@ -153,15 +158,6 @@ export abstract class CommandLineParameterBase {
     }
 
     if (this.environmentVariable) {
-      if (this.required) {
-        // TODO: This constraint is imposed only because argparse enforces "required" parameters, but
-        // it does not know about ts-command-line environment variable mappings.  We should fix this.
-        throw new Error(
-          `An "environmentVariable" cannot be specified for "${this.longName}"` +
-            ` because it is a required parameter`
-        );
-      }
-
       if (
         !this.allowNonStandardEnvironmentVariableNames &&
         !ENVIRONMENT_VARIABLE_NAME_REGEXP.test(this.environmentVariable)
@@ -200,6 +196,12 @@ export abstract class CommandLineParameterBase {
    * @internal
    */
   public abstract _setValue(data: unknown): void;
+
+  /**
+   * Called internally by CommandLineParameterProvider._processParsedData()
+   * @internal
+   */
+  public abstract _getValue(): unknown;
 
   /**
    * Returns additional text used by the help formatter.
--- a/libraries/ts-command-line/src/parameters/BaseClasses.ts
+++ b/libraries/ts-command-line/src/parameters/BaseClasses.ts
@@ -86,17 +86,17 @@ export abstract class CommandLineParameterBase {
   /**
    * @internal
    */
-  public _preParse: (() => void) | undefined;
+  public _preParse?: () => void;
 
   /**
    * @internal
    */
-  public _postParse: (() => void) | undefined;
+  public _postParse?: () => void;
 
   /**
    * @internal
    */
-  public _validateValue: ((value: unknown | undefined) => void) | undefined;
+  public _validateValue?: () => void;
 
   /** {@inheritDoc IBaseCommandLineDefinition.parameterLongName} */
   public readonly longName: string;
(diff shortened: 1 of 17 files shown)


libraries/ts-command-line/src/providers/CommandLineParameterProvider.ts
```diff
@@ -87,6 +87,12 @@ const LONG_NAME_GROUP_NAME: string = 'longName';
 const POSSIBLY_SCOPED_LONG_NAME_REGEXP: RegExp =
   /^--((?<scope>[a-z0-9]+(-[a-z0-9]+)*):)?(?<longName>[a-z0-9]+((-[a-z0-9]+)+)?)$/;
 
+interface IExtendedArgumentGroup extends argparse.ArgumentGroup {
+  // The types are incorrect - this function returns the constructed argument
```
D4N14L, line 91: What does this return according to the type?

libraries/ts-command-line/src/providers/CommandLineParameterProvider.ts
```diff
@@ -668,6 +684,9 @@ export abstract class CommandLineParameterProvider {
     for (const parameter of this._parameters) {
       const value: unknown = data[parameter._parserKey!];
       parameter._setValue(value);
+      if (parameter._postParseValidation) {
+        parameter._postParseValidation(parameter._getValue());
+      }
```
D4N14L, line 689: Can we `parameter._postParseValidation?.(parameter._getValue())` here to match the preparse styling?

libraries/ts-command-line/src/providers/CommandLineParameterProvider.ts
```diff
@@ -833,16 +854,31 @@ export abstract class CommandLineParameterProvider {
...
-    argumentGroup.addArgument(names, { ...argparseOptions });
+    const argparseArgument: argparse.ArgumentOptions = (argumentGroup as IExtendedArgumentGroup).addArgument(
+      names,
+      argparseOptions
+    );
+    if (required && environmentVariable) {
+      parameter._preParse = () => {
+        argparseArgument.required = false;
+      };
+      parameter._postParseValidation = (value: unknown | null | undefined) => {
```
D4N14L, line 874: Seems like an odd place to specify this... could we specify the pre/postParse on the base parameter instead? If not, this will work, but it does seem odd to be touching the parameter after registration

libraries/ts-command-line/src/providers/CommandLineParser.ts
```diff
@@ -239,6 +239,11 @@ export abstract class CommandLineParser extends CommandLineParameterProvider {
         }
       }
 
+      this._preParse();
+      for (const action of this.actions) {
+        action._preParse();
+      }
```
D4N14L, line 245: Preparse is handled here, why not postparse?

libraries/ts-command-line/src/providers/CommandLineParameterProvider.ts
iclanton: The types say it returns `void`

D4N14L: Oh that's silly

libraries/ts-command-line/src/providers/CommandLineParameterProvider.ts
iclanton: It only needs to do something in this particular case (env var+required), and it needs the argparse argument object, so defining it in the base class will be inconvenient and slower.

libraries/ts-command-line/src/parameters/BaseClasses.ts
```diff
@@ -67,6 +67,16 @@ export abstract class CommandLineParameter {
...
 
+  /**
+   * @internal
+   */
+  public _preParse: (() => void) | undefined;
+
+  /**
+   * @internal
+   */
+  public _postParseValidation: ((value: unknown | undefined) => void) | undefined;
```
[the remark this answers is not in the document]
iclanton: Added a `_postParse` that does something else.

D4N14L (approved):

libraries/ts-command-line/src/providers/CommandLineParameterProvider.ts
```diff
@@ -832,16 +866,58 @@ export abstract class CommandLineParameterProvider {
...
+    if (required && environmentVariable) {
+      // Add some special-cased logic to handle required parameters with environment variables
+      parameter._preParse = () => {
+        // Set the value as non-required before parsing. We'll validate it explicitly
+        argparseArgument.required = false;
+      };
+      parameter._postParse = () => {
+        // Reset the required value to make the usage text correct
+        argparseArgument.required = true;
+      };
```
D4N14L, line 891: These should probably call the original methods before performing the additional work

libraries/ts-command-line/src/providers/CommandLineParameterProvider.ts
```diff
@@ -832,16 +866,58 @@ export abstract class CommandLineParameterProvider {
...
+            }
+          };
+          break;
+        case CommandLineParameterKind.ChoiceList:
+        case CommandLineParameterKind.IntegerList:
+        case CommandLineParameterKind.StringList:
+          parameter._validateValue = function () {
+            if (this.values.length === 0) {
+              throwMissingParameterError();
+            }
```
D4N14L, line 916: Same for the validation

D4N14L (approved):

libraries/ts-command-line/src/providers/CommandLineParameterProvider.ts
```diff
@@ -881,19 +881,26 @@ export abstract class CommandLineParameterProvider {
     );
     if (required && environmentVariable) {
       // Add some special-cased logic to handle required parameters with environment variables
+
+      const originalPreParse: (() => void) | undefined = parameter._preParse?.bind(parameter);
       parameter._preParse = () => {
         // Set the value as non-required before parsing. We'll validate it explicitly
         argparseArgument.required = false;
+        originalPreParse?.();
```
D4N14L, line 889: Wondering if this should go before our tweaking? The latter one happens after we revert our tweaking, so to the implementor, nothing has changed

D4N14L (approved):

PR #4567, opened 2024-03-04
[rush] Fix an issue where deleted patches weren't deleted from `common/temp/patches`.

iclanton:
## Summary

There is currently an issue where if a patch is removed from `common/pnpm-patches` after `rush install` had already been run on a machine with that patch present, pnpm would try to continue applying the patch.

This PR fixes that issue by deleting the `common/temp/patches` folder if `common/pnpm-patches` doesn't exist, and deleting items from `common/temp/patches` if they no longer exist in `common/pnpm-patches`.

## How it was tested

Tested by running `rush install` in a repo with a patch, deleting the patch, and rerunning `rush install` and observing the output.

### DIFF
--- a/libraries/rush-lib/src/logic/base/BaseInstallManager.ts
+++ b/libraries/rush-lib/src/logic/base/BaseInstallManager.ts
@@ -13,7 +13,9 @@ import {
   NewlineKind,
   AlreadyReportedError,
   type FileSystemStats,
-  Path
+  Path,
+  type FolderItem,
+  Async
 } from '@rushstack/node-core-library';
 import {
   PrintUtilities,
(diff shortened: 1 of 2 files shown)


libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -405,17 +407,57 @@ export abstract class BaseInstallManager {
...
+          async (patch: FolderItem) => {
+            const name: string = patch.name;
+            if (!copiedPatchNames.has(name)) {
+              await FileSystem.deleteFileAsync(`${commonTempPnpmPatchesFolder}/${name}`);
+            }
+          },
+          { concurrency: 50 }
+        );
+      } else {
+        await FileSystem.deleteFolderAsync(commonTempPnpmPatchesFolder);
```
D4N14L, line 460: Wondering if we should just do this and then copy the existing patches into place instead of tracking which existing patches weren't copied in?

D4N14L (approved):

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
iclanton: I thought about doing that, but the Windows virus scanner sometimes doesn't like when you delete a file or folder and then immediately recreate it.

PR #4570, opened 2024-03-05
[build-cache] Update how cobuilds cache builds that succeed with warnings

### LINKED ISSUE #4558: [rush] Rush should not write to the build cache when build contains warnings
## Summary

Our monorepo has been experiencing issues with linting regressions making their way into the master branch. We believe it is due to the way that Rush continues to write to the build cache even when a build fails on warnings. This causes an edge case where our build will fail due to the linting error, but if the user makes another commit (e.g. rebasing master) that doesn't trigger changes in the project, it will build successfully since it will just pull from the build cache and assume that it is pulling a successful build from the build cache.

## Repro steps

1. Create a PR that has changes that contain a linting warning for a project (this build fails)
2. Push some changes that do not trigger the cache for that project (e.g. rebasing master) (this build succeeds)
3. Validate if the build for the latest commit succeeds and pulls from the build cache for the project that contains the linting error

  **Expected result:** When Rush experiences a warning during the build phase, it does not write to the build cache. 
  **Actual result:** When Rush experiences a warning during the build phase, it continues to write to the build cache.

## Standard questions

| Question | Answer |
| -------- | -------- |
| `rushVersion` from rush.json? | 5.109.1 |
| Operating system? | Mac |
| Node.js version (`node -v`)? | 18.17.0 |

aramissennyeydd:
## Summary

Fixes #4558.
We're been running into an issue intermittently where linting errors were getting put into the build cache in CI. This caused issues as running the CI a second time would cause the build to pass and if that code got to main, developer machines would pull that code from the build cache and get warnings for untouched code. This fix changes how cobuilds assign cacheIds to builds to separate `Success` from `SuccessWithWarnings` unless the user explicitly enables `allowWarningsOnSuccess`.

## Details

As I explain above, we've had a couple of issues with the build cache accepting builds that had warnings and then passing in later runs of the CI. Those build artifacts then get pulled into developer machines and start breaking if the developer were to run `rush rebuild`. We took an initial look at the code and for non-cobuilds everything looks good, but there's a case where we transfer the cobuild state back to the primary and write to the cache that seems to skip success with warnings and treats them as just success. From my understanding of the cobuild system, it's necessary to write the build artifact, but for non-passing outputs it shouldn't write to the same cacheID as a passing output. AFAIK there are no alternative approaches to fix this. This _could_ be a breaking change if users rely on this behavior, but I would expect users relying on this behavior to have set the `runner.warningsAreAllowed` option (`allowWarningsInSuccessfulBuild`).

## How it was tested

I monkey patched the rush version running on our CI with this fix and linting errors were not written to the build cache, subsequent retries of the same CI pipeline failed as expected.

### DIFF
--- a/common/changes/@microsoft/rush/fix-cobuild-build-cache_2024-03-05-17-28.json
+++ b/common/changes/@microsoft/rush/fix-cobuild-build-cache_2024-03-05-17-28.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "fixes an issue where cobuilds would write success with warnings as success cache entries unexpectedly.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/common/changes/@microsoft/rush/fix-cobuild-build-cache_2024-03-05-17-28.json
+++ b/common/changes/@microsoft/rush/fix-cobuild-build-cache_2024-03-05-17-28.json
@@ -2,8 +2,8 @@
   "changes": [
     {
       "packageName": "@microsoft/rush",
-      "comment": "fixes an issue where cobuilds would write success with warnings as success cache entries unexpectedly.",
-      "type": "patch"
+      "comment": "Fixes an issue where cobuilds would write success with warnings as successful cache entries.",
+      "type": "none"
     }
   ],
   "packageName": "@microsoft/rush"
--- a/common/changes/@microsoft/rush/fix-cobuild-build-cache_2024-03-05-17-28.json
+++ b/common/changes/@microsoft/rush/fix-cobuild-build-cache_2024-03-05-17-28.json
@@ -3,7 +3,7 @@
     {
       "packageName": "@microsoft/rush",
       "comment": "fixes an issue where cobuilds would write success with warnings as success cache entries unexpectedly.",
-      "type": "none"
+      "type": "patch"
     }
   ],
   "packageName": "@microsoft/rush"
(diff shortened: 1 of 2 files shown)


aramissennyeydd:
@microsoft-github-policy-service agree company="DoorDash"

common/changes/@microsoft/rush/fix-cobuild-build-cache_2024-03-05-17-28.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "fixes an issue where cobuilds would write success with warnings as success cache entries unexpectedly.",
+      "type": "patch"
```
iclanton, line 6:
```suggestion
      "comment": "Fixes an issue where cobuilds would write success with warnings as successful cache entries.",
      "type": "none"
```

The Rush release type is dictated by the version-policies.json file

iclanton (approved):

common/changes/@microsoft/rush/fix-cobuild-build-cache_2024-03-05-17-28.json
aramissennyeydd: Thanks 😅 ! New to the versioning side of things.

aramissennyeydd:
@iclanton Will this get picked up in a release this week? I'd love to update our instance to the new version and move away from our monkeypatch 😅

iclanton:
This was just published in Rush `5.117.6`.

aramissennyeydd:
Thanks!

PR #4592, opened 2024-03-21
[rush] Skip determining merge base if given git hash

### LINKED ISSUE #4589: [rush] Rush shouldn't need to determine the merge base if commit hash has been provided
## Summary

I was meeting issue like https://github.com/microsoft/rushstack/issues/2082

However, in this case, I already know the merge base commit (because it is provided as env variable by the CI Runner).

More specifically, in my setup:

I am running in Gitlab CI. CI runner is set to shallow clone with depth = 1. The scripts that I run are:

```sh
git fetch origin $CI_MERGE_REQUEST_DIFF_BASE_SHA
git fetch origin $CI_COMMIT_SHA
node common/scripts/install-run-rush.js install
node common/scripts/install-run-rush.js build --verbose --to git:$CI_MERGE_REQUEST_DIFF_BASE_SHA
```

The first two commands tells the CI runner to fetch the two specific commits (the base and the current commit). The values of both variables are provided by the CI runner.

Rush threw error on the 4th command:

```
The rush.json configuration requests Rush version 5.117.8
Invoking "rush build --verbose --to git:REDACTED_GIT_COMMIT_HASH"
---------------------------------------------------------------------------------
Rush Multi-Project Build Tool 5.117.8 - https://rushjs.io
Node.js version is 18.19.1 (LTS)
Starting "rush build"
Unable to determine merge base for branch "REDACTED_GIT_COMMIT_HASH". This can occur if the current clone is a shallow clone. If this clone is running in a CI pipeline, check your pipeline settings to ensure that the clone depth includes the expected merge base. If this clone is running locally, consider running "git fetch --deepen=<depth>".
```

## Details

When I investigated the rush source codes, I found that Rush seem to take the provided option as branch name. However, if I pass in a commit hash, Rush shouldn't need to determine the merge base anymore:

https://github.com/microsoft/rushstack/blob/0901985866434a113c68d141e969113817a21b0e/libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts#L225

Skipping this part where it determines the merge base will speed up the CI time too since we can fully use a shallow clone of the repository in CI.

Though... I wonder if there's a reliable way to determine whether a provided git expression is a commit or a branch name 🤔 🤔 

p.s. specific to my setup, I can do things like `git fetch origin $CI_COMMIT_SHA --deepen=10` and it will work-around the issue

## Standard questions

| Question | Answer |
| -------- | -------- |
| `@microsoft/rush` globally installed version? | 5.117.8 |
| `rushVersion` from rush.json? | 5.117.8 |
| Operating system? | MacOS |
| Node.js version (`node -v`)? | 18.19.1 |

kenrick95:
## Summary

Fixes #4589 

## Details

In `libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts`, I added a logic to skip determining merge base if the given "targetBranchName" is already a git commit hash

## How it was tested

I tested the local build against my own repro in #4589 and it works as expected. However I'd like some advise too on how to write test for this change.

## Impacted documentation

N/A

### DIFF
--- a/libraries/rush-lib/src/logic/Git.ts
+++ b/libraries/rush-lib/src/logic/Git.ts
@@ -595,4 +595,23 @@ export class Git {
       throw e;
     }
   }
+  /**
+   *
+   * @param ref Given a ref which can be branch name, commit hash, tag name, etc, check if it is a commit hash
+   */
+  public isRefACommit(ref: string): boolean {
+    const gitPath: string = this.getGitPathOrThrow();
+    try {
+      const output: string = this._executeGitCommandAndCaptureOutput(gitPath, ['rev-parse', '--verify', ref]);
+      const result: string = output.trim();
+
+      if (result === ref) {
+        return true;
+      }
+    } catch (e) {
+      // assume not a commit
+      return false;
+    }
+    return false;
+  }
 }
--- a/common/changes/@microsoft/rush/kenrick-skip-merge-base_2024-03-21-02-17.json
+++ b/common/changes/@microsoft/rush/kenrick-skip-merge-base_2024-03-21-02-17.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Skip determining merge base if given git hash",
+      "type": "none"
+    }
+  ],
+  "packageName": "@microsoft/rush"
+}
--- a/libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
+++ b/libraries/rush-lib/src/logic/ProjectChangeAnalyzer.ts
@@ -222,7 +222,10 @@ export class ProjectChangeAnalyzer {
     const gitPath: string = this._git.getGitPathOrThrow();
     const repoRoot: string = getRepoRoot(rushConfiguration.rushJsonFolder);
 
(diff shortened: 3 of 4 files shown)


libraries/rush-lib/src/logic/Git.ts
```diff
@@ -595,4 +595,23 @@ export class Git {
       throw e;
     }
   }
+  /**
+   *
+   * @param ref Given a ref which can be branch name, commit hash, tag name, etc, check if it is a commit hash
+   */
+  public isRefACommit(ref: string): boolean {
```
dmichon-msft, line 602: To unit test this, you can pass in `HEAD` for an input that will always be valid, and the result of `git rev-parse HEAD` for a valid commit id.

iclanton: @dmichon-msft - Do you know of a single git command that will work for both of these scenarios (finding a merge base or confirming that something is a commit SHA)?

dmichon-msft: Unfortunately not.

kenrick95: Added some unit tests, please help to review if that's satisfactory enough 🙏

dmichon-msft (approved):

octogonz (approved):

PR #4593, opened 2024-03-21
[rush] Add a policy to forbid sha1 hashes in pnpm-lock.yaml

L-Qun:
## Summary

Add support for `pnpm-lock.yaml` file validation mechanism.

## Details

Check the `pnpm-lock.yaml` against the rules defined in the `pnpm-lock-validation.json` file and displays error messages according to the definitions in the `custom-tips.json` file.

## How it was tested

First, add a new configuration item "pnpmLockfilePolicies": ["disallowInsecureSha1"], in the `common/config/rush/pnpm-config.json` file. For example:

```
{
  "$schema": "https://raw.githubusercontent.com/microsoft/rushstack/main/libraries/rush-lib/src/schemas/pnpm-config.schema.json",

  "pnpmLockfilePolicies": ["disallowInsecureSha1"],
}
```

Then,  running `rush update` in a Rush project with a `pnpm-lock.yaml` file containing integrity calculated using the SHA1 algorithm will result in an error, and an error message will be displayed according to the `custom-tips.json` file:

image

## Impacted documentation

- New config file setting
- New custom tip ID


L-Qun:
> @L-Qun please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 

@microsoft-github-policy-service agree company="TikTok"

libraries/rush-lib/src/api/PnpmLockValidationConfiguration.ts
```diff
@@ -0,0 +1,86 @@
...
+import type { Lockfile } from '@pnpm/lockfile-types';
+import type { ITerminal } from '@rushstack/terminal';
+
+import schemaJson from '../schemas/pnpm-lock-validation.schema.json';
+import type { RushConfiguration } from './RushConfiguration';
+import { CustomTipId } from './CustomTipsConfiguration';
+
+export interface IPnpmValidationJson {
+  pnpmLockValidationRules?: Array<{
+    ruleName: 'validateIntegritySha1';
```
octogonz, line 16: Consider a rule name that better indicates the goal, something like `requireSecureShaHashes` or `disallowInsecureSha1`

libraries/rush-lib/src/api/PnpmLockValidationConfiguration.ts
```diff
@@ -0,0 +1,86 @@
...
+    description: string;
+    pattern: string;
```
octogonz, line 18: If the purpose of this rule is to prevent SHA1 hashes, why should the user need to provide a RegExp?  It should be Rush's job to ensure that SHA1 is eliminated, and if the lockfile format changes, it should be Rush's responsibility to improve the RegExp.

On the other hand, we could instead consider a more flexible rule like `forbidRegExpInLockfile`, however flexibility is often a worse user experience.  Where possible, it's better for Rush to _**solve a problem well**_ rather than merely _**enabling users to solve the problem themselves.**_ The reason is that solving problems _**well**_ tends to be a lot of work. 😁

libraries/rush-lib/src/schemas/pnpm-lock-validation.schema.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Rush pnpm-lock-validation.schema.json config file",
+  "description": "This configuration file defines custom validation rules for pnpm-lock.yaml files in Rush projects. Each rule specifies a regular expression pattern that will be used to check the contents of pnpm-lock.yaml for compliance with project standards.",
+  "type": "object",
+  "properties": {
+    "pnpmLockValidationRules": {
```
octogonz, line 7: Although we may implement more validations in the future, I'm not sure their configuration will be complex enough to require its own config file. What if we just added this setting to the existing **pnpm-config.json** config file?

It is good to organize Rush's configuration into multiple config files, however each file adds some overhead that increases the startup time for Rush, so there is a balance.

libraries/rush-lib/src/schemas/pnpm-lock-validation.schema.json
```diff
@@ -0,0 +1,30 @@
...
+          "pattern": {
+            "type": "string",
+            "description": "The regular expression pattern to use for validation."
+          }
+        },
+        "required": ["ruleName", "pattern"]
+      }
+    }
+  },
+  "required": ["pnpmLockValidationRules"]
```
octogonz, line 29: Suggest to call this setting something like `lockfilePolicyRules` or `pnpmLockfilePolicies`.  Rush uses the term "policy" to mean extra opt-in requirements that prevent common problems.  The term "validation" can mean the same thing, but generally "validation" is more about detecting actual mistakes or misconfigurations.

libraries/rush-lib/src/schemas/pnpm-lock-validation.schema.json
```diff
@@ -0,0 +1,30 @@
+{
```
octogonz, line 1: Please update the `rush init` template to include your new setting:

https://github.com/microsoft/rushstack/blob/main/libraries/rush-lib/assets/rush-init/common/config/rush/pnpm-config.json

I can help with docs

octogonz:
Some context about this feature:

**pnpm-lock.yaml** stores hashes to verify the integrity of downloaded tarballs.  These hashes are received from the NPM registry.  NPM's original implement used SHA1 hashes, which is an insecure hashing mechanism.  "Insecure" means that the hash can be reversed -- for a given hash, it is possible to craft a malicious tarball that has the same hash as the real NPM version.  This security weakness does not affect the usefulness of SHA1 as a checksum to detect corrupted/truncated downloads. However it does pose some hypothetical security vulnerabilities:

- If a company's private NPM registry has private packages whose name is not registered in the public NPM registry, an attacker could publish a public package with the same name/version, hoping that a computer with a misconfigured `.npmrc` will accidentally download from the `npmjs.com` instead of the private registry.
- If an attacker could publish malicious NPM packages to a company's private NPM registry, then development environments would probably be quickly compromised because they upgrade very often.  However a strong SHA hash could protect a locked down release environment, which would prevent customers from being impacted by the attack.

For these reasons NPM later switched to a SHA512 hash.  However although modern NPM registries now serve SHA512 instead of SHA1, not all registries retroactively replace SHA1 in their database of previous versions. And caches often preserve SHA1 hashes. For this reason, even though its been years since the switch, SHA1 does frequently appear in large monorepo lockfiles.

This PR implements a security compliance policy to detect and prevent SHA1.

common/config/rush/browser-approved-packages.json
```diff
@@ -18,6 +18,10 @@
       "name": "@microsoft/rush-sdk",
       "allowedCategories": [ "tests" ]
     },
+    {
+      "name": "@pnpm/lockfile-types",
```
iclanton, line 22: This should go in nonbrowser-approved-packages

libraries/rush-lib/package.json
```diff
@@ -78,7 +78,8 @@
     "@types/tar": "6.1.1",
     "@types/uuid": "~8.3.4",
     "@types/webpack-env": "1.18.0",
-    "webpack": "~5.82.1"
+    "webpack": "~5.82.1",
+    "@pnpm/lockfile-types": "~5.1.5"
```
iclanton, line 82: Sort?

libraries/rush-lib/src/logic/base/BaseInstallManager.ts
```diff
@@ -227,8 +228,14 @@ export abstract class BaseInstallManager {
       ]);
 
       if (this.options.allowShrinkwrapUpdates && !shrinkwrapIsUpToDate) {
+        const committedShrinkwrapFilename: string = subspace.getCommittedShrinkwrapFilename();
+        PnpmLockfileConfiguration.pnpmLockValidation(
```
iclanton, line 232:
```suggestion
        PnpmLockfileConfiguration.validateLockfile(
```

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+// See LICENSE in the project root for license information.
+import { AlreadyReportedError } from '@rushstack/node-core-library';
+import fs from 'fs';
```
iclanton, line 4: Use FileSystem

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { RushConfiguration } from '../../api/RushConfiguration';
+import { CustomTipId } from '../../api/CustomTipsConfiguration';
+import type { PnpmLockfilePolicy } from './PnpmOptionsConfiguration';
+
+export class PnpmLockfileConfiguration {
+  /**
+   * Determine whether `pnpm-lock.yaml` complies with the rules specified in `common/config/rush/pnpm-config.schema.json`.
+   * @internal
+   */
+  public static pnpmLockValidation(
```
iclanton, line 18: If this is touching the filesystem, make it async.

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
```diff
@@ -0,0 +1,54 @@
...
+    terminal: ITerminal,
+    rushConfiguration: RushConfiguration,
+    filename: string
+  ): void {
+    const pnpmLockfilePolicies: PnpmLockfilePolicy[] | undefined =
+      rushConfiguration.pnpmOptions.pnpmLockfilePolicies;
+    if (pnpmLockfilePolicies && pnpmLockfilePolicies.length > 0) {
+      const lockfileRawContent: string = fs.readFileSync(filename, 'utf-8');
+      const lockfile: Lockfile = yaml.load(lockfileRawContent);
+      pnpmLockfilePolicies.forEach((policy) => this[policy](terminal, rushConfiguration, lockfile));
```
iclanton, line 28: Just explicitly call the policy validation functions instead of trying to get then from `this`.

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
```diff
@@ -0,0 +1,54 @@
...
+   * @internal
+   */
+  public static disallowInsecureSha1(
+    terminal: ITerminal,
+    rushConfiguration: RushConfiguration,
+    lockfile: Lockfile
+  ): void {
+    const { packages } = lockfile;
+    if (packages) {
+      Object.entries(packages).forEach(([, { resolution }]) => {
```
iclanton, line 43: Write this as a conventional loop to avoid so many stackframes.

Also consider using `Object.values`

libraries/rush-sdk/package.json
```diff
@@ -40,6 +40,7 @@
     "@rushstack/stream-collator": "workspace:*",
     "@rushstack/ts-command-line": "workspace:*",
     "@types/semver": "7.5.0",
-    "@types/webpack-env": "1.18.0"
+    "@types/webpack-env": "1.18.0",
+    "@pnpm/lockfile-types": "~5.1.5"
```
iclanton, line 44: Unused by this package?

common/reviews/api/rush-lib.api.md
```diff
@@ -674,6 +676,8 @@ export interface _IPnpmOptionsJson extends IPackageManagerOptionsJsonBase {
     globalPatchedDependencies?: Record<string, string>;
     // Warning: (ae-forgotten-export) The symbol "IPnpmPeerDependencyRules" needs to be exported by the entry point index.d.ts
     globalPeerDependencyRules?: IPnpmPeerDependencyRules;
+    // Warning: (ae-forgotten-export) The symbol "PnpmLockfilePolicy" needs to be exported by the entry point index.d.ts
```
iclanton, line 679: Fix this issue.

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
iclanton: Also write this as a conventional loop.

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
L-Qun: I think FileSystem.readFile is a synchronous function?

octogonz: `FileSystem.readFileAsync()` is the `async` version.  (Each API has both sync and async versions.)

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
```diff
@@ -0,0 +1,58 @@
...
+import type { RushConfiguration } from '../../api/RushConfiguration';
+import { CustomTipId } from '../../api/CustomTipsConfiguration';
+import type { PnpmLockfilePolicy } from './PnpmOptionsConfiguration';
+
+export class PnpmLockfileConfiguration {
+  /**
+   * Determine whether `pnpm-lock.yaml` complies with the rules specified in `common/config/rush/pnpm-config.schema.json`.
+   * @internal
+   */
+  public static validateLockfile(
```
dmichon-msft, line 18: Add your validation logic to the existing lockfile validator function here: https://github.com/microsoft/rushstack/blob/8fe0ea8f861a1cecd8b964bcdabc610f2a5574e3/libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts#L340

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
L-Qun: > `FileSystem.readFileAsync()` is the `async` version. (Each API has both sync and async versions.)

I see, but I use sync version, right?

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -311,6 +321,15 @@ export class PnpmOptionsConfiguration extends PackageManagerOptionsConfiguration
...
   public readonly jsonFilename: string | undefined;
 
+  /**
+   * The `pnpmLockfilePolicies` setting defines the policies that govern the `pnpm-lock.yaml` file.
+   *
+   * @remarks
+   * Available options:
+   *  - disallowInsecureSha1: Forbid sha1 hashes in `pnpm-lock.yaml`
+   */
+  public readonly pnpmLockfilePolicies: PnpmLockfilePolicy[] | undefined;
```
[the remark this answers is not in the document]
L-Qun: It's ok, I will change this

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
```diff
@@ -0,0 +1,58 @@
...
+  ): void {
+    const { packages } = lockfile;
+    if (packages) {
+      for (const { resolution } of Object.values(packages)) {
+        if ((resolution as { integrity: string }).integrity.startsWith('sha1')) {
+          rushConfiguration.customTipsConfiguration._showErrorTip(
+            terminal,
+            CustomTipId.TIP_PNPM_DISALLOW_INSECURE_SHA1
+          );
+          throw new AlreadyReportedError();
```
iclanton, line 53: Instead of throwing here, it seems useful to log issues for all policies and then throw afterward.

I've always found it annoying when a tool terminates after the first error.

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -32,6 +32,12 @@ export type PnpmStoreOptions = PnpmStoreLocation;
  */
 export type PnpmResolutionMode = 'highest' | 'time-based' | 'lowest-direct';
 
+/**
+ * Possible values for the `pnpmLockfilePolicies` setting in Rush's pnpm-config.json file.
+ * @public
+ */
+export type PnpmLockfilePolicy = 'disallowInsecureSha1';
```
iclanton, line 39: "disallowInsecureSha1" implies that there are secure SHA1 hashes.

Maybe call this just "disallowSha1"? @octogonz?

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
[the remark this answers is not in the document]
iclanton: I like that too.

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
```diff
@@ -0,0 +1,58 @@
...
+  public static disallowInsecureSha1(
+    terminal: ITerminal,
+    rushConfiguration: RushConfiguration,
+    lockfile: Lockfile
+  ): void {
+    const { packages } = lockfile;
+    if (packages) {
+      for (const { resolution } of Object.values(packages)) {
+        if ((resolution as { integrity: string }).integrity.startsWith('sha1')) {
+          rushConfiguration.customTipsConfiguration._showErrorTip(
```
iclanton, line 49: What happens if someone turns this option on but doesn't have a custom tip for it configured?

Seems like this should just be a standard error message.

iclanton (dismissed):

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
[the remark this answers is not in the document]
chengcyber: +1 for dictionary fashion here.

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
L-Qun: image
No output. Do you think we should give a default message when user doesn't config a custuom tip?

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
octogonz: I thought that "insecure" would more clearly explain why someone would want to disallow it

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
iclanton: Yeah. If someone turns this option on, there should be something printed if the rule is violated, even if they don't have their own message configured.

octogonz: Here's an example of how `TIP_RUSH_INCONSISTENT_VERSIONS ` does that:

https://github.com/microsoft/rushstack/blob/caab398bce0a67e71aac48af9a7642955c387abf/libraries/rush-lib/src/logic/versionMismatch/VersionMismatchFinder.ts#L147-L159

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
L-Qun: @dmichon-msft 

It seems that `validate` function will not be called when we run `rush update`


image

L-Qun: But we want to perform validation when the user runs `rush update`. Cc @octogonz

libraries/rush-lib/src/logic/pnpm/PnpmLockfileConfiguration.ts
octogonz: > *intended to provide additional detail for existing error messages*

Sorry this was unclear. In 97f17e7c6b08094e2780e6fdd672f08898d59f40 I've better documented that Rush's standard message is not replaced by custom tips.

The UX now looks like this:

image

octogonz (approved):

libraries/rush-lib/assets/rush-init/common/config/rush/pnpm-config.json
```diff
@@ -252,5 +252,31 @@
     },
     "pnpm": { "futurePnpmFeature": true }
     /*[END "HYPOTHETICAL"]*/
+  },
+
+  /**
+   * Defines the policies to be checked for the `pnpm-lock.yaml` file.
+   */
+  "pnpmLockfilePolicies": {
```
iclanton, line 260: Given that this is in `pnpm-config.json`, this could probably just be called `lockfilePolicies`.

libraries/rush-lib/assets/rush-init/common/config/rush/pnpm-config.json
```diff
@@ -252,5 +252,31 @@
...
+     * its main purpose is to detect corrupted or truncated network requests, this hash can also
+     * serve as a security fingerprint to protect against attacks that would substitute a
+     * malicious tarball, for example if a misconfigured .npmrc caused a machine to accidentally
+     * download a matching package name+version from npmjs.com instead of the private NPM registry.
+     * NPM originally used a SHA1 hash; this was insecure because an attacker can too easily craft
+     * a tarball with a matching fingerprint.  For this reason, NPM later deprecated SHA1 and
+     * instead adopted a cryptographically strong SHA512 hash.  Nonetheless, SHA1 hashes can
+     * occasionally reappear during "rush update", for example due to missing metadata fallbacks
+     * (https://github.com/orgs/pnpm/discussions/6194) or an incompletely migrated private registry.
+     * Enable `disallowInsecureSha1` to avoid security/compliance alerts about this issue.
```
iclanton, line 276:
```suggestion
     * Enable `disallowInsecureSha1` prevent Rush from recording these insecure hashes, which will avoid security/compliance alerts about this issue.
```

libraries/rush-lib/src/api/CustomTipsConfiguration.ts
```diff
@@ -327,27 +346,26 @@ export class CustomTipsConfiguration {
...
+        break;
+      case CustomTipSeverity.Warning:
+        writeFunction = terminal.writeWarningLine.bind(terminal);
+        prefix = Colorize.yellow('| ');
+        break;
+      default:
+        writeFunction = terminal.writeLine.bind(terminal);
+        prefix = '| ';
+        break;
+    }
```
iclanton, line 367: Why did this get moved? We can avoid doing this work if it's not going to get used for anything,

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -311,6 +329,15 @@ export class PnpmOptionsConfiguration extends PackageManagerOptionsConfiguration
 
   public readonly jsonFilename: string | undefined;
 
+  /**
+   * The `pnpmLockfilePolicies` setting defines the policies that govern the `pnpm-lock.yaml` file.
+   *
+   * @remarks
+   * Available options:
+   *  - disallowInsecureSha1: Forbid sha1 hashes in `pnpm-lock.yaml`
```
iclanton, line 337: It seems like this comment is likely to get outdated when people add things to `IPnpmLockfilePolicies`.

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -33,20 +33,34 @@ export type PnpmStoreOptions = PnpmStoreLocation;
 export type PnpmResolutionMode = 'highest' | 'time-based' | 'lowest-direct';
 
 /**
- * @beta
+ * Possible values for the `pnpmLockfilePolicies` setting in Rush's pnpm-config.json file.
+ * @public
+ */
+export interface IPnpmLockfilePolicies {
+  disallowInsecureSha1?: boolean;
```
iclanton, line 40:
```suggestion
  /**
   * Forbid sha1 hashes in `pnpm-lock.yaml`
   */
  disallowInsecureSha1?: boolean;
```

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -336,6 +337,68 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
+   * @internal
+   */
+  private _disallowInsecureSha1(
+    customTipsConfiguration: CustomTipsConfiguration,
+    terminal: ITerminal
+  ): boolean {
+    const { packages } = this._shrinkwrapJson;
+    if (packages) {
+      for (const { resolution } of Object.values(packages)) {
+        if ((resolution as { integrity: string }).integrity.startsWith('sha1')) {
```
iclanton, line 351: If the resolution property is missing, this will fail.

```suggestion
        if ((resolution?.integrity.startsWith('sha1')) {
```

This type is

```
  /** Information about the resolved package */
  resolution?: {
    /** The hash of the tarball, to ensure archive integrity */
    integrity: string;
    /** The name of the tarball, if this was from a TGX file */
    tarball?: string;
  };
```

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -336,6 +337,68 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
+          terminal.writeErrorLine(
```
iclanton, line 352: It'd be really useful to tell people which package this showed up in, and, ideally, tell them all of the packages that have problematic hashes instead of just the first one.

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -336,6 +337,68 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
+              );
+              if (isError) {
+                invalidPoliciesCount += 1;
+              }
+              break;
+            }
+
+            default: {
+              throw new Error(`Unknown pnpm lockfile policy "${policy}"`);
+            }
```
iclanton, line 391: This should never happen.

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -336,6 +337,68 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
+          }
+        }
+      }
```
iclanton, line 394: This is an awkward way to go through an object with a well-known type. This is easier to read:

```suggestion
      if (pnpmLockfilePolicies.disallowInsecureSha1) {
        const isError: boolean = this._disallowInsecureSha1(
          rushConfiguration.customTipsConfiguration,
          terminal
        );
        if (isError) {
          invalidPoliciesCount++;
        }
      }
```

And then as more policies are added, they'll get included as their own `if` statements after this one.

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
iclanton: @octogonz pointed out that someone can just search for "sha1" in the lockfile. Fair enough.

libraries/rush-lib/src/logic/pnpm/PnpmOptionsConfiguration.ts
```diff
@@ -311,6 +332,15 @@ export class PnpmOptionsConfiguration extends PackageManagerOptionsConfiguration
 
   public readonly jsonFilename: string | undefined;
 
+  /**
+   * The `pnpmLockfilePolicies` setting defines the policies that govern the `pnpm-lock.yaml` file.
+   *
+   * @remarks
+   * Available options:
+   *  - disallowInsecureSha1: Forbid sha1 hashes in `pnpm-lock.yaml`
```
iclanton, line 340:
```suggestion
```

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
octogonz: You can search for "sha1" in the YAML file pretty easily.

Actually the really valuable information to print would be the file path of the lockfile, since Rush Subspaces introduces a world of multiple `pnpm-lock.yaml` getting installed at once.  However unfortunately some refactoring is needed to obtain that information here, which is better done in a separate PR and probably as part of the subspace effort.

iclanton (approved):

octogonz:
Documentation PR: https://github.com/microsoft/rushstack-websites/pull/224

octogonz:
JSON Schema PR: https://github.com/microsoft/json-schemas/pull/200

PR #4611, opened 2024-03-28
[eslint-patch] Fix a few minor issues with `eslint-bulk prune`.

iclanton:
## Summary

This PR fixes two issues with `eslint-bulk prune`:
- Currently the `prune` action causes the tool to crash if files that previously contained suppressions had been deleted. This is now fixed
- Updates the `prune` action logic to delete the `.eslint-bulk-suppressions.json` file if all previously-suppresses issues have been resolved.
 
## How it was tested

Tested by:
1. creating a TS file with eslint issues
2. introducing eslint issues in an existing file
3. running `eslint-bulk suppress --all src`
4. deleting the new file
5. running `eslint-bulk prune` and observing that the suppressions were trimmed only from that file
6. reverting the changes to the existing file
7. running `eslint-bulk prune` and observing that the bulk suppressions file is deleted

## Impacted documentation

Probably none.

### DIFF
--- a/eslint/eslint-patch/src/eslint-bulk-suppressions/bulk-suppressions-file.ts
+++ b/eslint/eslint-patch/src/eslint-bulk-suppressions/bulk-suppressions-file.ts
@@ -25,11 +25,18 @@ const IS_RUNNING_IN_VSCODE: boolean = process.env[VSCODE_PID_ENV_VAR_NAME] !== u
 const TEN_SECONDS_MS: number = 10 * 1000;
 const SUPPRESSIONS_JSON_FILENAME: string = '.eslint-bulk-suppressions.json';
 
+function throwIfAnythingOtherThanNotExistError(e: NodeJS.ErrnoException): void | never {
+  if (e?.code !== 'ENOENT') {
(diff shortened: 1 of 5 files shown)


eslint/eslint-patch/src/eslint-bulk-suppressions/bulk-suppressions-file.ts
```diff
@@ -25,11 +25,18 @@ const IS_RUNNING_IN_VSCODE: boolean = process.env[VSCODE_PID_ENV_VAR_NAME] !== u
 const TEN_SECONDS_MS: number = 10 * 1000;
 const SUPPRESSIONS_JSON_FILENAME: string = '.eslint-bulk-suppressions.json';
 
+function throwIfAnythingOtherThanNotExistError(e: NodeJS.ErrnoException): void | never {
+  if (e?.code !== 'ENOENT') {
```
D4N14L, line 29: I believe you should be able to use the method from `FileSystem.isNotFoundError` to check this, though I think this script is intended to be lightweight so maybe bundling that would be a bit much.

D4N14L (approved):

PR #4613, opened 2024-03-28
Gracefully exit from workers instead of using process.exit(0)

dmichon-msft:
## Summary
Updates various workers to properly clean up their ports instead of using `process.exit(0)`

## Details
Minifier worker (used in the module minifier plugins)
Transpiler worker (heft-typescript-plugin)
Updated WorkerPool to use `once` instead of `on` for monitoring the exit event.

## How it was tested
Local repeated runs of the build test projects that use these components.

## Impacted documentation
None

### DIFF
--- a/libraries/module-minifier/src/MinifierWorker.ts
+++ b/libraries/module-minifier/src/MinifierWorker.ts
@@ -14,8 +14,8 @@ process.exitCode = 2;
 
 async function handler(message: IModuleMinificationRequest): Promise<void> {
   if (!message) {
-    parentPort?.off('postMessage', handler);
-    parentPort?.close();
+    parentPort!.off('postMessage', handler);
+    parentPort!.close();
     return;
   }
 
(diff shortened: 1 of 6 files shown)


libraries/module-minifier/src/MinifierWorker.ts
```diff
@@ -12,12 +12,19 @@ const terserOptions: MinifyOptions = workerData;
 // Set to non-zero to help debug unexpected graceful exit
 process.exitCode = 2;
 
-parentPort!.on('message', async (message: IModuleMinificationRequest) => {
+async function handler(message: IModuleMinificationRequest): Promise<void> {
   if (!message) {
-    process.exit(0);
+    parentPort?.off('postMessage', handler);
+    parentPort?.close();
```
iclanton, line 18: Optional chaining here, but non-null assertion farther down?

iclanton (approved):

Issue #4620, opened 2024-04-02
[api-extractor] Allow generation of API reports per release level


Josmithr:
## Use Case

My team has begun using API-Extractor in our mono-repository to help manage our API surfaces via release-tag-based trimming. We plan to begin leveraging the generated API report files to enforce special PR review requirements for API changes.

We have gone through the process of tagging all of our exports with the appropriate release tags based on our support commitments, but we have a known problem with APIs over-exposing members to support testing, etc. This has resulted in huge numbers of `@internal`exports. While we eventually want to reign these imports in, right now we are more focused on stabilizing our externally facing APIs.

Since API-Extractor generates only a single API report including all API members (including internal ones), it is difficult for us to enable PR gating requirements based on changes to them. Ideally, we would be able to differentiate between release levels, and apply the appropriate gating on a per-release-level basis, which we would likely leverage for preview APIs as well.

## Feature Proposal

Just as API-Extractor can surface separate rollups on a per-release-level basis, operating on release tags, we would like to be able to generate separate API reports for each release level. With a separate report per level, we can enforce different levels of gating depending on the support guarantees we have made for each.

As an added benefit, each generated report would end up mirroring the rollups generated by API-Extractor’s trimming logic. Separate reports would offer a nice at-a-glance view of what will be in each of the typing rollups we end up offering to consumers.

## Standard questions

| Question | Answer |
| -------- | -------- |
| Would you consider contributing a PR? | Yes  |

Josmithr:
A prototype implementation of this functionality can be found here: #4621

Josmithr:
https://github.com/microsoft/rushstack/pull/4621 was merged, so this is now implemented.

PR #4621, opened 2024-04-02
[api-extractor] Allow generating API reports for different release levels

Josmithr:
Adds API-Extractor functionality to generate per-release-level API reports.

Implementation of #4620


build-tests/api-extractor-lib2-test/config/api-extractor.json
```diff
@@ -4,7 +4,10 @@
   "mainEntryPointFilePath": "<projectFolder>/lib/index.d.ts",
 
   "apiReport": {
-    "enabled": true
+    "enabled": true,
+    "alphaReportFileName": "<unscopedPackageName>.alpha.api.md",
+    "betaReportFileName": "<unscopedPackageName>.beta.api.md",
+    "publicReportFileName": "<unscopedPackageName>.public.api.md"
```
octogonz, line 10: >`"description": "The filename for the API report files.  It will be combined with \"reportFolder\" or \"reportTempFolder\" to produce a full file path. The file extension should be \".api.md\", and the string should not contain a path separator such as \"\\\" or \"/\".",`

Design question: Is it really necessary to specify `.alpha.api.md`, `.beta.api.md` in the configuration?  If API Extractor enforces that `reportFilename` MUST end with `.api.md` then an alternative design might be to specify the variations that you want to generate (`alpha`,`beta`,`public`) and then `.api.md` can be transformed to `.alpha.api.md` automatically.  A benefit of this approach is that API Extractor could warn about the presence of deadwood files (e.g. `my-package.beta.api.md` appears in the target folder but is not a configured output).

apps/api-extractor/src/generators/ApiReportGenerator.ts
```diff
@@ -426,15 +496,50 @@ export class ApiReportGenerator {
...
+        : apiItemMetadata.effectiveReleaseTag;
+
+    // If the declaration has a release tag that is not in scope, omit it from the report.
+    switch (releaseLevel) {
+      case ApiReportReleaseLevel.Untrimmed:
+        return true;
+      case ApiReportReleaseLevel.Alpha:
+        return releaseTag >= ReleaseTag.Alpha;
+      case ApiReportReleaseLevel.Beta:
+        return releaseTag >= ReleaseTag.Beta;
```
octogonz, line 537: Design question: Is it beneficial to duplicate the `@public` signatures in the `.beta.api.md` and `.alpha.md` files? From an approval standpoint, if I already approved the `.public.api.md` file then why do I need to reexamine the exact same API's in two other files?  Unlike a .d.ts rollup, there is no technical requirement for the .api.md report to include all of its dependent types.  Actually its main design goal is to avoid triggering spurious CODEOWNERS approvals.

apps/api-extractor/src/generators/ApiReportGenerator.ts
```diff
@@ -426,15 +496,50 @@ export class ApiReportGenerator {
...
+
+    // No specified release tag is considered the same as `@public`.
+    const releaseTag: ReleaseTag =
+      apiItemMetadata.effectiveReleaseTag === ReleaseTag.None
+        ? ReleaseTag.Public
+        : apiItemMetadata.effectiveReleaseTag;
+
+    // If the declaration has a release tag that is not in scope, omit it from the report.
+    switch (releaseLevel) {
+      case ApiReportReleaseLevel.Untrimmed:
```
octogonz, line 532: The term "trimming" has always referred to .d.ts rollups.  It may be confusing to introduce this term for .api.md files, because even though it involves deleting unnecessary declarations, the motivation and details are rather different in this context.

Compare with an alternative possible design like this:
```js
  // the base filename
  "reportFileName": "<unscopedPackageName>.api.md",
  
   // magically transforms to produce my-app.alpha.api.md and my-app.beta+public.api.md
  "reportVariations": [ "alpha", "beta+public" ]
```

build-tests/api-extractor-lib2-test/config/api-extractor.json
Josmithr: Yeah, if we wanted to enforce that all of these reports end in `api.md` (which doesn't seem unreasonable to me), that would simplify things some. It deviates from the existing pattern for the single report, but the benefits seem worth it. I'm in favor of that if you are.

apps/api-extractor/src/generators/ApiReportGenerator.ts
Josmithr: It's an interesting semantics question. I was operating with the following goals in mind:
1. To know what appears in a given API release level, you can look at a single report.
2. Each report mirrors its corresponding trimmed rollup, contents-wise.

I think these are useful invariants, but they do come at the cost of some duplication. We could also make this configurable - allow users to specify if they want all API items that would be included at the release level XOR only include items with the specific tag.

As for the design goal, this pattern still addresses the needs, because we can put different (looser) requirements for changes to `beta.api.md` than `public.api.md`.

apps/api-extractor/src/generators/ApiReportGenerator.ts
Josmithr: I think the "trimming" terminology may make sense if there is intentional parallelism between the reports at a given release level and the rollup at the same level. But I'm not tied to it. 

I do like the simpler config model, though. I'll proceed with that change while we continue to iterate on report structure / semantics.

apps/api-extractor/src/api/Extractor.ts
```diff
@@ -285,100 +286,23 @@ export class Extractor {
       });
     }
 
-    let apiReportChanged: boolean = false;
```
Josmithr, line 288: Since we now (potentially) generate multiple report files, this logic was extracted into a separate static method.

apps/api-extractor/src/api/Extractor.ts
```diff
@@ -434,12 +358,134 @@ export class Extractor {
...
+    collector: Collector,
+    extractorConfig: ExtractorConfig,
+    messageRouter: MessageRouter,
+    reportTempDirectoryPath: string,
+    reportDirectoryPath: string,
+    reportFileName: string,
+    reportVariant: ApiReportVariant,
+    localBuild: boolean
+  ): boolean {
+    let apiReportChanged: boolean = false;
```
Josmithr, line 389: This logic was moved mostly as is. The only changes made are with regards to how `reportVariant` is leveraged.

build-tests/api-extractor-lib2-test/etc/api-extractor-lib2-test.alpha.api.md
```diff
@@ -0,0 +1,22 @@
+## ALPHA API Report File for "api-extractor-lib2-test"
```
Josmithr, line 1: Added "Public" and "Alpha" configurations to this test suite

build-tests/api-extractor-lib1-test/config/api-extractor.json
```diff
@@ -4,7 +4,8 @@
   "mainEntryPointFilePath": "<projectFolder>/lib/index.d.ts",
 
   "apiReport": {
-    "enabled": true
+    "enabled": true,
+    "reportFileName": "<unscopedPackageName>.api-report.md"
```
Josmithr, line 8: Updated report name pattern here to test extension preservation logic.

apps/api-extractor/src/api/IConfigFile.ts
```diff
@@ -63,14 +68,37 @@ export interface IConfigApiReport {
   enabled: boolean;
 
   /**
-   * The filename for the API report files.  It will be combined with `reportFolder` or `reportTempFolder` to produce
-   * a full output filename.
+   * The filename for the "complete" API report file, which will encompass all API items, regardless of release tag.
    *
    * @remarks
-   * The file extension should be ".api.md", and the string should not contain a path separator such as `\` or `/`.
+   * It will be combined with {@link IConfigApiReport.reportFolder} and {@link IConfigApiReport.reportTempFolder} to
```
Josmithr, line 74: Updated the docs here to reflect the new contract. The contract itself was updated to preserve backwards compatibility, while also offering a policy that is easy to reason about.

Josmithr:
@octogonz Another policy question related to this: Do we generate report files when there are no package exports for the given variant?

To emulate the current behavior, the answer to this would be "yes". And I think that seems like the best default policy to me either way. But I could see offering a config option to not generate a report file when there are no exports. This could be tracked/implemented as a separate feature.

apps/api-extractor/src/schemas/api-extractor.schema.json
```diff
@@ -68,8 +68,17 @@
         },
 
         "reportFileName": {
-          "description": "The filename for the API report files.  It will be combined with \"reportFolder\" or \"reportTempFolder\" to produce a full file path. The file extension should be \".api.md\", and the string should not contain a path separator such as \"\\\" or \"/\".",
-          "type": "string"
+          "description": "The filename for the complete (\"untrimmed\") API report files.  It will be combined with \"reportFolder\" or \"reportTempFolder\" to produce a full file path. The file extension should be \".api.md\", and the string should not contain a path separator such as \"\\\" or \"/\".",
+          "type": ["string", "null"]
```
octogonz, line 72: We don't ever use `null` in JSON files, due to a philosophical stance that `null` is [best avoided](https://github.com/microsoft/rushstack/tree/6398f7bf4440a82fb22eca3286519f5eef8d6438/eslint/eslint-plugin#rushstackno-new-null) in JavaScript code.

common/changes/@microsoft/api-extractor/emit-report-per-release-level_2024-04-02-22-39.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/api-extractor",
+      "comment": "Support creating API reports for each release level, in addition to a complete, \"untrimmed\" report",
```
octogonz, line 5:
```suggestion
      "comment": "Add support for \"variants\" of API reports which include or exclude items by release tag",
```

For a newcomer, it's not at all obvious what "variant" means in this context.  We should use the term "variant" consistently everywhere this feature is mentioned, and probably avoid the term "trimming" which refers to a fairly unrelated feature involving .d.ts rollups.

apps/api-extractor/src/schemas/api-extractor.schema.json
```diff
@@ -68,8 +68,17 @@
         },
 
         "reportFileName": {
-          "description": "The filename for the API report files.  It will be combined with \"reportFolder\" or \"reportTempFolder\" to produce a full file path. The file extension should be \".api.md\", and the string should not contain a path separator such as \"\\\" or \"/\".",
-          "type": "string"
+          "description": "The filename for the complete (\"untrimmed\") API report files.  It will be combined with \"reportFolder\" or \"reportTempFolder\" to produce a full file path. The file extension should be \".api.md\", and the string should not contain a path separator such as \"\\\" or \"/\".",
```
octogonz, line 71:
```suggestion
          "description": "The base filename for the API report files.  It will be combined with \"reportFolder\" or \"reportTempFolder\" to produce a full file path. The file extension should be \".api.md\", and the string should not contain a path separator such as \"\\\" or \"/\".  The \"complete\" report variant uses this base filename directly; other variants insert a name such as \"my-report.alpha.api.md\" or \"my-report.beta.api.md\" for alpha and beta variants.",
```

apps/api-extractor/src/schemas/api-extractor.schema.json
```diff
@@ -68,8 +68,17 @@
...
+          "type": ["string", "null"]
+        },
+
+        "reportVariants": {
+          "description": "The set of report variants to generate. Default: [\"complete\"]",
```
octogonz, line 76:
```suggestion
          "description": "To support different approval requirements for different API release tags, multiple variants of the API report can be generated.  The \"reportVariants\" setting specifies a list of variants to be generated.  If omitted, by default only the \"complete\" variant will be generated, which includes all @alpha, @beta, and @public items.  Other possible variants are \"alpha\" (@alpha release tag only), \"beta\" (@beta only), and \"public\" (@public only).",
```

apps/api-extractor/src/schemas/api-extractor.schema.json
Josmithr: Yup, happy to conform to practices followed by the repo.

My rationale here was that (IMO) introducing a special string value to avoid null is worse than using null. Likely moot in this particular case, but it reserves a potentially valid string value to mean something not captured by the type-system.

apps/api-extractor/src/api/ExtractorConfig.ts
```diff
@@ -856,42 +888,81 @@ export class ExtractorConfig {
...
+      let reportDirectoryPath: string = tokenContext.projectFolder;
+      let reportTempDirectoryPath: string = tokenContext.projectFolder;
+      const reportConfigs: IApiReportConfig[] = [];
+      if (apiReportEnabled) {
+        // Undefined case checked above where we assign `apiReportEnabled`
+        const apiReportConfig: IConfigApiReport = configObject.apiReport!;
+
+        let reportFileNameBase: string;
+
+        // The config spec allows the user to optionally include an extension postfix in the report file name.
```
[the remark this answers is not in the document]
Josmithr: It was somewhat ambiguous before. It was documented as being required, but the code didn't validate this. I believe it was happy to let you use any extension or no extension.

I figured for strict backwards compat, maintaining the flexibility to some degree made sense, and I updated the docs to reflect the implemented contract.

If we would prefer a stricter contract (with enforcement), I can change things here. That said, it seems odd to me to require someone specify a specific file extension. If we only support one kind of file output, we can generate that for them. 🤷‍♂️

apps/api-extractor/src/schemas/api-extractor.schema.json
Josmithr: This isn't accurate, though. As implemented, `alpha` will include anything `@alpha` or "higher" (i.e. `@alpha`, `@beta`, or `@public`). This creates a parallel between the reports and corresponding trimmed `d.ts` files.

https://github.com/microsoft/rushstack/pull/4621/files/0209edd7e0af2576df4e4bf987c2eb4ebb82c2cd#r1548764854

Will update the phrasing with your feedback though, while we settle on the semantics in the above thread.

apps/api-extractor/src/generators/ApiReportGenerator.ts
octogonz: That makes sense. And your "variants" design leaves open the possibility of introducing `beta-only` or `public-only` variants in the future if someone wants that, right?

apps/api-extractor/src/api/ExtractorConfig.ts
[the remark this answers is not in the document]
octogonz: > If we would prefer a stricter contract (with enforcement), I can change things here. 

I'm kind of in favor of that, for two reasons:

1. The configuration is already a bit confusing:  This setting originally "required" the file extension because `"reportFileName": "<unscopedPackageName>.api.md"` seemed more intuitive than `"reportMeaninglessPathFragment": "<unscopedPackageName>"` for a person who is scanning the config file trying to determine how `my-library.api.md` got named.  But now we are introducing a transform that converts that string to `my-library.public.api.md`, and in fact the original name `my-library.api.md` may not get written to disk at all (if its variant is unused).  So the situation is already not super intuitive. (It really should be something like `"<unscopedPackageName>.<variant>.api.md"` but that has some other tradeoffs, e.g. do we get `..` with the default variant, etc.)

   Therefore it seems the system will be more learnable and less error-prone if we are strict about the input format.

2. `.api.md` is our standard file extension. Is there a compelling reason why someone needs to create PDF files with `.fdp` file extension or PNG files without the file extension? It seems like a somewhat unlikely requirement.

apps/api-extractor/src/generators/ApiReportGenerator.ts
Josmithr: Yeah, we can add whatever flexibility we want by adding new "enum" options and adding handler logic for them.

apps/api-extractor/src/api/ExtractorConfig.ts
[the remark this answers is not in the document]
Josmithr: > 1. The configuration is already a bit confusing:  This setting originally "required" the file extension because `"reportFileName": "<unscopedPackageName>.api.md"` seemed more intuitive than `"reportMeaninglessPathFragment": "<unscopedPackageName>"` for a person who is scanning the config file trying to determine how `my-library.api.md` got named.  But now we are introducing a transform that converts that string to `my-library.public.api.md`, and in fact the original name `my-library.api.md` may not get written to disk at all (if its variant is unused).  So the situation is already not super intuitive. (It really should be something like `"<unscopedPackageName>.<variant>.api.md"` but that has some other tradeoffs, e.g. do we get `..` with the default variant, etc.)
>    Therefore it seems the system will be more learnable and less error-prone if we are strict about the input format.

I guess I personally don't find the idea of "reportFileName" being consumed as an extension-less "fragment" to be all too confusing, especially if we discourage or disallow specifying the extension pattern (though this would be a breaking change).

I'm not opposed to re-adding the `.api.md` requirement, but if we add enforcement for it we would be introducing a functional breaking change, despite the existing documented contract. But as you mentioned, this almost seems more confusing given that we would be splicing in the `<variant>` string. The `<unscopedPackageName>.<variant>.api.md` idea is interesting, but I agree that there is some ambiguity there.

All of this leads me to think that the ideal would be to _not_ recommend providing the `.api.md` postfix, and document how the extensionless name is consumed. Then we will _always_ output files ending in `.api.md`, which I would argue is less confusing than asking for a specific extension that we would always want to output anyways. For_ backwards compatibility, we can continue to support the `.api.md` postfix when we see it, but probably leave that support undocumented to encourage users to just specify the extension-less name component.

Curious if you feel strongly otherwise here.

> 2. `.api.md` is our standard file extension. Is there a compelling reason why someone needs to create PDF files with `.fdp` file extension or PNG files without the file extension? It seems like a somewhat unlikely requirement.

No, my point above was just that the code didn't strictly enforce that the extension was `.api.md`. The docs asked for it, but the code would happily use whatever the user specified. So if they did specify `foo.pdf`, we would generate that file with markdown formatted contents.

I agree that we only output markdown, so we should always output `.md` files. This is why I find it odd that ask the user to specify the extension at all. I get your point about `reportMeaninglessPathFragment`, but I think calling it `reportFileName` and documenting what that means and how it gets consumed is a pretty reasonable compromise that users will be able to understand pretty easily. If there is any ambiguity (or if they don't read the docs), that will be clarified after they run the tool for the first time and see the output.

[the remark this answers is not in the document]
octogonz: Just to clarify, your proposal is:

1. Update the docs everywhere to stipulate that `"reportFileName"` should _never_ include a file extension.  It will be automatically appended, as well as a suffix like `.public.api.md` for variants.
2. However, if the filename happens have the specific extension `.api.md` then it will be silently removed (and re-added) to avoid breaking backwards compatibility with existing config files.

This can work, although we will forevermore have config files everywhere with `"reportFileName": "thing.api.md"` which might undermine that definition.

(Another idea would be to introduce a new setting like `"reportName": "<unscopedPackageName>"`, deprecating the old setting and planning to remove it in the next major release, and requiring the new setting if the variants feature is used.  However, this idea gets messy when we consider `"$extends"` and handling of default values.  So I think your design is maybe superior.)

And what if someone really really needs a different file extension? Then we could always introduce a `reportFileExtension` setting.

One final consideration is why is this designed differently from the quite straightforward `dtsRollup.untrimmedFilePath`, `dtsRollup.betaTrimmedFilePath`, `dtsRollup.publicTrimmedFilePath` settings?  I guess the answer is that we expect there to be a small number of trimming possibilities compared to variant profiles?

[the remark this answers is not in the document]
octogonz: > One final consideration is why is this designed differently from the quite straightforward `dtsRollup.untrimmedFilePath`, `dtsRollup.betaTrimmedFilePath`, `dtsRollup.publicTrimmedFilePath` settings? I guess the answer is that we expect there to be a small number of trimming possibilities compared to variant profiles?

Actually I remember the answer to this question now:

- For .d.ts rollups, the user often needs fairly precise control over where each file is written and what is its name, because they can be import paths that are part of a package's API contract.
- Whereas for .api.md reports, I had been thinking about a potential mistake where e.g. the `alpha` variant was being generated but the latest configuration removed that, and so `my-library.alpha.api.md` accidentally is being tracked by Git but no longer getting generated.  To prevent such confusion, we could introduce a feature something like `disallowUnusedVariantFiles` where API Extractor could scan the target folder for such files, which requires more strict control about how they are named.

With these thoughts, I'm 👍 👍 for your proposal.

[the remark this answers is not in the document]
Josmithr: > This can work, although we will forevermore have config files everywhere with "reportFileName": "thing.api.md" which might undermine that definition.

At least until the next major version rev, yes. We could basically note that providing a file extension is deprecated.

Do the maintainers of this repo have a place where they track such follow-up work? I'm happy to add something there as needed.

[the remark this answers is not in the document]
Josmithr: Updated the code and docs to match the following contract:
* Don't specify extension
* `reportFileName` is enterpreted as the body component of `<body>.<variant>.api.md` _except_ we will (as a special case) search for and strip out `.api.md` if we see it. This special-casing should be removed in version 8.

common/reviews/api/api-extractor.api.md
```diff
@@ -74,8 +78,9 @@ export class ExtractorConfig {
     readonly projectFolder: string;
     readonly projectFolderUrl: string | undefined;
     readonly publicTrimmedFilePath: string;
-    readonly reportFilePath: string;
-    readonly reportTempFilePath: string;
```
octogonz, line 78: Removing `reportFilePath` and `reportTempFilePath` would be a breaking change to the API. 

In the existing design, they return an empty string when the API report feature is disabled.  Maybe we could just change these settings to return the `"complete"` variant, only if that variant is enabled.  We could also mark them as `@deprecated`.

common/reviews/api/api-extractor.api.md
```diff
@@ -163,6 +168,12 @@ export class ExtractorResult {
     readonly warningCount: number;
 }
 
+// @public
+export interface IApiReportConfig {
```
octogonz, line 172: It's confusing to have both `IConfigApiReport` and `IApiReportConfig` in the same API.

Suggest to rename `IApiReportConfig` -> `IExtractorConfigReport` since it is conceptually just a nested part of the ExtractorConfig.

common/reviews/api/api-extractor.api.md
```diff
@@ -163,6 +168,12 @@ export class ExtractorResult {
...
+    fileName: string;
```
octogonz, line 173: It is slightly confusing that `fileName` includes the file extension but `IConfigApiReport.fileName` now excludes it.


Following [path.parse](https://nodejs.org/api/path.html#pathparsepath), maybe we can call it `baseFileName` or `baseName`:
> image

octogonz: I might be overthinking this however. I was surprised to find that Heft's [ApiExtractorRunner.ts](https://github.com/microsoft/rushstack/blob/main/heft-plugins/heft-api-extractor-plugin/src/ApiExtractorRunner.ts) doesn't use these API's at all. I guess in most use cases API Extractor handles copying the files and printing warnings about report files. (?)

apps/api-extractor/src/api/ExtractorConfig.ts
```diff
@@ -158,8 +182,9 @@ interface IExtractorConfigParameters {
   overrideTsconfig: {} | undefined;
   skipLibCheck: boolean;
   apiReportEnabled: boolean;
-  reportFilePath: string;
-  reportTempFilePath: string;
+  reportConfigs: readonly IApiReportConfig[];
+  reportDirectoryPath: string;
+  reportTempDirectoryPath: string;
```
octogonz, line 187: Rush Stack typically uses the word `folder` instead of `directoryPath` because it is much more concise.

octogonz: (And `filePath` is normally a filename that may include a relative or absolute path or not. And `fileName` is generally the name+extension part of the path.)

octogonz (approved):

common/reviews/api/api-extractor.api.md
Josmithr: Oh, I missed that this class was exported. Yeah, I'll revert the changes here.

Josmithr: Question: are consumers expected to extend this class at all? Would it be safe to add legacy getter properties (possibly marked as deprecated)?

Also wondering if this class should be marked as `@sealed`.

Josmithr: I went ahead and implemented the getter property route but let me know if you think that is insufficient. I'm also happy to mark the class as `@sealed` while I'm here, if we think that's the right call.

common/reviews/api/api-extractor.api.md
Josmithr: Yeah, as far as I can tell that's strictly handled in `Extractor.ts`

apps/api-extractor/src/api/ExtractorConfig.ts
```diff
@@ -856,42 +908,79 @@ export class ExtractorConfig {
...
-
-        const reportFilename: string = ExtractorConfig._expandStringWithTokens(
-          'reportFileName',
-          configObject.apiReport.reportFileName || '',
-          tokenContext
-        );
+      const apiReportEnabled: boolean = configObject.apiReport?.enabled ?? false;
+      const apiReportIncludeForgottenExports: boolean =
+        configObject.apiReport?.includeForgottenExports ?? false;
+      let reportDirectoryPath: string = tokenContext.projectFolder;
```
Josmithr, line 914: TODO: rename vars

common/reviews/api/api-extractor.api.md
octogonz: >  are consumers expected to extend this class at all?

Consumers can't inherit `ExtractorConfig` because the constructor is private.

> Also wondering if this class should be marked as `@sealed`.

Sure

apps/api-extractor/src/api/ExtractorConfig.ts
```diff
@@ -856,42 +888,79 @@ export class ExtractorConfig {
...
+          ) {
+            throw new Error(
+              `The "reportFileName" setting contains invalid characters: "${apiReportConfig.reportFileName}"`
+            );
+          }
 
-        if (!reportFilename) {
-          // A merged configuration should have this
-          throw new Error('The "reportFilename" setting is missing');
+          const iSuffix: number = apiReportConfig.reportFileName.indexOf(reportFileNameSuffix);
```
octogonz, line 933: Rename `iSuffix` to `suffixIndex` following our coding convention that identifiers should be comprised of full English words

common/reviews/api/api-extractor.api.md
octogonz: > Would it be safe to add legacy getter properties (possibly marked as deprecated)?

Yes this is fine. Since it's a `class` we can use logic to defer computation.  (That is also a potential reason for `IExtractorConfigApiReport` to be a `class`, but I didn't bring this up since it seems probably immaterial.)

However according to our coding conventions, property getters should not normally throw exceptions.

Josmithr: > Consumers can't inherit ExtractorConfig because the constructor is private.

Ah, missed that the constructor was private. Excellent.

> However according to our coding conventions, property getters should not normally throw exceptions.

None of what's in the getter implementations should throw. Can you elaborate?

octogonz: > None of what's in the getter implementations should throw. Can you elaborate?

The code's fine. ( This comment was like when GPT keeps blathering about stuff you didn't ask and you have to press the ⏹️ button. 😄 )

PR #4638, opened 2024-04-13
[api-documenter] Fix return block rendering for page markdown

colincasey:
## Summary

Fixes rendering of the "\*\*Returns:\*\*" markdown in generated page documentation for a function.

## Details

Noticed that a missing line break was causing a rendering of the `Returns:` section on the markdown output for a function.  You can see an example of this [here](https://github.com/salesforce/tough-cookie/blob/v5-docs/api/docs/tough-cookie.canonicaldomain.md).

The markdown is correct `**Returns:**` but because there is no line break between the end of the parameter table and the aforementioned markdown content, it is not treated as a new paragraph and is rendered as "\*\*Returns:\*\*" instead of "**Returns:**".

This is fixed by adding the required line break to the generated markdown content.

## How it was tested

I rebuild the `api-documenter` with the included code change and regenerated the markdown pages for my project and visually compared them.

## Impacted documentation

Not applicable.


colincasey:

PR #4642, opened 2024-04-14
Hotfix an issue with rush install with subspaces

william2958:


common/changes/@microsoft/rush/will-subspace-patch-16_2024-04-14-23-25.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Fixes an issue where rush install fails in monorepos with subspaces enabled",
```
iclanton, line 5: Always? Or are there specific scenarios that don't work?

iclanton (commented):
Can you fill out the PR description?

PR #4648, opened 2024-04-16
feat: Support skipping validation for a specific package during the c…

L-Qun:
## Summary

Support skipping validation for a specific package during the check of the pnpm-lock.yaml file.

## Details

When we enable the `disallowInsecureSha1` policy, an error is reported if sha1 hashes is included in the `pnpm-lock.yaml `file. However, sometimes we are temporarily unable to fix this issue. Therefore, the `skipCheckPackages` field is provided to skip validation for these packages that cannot be fixed temporarily: 

image

Then we can add this package and version to `exmeptPackageVersions` field:

image

Rush will skip the checks for these packages and will not report errors.

## Impacted documentation

- New config file setting


common/reviews/api/rush-lib.api.md
```diff
@@ -670,7 +670,10 @@ export interface IPhasedCommand extends IRushCommand {
 
 // @public
 export interface IPnpmLockfilePolicies {
-    disallowInsecureSha1?: boolean;
+    disallowInsecureSha1?: {
+        enabled: boolean;
+        skipCheckPackages: Record<string, boolean>;
```
chengcyber, line 675: 1. How about "exemptPackageVersions"?
2. The schema could be
```
"exmeptPackageVersions": {
  "package_name": ["version1", "version2"]
}

Comment (L-Qun) [reply common/reviews/api/rush-lib.api.md]: done

File: libraries/rush-lib/assets/rush-init/common/config/rush/pnpm-config.json [hunk 2]
```diff
@@ -144,7 +144,10 @@
      *
      * The default value is false.
      */
-    /*[LINE "HYPOTHETICAL"]*/ "disallowInsecureSha1": true
+    /*[LINE "HYPOTHETICAL"]*/ "disallowInsecureSha1": {
+      "enabled": true,
+      "skipCheckPackages": {}
```
chengcyber, line 149: It seems this property name is not updated

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -451,6 +441,20 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
+    let depPath: string = pkgPath;
+    if (this.shrinkwrapFileMajorVersion >= 6) {
+      const index: number = pkgPath.indexOf('@', pkgPath.indexOf('/@') + 2);
+      const suffixIndex: number | undefined = pkgPath.includes('(')
+        ? dp.indexOfPeersSuffix(pkgPath)
+        : undefined;
+      depPath = `${pkgPath.slice(0, index)}/${pkgPath.slice(index + 1, suffixIndex)}`;
+    }
+
+    const pkgInfo: ReturnType<typeof dp.parse> = dp.parse(depPath);
```
octogonz, line 454: Our coding convention is to spell out words instead of using arbitrary abbreviations, something like this:

- `import * as dp from '@pnpm/dependency-path';` -> `import * as dependencyPath`
- `_parseDepPath` -> `_parseDependencyPath`
- `pkgPath` -> `packagePath`

The underlying idea is that **Reading &gt; Writing**.  Being able to write source code easily is less important than other people being able to read it easily.

common/reviews/api/rush-lib.api.md
```diff
@@ -670,7 +670,10 @@ export interface IPhasedCommand extends IRushCommand {
 
 // @public
 export interface IPnpmLockfilePolicies {
-    disallowInsecureSha1?: boolean;
+    disallowInsecureSha1?: {
+        enabled: boolean;
+        exmeptPackageVersions: Record<string, Array<string>>;
```
octogonz, line 675: The spelling is incorrect.  It should be `exemptPackageVersions`.

Also, instead of `Array<string>` use `string[]` because it is idiomatic.

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -343,59 +344,48 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
-    if (packages) {
-      for (const { resolution } of Object.values(packages)) {
-        if (resolution?.integrity.startsWith('sha1')) {
-          terminal.writeErrorLine(
-            'Error: An integrity field with "sha1" was found in pnpm-lock.yaml;' +
-              ' this conflicts with the "disallowInsecureSha1" policy from pnpm-config.json.\n'
-          );
+    const exmeptPackageList: Map<string, boolean> = new Map();
+    for (const [pkgName, versions] of Object.entries(exmeptPackageVersions)) {
+      versions.forEach((version) => exmeptPackageList.set(this._getPackageId(pkgName, version), true));
```
octogonz, line 352: use `for (const version of versions` to avoid an unnecessary closure and make it easier to step through the loop in a debugger.

libraries/rush-lib/src/schemas/pnpm-config.schema.json
```diff
@@ -183,8 +183,25 @@
...
+            },
+            "exemptPackageVersions": {
+              "description": "Packages that should skip checks.",
+              "type": "object",
+              "additionalProperties": {
+                "type": "array",
+                "items": {
+                  "type": "string"
+                },
+                "description": "An array of version string for the package."
```
octogonz, line 200:
```suggestion
                "description": "An array of exempted versions for this package."
```

libraries/rush-lib/src/schemas/pnpm-config.schema.json
```diff
@@ -183,8 +183,25 @@
...
         "disallowInsecureSha1": {
+          "type": "object",
           "description": "Forbid sha1 hashes in `pnpm-lock.yaml`.",
-          "type": "boolean"
+          "properties": {
+            "enabled": {
+              "type": "boolean"
+            },
+            "exemptPackageVersions": {
+              "description": "Packages that should skip checks.",
```
octogonz, line 193:
```suggestion
              "description": "A list of specific package versions to be exempted from the \"disallowInsecureSha1\" policy",
```

common/changes/@microsoft/rush/main_2024-04-16-06-52.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Support skipping validation for a specific package during the check of the pnpm-lock.yaml file.",
```
octogonz, line 5:
```suggestion
      "comment": "(BREAKING CHANGE) Improve the `disallowInsecureSha1` policy to support exemptions for certain package versions. This is a breaking change for the `disallowInsecureSha1` field in pnpm-config.json since Rush 5.119.0.",
```

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -451,6 +441,20 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
     }
   }
 
+  private _parseDepPath(pkgPath: string): string {
```
octogonz: Add a `/** */` comment explaining the purpose of this function.

libraries/rush-lib/src/logic/pnpm/PnpmShrinkwrapFile.ts
```diff
@@ -451,6 +441,20 @@ export class PnpmShrinkwrapFile extends BaseShrinkwrapFile {
...
+    let depPath: string = pkgPath;
+    if (this.shrinkwrapFileMajorVersion >= 6) {
+      const index: number = pkgPath.indexOf('@', pkgPath.indexOf('/@') + 2);
+      const suffixIndex: number | undefined = pkgPath.includes('(')
+        ? dp.indexOfPeersSuffix(pkgPath)
+        : undefined;
+      depPath = `${pkgPath.slice(0, index)}/${pkgPath.slice(index + 1, suffixIndex)}`;
```
octogonz: This logic is difficult to understand.  It would be very helpful to include `//` comments showing examples of the input and output strings, something like this (fill in your own examples - I'm just guessing 😄 ):

```ts
    let depPath: string = pkgPath;
    if (this.shrinkwrapFileMajorVersion >= 6) {
      // Example rewrites:
      //    "/@scope/pkg@1.0.0/peer@1.0.0" --> "/@scope/pkg/peer@1.0.0"
      //    "/@scope/pkg@1.0.0/peer@1.0.0 (foo)" --> "/@scope/pkg@1.0.0/peer@1.0.0 (foo)"
      const index: number = pkgPath.indexOf('@', pkgPath.indexOf('/@') + 2);
      const suffixIndex: number | undefined = pkgPath.includes('(')
        ? dp.indexOfPeersSuffix(pkgPath)
        : undefined;
      depPath = `${pkgPath.slice(0, index)}/${pkgPath.slice(index + 1, suffixIndex)}`;
```

Also, if `indexOf()` does not find the string, it returns `-1`.  It's unclear whether the `pkgPath.indexOf('/@') + 2` math will work correctly in that case, nor is it obvious that `-1` is impossible.

In other words, from the perspective of another person reading this code, it is not "obviously correct." I would have to think about it to convince myself that it's right.  Such a situation is inevitable for complicated problems, but for a relatively straightforward problem like `_parseDepPath()`, we can make the correctness more obvious.  Adding some `if` statements and comments is probably sufficient.

L-Qun: I copied it from the pnpm source code:

https://github.com/pnpm/pnpm/blob/73ebfc94e06d783449579cda0c30a40694d210e4/lockfile/lockfile-file/src/experiments/inlineSpecifiersLockfileConverters.ts#L162

L-Qun: image

If position is less than zero, the method behaves as it would if position were 0.

octogonz (approved):

PR #4652, opened 2024-04-16
[rush] add support for sharding phases

aramissennyeydd:
## Summary

This PR adds support for sharding to rush phases. This allows plugins that support sharding, like jest, to be split into multiple shards and run independently. It does this by adding a new set of options to `rush-project.json` under a new `sharding` key. example:
```json
{
    "operationSettings": {
        {
            "operationName": "_phase:test",
            "outputFolderNames": ["coverage", "temp/coverage"],
            "sharding": {
                "count": 6,
                // Defaults to `--shard={shardIndex}/{shardCount}`
                "shardArgumentFormat": "--shard-format={shardIndex}-{shardCount}"
            }
        }
    }
}
```

## Details

This is the initial chunk of work to support sharding in the operation graph. It includes both the sharding nodes as well as a collator node that can run a script after all of the shard nodes are complete.

I originally attempted this with heft plugins, however tying into rush parallelism + cobuilds is one of our end goals with this work and heft plugins at the moment don't allow that. 

## How it was tested

I've been locally testing with `node apps/rush/lib/start-dev test --to heft-jest-shards-test -p 6` and varying parallelism flags. The tests in `heft-jest-shards-test` run for 10 seconds and then pass and there are 6 of those files, so `-p 6` should run in ~10 seconds, `-p 3` in 20 seconds and so on. I also added a `sharded-repo` to the existing cobuild suite to ensure this works with cobuilds.

TODO: 
- [x] retest cobuilds after the log filename is updated.

## Impacted documentation

- https://rushjs.io/pages/configs/rush-project_json/
- Not sure where else this fits in? Should it get its own page under "Maintainer tutorials"?


libraries/rush-lib/src/logic/operations/OperationExecutionRecord.ts
```diff
@@ -145,6 +146,7 @@ export class OperationExecutionRecord implements IOperationRunnerContext {
   public get collatedWriter(): CollatedWriter {
     // Lazy instantiate because the registerTask() call affects display ordering
     if (!this._collatedWriter) {
+      console.log(this.name);
```
octogonz, line 149: ?

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/e/config/rush-project.json
```diff
@@ -0,0 +1,13 @@
...
+      "operationName": "_phase:build",
+      "outputFolderNames": ["dist"],
+      "shards": 100
+    },
+    {
+      "operationName": "cobuild",
+      "outputFolderNames": ["dist"]
+    }
+  ]
+}
```
octogonz, line 13: ## Design idea 1 discussed today:

**package.json**
```js
{
  "scripts": {
    "_phase:build": "invoke-jest",
    "_phase:build:after-shards": "./collate-coverage.js",
  }
}
```

**collate-coverage.js**
```js
const shardCount = parseInt(process.env._RUSH_SHARDS_COUNT);
for (let i: number  = 0; i<shardCount; ++i) {
  // Its build cache entry reads from "_phase:build" output folder
 `${process.env._RUSH_SHARDS_FOLDER}/${index}`
}
```

## Design idea 2

Design idea discussed today:

**package.json**
```js
{
  "scripts": {
    // 
    "_phase:build:shard": "invoke-jest",
    "_phase:build": "./collate-coverage.js"
  }
}
```

**collate-coverage.js**
```js
const shardCount = parseInt(process.env._RUSH_SHARDS_COUNT);
for (let i: number  = 0; i<shardCount; ++i) {
  // Its build cache entry reads from "_phase:build" output folder
 `${process.env._RUSH_SHARDS_FOLDER}/${index}`
}
```

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,85 @@
...
+  return existingOperations;
+}
+
+async function getShardConfig(
+  phase: IPhase,
+  project: RushConfigurationProject,
+  terminal: ITerminal
+): Promise<IRushPhaseSharding | undefined> {
+  const rushProjectConfiguration: RushProjectConfiguration | undefined =
+    await RushProjectConfiguration.tryLoadForProjectAsync(project, terminal);
```
dmichon-msft, line 81: This is available as `context.projectConfigurations.get(project)`, so you don't need anything async.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,85 @@
...
+      const shards: number = shardConfig.count;
+      existingOperations.delete(operation);
+      for (const shard of Array.from({ length: shards }, (element, index) => index + 1)) {
+        const shardOperation: Operation = new Operation({
+          project,
+          phase,
+          shard: {
+            current: shard,
+            total: shards
+          },
```
dmichon-msft, line 55: This belongs in the runner instance, which is the entity that owns the command line and the child process invocation.

libraries/rush-lib/src/logic/operations/Operation.ts
```diff
@@ -23,6 +23,11 @@ export interface IOperationOptions {
    * running the operation.
    */
   runner?: IOperationRunner | undefined;
+
+  shard?: {
+    current: number;
+    total: number;
+  };
```
dmichon-msft, line 30: I'm fine with promoting `outputFolderNames` to be a property of `Operation` if you need a place that is accessible to the `CacheableOperationPlugin`, but `Operation` has no business knowing anything about the concept of shards. That's a runner detail.

libraries/rush-lib/src/logic/operations/OperationMetadataManager.ts
```diff
@@ -45,19 +47,24 @@ export class OperationMetadataManager {
...
   public constructor(options: IOperationMetadataManagerOptions) {
-    const { rushProject, phase } = options;
+    const { rushProject, phase, operation } = options;
     const { projectFolder } = rushProject;
 
     const identifier: string = phase.logFilenameIdentifier;
-    this._metadataFolder = `${RushConstants.projectRushFolderName}/${RushConstants.rushTempFolderName}/operation/${identifier}`;
+
+    const operationPath: string | undefined = operation.shard
+      ? `${operation.name}_${operation.shard.current}`
```
dmichon-msft, line 56: The shard should be part of the operation name, which is in turn just the forwarded runner name.

libraries/rush-lib/src/logic/operations/ShellOperationRunnerPlugin.ts
```diff
@@ -40,13 +40,21 @@ function createShellOperations(
...
       // to specify a runner type requested in rush-project.json
-      const customParameterValues: ReadonlyArray<string> = getCustomParameterValuesForPhase(phase);
+      let customParameterValues: ReadonlyArray<string> = getCustomParameterValuesForPhase(phase);
+
+      if (operation.shard) {
+        customParameterValues = [
+          ...customParameterValues,
+          `--shard=${operation.shard.current}/${operation.shard.total}`
+        ];
+      }
```
dmichon-msft, line 50: Instantiate your own `ShellOperationRunner` instances inside of `ShardedPhaseOperationPlugin` for sharded operations. The `ShellOperationRunnerPlugin` has no business knowing anything about sharding.

libraries/rush-lib/src/logic/operations/ShellOperationRunnerPlugin.ts
```diff
@@ -130,12 +139,18 @@ export function formatCommand(rawCommand: string, customParameterValues: Readonl
...
+  operation: Operation
+): string {
   if (phase.isSynthetic) {
     // Because this is a synthetic phase, just use the project name because there aren't any other phases
     return project.packageName;
   } else {
     const phaseNameWithoutPrefix: string = phase.name.slice(RushConstants.phaseNamePrefix.length);
-    return `${project.packageName} (${phaseNameWithoutPrefix})`;
+    return `${project.packageName} (${phaseNameWithoutPrefix})${
+      operation.shard ? ` - shard ${operation.shard.current}` : ''
```
dmichon-msft, line 153: The `ShardedPhaseOperationPlugin` should just append the shard suffix to the operation name when creating the runner, rather than modifying this function.

libraries/rush-lib/src/logic/operations/CacheableOperationPlugin.ts
```diff
@@ -591,17 +594,25 @@ export class CacheableOperationPlugin implements IPhasedCommandPlugin {
...
+      const { operationSettings, projectChangeAnalyzer, shard } = buildCacheContext;
       if (!operationSettings || !buildCacheConfiguration) {
         // Unreachable, since this will have set `cacheDisabledReason`.
         return;
       }
 
-      const projectOutputFolderNames: ReadonlyArray<string> = operationSettings.outputFolderNames || [];
+      const projectOutputFolderNames: ReadonlyArray<string> = [
+        ...(operationSettings.outputFolderNames || [])
+      ];
```
dmichon-msft, line 605: To facilitate this being customized by plugins, I'd prefer to forward this as a property on `Operation` or `IOperationRunner`. That way the `CacheableOperationPlugin` doesn't need to understand anything about shards.

That also has the eventual benefit of allowing us to give Rush as stronger concept of `clean`, since the output folder names are what are necessary to do that.

aramissennyeydd: Do we want a flat config on the Operation or IOperationRunner? ie adding `outputFolderNames` versus `operationSettings`. I added the latter as the other properties might be useful later on?

libraries/rush-lib/src/logic/operations/ShellOperationRunnerPlugin.ts
aramissennyeydd: Done

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Updated in [7d34e74](https://github.com/microsoft/rushstack/pull/4652/commits/7d34e74bc94de34fb1bafe0c89cde48b7d01c7de)

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Updated in [7d34e74](https://github.com/microsoft/rushstack/pull/4652/commits/7d34e74bc94de34fb1bafe0c89cde48b7d01c7de)

libraries/rush-lib/src/logic/operations/Operation.ts
aramissennyeydd: Definitely, updated in [6c8cf5c](https://github.com/microsoft/rushstack/pull/4652/commits/6c8cf5ce262653513436729cb978a815179ea525)

libraries/rush-lib/src/logic/operations/OperationMetadataManager.ts
aramissennyeydd: Updated in [7d34e74](https://github.com/microsoft/rushstack/pull/4652/commits/7d34e74bc94de34fb1bafe0c89cde48b7d01c7de)

libraries/rush-lib/src/logic/operations/ShellOperationRunnerPlugin.ts
aramissennyeydd: Updated in [7d34e74](https://github.com/microsoft/rushstack/pull/4652/commits/7d34e74bc94de34fb1bafe0c89cde48b7d01c7de).

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/e/config/rush-project.json
aramissennyeydd: I am leaning towards the first, this PR uses `_phase:build` and `_phase:build:collate` or `_phase:test` and `_phase:test:collate`. I think that makes more sense as you need to add a new command to reconcile the phases and your existing commands read the same, so `_phase:test` maps to jest and you have a new `_phase:test:collate` command to collate the coverage.

libraries/rush-lib/src/logic/operations/OperationMetadataManager.ts
aramissennyeydd: Follow up question here, this currently only depends on `phase.logFilenameIdentifier` for the directory name. Should I move that into Operation instead or just use the operation name here? If we don't use the operation name, shards aren't differentiated here - but this might be a breaking change for projects that upgrade and expect the old path to restore correctly.

aramissennyeydd:
Also, I think CI is failing due to mismatched versions of rush - the `sharding` option isn't available in the version of `install-run-rush`.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,162 @@
...
+          rushConfiguration,
+          rushProject: project,
+          operationSettings
+        });
+        collatorNode.runner = collatorRunner;
+      } else {
+        collatorNode.runner = new NullOperationRunner({
+          name: collatorDisplayName,
+          result: OperationStatus.NoOp,
+          silent: phase.missingScriptBehavior === 'silent'
```
dmichon-msft, line 76: We might need to have separate `missingScriptBehavior` values for the phase and for the collator. If you have a required phase that doesn't require collation, for example.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,162 @@
...
+        });
+      }
+      for (const dependent of operation.consumers) {
+        dependent.addDependency(collatorNode);
+        dependent.deleteDependency(operation);
+      }
+
+      const shards: number = operationSettings.sharding.count;
```
dmichon-msft, line 84: This value was fetched for the comparison on line 51; since we need it multiple times I usually try to extract it before the first usage.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,162 @@
...
+      shardableOperations.add(operation);
+    }
+  }
+  const getCustomParameterValuesForPhase: (phase: IPhase) => ReadonlyArray<string> =
+    getCustomParameterValuesByPhase();
+
+  for (const operation of shardableOperations) {
+    const { associatedPhase: phase, associatedProject: project } = operation;
+    const operationSettings: IOperationSettings | undefined = getOperationSettings(phase, project);
+    if (phase && project && operationSettings?.sharding && operationSettings.sharding.count > 1) {
```
dmichon-msft, line 51: It isn't safe to assume that just because the operation is configured to have exactly 1 shard that it doesn't need collation. It could have a build script for which the sharding arguments are mandatory.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,162 @@
...
+      for (const shard of Array.from({ length: shards }, (element, index) => index + 1)) {
+        let customParameters: readonly string[] = getCustomParameterValuesForPhase(phase);
+
+        // Add the shard argument to the custom parameters, replacing any templated values.
+        const shardArgumentFormat: string =
+          operationSettings.sharding.shardArgumentFormat ?? '--shard={shardIndex}/{shardCount}';
+        const outputDirectory: string = `.rush/shards/${shard}`;
+        const shardArgument: string = shardArgumentFormat
+          .replace('{shardIndex}', shard.toString())
+          .replace('{shardCount}', shards.toString());
```
dmichon-msft, line 94: Use `replaceAll` just to be safe. Technically the argument format could place one of the values in the arguments list multiple times, even if that would be strange.

Also, these tokens should be constants, since we will want to have a step that verifies for their existence in the `shardArgumentFormat` property as part of validating the config.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,162 @@
...
+        const outputDirectoryArgument: string = `--shard-output-directory="${outputDirectory}"`;
+        customParameters = [...customParameters, shardArgument, outputDirectoryArgument];
+
+        const commandToRun: string | undefined = getScriptToRun(
+          project,
+          phase.name,
+          customParameters,
+          phase.shellCommand
+        );
```
dmichon-msft, line 103: If you follow my suggestion to move the call to `formatCommand` out of `getScriptToRun`, this call can be hoisted outside of the shard loop.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,162 @@
...
+        const outputDirectory: string = `.rush/shards/${shard}`;
+        const shardArgument: string = shardArgumentFormat
+          .replace('{shardIndex}', shard.toString())
+          .replace('{shardCount}', shards.toString());
+        const outputDirectoryArgument: string = `--shard-output-directory="${outputDirectory}"`;
+        customParameters = [...customParameters, shardArgument, outputDirectoryArgument];
+
+        const commandToRun: string | undefined = getScriptToRun(
+          project,
+          phase.name,
```
dmichon-msft, line 100: I thought the last state we left the design in during the call was that the collator node was `_phase:foo` and the shards were `_phase:foo:shard`? That was `_phase:foo` is the node that exits the original phase, and `_phase:foo:shard` is the one that gets special arguments that aren't part of the default phase definition.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,162 @@
...
+        phase,
+        project
+      });
+      existingOperations.add(collatorNode);
+      const collatorDisplayName: string = `${getDisplayName(phase, project)} - collate`;
+      const rawScript: string | undefined = project.packageJson.scripts?.[`${phase.name}:collate`];
+
+      if (rawScript) {
+        const collatorRunner: ShellOperationRunner = new ShellOperationRunner({
+          commandToRun: rawScript,
```
dmichon-msft, line 64: The collator script also needs to be passed the parent folder of the shard output directory, unless we are enforcing that it will always be the value currently in this PR.

Alternatively, we could make the shard output folder configurable?

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,162 @@
...
+      }
+
+      const shards: number = operationSettings.sharding.count;
+      for (const shard of Array.from({ length: shards }, (element, index) => index + 1)) {
+        let customParameters: readonly string[] = getCustomParameterValuesForPhase(phase);
+
+        // Add the shard argument to the custom parameters, replacing any templated values.
+        const shardArgumentFormat: string =
+          operationSettings.sharding.shardArgumentFormat ?? '--shard={shardIndex}/{shardCount}';
+        const outputDirectory: string = `.rush/shards/${shard}`;
```
dmichon-msft, line 91: This folder name _must_ contain something that is a function of the phase, since you are allowed to have multiple sharded phases in the same project.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: See my comment [here](https://github.com/microsoft/rushstack/pull/4652#discussion_r1581247792), I think the `{phase.name}:collate` format makes it a little clearer what is happening and requires fewer changes for people adopting this script, ie "`_phase:test` is still testing but I need a collate command to collate coverage"

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Good point, I'll add `phase.logFileName` to the directory path.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: I'll adjust to a separate `collatorScript` key in the sharding config in `rush-project.json` that lets you specify this.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Agreed, looks like `replaceAll` is an es2021 feature, is there a rush polyfill?

libraries/rush-lib/src/logic/operations/CacheableOperationPlugin.ts
aramissennyeydd: Fixed

libraries/rush-lib/src/api/CommandLineJson.ts
```diff
@@ -106,6 +106,11 @@ export interface IPhaseJson {
    * By default, Rush returns a nonzero exit code if errors or warnings occur during a command. If this option is set to \"true\", Rush will return a zero exit code if warnings occur during the execution of this phase.
    */
   allowWarningsOnSuccess?: boolean;
+
+  sharding?: {
+    count: number;
+    shardArgumentFormat: string;
+  };
```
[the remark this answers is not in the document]
aramissennyeydd: I was mistaken here, don't need this. Removing.

libraries/rush-lib/src/api/RushProjectConfiguration.ts
```diff
@@ -92,6 +97,14 @@ export interface IOperationSettings {
...
    */
   dependsOnAdditionalFiles?: string[];
+
+  /**
+   * An optional config object for sharding the operation. If specified, the operation will be sharded
+   * into multiple invocations. The `count` property specifies the number of shards to create. The
+   * `shardArgumentFormat` property specifies the format of the argument to pass to the command to
+   * indicate the shard index and count. The default value is `--shard={shardIndex}/{shardCount}`.
+   */
+  sharding?: IRushPhaseSharding;
```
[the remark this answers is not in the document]
aramissennyeydd: command-line.json addition was a mistake, this should only be specified per operation.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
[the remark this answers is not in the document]
aramissennyeydd: TIL, that's what I was concerned about.

aramissennyeydd: Isn't sharding of 1 and then collating the same as having 2 phases where one depends on the other? I removed the check already, but this feels like it this would be an incorrect usage of shards. You could just call the phases with `--shard 1/1` and `--output-directory my-output-directory` manually.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Updated to make the output folder configurable and passed it through environment variables.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Added in [67598bf](https://github.com/microsoft/rushstack/pull/4652/commits/67598bf13fa258cef6cb08d8a3b6958ef958ca90)

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Thinking on this some more, I'll adjust to the `:shard` model. The existing config like output folders and phase name should consistently refer to something, where currently they don't because shard is the primary operation, so `outputFolders` is used for the collation but `operationName` and config from `command-line.json` refers to the sharding operation.

libraries/rush-lib/src/logic/operations/OperationMetadataManager.ts
aramissennyeydd: Adjusted to a new `operationMetadataManager.logFilenameIdentifier` which is used for all cacheable operation logs

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Updated in [058eee7](https://github.com/microsoft/rushstack/pull/4652/commits/058eee73433bbcb406f0d9798642b575f0c3904b)

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,212 @@
...
+
+      if (collatorRawScript === undefined && phase.missingScriptBehavior === 'error') {
+        throw new Error(
+          `The project '${project.packageName}' does not define a '${phase.name}' command in the 'scripts' section of its package.json`
+        );
+      }
+
+      if (collatorRawScript) {
+        const collatorRunner: ShellOperationRunner = new ShellOperationRunner({
+          commandToRun: collatorRawScript,
```
dmichon-msft, line 99:
```suggestion
          commandToRun: formatCommand(collatorRawScript, customParameters),
```
Rush custom parameters for the phase should still get forwarded to the collator. They may alter the behavior of the shards in ways that the collator needs to know about.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,212 @@
...
+          result: OperationStatus.NoOp,
+          silent: phase.missingScriptBehavior === 'silent'
+        });
+      }
+      for (const dependent of operation.consumers) {
+        dependent.addDependency(collatorNode);
+        dependent.deleteDependency(operation);
+      }
+
+      const customParameters: readonly string[] = getCustomParameterValuesForPhase(phase);
```
dmichon-msft, line 124: This needs to be moved earlier since the collator needs them as well.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,212 @@
...
+        });
+        collatorNode.runner = collatorRunner;
+      } else {
+        collatorNode.runner = new NullOperationRunner({
+          name: collatorDisplayName,
+          result: OperationStatus.NoOp,
+          silent: phase.missingScriptBehavior === 'silent'
+        });
+      }
+      for (const dependent of operation.consumers) {
```
dmichon-msft, line 119:
```suggestion
      for (const consumer of operation.consumers) {
```
Consistent terminology. The reason the apis use `consumer` instead of `dependent` is because the difference between `consumer` and `dependency` is much more obvious than between `dependent` and `dependency`.

libraries/rush-lib/src/schemas/rush-project.schema.json
```diff
@@ -72,6 +72,49 @@
...
+                "type": "string",
+                "description": "A template string that specifies the command-line argument to pass to the operation for each shard. The string may contain the following placeholders: {shardIndex} {shardCount}. Defaults to --shard=\"{shardIndex}/{shardCount}\""
+              },
+              "outputFolderArgument": {
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                  "parentFolderNameFormat": {
+                    "type": "string",
+                    "description": "The name of the parent folder that contains the output folders for each shard. The string may contain the following placeholders: {phaseName}. Defaults to \".rush/operation/{phaseName}/shards/{shardIndex}\""
```
dmichon-msft, line 97: In practice there is no need to have a placeholder for `phaseName` because this is defined inside of the object where `phaseName` is defined. Therefore it is probably better practice for developers to simply hardcode the name into this string.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,212 @@
...
+      for (let shard: number = 1; shard <= shards; shard++) {
+        const shardOperation: Operation = new Operation({
+          project,
+          phase
+        });
+
+        // Add the shard argument to the custom parameters, replacing any templated values.
+        const shardArgumentFormat: string =
+          operationSettings.sharding.shardArgumentFormat ??
+          `--shard=${TemplateStrings.SHARD_INDEX}/${TemplateStrings.SHARD_COUNT}`;
```
dmichon-msft, line 148: Calculate this outside of the loop.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,212 @@
...
+
+        const shardArgument: string = shardArgumentFormat
+          .replace(TemplateStrings.SHARD_INDEX, shard.toString())
+          .replace(TemplateStrings.SHARD_COUNT, shards.toString());
+
+        const outputFolderArgumentFlag: string =
+          operationSettings.sharding.outputFolderArgument?.argumentFlag ?? '--shard-output-directory';
+        const outputDirectory: string = path.join(parentFolder, shard.toString());
```
dmichon-msft, line 156:
```suggestion
        const outputDirectory: string = `${parentFolder}/${shard}`;
```
This preserves use of `/` as the directory separator, which prevents the cache key from changing between Windows and Linux.

libraries/heft-config-file/package.json
```diff
@@ -16,7 +16,7 @@
   "license": "MIT",
   "scripts": {
     "build": "heft build --clean",
-    "start": "heft test --clean --watch",
+    "start": "heft build --clean --watch",
```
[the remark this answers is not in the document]
aramissennyeydd: Yup, noticed this while checking for start scripts before I jumped into the meat of the PR.

libraries/rush-lib/src/schemas/rush-project.schema.json
aramissennyeydd: Yeah, good point. I'll remove this.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,212 @@
...
+  formatCommand,
+  getCustomParameterValuesByPhase,
+  getDisplayName,
+  getScriptToRun
+} from './ShellOperationRunnerPlugin';
+import { NullOperationRunner } from './NullOperationRunner';
+import { OperationStatus } from './OperationStatus';
+import * as path from 'path';
+import { EnvironmentVariableNames } from '../../api/EnvironmentConfiguration';
+import { OperationMetadataManager } from './OperationMetadataManager';
```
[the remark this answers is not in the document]
aramissennyeydd: Another TIL, done!

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,212 @@
...
+      for (const dependent of operation.consumers) {
+        dependent.addDependency(collatorNode);
+        dependent.deleteDependency(operation);
+      }
+
+      const customParameters: readonly string[] = getCustomParameterValuesForPhase(phase);
+      const baseCommand: string | undefined = getScriptToRun(
+        project,
+        `${phase.name}:shard`,
+        phase.shellCommand
```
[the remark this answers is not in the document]
aramissennyeydd: sounds good, I think it makes sense from the view of the collator takes up the original phase and if needed you can pass additional properties to the shard.

libraries/rush-lib/src/logic/operations/CacheableOperationPlugin.ts
```diff
@@ -124,6 +124,7 @@ export class CacheableOperationPlugin implements IPhasedCommandPlugin {
             }
 
             const operationSettings: IOperationSettings | undefined =
+              operation.runner?.operationSettings ??
               projectConfiguration?.operationSettingsByOperationName.get(phaseName);
```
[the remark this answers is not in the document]
aramissennyeydd: Hoisted it to `Operation`, which is probably where I should have put it initially.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,212 @@
...
+
+      // Create a new one to avoid moving this into `beforeCreateOperation`.
+      const operationMetadataManager: OperationMetadataManager = new OperationMetadataManager({
+        phase,
+        rushProject: project,
+        operation
+      });
+      const parentFolder: string = parentFolderFormat.replace(
+        TemplateStrings.PHASE_NAME,
+        operationMetadataManager.logFilenameIdentifier
```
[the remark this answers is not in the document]
aramissennyeydd: Not sure I'm 100% following, but I moved all normalizing file functions to a single function in `OperationMetadataManager`

heft-plugins/heft-jest-plugin/src/JestPlugin.ts
```diff
@@ -591,9 +607,13 @@ export default class JestPlugin implements IHeftTaskPlugin<IJestPluginOptions> {
     } else {
       // If "silent" is specified via IJestPluginOptions, that takes precedence over jest.config.json
       silent = options.silent ?? jestConfig.silent ?? false;
+      shard = options.shard;
     }
 
     const jestArgv: Config.Argv = {
+      shard,
+      coverageDirectory: options.coverageDirectory,
```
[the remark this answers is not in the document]
aramissennyeydd: Any recommendations on how to solve for this? Seems like there's additional complexity around the cache folder being reachable for heft cleans (AFAIK those don't clean the `.rush` folder)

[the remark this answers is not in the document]
dmichon-msft: This gets extra messy because Heft's temp folder is not safe to use for the shards, since it'll get cleaned. Heft itself will need to most likely need to have the concept of shards baked in completely, so that it can scope the temp folder to be shard-specific.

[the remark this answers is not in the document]
aramissennyeydd: Do we want to split this work (heft/heft-jest-plugin) out into a separate PR? Sounds like there's some heft specific work that also needs to happen.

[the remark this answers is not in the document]
dmichon-msft: Yeah, we should do the Heft side as a separate PR. For testing just make a project with a phase script that logs the phase input parameters.

aramissennyeydd:
@dmichon-msft I trimmed out the heft changes - this should be good for another 👀

octogonz:
@iclanton @dmichon-msft Are we ready to merge this?

common/reviews/api/rush-lib.api.md
```diff
@@ -621,6 +626,8 @@ export interface IOperationSettings {
     disableBuildCacheForOperation?: boolean;
     operationName: string;
     outputFolderNames?: string[];
+    // Warning: (ae-forgotten-export) The symbol "IRushPhaseSharding" needs to be exported by the entry point index.d.ts
```
dmichon-msft, line 629: Add this export? Mark @internal if needed.

libraries/rush-lib/src/logic/operations/PhasedOperationPlugin.ts
```diff
@@ -71,10 +73,14 @@ function createOperations(
   function getOrCreateOperation(phase: IPhase, project: RushConfigurationProject): Operation {
     const key: string = getOperationKey(phase, project);
     let operation: Operation | undefined = operations.get(key);
+    const operationSettings: IOperationSettings | undefined = projectConfigurations
+      .get(project)
+      ?.operationSettingsByOperationName.get(phase.name);
```
dmichon-msft, line 78: Move this inside of the `if`, since we don't need to refetch when returning from cache.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,176 @@
...
+      const { count: shards, shardScriptConfiguration } = operationSettings.sharding;
+
+      existingOperations.delete(operation);
+
+      const collatorNode: Operation = new Operation({
+        phase,
+        project,
+        settings: operationSettings
+      });
+      existingOperations.add(collatorNode);
```
dmichon-msft, line 63: `collatorNode` is literally `operation` with its dependencies redefined; we could just reuse the existing operation and remove and relocate the dependencies.

To reduce the number of edges in the graph we could also inject a noop that keeps all the dependencies and that all the shards depend on, i.e.:

```
depA -\          /- shard 1 -\
depB -- > noop < -- shard 2 -- > collator (reused operation)
depC -/          \- shard 3 -/
```

dmichon-msft (approved):
Couple nits for performance, but otherwise looks about good to go.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/a/config/rush-project.json
```diff
@@ -0,0 +1,22 @@
...
+        },
+        "shardScriptConfiguration": {
+          "missingScriptBehavior": "silent"
+        }
+      }
+    },
+    {
+      "operationName": "cobuild",
+      "outputFolderNames": ["dist"]
+    }
```
iclanton, line 20: This isn't defined in `command-line.json`.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/a/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "a",
+  "version": "1.0.0",
+  "scripts": {
+    "cobuild": "node ../build.js a",
+    "build": "node ../build.js a",
+    "__phase:build": "exit 1",
```
iclanton, line 7: ?

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/b/config/rush-project.json
```diff
@@ -0,0 +1,18 @@
...
+        "count": 15,
+        "outputFolderArgument": {
+          "argumentFlag": "--output-directory"
+        }
+      }
+    },
+    {
+      "operationName": "cobuild",
+      "outputFolderNames": ["dist"]
+    }
```
iclanton, line 16: Same here.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/build.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-env es6 */
+const path = require('path');
+const { FileSystem } = require('@rushstack/node-core-library');
```
iclanton, line 3:
```suggestion
const { FileSystem, Async } = require('@rushstack/node-core-library');
```

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/build.js
```diff
@@ -0,0 +1,25 @@
...
+const shardOutputDir = getArgument('--shard-output-directory');
+
+const outputDirectory = shard ? shardOutputDir ?? outputDir : undefined;
+
+setTimeout(() => {
+  const outputFolder = shard ? path.resolve(outputDirectory) : path.resolve('dist');
+  const outputFile = path.resolve(outputFolder, 'output.txt');
+  FileSystem.ensureFolder(outputFolder);
+  FileSystem.writeFile(outputFile, `Hello world! ${args.join(' ')}`);
+}, 500);
```
iclanton, line 25:
```suggestion
async function runAsync() {
  await Async.sleepAsync(500);
  const outputFolder = shard ? path.resolve(outputDirectory) : path.resolve('dist');
  const outputFile = path.resolve(outputFolder, 'output.txt');
  FileSystem.writeFile(outputFile, `Hello world! ${args.join(' ')}`, { ensureFolderExists: true });
} 

void runAsync().catch(() => process.exit(1));
```

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/c/config/rush-project.json
```diff
@@ -0,0 +1,12 @@
+{
+  "operationSettings": [
+    {
+      "operationName": "_phase:build",
+      "outputFolderNames": ["dist"]
+    },
+    {
+      "operationName": "cobuild",
+      "outputFolderNames": ["dist"]
+    }
```
iclanton, line 10: Same here.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/collate.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-env es6 */
+const path = require('path');
+const { FileSystem } = require('@rushstack/node-core-library');
```
iclanton, line 3:
```suggestion
const { FileSystem, Async } = require('@rushstack/node-core-library');
```

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/collate.js
```diff
@@ -0,0 +1,30 @@
...
+    const outputFile = path.resolve(outputFolder, 'output.txt');
+    FileSystem.ensureFolder(outputFolder);
+    output += FileSystem.readFile(outputFile, 'utf-8');
+    output += '\n';
+  }
+  const finalOutputFolder = path.resolve('coverage');
+  const outputFile = path.resolve(finalOutputFolder, 'output.txt');
+  FileSystem.ensureFolder(finalOutputFolder);
+  FileSystem.writeFile(outputFile, output);
+}, 500);
```
iclanton, line 30:
```suggestion
async function runAsync() {
  await Async.sleepAsync(500);

  const parentFolder = process.env.RUSH_SHARD_PARENT_FOLDER;
  const shards = +process.env.RUSH_SHARD_COUNT;
  let output = '';
  for (let i = 1; i <= shards; i++) {
    const outputFolder = path.resolve(parentFolder, `${i}`);
    const outputFile = path.resolve(outputFolder, 'output.txt');
    FileSystem.ensureFolder(outputFolder);
    output += FileSystem.readFile(outputFile, 'utf-8');
    output += '\n';
  }
  const finalOutputFolder = path.resolve('coverage');
  const outputFile = path.resolve(finalOutputFolder, 'output.txt');
  FileSystem.writeFile(outputFile, output, { ensureFolderExists: true });
}

void runAsync().catch(() => process.exit(1));
```

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/d/config/rush-project.json
```diff
@@ -0,0 +1,12 @@
+{
+  "operationSettings": [
+    {
+      "operationName": "_phase:build",
+      "outputFolderNames": ["dist"]
+    },
+    {
+      "operationName": "cobuild",
+      "outputFolderNames": ["dist"]
+    }
```
iclanton, line 10: Same here.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/e/config/rush-project.json
```diff
@@ -0,0 +1,18 @@
...
+        "count": 75,
+        "shardScriptConfiguration": {
+          "missingScriptBehavior": "error"
+        }
+      }
+    },
+    {
+      "operationName": "cobuild",
+      "outputFolderNames": ["dist"]
+    }
```
iclanton, line 16: Same here.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/f/config/rush-project.json
```diff
@@ -0,0 +1,13 @@
...
+  "disableBuildCacheForProject": true,
+  "operationSettings": [
+    {
+      "operationName": "_phase:build",
+      "outputFolderNames": ["dist"]
+    },
+    {
+      "operationName": "cobuild",
+      "outputFolderNames": ["dist"]
+    }
```
iclanton, line 11: Same here.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/h/config/rush-project.json
```diff
@@ -0,0 +1,15 @@
...
+      "operationName": "_phase:build",
+      "outputFolderNames": ["dist"],
+      "sharding": {
+        "count": 50
+      }
+    },
+    {
+      "operationName": "cobuild",
+      "outputFolderNames": ["dist"]
+    }
```
iclanton, line 13: Same here.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/pre-build.js
```diff
@@ -0,0 +1,11 @@
...
+const path = require('path');
+const { FileSystem } = require('@rushstack/node-core-library');
+
+setTimeout(() => {
+  const outputFolder = path.resolve(process.cwd(), 'dist');
+  const outputFile = path.resolve(outputFolder, 'pre-build');
+  FileSystem.ensureFolder(outputFolder);
+  FileSystem.writeFile(outputFile, `Hello world!`);
+  console.log('done');
+}, 2000);
```
iclanton, line 11:
```suggestion
const { FileSystem, Async } = require('@rushstack/node-core-library');

async function runAsync() {
  await Async.sleepAsync(500);
  
  const outputFolder = path.resolve(process.cwd(), 'dist');
  const outputFile = path.resolve(outputFolder, 'pre-build');
  FileSystem.writeFile(outputFile, `Hello world!`, { ensureFolderExists: true });
  console.log('done');
}

void runAsync().catch(() => process.exit(1));
```

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/rush.json
```diff
@@ -0,0 +1,41 @@
+{
+  "rushVersion": "5.80.0",
```
iclanton, line 2: This is a pretty old version.

common/changes/@microsoft/rush/jest-sharding_2024-04-26-17-58.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft/rush",
+      "comment": "Adds support for sharding operations during task execution.",
```
iclanton, line 5: Does this require cobuilds?

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/rush.json
```diff
@@ -0,0 +1,41 @@
+{
```
iclanton, line 1: Does anything run this during build, or is this just for local testing?

A README in `build-tests/rush-redis-cobuild-plugin-integration-test` would be helpful.

libraries/rush-lib/src/api/EnvironmentConfiguration.ts
```diff
@@ -208,7 +208,19 @@ export const EnvironmentVariableNames = {
...
+   * Passed to a shard collating script to indicate the number of shards that were executed. This can be combined
+   *  with `RUSH_SHARD_PARENT_FOLDER` to determine the location of the individual shard output.
+   */
+  RUSH_SHARD_COUNT: 'RUSH_SHARD_COUNT',
+
+  /**
+   * Passed to a shard collating script to indicate the parent folder where shard output is stored.
+   * This can be combined with `RUSH_SHARD_COUNT` to determine the location of the individual shard output.
+   */
+  RUSH_SHARD_PARENT_FOLDER: 'RUSH_SHARD_PARENT_FOLDER'
```
iclanton, line 223: The environment variables listed here are generally things that users would pass into Rush. Perhaps these should enumerated elsewhere, and maybe they should have a different prefix? @octogonz - thoughts?

libraries/rush-lib/src/api/RushProjectConfiguration.ts
```diff
@@ -42,6 +42,47 @@ export interface IRushProjectJson {
...
 
+export interface IRushPhaseSharding {
+  /**
+   * The number of shards to create.
+   */
+  count: number;
+
+  /**
+   * The format of the argument to pass to the command to indicate the shard index and count.
+   * The default value is `--shard={shardIndex}/{shardCount}`.
```
iclanton, line 53:
```suggestion
   *
   * @defaultValue `--shard={shardIndex}/{shardCount}`
```

libraries/rush-lib/src/api/RushProjectConfiguration.ts
```diff
@@ -42,6 +42,47 @@ export interface IRushProjectJson {
...
+  shardArgumentFormat?: string;
+
+  /**
+   * An optional argument to pass to the command to indicate the output folder for the shard.
+   * The default argument is `--shard-output-folder=.rush/operations/{phaseName}/shards/{shardIndex}`.
+   */
+  outputFolderArgument?: {
+    /**
+     * The format of the parent folder name to output shards to. Child folders will be created for each
+     *  shard index, ie `{parentFolder}/1`, `{parentFolder}/2`. The default value is `.rush/operations/{phaseName}/shards`.
```
iclanton, line 64:
```suggestion
     *  shard index, ie `{parentFolder}/1`, `{parentFolder}/2`. 
     *
     * @defaultValue `.rush/operations/{phaseName}/shards`
```

libraries/rush-lib/src/api/RushProjectConfiguration.ts
```diff
@@ -42,6 +42,47 @@ export interface IRushProjectJson {
...
+     */
+    parentFolderName?: string;
+
+    /**
+     * The name of the argument to pass to the command to indicate the output folder for the shard.
+     * The default value is `--shard-output-folder`. Does not accept templating.
+     */
+    parameterLongName: string;
+  };
```
iclanton, line 73: For consistency with `shardArgumentFormat`, why not: 

```suggestion
  /**
   * An optional argument to pass to the command to indicate the output folder for the shard.
   *
   * @defaultValue `--shard-output-folder=.rush/operations/{phaseName}/shards/{shardIndex}`
   */
  outputFolderArgumentFormat?: string;
```

iclanton: Call out that this path is relative to the project root.

libraries/rush-lib/src/api/RushProjectConfiguration.ts
```diff
@@ -42,6 +42,47 @@ export interface IRushProjectJson {
...
+
+  /**
+   * Configuration for the collator operation that combines the shards.
+   */
+  shardScriptConfiguration?: {
+    /**
+     * What should happen if the "_phase:<name>:shard" script is not defined in a project's package.json scripts field. Default is "error".
+     */
+    missingScriptBehavior?: 'silent' | 'log' | 'error';
```
iclanton, line 82:
```suggestion
    missingScriptBehavior?: PhaseBehaviorForMissingScript;
```

libraries/rush-lib/src/api/RushProjectConfiguration.ts
```diff
@@ -42,6 +42,47 @@ export interface IRushProjectJson {
...
+  };
```
iclanton, line 83: Why not just use the operation's `missingScriptBehavior`?

libraries/rush-lib/src/cli/scriptActions/PhasedScriptAction.ts
```diff
@@ -47,6 +47,7 @@ import { RushProjectConfiguration } from '../../api/RushProjectConfiguration';
 import { LegacySkipPlugin } from '../../logic/operations/LegacySkipPlugin';
 import { ValidateOperationsPlugin } from '../../logic/operations/ValidateOperationsPlugin';
 import type { ProjectWatcher } from '../../logic/ProjectWatcher';
+import { ShardedPhasedOperationPlugin } from '../../logic/operations/ShardedPhaseOperationPlugin';
```
iclanton, line 50:
```suggestion
import { ShardedPhasedOperationPlugin } from '../../logic/operations/ShardedPhaseOperationPlugin';
import type { ProjectWatcher } from '../../logic/ProjectWatcher';
```

To keep the plugins together.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,176 @@
...
+export const PLUGIN_NAME: 'ShardedPhasedOperationPlugin' = 'ShardedPhasedOperationPlugin';
+
+// eslint-disable-next-line @typescript-eslint/typedef
+const TemplateStrings = {
+  SHARD_INDEX: '{shardIndex}',
+  SHARD_COUNT: '{shardCount}'
+} as const;
+
+// eslint-disable-next-line @typescript-eslint/typedef
+const TemplateStringRegex = {
```
iclanton, line 31:
```suggestion
const TemplateStringRegexes = {
```

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,176 @@
...
+
+function spliceShards(existingOperations: Set<Operation>, context: ICreateOperationsContext): Set<Operation> {
+  const { rushConfiguration } = context;
+
+  const getCustomParameterValuesForPhase: (phase: IPhase) => ReadonlyArray<string> =
+    getCustomParameterValuesByPhase();
+
+  for (const operation of existingOperations) {
+    const { associatedPhase: phase, associatedProject: project, settings: operationSettings } = operation;
+    if (phase && project && operationSettings?.sharding && !operation.runner) {
```
iclanton, line 53: So if an operation runner is already defined, the sharding operation settings are ignored?

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,176 @@
...
+      const parentFolderFormat: string =
+        operationSettings.sharding.outputFolderArgument?.parentFolderName ??
+        `.rush/operations/${phase.logFilenameIdentifier}/shards`;
+
+      const parentFolder: string = parentFolderFormat;
+
+      const collatorDisplayName: string = `${getDisplayName(phase, project)} - collate`;
+      const collatorRawScript: string | undefined = getScriptToRun(project, phase.name, undefined);
+
+      if (collatorRawScript === undefined && phase.missingScriptBehavior === 'error') {
```
iclanton, line 74: Handle the `phase.missingScriptBehavior === 'log'` case?

iclanton: Ah I see line 99. Is that what `'log'` does for non-sharded operations?

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,176 @@
...
+      }
+      for (const consumer of operation.consumers) {
+        consumer.addDependency(collatorNode);
+        consumer.deleteDependency(operation);
+      }
+
+      const baseCommand: string | undefined = getScriptToRun(project, `${phase.name}:shard`, undefined);
+
+      const shardMissingScriptBehavior: string = shardScriptConfiguration?.missingScriptBehavior ?? 'error';
+      if (baseCommand === undefined && shardMissingScriptBehavior === 'error') {
```
iclanton, line 110: Same here for the `=== 'log'` case.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,176 @@
...
+          phase,
+          settings: {
+            ...operationSettings,
+            outputFolderNames: [outputDirectory]
+          }
+        });
+
+        const shardArgument: string = shardArgumentFormat
+          .replace(TemplateStringRegex.SHARD_INDEX, shard.toString())
+          .replace(TemplateStringRegex.SHARD_COUNT, shards.toString());
```
iclanton, line 136: Is validation that the tokens are present necessary?

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,176 @@
...
+        const collatorRunner: ShellOperationRunner = new ShellOperationRunner({
+          commandToRun: formatCommand(collatorRawScript, customParameters),
+          displayName: collatorDisplayName,
+          phase,
+          rushConfiguration,
+          rushProject: project,
+          environment: {
+            ...process.env,
+            [EnvironmentVariableNames.RUSH_SHARD_PARENT_FOLDER]: parentFolder,
+            [EnvironmentVariableNames.RUSH_SHARD_COUNT]: shards.toString()
```
iclanton, line 91: Seems slightly odd that these are passed via env vars, but other values are passed via CLI.

iclanton (changes requested):

libraries/rush-lib/src/api/RushProjectConfiguration.ts
aramissennyeydd: Hmm, I think this shouldn't be configurable in the first place. Collation is optional but sharding should be required. This was from a previous iteration where the operation referenced the shard and not the collation. I'll remove this.

libraries/rush-lib/src/api/RushProjectConfiguration.ts
aramissennyeydd: I think we want the argument name to be configurable so you can pass it easily into existing commands and I couldn't think of a good way to make that happen without separating the parameter name and the value.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/rush.json
aramissennyeydd: I copied the existing cobuild sandbox repo for this. It uses the locally installed version of rush, though the start-dev script and doesn't use this version AFAIK.

common/changes/@microsoft/rush/jest-sharding_2024-04-26-17-58.json
aramissennyeydd: It shouldn't, there's nothing in the implementation that requires cobuilds to be set up.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Yeah, I was following what [`ShellOperationRunnerPlugin`](https://github.com/aramissennyeydd/rushstack/blob/42f63c2d88fea68022699345055e59b1036c9789/libraries/rush-lib/src/logic/operations/ShellOperationRunnerPlugin.ts#L40) and [`IPCOperationRunnerPlugin`](https://github.com/aramissennyeydd/rushstack/blob/42f63c2d88fea68022699345055e59b1036c9789/libraries/rush-lib/src/logic/operations/IPCOperationRunnerPlugin.ts#L50) do.

libraries/rush-lib/src/api/RushProjectConfiguration.ts
aramissennyeydd: I'll do a followup PR to align sharding with the new weighting, as I'd like to add weights to individual shards in my CI for jest tests.

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: Updated in [42f63c2](https://github.com/microsoft/rushstack/pull/4652/commits/42f63c2d88fea68022699345055e59b1036c9789)

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
aramissennyeydd: I believe so, I adjusted this to a single method in `ShellOperationRunnerPlugin` that handles this.

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/rush.json
aramissennyeydd: I'd like to hook this up to run during build but I'm not 100% sure how to. Currently, this is for local testing like the `sandbox/repo` folder for cobuilds.

aramissennyeydd:
The conversations that I've left unresolved above I have open questions around.

Current state of things, 
1. Shards are spliced into the current operation graph, with a pre-shard node that does nothing, a set of N shard nodes that do the sharding work (`_phase:${name}:shard`) and a single collate node that runs work over multiple shards (`_phase:${name}`)
2. Both shard + collate operations use the overall phase `missingScriptBehavior` configuration.
3. Operation weighting works with sharding as well, in the below timeline b shards have `weight=10` and a shards have weight 4. Parallelism is set to 10. As expected, only 1 b build is picked up at a time, and  3 a builds are picked up at once.
```
b (build) - shard 15/15 ###----------------------------------------------------------------------------- 2.1s
b (build) - shard 14/15 --####-------------------------------------------------------------------------- 2.1s
b (build) - shard 13/15 -----###------------------------------------------------------------------------ 2.1s
b (build) - shard 12/15 -------####--------------------------------------------------------------------- 2.1s
b (build) - shard 11/15 ----------####------------------------------------------------------------------ 2.1s
b (build) - shard 10/15 -------------###---------------------------------------------------------------- 2.1s
 b (build) - shard 9/15 ---------------####------------------------------------------------------------- 2.2s
 b (build) - shard 8/15 ------------------####---------------------------------------------------------- 2.2s
 b (build) - shard 7/15 ---------------------###-------------------------------------------------------- 2.2s
 b (build) - shard 6/15 -----------------------####----------------------------------------------------- 2.2s
 b (build) - shard 5/15 --------------------------####-------------------------------------------------- 2.2s
 b (build) - shard 4/15 -----------------------------###------------------------------------------------ 2.2s
 b (build) - shard 3/15 -------------------------------####--------------------------------------------- 2.2s
 b (build) - shard 2/15 ----------------------------------####------------------------------------------ 2.2s
 b (build) - shard 1/15 -------------------------------------###---------------------------------------- 2.1s
    b (build) - collate ---------------------------------------##--------------------------------------- 0.7s
  a (build) - shard 3/3 ---------------------------------------####------------------------------------- 2.2s
  a (build) - shard 2/3 ---------------------------------------####------------------------------------- 2.2s
  a (build) - shard 1/3 ---------------------------------------####------------------------------------- 2.1s
```
4. I adjusted the collate script to also use CLI parameters, `--shard-parent-folder` and `--shard-count`.
5. I also verified that the new build + collate scripts work as expected, example collate output below,
```
Hello world! b --shard=1/15 --output-directory=.rush/operations/_phase_build/shards/1
Hello world! b --shard=2/15 --output-directory=.rush/operations/_phase_build/shards/2
Hello world! b --shard=3/15 --output-directory=.rush/operations/_phase_build/shards/3
Hello world! b --shard=4/15 --output-directory=.rush/operations/_phase_build/shards/4
Hello world! b --shard=5/15 --output-directory=.rush/operations/_phase_build/shards/5
Hello world! b --shard=6/15 --output-directory=.rush/operations/_phase_build/shards/6
Hello world! b --shard=7/15 --output-directory=.rush/operations/_phase_build/shards/7
Hello world! b --shard=8/15 --output-directory=.rush/operations/_phase_build/shards/8
Hello world! b --shard=9/15 --output-directory=.rush/operations/_phase_build/shards/9
Hello world! b --shard=10/15 --output-directory=.rush/operations/_phase_build/shards/10
Hello world! b --shard=11/15 --output-directory=.rush/operations/_phase_build/shards/11
Hello world! b --shard=12/15 --output-directory=.rush/operations/_phase_build/shards/12
Hello world! b --shard=13/15 --output-directory=.rush/operations/_phase_build/shards/13
Hello world! b --shard=14/15 --output-directory=.rush/operations/_phase_build/shards/14
Hello world! b --shard=15/15 --output-directory=.rush/operations/_phase_build/shards/15
```

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/rush.json
iclanton: Fair enough.

libraries/rush-lib/src/api/RushProjectConfiguration.ts
iclanton: Do you want to do this before we publish? Will that be a breaking change if this is published first?

aramissennyeydd: It wouldn't be, bit of chaos in the comments here. I ended up getting the weights to work with sharding - it wasn't quite as bad as I thought it would be. Added in [791097f](https://github.com/microsoft/rushstack/pull/4652/commits/791097f8432104666a3d44d05e27092f2eae1a6b)

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/build.js
```diff
@@ -0,0 +1,30 @@
...
+const shard = getArgument('--shard');
+
+const outputDir = getArgument('--output-directory');
+
+const shardOutputDir = getArgument('--shard-output-directory');
+
+const outputDirectory = shard ? (shardOutputDir ? shardOutputDir : outputDir) : undefined;
+
+async function runAsync() {
+  await Async.sleepAsync(2000);
```
iclanton, line 21: Why so long?

common/reviews/api/rush-lib.api.md
```diff
@@ -776,6 +781,21 @@ export interface IRushCommandLineSpec {
     actions: IRushCommandLineAction[];
 }
 
+// Warning: (ae-internal-missing-underscore) The name "IRushPhaseSharding" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
```
iclanton, line 786: Mark this as `@alpha`

libraries/rush-lib/src/logic/operations/ShardedPhaseOperationPlugin.ts
```diff
@@ -0,0 +1,177 @@
...
+      existingOperations.add(preShardOperation);
+
+      for (const dependency of operation.dependencies) {
+        preShardOperation.addDependency(dependency);
+        operation.deleteDependency(dependency);
+      }
+
+      const parentFolderFormat: string =
+        operationSettings.sharding.outputFolderArgument?.parentFolderName ??
+        `.rush/operations/${phase.logFilenameIdentifier}/shards`;
```
iclanton, line 83:
```suggestion
        `${RushConstants.projectRushFolderName}/operations/${phase.logFilenameIdentifier}/shards`;
```

build-tests/rush-redis-cobuild-plugin-integration-test/sandbox/sharded-repo/projects/build.js
aramissennyeydd: Adjusted this to help me see if sharding is working as I would expect, made it a little clearer in timeline view and during cobuild logging if it was working.

iclanton (approved):

